PR #190, opened 2014-12-10
lints!

seanmonstar:
Specifically, this will prevent 2 code smells:
1. glob: `use foo::*` is by default denied. `use self::sub::*` is fine, as way to uplift sub modules. For uses of bad globs, they must be covered with a `#[glob = "explanation"]` attribute.
2. str_to_string: Micro benchmarks show it's 5x slower than `&str.into_string()`. So this lint denies such usage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: rust
 
 script:
-    - cargo build
-    - cargo test
+    - cargo build --features hyperlint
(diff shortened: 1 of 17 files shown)


Manishearth:
The tostring lint has been upstreamed into Clippy. I think the glob one would be nice to have as well, PRs welcome! :)

PR #200, opened 2014-12-18
Split hyper into smaller subpackages

DiamondLovesYou:
hyper-core, hyper-protocol, hyper-net, hyper-client, and hyper-server.

Existing ```extern crate hyper;``` behaviors remain the same, though obviously some rust paths are now different.
Also bump the version major because this is quite a breaking change.

I haven't checked the tests yet; I'll be able to get to that sometime after my chem final. I created the PR anyway to let you look it over while I'm so occupied.


reem:
It looks like progress and discussion on this PR has stalled, and it will require some serious rebasing to be relevant again. It seems that we don't need this change right now, but are open to it if done in a way that keeps maintenance and use simple. As a result, I'm closing this PR.

Feel free to open another PR with the changes updated or if you think further discussion is warranted.

PR #208, opened 2014-12-24
Other approach to implementing server SSL support

cactorium:
So I wrote another possible way to implement SSL! Lots of good Rust practice. So this approach uses generics and implemented the SSL layer as a wrapper that can be applied to another listener/acceptor/stream. This avoids breaking the `NetworkListener` API, because it instead breaks the `Server` API; there ends up being two possible values returned, based on whether the acceptor ends up being wrapped in the TLS wrappers or not, and I'm not sure how it should be cleaned up. Thoughts? Sorry for bothering you so much!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "hyper"
-version = "0.0.18"
+version = "0.0.19"
 description = "A modern HTTP library."
 readme = "README.md"
 documentation = "http://hyperium.github.io/hyper/hyper/index.html"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "hyper"
-version = "0.0.17"
+version = "0.0.18"
 description = "A modern HTTP library."
 readme = "README.md"
 documentation = "http://hyperium.github.io/hyper/hyper/index.html"
--- a/src/header/common/cache_control.rs
+++ b/src/header/common/cache_control.rs
@@ -109,7 +109,7 @@ impl FromStr for CacheDirective {
             "proxy-revalidate" => Some(ProxyRevalidate),
             "" => None,
             _ => match s.find('=') {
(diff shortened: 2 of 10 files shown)


cactorium:
What if the listeners were created using a factory:
```
pub trait ListenerFactory<L> {
    pub fn bind(...) -> Box<NetworkListener>;
    pub fn new() -> Self;
}

pub enum HttpListenerFactory<L> {
    Http,
    Https(Path, Path)
}

pub impl ListenerFactory<L> for HttpListenerFactory<L> {
    ...
}

pub struct IdentityListenerFactory<L> { }
pub impl ListenerFactory<L> for ... {
    ...
}

pub struct Server {
    // ...
    listener: Box<ListenerFactory<L>>,
}

impl Server {
    pub fn http(...) -> Server {
        Server::with_listener_factory(..., box HttpListenerFactory::<L>::Http)
    }

    pub fn https(..., key, cert) -> Server {
        Server::with_listener_factory(..., box HttpListenerFactory::<L>::Https(key, cert))
    }

    pub fn with_listener(...) -> Server {
        Server::with_listener_factory(..., box IdentityListenerFactory::<L>::new())
    }

    pub fn with_listener_factory(ip, port, listener: Box<NetworkListener>) -> Server {
        Server::with_listener_factory(..., box HttpListenerFactory::Https(key, cert))
    }
}
```

That way the factory could be used to choose between a custom listener, or the standard Http/Https one, and NetworkListener would work basically like how it does now. Or would that be too complicated?

seanmonstar:
I'm currently switching the traits over to use associated types. I think that will ease this is in.

seanmonstar:
Example:
```rust
trait NetworkListener {
    type Acceptor: NetworkAcceptor;
    // ...
}
```

reem:
Oh man - that will be such an improvement.

seanmonstar:
@cactorium I recently redesigned those traits as I said I would, can you where to plug in the Ssl pieces?

cactorium:
Wow, the code looks a lot cleaner! Gonna read through all of it and see if I can figure it out

PR #216, opened 2014-12-30
Refactor and add Accept-Encoding header

pyfisch:
Refactored the headers a bit and added the `Accept-Encoding` header. All tests pass, the code is ready for merge. I could do more cleanup in this request but I do not want to make this even bigger.


seanmonstar:
Excellent work! I'm going to toy with a `IntoQualityItem` trait to see how that looks, but that doesn't stop this. Just needs a rebase and I can merge it in.

pyfisch:
Rebased.

By the way, will Hyper use Rust 1.0.0-alpha, after it is released? And after that use the stable versions? It really sucks that the build fails so often.

reem:
@pyfisch Most likely yes, all of the feature gates that hyper uses extensively are slated to be stabilized.

seanmonstar:
@pyfisch I merged this manually, cleaning up the commits, as it seems somehow the commit history in this PR duplicated at some point.

PR #219, opened 2015-01-02
Use renamed unsafe_get and concat_vec methods.

mkovalchik:
Changed calls to unsafe_get to use get_unchecked and calls to concat_vec
to use concat to avoid deprecation lints.

These methods were renamed by commits [4f863a3](https://github.com/rust-lang/rust/commit/4f863a338e0a7c33f81a8ac138103f1a0e8b33c5) and [e91d810](https://github.com/rust-lang/rust/commit/e91d810b9b36d6bb163970cd0e8bbf4692f704bb) in rust-lang/rust


znewman01:
I found one you missed: https://github.com/hyperium/hyper/blob/master/src/header/mod.rs#L576 :smiley:

PR #220, opened 2015-01-02
`mark_dead` NetworkStream method to support client keep-alives.

jamwt:
When the `with_connector` client constructor is used, and those connections are being
retrieved from a pool, and re-used as keep-alive connections, the stream
needs some way to be notified that it is going to be closed by the
server because "Connection: close" was provided in the
response.

Then, the stream (which will be closed by the server and the next
request would fail) can be `drop()`-ped instead of being retained in the
pool.

(Also, brought current with some rust nightly changes.)

### DIFF
--- a/benches/client_mock_tcp.rs
+++ b/benches/client_mock_tcp.rs
@@ -87,6 +87,8 @@ impl net::NetworkStream for MockStream {
     fn peer_name(&mut self) -> IoResult<SocketAddr> {
         Ok(from_str("127.0.0.1:1337").unwrap())
     }
+    fn mark_dead(&mut self) {
+    }
 }
 
 struct MockConnector;
--- a/benches/client_mock_tcp.rs
+++ b/benches/client_mock_tcp.rs
@@ -87,8 +87,7 @@ impl net::NetworkStream for MockStream {
     fn peer_name(&mut self) -> IoResult<SocketAddr> {
         Ok(from_str("127.0.0.1:1337").unwrap())
     }
-    fn mark_dead(&mut self) {
-    }
+    fn mark_dead(&mut self) {}
 }
 
 struct MockConnector;
--- a/src/net.rs
+++ b/src/net.rs
@@ -48,6 +48,10 @@ pub trait NetworkAcceptor<S: NetworkStream>: Acceptor<S> + Clone + Send {
 pub trait NetworkStream: Stream + Any + StreamClone + Send {
     /// Get the remote address of the underlying connection.
     fn peer_name(&mut self) -> IoResult<SocketAddr>;
+
+    /// During a client request/response, the remote server did not want to keep-alive
+    /// the connection, so the stream cannot be reused for subsequent requests.
+    fn mark_dead(&mut self);
 }
 
 #[doc(hidden)]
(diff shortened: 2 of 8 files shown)


benches/client_mock_tcp.rs
```diff
@@ -87,6 +87,8 @@ impl net::NetworkStream for MockStream {
     fn peer_name(&mut self) -> IoResult<SocketAddr> {
         Ok(from_str("127.0.0.1:1337").unwrap())
     }
+    fn mark_dead(&mut self) {
```
reem: nit: style is `fn mark_dead(&mut self) {}`

jamwt: Okay, can do.

jamwt: (I can't wait for `rustfmt`; good way to learn idioms is to have them thrust upon you.)

src/net.rs
```diff
@@ -236,6 +240,9 @@ impl NetworkStream for HttpStream {
             Https(ref mut inner) => inner.get_mut().peer_name()
         }
     }
+
+    fn mark_dead(&mut self) {
```
reem: Shouldn't there be something here, or am I misunderstanding?

jamwt: No, the "built-in" `HttpStream` is created and then dropped when used by the client currently, so it doesn't particularly care about the state of the response header; it's going to close anyhow.

This is for custom `NetworkStream`s that are part of a pool and have additional fields that track health.

jamwt: (At least, so far; it's possible at some future date when client keep-alives are built into hyper in some fashion, that this method will be interesting again... but I'm treating that as out of scope of this change.)

reem: Ah, ok. You're saying this would be used if hyper had client-side keep-alive built-in.

reem:
@jamwt I'm going to wait and let @seanmonstar merge this since he's worked on the client more and has a better idea of if this is appropriate.

seanmonstar:
Why use mark_dead instead of handling this outside of the Request class?

jamwt:
So, you guys facilitate providing a connector which retrieves an stream.  Those streams are coming out of a pool.

This stream is a wrapper as part of the pool implementation that attaches to drop so the stream can be re-added to the pool when it is done--if the stream is still "healthy".  This has different definitions for mysql connections vs. http ones, etc, but the general gist is "can this stream be reused for the next request?"

So the connection trait has a method called "is_healthy(&mut self) -> bool".

Here's some application code using these abstractions:

    {
        // client is hyper::client::Client, grabbed one from the pool which may or may not be reused.
        // Easy to do this using `with_connector`.
        let mut client = pool.client();
        let response = client.get(/* something */).send().unwrap();
        /* do something with response, which is regular ol' hyper::client::Response */
    }
    // drop()

When the response drops, the stream wrapper drops, and it checks to see if it is "healthy" according to the definition of HTTP (keep-alive), and if it should be re-added to the pool for reuse.  This information is derived from the response headers.

Unfortunately, it's not clear to me how else to get access to those response headers without some support from the HTTP framework, or without having to force the user (i.e., the application author) to remember to always explicitly pass the response into something...

Or to completely replace the hyper client/request builder/response layer with wrappers all the way down, so that we were using a custom response object that we could implement Drop and have it check the inner response (the hyper one) for headers, and use into_inner to extract the stream, and then unsafe transmute the stream into the concrete type that we know it is... etc.  That seems like overkill, when this is the one piece of information missing (given `with_connector`) that's necessary to integrate with connection pools.

PR #225, opened 2015-01-04
Added AccessControl* family of headers.

### LINKED ISSUE #169: add AccessControl* family of headers
See https://github.com/seanmonstar/servo/blob/hyper/components/script/cors.rs#L385-L526

pyfisch:
Fixes #169


src/header/common/access_control/allow_headers.rs
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> Option<AccessControlAllowHeaders> {
+        shared::from_comma_delimited(raw).map(AccessControlAllowHeaders)
+    }
+}
+
+impl header::HeaderFormat for AccessControlAllowHeaders {
+    fn fmt_header(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let AccessControlAllowHeaders(ref parts) = *self;
```
reem: We have tuple indexing, so you can do `self.0` to get rid of this now!

src/header/common/access_control/allow_headers.rs
```diff
@@ -0,0 +1,25 @@
+use std::fmt::{mod};
+
+use header;
+use header::shared;
+
+#[deriving(Clone)]
+struct AccessControlAllowHeaders(pub Vec<String>);
```
reem: @seanmonstar should these be deref also?

src/header/common/access_control/allow_headers.rs
pyfisch: Hyper has many variants to do the common task of "how to get the inner value of the header struct for formatting"

* `self[]` (used by `Accept-Encoding` and `Accept`, requires `deref!`)
* `let tm = **self;` (used by `Date`)
* `self.0` (used by `Authorization`)

I personally prefer the first form `self[]` but also `self.0` is good. We just should make sure we use the same everywhere in the code.

Try it out: http://is.gd/ZkmjYV

src/header/common/access_control/allow_headers.rs
pyfisch: Probably it should be there.
But I do not exactly know what the `deref!` is doing and for what we need it.

pyfisch:
If this code is not perfect, please merge it anyway (after we are able to test it). All headers code needs a cleanup, I want to do. But I think it is better to this after the important headers are added.

reem:
@pyfisch that seems reasonable. I will merge this when we get hyper building again.

pyfisch:
@reem This is not finished yet probably Rust will complain about missing docs and other stuff.

pyfisch:
Hyper ~~is~~ **was**  again. The code compiled with `rustc 0.13.0-nightly (ad9e75938 2015-01-05 00:26:28 +0000)` for me.

So hopefully `url` is fixed soon so this can get merged.

src/header/common/access_control/allow_headers.rs
seanmonstar: - So, `self[]` is fine if passing a slice, like `&[T]` or `&str`. Examples: `struct UserAgent(pub String)`.
- `**self` **does** require `Deref`

PR #226, opened 2015-01-05
Replaced deprecated reference to trim_chars with trim_matches

alexandrubordei:
Just a minor replacement of a deprecated reference to trim_chars.


seanmonstar:
thanks!

PR #228, opened 2015-01-06
Fix chunk size parsing: handle invalid chunk sizes

mlalic:
Currently, when the chunk size is an invalid hex number (i.e. contains a non-hex digit), hyper would halt its parsing at the last valid hex digit and report this as the chunk size.

For example, the following would be the chunk sizes reported for some strings:

    "X" => 0
    "1X" => 1
    "aY" => 10
    "a;a" => 10 (only this should be correct!)

This fix makes it so that any chunk size, which cannot be parsed as a valid hex number, causes an `IoError`.

Additionally, the case where the chunk size is followed by a chunk extension is also handled -- the ";" token is allowed to immediately follow the chunk size, as it represents the delimiter between the size and the extensions.

The chunk extension parsing is (still) not fully compliant with the official syntax spec and accepts extensions deviating from it (or rather no error is raised). For example, the following shouldn't be a valid chunk size + extension (because the name needs to be a single token preceding the "=" separator):

    "A;name with spaces=value"

but it will not cause an error and will return `10` as the chunk size.

However, octets which should be part of the chunk extension are identified, meaning that if in the future the extension should be fully parsed and exposed by some API, the logic can be plugged in. Although, AFAIK, most HTTP parser implementations disregard the extensions in a similar manner as done here.

Regression tests for the `get_chunk_size` function and the parsing of requests in the `server` and responses in the `client` module are included.

### DIFF
--- a/src/client/response.rs
+++ b/src/client/response.rs
@@ -100,6 +100,8 @@ mod tests {
     use std::io::BufferedReader;
 
     use header::Headers;
+    use header::common::TransferEncoding;
+    use header::common::transfer_encoding::Encoding;
     use http::HttpReader::EofReader;
     use http::RawStatus;
     use mock::MockStream;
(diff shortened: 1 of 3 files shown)


seanmonstar:
This is amazing work. Thanks!

PR #229, opened 2015-01-07
macro syntax changes

globin:


src/lib.rs
```diff
@@ -1,6 +1,4 @@
-#![feature(macro_rules, phase, default_type_params,
-           slicing_syntax, globs, associated_types,
-           old_orphan_check)]
+#![feature(slicing_syntax, old_orphan_check)]
```
reem: +1000

globin:
Btw builds with mucell master, and carllerche/curl-rust#53.

src/lib.rs
seanmonstar: woo!

PR #231, opened 2015-01-08
WIP: Headers cleanup.

pyfisch:


seanmonstar:
This looks mostly awesome. Just a couple concerns around ease of use.

- `use hyper::header::LastModified` feels nicer than `use hyper::header::common::LastModified`.
- `Encoding` and `ConnectionOption` should be importable from the same place as their headers. One should be able to do `use hyper::header::{Connection, ConnectionOption}` or similar. Instead of one to `common` and another to `shared`.

src/header/common/mod.rs
```diff
@@ -1,13 +1,124 @@
 //! A Collection of Header implementations for common HTTP Headers.
 //!
-//! ## Mime
-//!
 //! Several header fields use MIME values for their contents. Keeping with the
-//! strongly-typed theme, the [mime](http://seanmonstar.github.io/mime.rs) crate
+//! strongly-typed theme, the [mime](//seanmonstar.github.io/mime.rs) crate
 //! is used, such as `ContentType(pub Mime)`.
```
reem: This should be changed to `such as in`

src/header/common/mod.rs
```diff
@@ -1,13 +1,124 @@
...
+//!
+//! Hyper aims to provide all common headers. This list shows which headers
```
reem: This is pretty awesome, but we will need to remember to update it.

pyfisch:
I agree that it minimally less convenient to have to import things from two places. But I think it is worth it, because it makes hyper::header::common very clean. It only contains headers. Since header names must be unique, there can not be collisions between different headers. Currently the struct for the `Cookie` header is called `Cookies`, not very easy to know. This is because the name `Cookie` already used by https://github.com/alexcrichton/cookie-rs If we only have headers in common there are no such problems and we can rename `Cookies` to `Cookie`.

The rust guide recommends to import modules and no individual items (http://doc.rust-lang.org/guide.html#standard-input "However, it's considered better practice to not import individual functions, but to import the module, and only use one level of qualification"), I think you should anyway not do `use hyper::header::{Connection, ConnectionOption}` but `use hyper::header`. (I know hyper does not yet follow this practice at most places, but I think it would be a very good thing if it did)

It may be a good idea to reexport all contents of `hyper::header::shared` to hyper::header`. If common would be renamed to `c` or something else very short you could do `use hyper::header::{self, c}` and then do `header::ConnectionOption` and `c::Connection`, which would be very easy to use, or don't you think so?

pyfisch:
I am thinking about suffixing all headers with `Header` this would also make clear what structs and enums are headers and prevent collisions. This will break most existing programs and make the names longer. @seanmonstar What do you think?

reem:
I am not a fan of that idea - `AccessControlAllowOrigin` is long enough already :/

seanmonstar:
Me neither. No need for suffixes or prefixes when they come from a `header` module.

PR #232, opened 2015-01-10
Update for latest Rust nightly

cyderize:
This allows hyper to compile on the Rust alpha release.

I've just done this as quickly as I could, so there are probably things that are done the wrong way, since I was more concerned about getting it to compile successfully and pass tests.

Also, what's the need for typeable? I don't see it being used anywhere, so I've removed it from the dependencies.


seanmonstar:
Amazing! I've been working on it too, but converting to associated types. I'll merge this, so that people can build, and then rebase against my assoc types work.

PR #233, opened 2015-01-10
Make NetworkConnector uses associated types.

saneyuki:


reem:
@seanmonstar You said you had an in-progress PR with a lot of changes like this.

seanmonstar:
This was taken care of as part of a larger rust upgrade. Thanks for the intent, tho!

PR #234, opened 2015-01-10
fixes invalid format of mime types

kstep:
See also hyperium/mime.rs#14.


kstep:
Seems like curl is out of date and makes the travis build to fail =(

reem:
@seanmonstar would you be against just dropping curl from the benches to make our builds more reliable until 1.0?

seanmonstar:
I dropped it myself in my current branch.

PR #235, opened 2015-01-10
Add `Referer` header definition

kstep:


src/header/common/referer.rs
```diff
@@ -0,0 +1,35 @@
...
+use header::shared::util::from_one_raw_str;
+
+/// The `Referer` header.
+///
+/// The Referer header is used by user agents to inform server about
+/// the page URL user has came from.
+///
+/// See alse [RFC 1945, section 10.13](http://tools.ietf.org/html/rfc1945#section-10.13).
+///
+/// Currently just a string, but maybe better replace it with url::Url or something like it.
```
reem: Could this be replaced with `url::Url`? It doesn't seem like a lot of additional effort would be needed.

kstep: Actually I used `Location` header definition as a model. And yes, I think it would be nice to make it Url, but I'm not sure it will work for every possible value, provided by, e.g., HTTP clients.

pyfisch: Actually it would be a good idea if header parsers could access the request to know the base URL. Like that the relative URL could be resolved immediately and we could use `url::Url`

kstep: But then headers must have access not only to a request, but to a response too. We need to generalize them if we want to do it this way, for example create `HttpContext` trait with common methods and implement it for both `HttpRequest` and `HttpResponse`.

seanmonstar: Is there need for more than just the Url?

pyfisch: Currently I do not think so. Maybe in the future?

kstep: This is the case, which makes me want for Rust to support implicits, like Scala.

Anyway, I wouldn't like to make special exception for base URL in respect to passing to around to header parsers, as it looks not generic enough for me.

kstep: And I think, if `Referer` should wrap around `Url`, so should do `Location`. And it's a place for another issue.

seanmonstar:
I'll merge this, so the header exists. Changing to a Url should happen the same time as Location.

kstep:
Thanks!

PR #236, opened 2015-01-11
Rust up for nightly

seanmonstar:


seanmonstar:
@reem this is mostly it, I need to clean up the benches to not use curl, and write the commit message, but I'm on my phone, so it'll have to wait till I get home.

reem:
@seanmonstar that's fine, I'll start reviewing.

Cargo.toml
```diff
@@ -11,16 +11,17 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
...
+openssl = "*"
 rustc-serialize = "*"
+time = "*"
+unicase = "*"
+unsafe-any = "*"
+url = "*"
 
 [dev-dependencies]
-curl = "*"
+#curl = "*"
```
reem: Should be deleted, not just commented.

src/client/mod.rs
```diff
@@ -243,15 +242,16 @@ pub enum Body<'a> {
     /// A Reader does not necessarily know it's size, so it is chunked.
     ChunkedBody(&'a mut (Reader + 'a)),
     /// For Readers that can know their size, like a `File`.
-    SizedBody(&'a mut (Reader + 'a), usize),
+    SizedBody(&'a mut (Reader + 'a), u64),
```
reem: Should this really be u64 and not usize?

src/client/request.rs
```diff
@@ -46,11 +45,14 @@ impl Request<Fresh> {
     }
 
     /// Create a new client request with a specific underlying NetworkStream.
-    pub fn with_connector<C: NetworkConnector<S>, S: NetworkStream>(method: method::Method, url: Url, connector: &mut C) -> HttpResult<Request<Fresh>> {
-        debug!("{:?} {:?}", method, url);
+    pub fn with_connector<
+        C: NetworkConnector<Stream=S>,
```
reem: Use a where clause instead, it's much cleaner.

src/client/mod.rs
seanmonstar: Yes. It could be from an fstat. usize on 32bit system will limit you to files less than 4gb.

src/header/mod.rs
```diff
@@ -5,23 +5,23 @@
...
-use std::{hash, mem, raw};
+use std::{mem, raw};
 
 use mucell::MuCell;
 use uany::{UnsafeAnyExt};
+use unicase::UniCase;
 
-use http::{self, LineEnding};
+use http;
 use {HttpResult};
```
reem: this line and the one above could be combined

src/http.rs
```diff
@@ -196,7 +196,7 @@ pub enum HttpWriter<W: Writer> {
     /// A Writer for when Content-Length is set.
     ///
     /// Enforces that the body is not longer than the Content-Length header.
-    SizedWriter(W, usize),
+    SizedWriter(W, u64),
```
reem: Here too, not usize for the size?

src/net.rs
```diff
@@ -26,29 +24,49 @@ pub struct Fresh;
...
+    fn accept(&mut self) -> IoResult<Self::Stream>;
+
+    /// Get the address this Listener ended up listening on.
+    fn socket_name(&self) -> IoResult<SocketAddr>;
+
     /// Closes the Acceptor, so no more incoming connections will be handled.
     fn close(&mut self) -> IoResult<()>;
+
+    /// Returns an iterator over incoming connections.
+    fn incoming<'a>(&'a mut self) -> NetworkConnections<'a, Self> {
```
reem: Shouldn't need this explicit lifetime.

src/net.rs
```diff
@@ -161,50 +180,58 @@ impl NetworkStream {
...
-        self.inner.close_accept()
+        match *self {
+            HttpAcceptor::Http(ref mut tcp) => tcp.close_accept(),
+            HttpAcceptor::Https(ref mut tcp) => tcp.close_accept(),
+        }
+    }
+
+    #[inline]
+    fn socket_name(&self) -> IoResult<SocketAddr> {
+        unimplemented!()
```
reem: Not ideal

src/server/mod.rs
```diff
@@ -43,29 +44,28 @@ macro_rules! try_option(
 impl Server<HttpListener> {
     /// Creates a new server that will handle `HttpStream`s.
     pub fn http(ip: IpAddr, port: Port) -> Server {
+        Server::with_listener(ip, port, HttpListener::Http)
+    }
+}
+
+impl<
+L: NetworkListener<Acceptor=A> + Send,
```
reem: where clause here too

reem:
Looks awesome except those nits. Associated types are a major improvement.

src/http.rs
seanmonstar: Likewise, if your server has files bigger than 4gb (which is becoming more common), then a 32bit server won't be able to serve them.

src/client/request.rs
seanmonstar: I don't see how it much cleaner, it looks the same to me except one is before the parens, one is after. But I moved it cause people seem to like it.

src/server/mod.rs
seanmonstar: Trying to use a where clause here on the impl gave errors that down below `L::Acceptor` couldn't be resolved:

```
/home/sean/code/hyper/src/server/mod.rs:64:101: 64:112 error: associated type `Acceptor` not found for type parameter `L`
/home/sean/code/hyper/src/server/mod.rs:64     pub fn listen_threads<H: Handler>(mut self, handler: H, threads: usize) -> HttpResult<Listening<L::Acceptor>> {
                                                                                                                                               ^~~~~~~~~~~
/home/sean/code/hyper/src/server/mod.rs:138:73: 138:84 error: associated type `Acceptor` not found for type parameter `L`
/home/sean/code/hyper/src/server/mod.rs:138     pub fn listen<H: Handler>(self, handler: H) -> HttpResult<Listening<L::Acceptor>> {
```

This should like be filed as an issue of Rust.

PR #240, opened 2015-01-12
Added TLS/SSL support serverside

cactorium:
Try three! Thoughts on whether more stuff should be configurable? Also, sorry for being too dense to figure out the refactoring you guys were talking about on the last PR

Tested with
```
extern crate hyper;

use std::io::net::ip::Ipv4Addr;
use std::path::posix::Path;
use hyper::server::{Request, Response};

static PHRASE: &'static [u8] = b"Hello World!";

fn hello(_: Request, res: Response) {
    let mut res = res.start().unwrap();
    res.write(PHRASE).unwrap();
    res.end().unwrap();
}

fn main() {
    hyper::Server::https(Ipv4Addr(127, 0, 0, 1), 3000,
            Path::new("./cert.pem"), Path::new("./key.pem"))
            .listen(hello).unwrap();
    println!("Listening on http://127.0.0.1:3000");
}
```

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -8,12 +8,21 @@ use std::io::net::ip::{SocketAddr, ToSocketAddr, Port};
 use std::io::net::tcp::{TcpStream, TcpListener, TcpAcceptor};
 use std::mem;
 use std::raw::{self, TraitObject};
+use std::sync::Arc;
 
 use uany::UnsafeAnyExt;
 use openssl::ssl::{Ssl, SslStream, SslContext, VerifyCallback};
-use openssl::ssl::SslVerifyMode::SslVerifyPeer;
+use openssl::ssl::SslVerifyMode::{SslVerifyPeer, SslVerifyNone};
 use openssl::ssl::SslMethod::Sslv23;
 use openssl::ssl::error::{SslError, StreamError, OpenSslErrors, SslSessionClosed};
+use openssl::x509::X509FileType;
+
+macro_rules! try_some {
+    ($expr:expr) => (match $expr {
+        Some(val) => { return Err(val); },
+        _ => {}
+    })
+}
 
 /// The write-status indicating headers have not been written.
 #[allow(missing_copy_implementations)]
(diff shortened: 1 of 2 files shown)


src/net.rs
```diff
@@ -218,7 +233,12 @@ impl NetworkAcceptor for HttpAcceptor {
     fn accept(&mut self) -> IoResult<HttpStream> {
         Ok(match *self {
             HttpAcceptor::Http(ref mut tcp, _) => HttpStream::Http(try!(tcp.accept())),
-            HttpAcceptor::Https(ref mut _tcp, _) => unimplemented!(),
+            HttpAcceptor::Https(ref mut _tcp, _, ref ssl_context) => {
```
seanmonstar: The `_tcp` before was to remove the unused warnings. Since it's used, the `_` can be removed.

src/net.rs
```diff
@@ -197,7 +199,20 @@ impl NetworkListener for HttpListener {
         let addr = try!(tcp.socket_name());
         Ok(match *self {
             HttpListener::Http => HttpAcceptor::Http(try!(tcp.listen()), addr),
-            HttpListener::Https => unimplemented!(),
+            HttpListener::Https(ref cert, ref key) => {
+                let mut ssl_context = try!(SslContext::new(Sslv23).map_err(lift_ssl_error));
+                if let Some(err) = ssl_context.set_cipher_list("DEFAULT") {
```
seanmonstar: So as to keep this in more of a straight line, how about each `if let` becoming `try!(ssl_context.something(...).map_err(lift_ssl_error);`?

seanmonstar:
Excellent! Along with fixing the nits, mind adjusting the commit message to match the new guidelines? https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md

It'll help me produce a nice changelog from now on.

cactorium:
Fixed! The `ssl_context` functions return a `Option`s instead of a `Result`s, so I made a macro called `try_some!` to act like `try!`

PR #241, opened 2015-01-12
Don't use Show to write UserAgent header

retep998:


seanmonstar:
Doh! Thanks for the fix. Mind adding a simple test in that same module testing that adding a UserAgent to a `Headers` and then checking the `to_string()` to be correct? 

And also, adjusting the commit message to match https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md ? In this case, it'd be something like `fix(headers): don't use Show to write UserAgent header`.

retep998:
Updated with test.

PR #251, opened 2015-01-17

pyfisch:
@seanmonstar and @reem, all contents of headers will be exposed at header
directly.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -6,7 +6,7 @@ use std::io::util::copy;
 use std::io::net::ip::Ipv4Addr;
 
 use hyper::{Get, Post};
-use hyper::header::common::ContentLength;
+use hyper::header::ContentLength;
 use hyper::server::{Server, Request, Response};
 use hyper::uri::RequestUri::AbsolutePath;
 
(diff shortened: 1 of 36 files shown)


reem:
The only thing I would change about this design is that the parsing helpers (`from_comma_delimited`, etc.) should be in their own module, since they feel like clutter here. Perhaps `header::parsing`?

pyfisch:
I will change it like you propose.

reem:
Hmm. Right now macros are exported at the crate level. I think it's fine to export them, since you only get them if you `#[macro_use]` anyway.

reem:
Added a bunch of nits.

pyfisch:
Fixed all nits. Moved parsing to its own file. (Rustoc did not like it otherwise, and we also use separate files for everything else)

reem:
Waiting on deps to run this on travis.

pyfisch:
Can someone please rerun this on Travis, or do I need to change something in the code?

seanmonstar:
I can't restart it on my phone, the Travis site sucks on mobile. The button
doesn't even render.

I can give it a kick when I get home.

seanmonstar:
Apparently, it doesn't work on desktop either. Try opening a new PR.

reem:
Ya, the travis site is bugging out for me too.

PR #259, opened 2015-01-20

pyfisch:
@seanmonstar and @reem, all contents of headers will be exposed at `header`
directly. Parsing utilities will be exposed at `header::parsing`. Header
macros can now be used from other crates.


GitCop:

 * Commit: 48186f4883222297f4840961edb82049e0b84d1b

---

PR #260, opened 2015-01-20

pyfisch:
@seanmonstar and @reem, all contents of headers will be exposed at `header`
directly. Parsing utilities will be exposed at `header::parsing`. Header
macros can now be used from other crates.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -6,7 +6,7 @@ use std::io::util::copy;
 use std::io::net::ip::Ipv4Addr;
 
 use hyper::{Get, Post};
-use hyper::header::common::ContentLength;
+use hyper::header::ContentLength;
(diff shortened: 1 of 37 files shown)


pyfisch:
Can you please merge it?

reem:
@pyfisch Thanks for pushing this though, it's a very worthwhile change.

rsolomo:
@pyfisch and @reem, I think the Basic scheme for Authorization headers is no longer publicly available after this change, unless I'm missing something?

reem:
It looks like we accidentally dropped a few things from public exports - @pyfisch think you could do a pass to make sure we get everything and PR the fixes?

pyfisch:
I will look if there are things I have forgotten.

PR #264, opened 2015-01-21
Make CacheDirective public

ProtectedMode:
This is needed to allow compilation of Iron's static as far as I can see, but maybe @reem has an alternative?


reem:
Looks like an oversight. Thanks for the PR.

PR #265, opened 2015-01-22
fix(headers): make Schemes, Basic, Protocol public

cyderize:
Expose Authorization header Scheme trait and Basic authorization
Expose the Upgrade header Protocol
Refactor ConnectionOption export to be more similar to CacheDirective export for consistency

This is required for people to use the Authorization and Upgrade headers.


seanmonstar:
Thanks!

PR #267, opened 2015-01-22
feat(headers): Add `Pragma` header field

### LINKED ISSUE #237: add Pragma header
This is needed for https://github.com/servo/servo/pull/4117/. 
Spec: https://tools.ietf.org/html/rfc7234#section-5.4

pyfisch:
Add the HTTP/1.0 `Pragma` header field used to prevent older Caches, that
do not understand the `Cache-Control` header field from caching the ressource.

Closes #237


src/header/common/pragma.rs
```diff
@@ -0,0 +1,62 @@
...
+
+impl Header for Pragma {
+    fn header_name(_: Option<Pragma>) -> &'static str {
+        "Pragma"
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> Option<Pragma> {
+        parsing::from_one_raw_str(raw).and_then(|s: String| {
+            let c = s.to_ascii_lowercase();
+            let slice = &c.clone()[];
```
seanmonstar: Why the clone here?

src/header/common/pragma.rs
```diff
@@ -0,0 +1,62 @@
...
+/// > HTTP/1.1).  When the Cache-Control header field is also present and
+/// > understood in a request, Pragma is ignored.
+
+/// > In HTTP/1.0, Pragma was defined as an extensible field for
+/// > implementation-specified directives for recipients.  This
+/// > specification deprecates such extensions to improve interoperability.
+///
+/// Spec: https://tools.ietf.org/html/rfc7234#section-5.4
+#[derive(Clone, PartialEq, Show)]
+pub enum Pragma{
```
reem: missing space before the {

src/header/common/pragma.rs
pyfisch: Ok, bug in code. I want to preserve the upper/lowercase of the value, but the matching should be case insensitive. You will see it in the new code.

src/header/common/pragma.rs
pyfisch: thx

src/header/common/pragma.rs
seanmonstar: If you have a change, you can use `git commit --amend` to add code to one commit, or `git rebase -i master` to smash multiple commits into one.

I imagine you can just remove the `.clone()`, and the rest should work, no?

src/header/common/pragma.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(Clone, PartialEq, Show)]
+pub enum Pragma {
+    /// Corresponds to the `no-cache` value.
+    NoCache,
+    /// Every value other than `no-cache`.
+    Ext(String),
+}
+
+impl Header for Pragma {
+    fn header_name(_: Option<Pragma>) -> &'static str {
```
seanmonstar: This has been changed in master, the `_` marker argument is no longer required.

src/header/common/pragma.rs
```diff
@@ -0,0 +1,62 @@
...
+/// > that they will understand (as Cache-Control was not defined until
+/// > HTTP/1.1).  When the Cache-Control header field is also present and
+/// > understood in a request, Pragma is ignored.
+
+/// > In HTTP/1.0, Pragma was defined as an extensible field for
+/// > implementation-specified directives for recipients.  This
+/// > specification deprecates such extensions to improve interoperability.
+///
+/// Spec: https://tools.ietf.org/html/rfc7234#section-5.4
+#[derive(Clone, PartialEq, Show)]
```
seanmonstar: Most recent rustup changed all `Show` to `Debug`, can you do the same here?

src/header/common/pragma.rs
pyfisch: You were right `.clone()` can be simply removed. It was needed for an older version that did not convert the string to lowercase.

pyfisch:
Rebased and fixed everything.

reem:
LGTM.

PR #268, opened 2015-01-23
refactor(headers): remove marker from header_name method

seanmonstar:
It is no longer required, as we can use `<H as Header>::header_name()`.

BREAKING CHANGE: Implementations of Header will need to adjust the
    header_name method. It no longer takes any arguments.


reem:
Rejoice!

PR #270, opened 2015-01-24
fix(headers): make ConnectionHeader and search for websocket protocol unicase

cyderize:
This makes ConnectionHeader case-insensitive since HTTP header field names are case-insensitive.

RFC6455 requires the Upgrade Protocol to search case-insensitively for "websocket". At the moment only lower-case works (I hit this when trying to get Autobahn to work with my websocket library).

Other protocol values may be case-sensitive, however, so ProtocolExt is still a normal case-sensitive String.


seanmonstar:
awesome!

PR #271, opened 2015-01-24
refactor(headers): Rename `Cookies` header to `Cookie`

pyfisch:
`Cookie` is the actual name of the header and since all other header structs
use the exact camel-cased version of their name using a different name here
is very inconvienient and confusing. You will encounter weird errors if you
try to use `Cookie` as the header. For this reason rename `Cookies` as
discussed on IRC with @seanmonstar and @reem and use `CookiePair` for real
cookies.

Breaks code using `Cookie` header.

### DIFF
--- a/src/header/common/cookie.rs
+++ b/src/header/common/cookie.rs
@@ -2,7 +2,7 @@ use header::{Header, HeaderFormat};
 use std::fmt;
 use std::str::from_utf8;
 
-use cookie::Cookie;
+use cookie::Cookie as CookiePair;
 use cookie::CookieJar;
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
Looks great. Could you change the commit message to end with "BREAKING CHANGE: desc"?

pyfisch:
This is not exactly what CONTRIBUTING.MD asks for:

>The footer should contain any information about Breaking Changes and is also the place to reference GitHub issues that this commit Closes.

And I wrote it in the footer. So change the guide or the message?

seanmonstar:
If that's not mentioned in contributing, I'll update it. It is part of the angular changelog scheme, which will generate a nice list including breaking changes.

PR #274, opened 2015-01-26
chore(warnings): Allow direct recursion for type enforcement.

reem:


reem:
Ugh multirust gave me a newer rustc than travis...

reem:
Seems like this was fixed in #280

PR #276, opened 2015-01-27
refactor(server): Add explicit lifetime annotations for Handler trait

Ryman:
It's different from the elided lifetimes, but having at least one explicit shared lifetime has allowed me to actually get the compiler to trust me on some dependent code. 

Any concerns on this being too restrictive?

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -175,7 +175,7 @@ pub trait Handler: Sync + Send {
     /// Receives a `Request`/`Response` pair, and should perform some action on them.
     ///
     /// This could reading from the request, and writing to the response.
-    fn handle(&self, Request, Response<Fresh>);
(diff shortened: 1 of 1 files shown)


reem:
Should be totally fine, especially since it seems you can elide these lifetimes on implementations with no problem.

seanmonstar:
I wonder if this is an example of https://github.com/rust-lang/rfcs/pull/738 ?

reem:
I had the same thought - I'm pretty sure it is, but I'm not certain.

reem:
Anyway I think this is good to go, since the lifetimes are actually always the same.

flubba86:
Oh wow, thank you so much for adding this.
I was maintaining a local modified version of Hyper with exactly this change in Handler, specifically so I could pass the request/response into my system entry point.

seanmonstar:
@flubba86 don't keep those issues to yourself! We wanna fix em.

PR #285, opened 2015-01-28
docs(version): Add fixup scope to contributing guidelines

stchris:
Sorry for abusing the 'version' scope :)


pyfisch:
The scope is (as far as I know) not limited to specific values.

seanmonstar:
True, the scope isn't checked for specific values. The ones in that list are suggested, as they will cover most cases.

seanmonstar:
Also, I'd use `rustup` personally.

PR #289, opened 2015-02-02
refactor(headers): switch from MuCell to OptCell

### LINKED ISSUE #252: Mucell often breaks Hyper
Mucell breaks Hyper really often. Like for example just now: https://travis-ci.org/hyperium/hyper/builds/47357537 I do not really know what mucell makes it superior to the standard library types, but for me it looks like it only provides a minor speed increasement. So do we really need it or could we switch to another builtin type?

seanmonstar:
It turns out, we don't need capabilities of MuCell (or even RefCell).
We don't to have sophisticated interior mutability. We don't ever lend
out references that may be mutated later. Instead, if there is no value
to lend, then we generate the value, require interior mutability to save
the value internally, and then we return a reference. We never ever
change a value once it's been generated. It can be changed, but only via
&mut self methods, where we can safely reason about mutability.

This means we don't need keep borrow checking code at runtime, which
helps performance. We also are able to reduce the amount of unsafe
transmutes. The internal API more safely constrains the usage of unsafe,
enforcing our invariants, instead of shotgunning unsafe usage with Refs
and promising we're doing it correctly.

On a machine with lower memory (1GB):

    Before:
    test bench_mock_hyper ... bench:    251772 ns/iter (+/- 128445)

    After:
    test bench_mock_hyper ... bench:    152603 ns/iter (+/- 25928)

On a machine with more memory, weaker CPU:

    Before:
    test bench_mock_hyper ... bench:    129398 ns/iter (+/- 51740)

    After:
    test bench_mock_hyper ... bench:    115935 ns/iter (+/- 28555)

Closes #252 

/cc @reem @chris-morgan

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ keywords = ["http", "hyper", "hyperium"]
 cookie = "*"
 log = ">= 0.2.0"
 mime = "*"
-mucell = "*"
 openssl = "*"
 rustc-serialize = "*"
 time = "*"
--- a/src/header/cell.rs
+++ b/src/header/cell.rs
@@ -1,0 +1,41 @@
+use std::cell::UnsafeCell;
+use std::ops::Deref;
+
+pub struct OptCell<T>(UnsafeCell<Option<T>>);
+
+impl<T> OptCell<T> {
+    #[inline]
+    pub fn new(val: Option<T>) -> OptCell<T> {
+        OptCell(UnsafeCell::new(val))
+    }
+
+    #[inline]
+    pub fn set(&self, val: T) {
+        unsafe {
+            let opt = self.0.get();
+            debug_assert!((*opt).is_none());
+            *opt = Some(val)
+        }
+    }
+
+    #[inline]
+    pub unsafe fn get_mut(&mut self) -> &mut T {
+        let opt = &mut *self.0.get();
+        opt.as_mut().unwrap()
+    }
+}
+
+impl<T> Deref for OptCell<T> {
+    type Target = Option<T>;
+    #[inline]
+    fn deref<'a>(&'a self) -> &'a Option<T> {
+        unsafe { &*self.0.get() }
+    }
+}
+
+impl<T: Clone> Clone for OptCell<T> {
+    #[inline]
+    fn clone(&self) -> OptCell<T> {
+        OptCell::new((**self).clone())
+    }
+}
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -16,15 +16,16 @@ use std::iter::FromIterator;
 use std::borrow::IntoCow;
 use std::{mem, raw};
 
-use mucell::MuCell;
 use uany::{UnsafeAnyExt};
 use unicase::UniCase;
 
+use self::cell::OptCell;
 use {http, HttpResult, HttpError};
 
 pub use self::shared::{Encoding, QualityItem, qitem};
 pub use self::common::*;
 
+mod cell;
 mod common;
 mod shared;
 pub mod parsing;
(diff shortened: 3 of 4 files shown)


chris-morgan:
(I believe a certain bluntness is needed to be effective in this case. Pray take no offence, any one who should read this.)

Speaking of the *status quo* with no reference to the changes enacted in this PR:

I’ve never looked closely at what hyper did with its header representation and mucell and so forth. I took a little closer look just now and I am not at all pleased at what I see.

Header reparsing (accessing it as multiple types) is not permitted, which while being more restrictive in corner cases does simplify various things. Most notably, it does indeed mean that mutation can be constrained to only occur through a mutable reference, and so indeed mucell’s niftiness is wasted. I deemed the misbehaviour in corner cases as unacceptable in Teepee, which is why I wrote mucell—but it does have additional costs, most notably that you cannot use plain references any more but must propagate new reference types that implement `Deref`. If you consider the corner-case misbehaviour acceptable, then you might as well bypass the small costs of mucell.

To the author or authors of all this code: your license to use the `unsafe` keyword and more importantly the `transmute` function has been revoked for 30 days and you will be on probation for 180 days thereafter. Injudicious usage of transmutation has left the header representation subsystem with notable memory safety bugs. This is in no manner the fault of mucell but rather the cruel and unusual treatment it has received (e.g. `mem::transmute::<&H, &H>(downcast(&*item.borrow()))`, ruining the reference counting, about which I may furthermore add that the `downcast(&*item.borrow())` part should not compile due to a lifetime conflict). As mentioned above, if you’re going to use mucell returning plain references is not an option; new reference types to handle the dropping of the reference count must be used, as is demonstrated in Teepee, the header representation scheme of which I believe to be devoid of memory-safety bugs.

I have not extensively assessed the impact of this PR, but I do know that my judgement on the issue is that the entire header module needs careful auditing and probably rewriting and I believe all unsafe code should be removed from the module with prejudice for the moment.

I will leave you with just one comment about the PR itself:

Bear in mind also that this `OptCell` is *not* a safe type; a reference may exist through `Deref` when `set` is called. `set` should therefore be marked an unsafe function.

chris-morgan:
I mostly retract my last paragraph; because `set` can only be called when the wrapped value is `None` (in debug mode, anyway) the worst that can happen is that a supposedly immutable `&None` changes underneath the variable binding to `&Some(…)`. This is bad, but does not affect memory safety as there can be no references to inside a `None`.

seanmonstar:
> Header reparsing (accessing it as multiple types) is not permitted, which while being more restrictive in corner cases does simplify various things.

I took the road of "we don't need it yet, and it grossly complicates things. Let's not have it for now, and if that's a terrible thing, people will file issues."

> To the author or authors of all this code: your license to use the unsafe keyword and more importantly the transmute function has been revoked for 30 days and you will be on probation for 180 days thereafter.

This was the theatrics I mentioned. The rest of the critique is perfectly fine.

> Injudicious usage of transmutation has left the header representation subsystem with notable memory safety bugs.

The unsafety is not as dire as it seems. Parsing can only happen once, and so the &H received cannot be changed from under you. The only way to change the value is by using a &mut method, which the borrow checker secures for us.

If an actual example of the unsafety can be shown, I'd love to learn from it.

> I have not extensively assessed the impact of this PR, but I do know that my judgement on the issue is that the entire header module needs careful auditing and probably rewriting and I believe all unsafe code should be removed from the module with prejudice for the moment.

At this point, the unsafe code that exists at this point is: a) downcast_unchecked, b) OptCell, exposing a safe exterior, and c) one lifetime transmute to handle an annoyance with `HashMap<CowString<'static>, V>`.

> Bear in mind also that this OptCell is not a safe type; a reference may exist through Deref when set is called. set should therefore be marked an unsafe function.

`set()` can only be called once. An assertion exists that the `Option.is_none()` before setting. You could take a reference to the `None`, but we don't. The only references returned are if the value is `Some`.

Any mutable access to the reference requires both an `&mut` method, and resetting the opposite value (raw or typed) so they stay in sync.

seanmonstar:
Merging since this doesn't change the semantics of the design, makes things faster and safer, and doesn't prevent #293. r=Manish

PR #290, opened 2015-02-03
fix(headers): add limit to maximum headers that should be parsed

### LINKED ISSUE #256: DOS vulnerability by sending a request with many many headers
We currently only check for the size of a single header field rather than all headers combined, this means a malicious client can send a request with thousands of headers and quickly consume a lot of memory on the server before the user of hyper can even access the request to reject it.

seanmonstar:
Closes #256

### DIFF
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -20,7 +20,7 @@ use mucell::MuCell;
 use uany::{UnsafeAnyExt};
 use unicase::UniCase;
 
-use {http, HttpResult};
+use {http, HttpResult, HttpError};
 
 pub use self::shared::{Encoding, QualityItem, qitem};
 pub use self::common::*;
@@ -118,6 +118,10 @@ pub struct Headers {
     data: HashMap<HeaderName, MuCell<Item>>
 }
 
+// To prevent DOS from a server sending a never ending header.
+// The value was copied from curl.
+const MAX_HEADERS_LENGTH: u32 = 100 * 1024;
+
 impl Headers {
 
     /// Creates a new, empty headers map.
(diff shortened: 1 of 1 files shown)


src/header/mod.rs
```diff
@@ -118,6 +118,8 @@ pub struct Headers {
     data: HashMap<HeaderName, MuCell<Item>>
 }
 
+const MAX_HEADERS_COUNT: u8 = 100;
```
reem: This feels a bit low. How about 256?

s-panferov: Maybe it's better to restrict the total size instead of the count?

seanmonstar: I noticed other parsers did indeed restrict the total size. I thought that would be complicated, but I just thought of a way.

s-panferov: We need to restrict the total size because actually it's possible to consume a lot of memory just by *one* header, e.g. Cookie.

seanmonstar: Well, that is already checked in `read_header`.

seanmonstar:
Updated to count characters instead of headers.

s-panferov:
:+1: Can we make it configurable somehow in the future?

src/header/mod.rs
```diff
@@ -118,6 +118,10 @@ pub struct Headers {
     data: HashMap<HeaderName, MuCell<Item>>
 }
 
+// To prevent DOS from a server sending a never ending header.
+// The value was copied from curl.
+const MAX_HEADERS_COUNT: u32 = 100 * 1024;
```
s-panferov: Maybe it's better to rename it? How about MAX_HEADERS_LENGTH?

seanmonstar:
Hm, I guess it could be configurable by passing it as an argument to `from_raw`...

src/header/mod.rs
reem: Agreed.

PR #295, opened 2015-02-06
feat(client): add connection timeout

ok32:


seanmonstar:
@ok32 thanks a lot for this! Unfortunately, it looks like the timeout methods are going away for a little while, as seen in the std::net RFC: https://github.com/rust-lang/rfcs/pull/807

As such, I'm hesitant to add an API that will soon disappear. Once it's re-added std::net, I'd certainly welcome the addition!

PR #300, opened 2015-02-07
feat(headers): add AcceptCharset header

hugoduncan:
Adds support for the Accept-Charset header.  Encodes the charset as a string.

### DIFF
--- a/src/header/shared/charset.rs
+++ b/src/header/shared/charset.rs
@@ -1,0 +1,150 @@
+use std::fmt::{self,Display};
+use std::str::FromStr;
+use std::ascii::AsciiExt;
+
+use self::Charset::*;
+
+/// A Mime charset.
+///
+/// The string representation is normalised to upper case.
+///
+/// See http://www.iana.org/assignments/character-sets/character-sets.xhtml
+#[derive(Clone,Debug,PartialEq)]
+#[allow(non_camel_case_types)]
+pub enum Charset{
+    /// US ASCII
+    Us_Ascii,
+    /// ISO-8859-1
+    Iso_8859_1,
+    /// ISO-8859-2
+    Iso_8859_2,
+    /// ISO-8859-3
+    Iso_8859_3,
+    /// ISO-8859-4
+    Iso_8859_4,
+    /// ISO-8859-5
+    Iso_8859_5,
+    /// ISO-8859-6
+    Iso_8859_6,
+    /// ISO-8859-7
+    Iso_8859_7,
+    /// ISO-8859-8
+    Iso_8859_8,
+    /// ISO-8859-9
+    Iso_8859_9,
+    /// ISO-8859-10
+    Iso_8859_10,
+    /// Shift_JIS
+    Shift_Jis,
+    /// EUC-JP
+    Euc_Jp,
+    /// ISO-2022-KR
+    Iso_2022_Kr,
+    /// EUC-KR
+    Euc_Kr,
+    /// ISO-2022-JP
+    Iso_2022_Jp,
+    /// ISO-2022-JP-2
+    Iso_2022_Jp_2,
+    /// ISO-8859-6-E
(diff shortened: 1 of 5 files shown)


pyfisch:
Looks good to me, although I would prefer a stronger typed charset type.

hugoduncan:
I am happy to add an enum for the charset type, but I think we would need to agree which list of charsets to include (and have the capability of specifying a string).  Would the list in the "Preferred MIME Name" column of the [IANA rgistry](http://www.iana.org/assignments/character-sets/character-sets.xhtml) (as suggested by [rfc2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.4)) be suitable?

seanmonstar:
I'd prefer stronger typed also, but that list is quite long. Can either keep as String for now, or try an enum with 5-10 of the most common, and then a `Ext(String)` variant.

pyfisch:
Actually it is not that long. It can be a separate crate.

hugoduncan:
The suggested list has 24 entries, which seems reasonable to me.  @pyfisch Would you be interested in owning such a crate (seems it would go well alongside [rust-language-tags](https://github.com/pyfisch/rust-language-tags))? @seanmonstar would you be OK adding such a dependency?

reem:
24 entries is short enough to live in hyper imo, but I wouldn't mind another crate in hyperium to help split things up.

seanmonstar:
24 can fit in hyper. I made mime a separate crate as it seemed useful for others.

hugoduncan:
Just noticed https://github.com/lifthrasiir/rust-encoding.

seanmonstar:
Oh nice. That'd be rad if we could integrate that, since we already depend on encoding (via rust-url). @lifthrasiir thoughts about a type that we could use in the headers? A label?

lifthrasiir:
@seanmonstar lifthrasiir/rust-encoding#64 is probably relevant. I'm not sure whether this is truly desirable or not though.

seanmonstar:
@lifthrasiir yep, that sounds exactly what would be best here...

src/header/shared/charset.rs
```diff
@@ -0,0 +1,150 @@
...
+            "ISO-8859-2" => Iso_8859_2,
+            "ISO-8859-3" => Iso_8859_3,
+            "ISO-8859-4" => Iso_8859_4,
+            "ISO-8859-5" => Iso_8859_5,
+            "ISO-8859-6" => Iso_8859_6,
+            "ISO-8859-7" => Iso_8859_7,
+            "ISO-8859-8" => Iso_8859_8,
+            "ISO-8859-9" => Iso_8859_9,
+            "ISO-8859-10" => Iso_8859_10,
+            "Shift-JIS" => Shift_Jis,
```
SBSTP: Shouldn't that be "SHIFT-JIS" since you're using uppercase?

seanmonstar:
merged manually

SBSTP:
I'm pretty sure the `Shift-JIS` bug is still there.
I added a test, `assert_eq!(Shift_Jis,"Shift-JIS".parse().unwrap());` and it doesn't work.
The line `"Shift-JIS" => Shift_Jis,` never matches and it goes straight to `Ext("SHIFT-JIS")`.

seanmonstar:
@SBSTP oh! i see what you mean, since `to_ascii_uppercase()` is called just before it. I was quite confused at first, since it's `Shift-JIS` in both `from_str` and `name`.

SBSTP:
Do you want me to PR a fix?

seanmonstar:
Sure. Though, if the name is supposed to mixed case like that, then the fix
must be about using to_uppercase. If it's not supposed to, then your
suggestion works perfectly.

SBSTP:
I was only going to fix the parsing part, since uppercase should handle mixed case.

PR #301, opened 2015-02-07
feat(headers): add IfMatch header

hugoduncan:
Add support for the If-Match http header.


pyfisch:
typo "IfMtach"

src/header/common/if_match.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl FromStr for IfMatch {
+    type Err = ();
+    fn from_str(s: &str) -> Result<IfMatch, ()> {
+        if s.trim()=="*" {
+            Ok(IfMatch(EntityTagMatch::Star))
+        } else {
+            let mut v : Vec<u8> = Vec::new();
+            v.push_all(s.as_bytes());
```
hugoduncan: I'm not happy with this, but haven't found an alternative yet.

pyfisch: Looks more like a problem with Hyper (only accepting Vec<Vec<u8>>) than your fault. I think it can stay like that until we rewrite hyper. :grin:

seanmonstar: I'd be fine with `parsing::from_comma_delimited` to accept `&[u8]` instead of `&[Vec<u8>]`.

src/header/common/if_match.rs
```diff
@@ -0,0 +1,82 @@
...
+
+deref!(IfMatch => EntityTagMatch);
+
+impl Header for IfMatch {
+    fn header_name() -> &'static str {
+        "If-Match"
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> Option<IfMatch> {
+        if raw[0][0]==b'*' {
```
seanmonstar: There isn't a promise that there is always a value. It may be empty, if it was just whitespace and then a newline, for instance. So, this could panic.

seanmonstar:
Mind squashing into a single commit? No real need to include "fixes" in the changelog for something that came in a feature in the same release.

src/header/common/if_match.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(PartialEq, Clone, Debug)]
+pub enum EntityTagMatch{
+    EntityTags(Vec<EntityTag>),
+    Star
+}
+
+/// The `If-Match` header
+///
+/// The `If-Match` header can be used by clients to indicate what
+/// response matchs they if.
```
reem: this sentence doesn't make sense, probably some typos

src/header/common/if_match.rs
```diff
@@ -0,0 +1,82 @@
...
+impl Header for IfMatch {
+    fn header_name() -> &'static str {
+        "If-Match"
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> Option<IfMatch> {
+        if raw.get(0).and_then(|x| x.get(0).map(|y| y == &b'*')).unwrap_or(false) {
+            Some(IfMatch(EntityTagMatch::Star))
+        } else {
+            match parsing::from_comma_delimited(raw).map(EntityTagMatch::EntityTags) {
```
reem: .map is preferable to this match

src/header/common/if_match.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+impl FromStr for IfMatch {
+    type Err = ();
+    fn from_str(s: &str) -> Result<IfMatch, ()> {
+        if s.trim()=="*" {
+            Ok(IfMatch(EntityTagMatch::Star))
+        } else {
+            let mut v : Vec<u8> = Vec::new();
```
reem: Can just replace this with s.as_bytes().to_vec()

Also, just FYI no need for annotation here and the style is `let mut v: Vec<u8>` (no spaces around `:`).

src/header/common/if_match.rs
```diff
@@ -0,0 +1,82 @@
...
+            v.push_all(s.as_bytes());
+            match parsing::from_comma_delimited(&[v]).map(EntityTagMatch::EntityTags) {
+                Some(tag) => Ok(IfMatch(tag)),
+                None => Err(())
+            }
+        }
+    }
+}
+
+#[test]
```
reem: Can you add some benches too like all the other headers? They are generated using a macro.

GitCop:

 * Commit: fc776e815e1e5de7ae550014d254b23fdf957fc9

 * Commit: e9a791f5d59e31fccc72c25a9ab674d51f69a9ed

 * Commit: 689a842a3d0e0ac10de66eb2f0a53489875d9abf

---

hugoduncan:
I addressed @reem's comments and squashed.

hugoduncan:
I see `IfNoneMatch` was added.  For consistency, I would have no objection to @fmendez adding an `IfMatch` that paralleled his `IfNoneMatch` implementation.

src/header/common/if_match.rs
```diff
@@ -0,0 +1,88 @@
...
+/// The `If-Match` header
+///
+/// The `If-Match` request-header field is used with a method to make
+/// it conditional.  The client provides a list of entity tags, and
+/// the request is only executed if one of those tags matches the
+/// current entity.
+///
+/// See http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.24
+#[derive(Clone, PartialEq, Debug)]
+pub struct IfMatch(pub EntityTagMatch);
```
seanmonstar: There's no need for the wrapper struct here. Just name the enum above "IfMatch".

hugoduncan: I added the wrapper with the thought that it could be shared between `IfNoneMatch` and `IfMatch`, and to have something to implement a function on for checking if an entity matched.  Since `IfNoneMatch` and `IfMatch` need to use different versions of entity tag matching anyway (weak vs strong), maybe that is less important.

PR #302, opened 2015-02-07
Add If-None-Match header

### LINKED ISSUE #238: add If-None-Match header
This is needed for https://github.com/servo/servo/pull/4117/.
Spec: https://tools.ietf.org/html/rfc7232#section-3.2

fmendez:
feat(headers): Add `If-None-Match` header field

Add the HTTP/1.1 `If-None-Match` header field makes the request method conditional
on a recipient cache or origin server either not having any current
representation of the target resource, when the field-value is "*",
or having a selected representation with an entity-tag that does not
match any of those listed in the field-value.

Closes #238


seanmonstar:
Thanks!

fmendez:
My pleasure!! :smile:

Issue #177, opened 2014-12-05
Support arbitrary status codes


Manishearth:
While the spec goes up till status code 599, we should probably support arbitrary three-digit codes.

[This test](https://github.com/w3c/web-platform-tests/blob/master/XMLHttpRequest/status-error.htm) [fails now](http://build.servo.org/builders/mac1/builds/377/steps/shell_1/logs/stdio) because of it (I don't mind allowing that failure for now, but I'd like a solution where hyper doesn't outright reject the code).

pyfisch:
Maybe we should just ask if they can fix their test? Just my 2 cents.
As I understand it a StatusCode higher thang 599 is just forbidden by all relevant specs.

Manishearth:
Not exactly. The web platform specs don't forbid it explicitly; and this means that there will be apps out there using it. Well, it's the other way around probably -- webapps were using higher status codes and the specs have to be written such that the web isn't broken just because the specs are too strict.

So the test is correct; not something to be "fixed".

pyfisch:
Ok, does anybody think that there are status codes out there greater 999 or smaller 100? In this case I would add the variants to the enum.

Manishearth:
Not that I know of. Perhaps ask in #whatwg on freenode

pyfisch:
I asked on #whatwg and @Ms2ger answered.

* XHR defines status codes as 16-bit attributes (unsigned :grinning: )
* HTTP/1.1 [RFC7231](https://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7231.html#status.codes) defines status codes as "three-digit integer code"
 * We were not sure if `042` is a "valid" status code
* [HTTP/2](https://http2.github.io/http2-spec/#HttpResponse) simply refers to HTTP/1.1 for a definition.

Manishearth:
Welcome to the world of web specs. We hope you stay, but you probably should get out while you can :joy_cat:

pyfisch:
I think it is hopeless :joy_cat:

Two more interesting facts:
* Both Chromium and Firefox understand 42 and 1024
* Only Chromium understands 2147483647, firefox thinks it is 200

Who wants to test IE, Safari and Opera?

seanmonstar:
Ps, my experiment with using a newtype around u16 stalled. I wanted to keep
similar usage to the enum now, but can't have a mod and a struct with the
same name. Associated constants would be needed to finish that.

pyfisch:
[...]

PR #303, opened 2015-02-08
feat(header): Support arbitary status codes

pyfisch:
As discussed in #177 hyper must support status code outside the standard range for compatibility reasons.

BREAKING CHANGE: This removes unregistered status codes from the enum. Use FromPrimitive methods to create them now.
StatusCode and StatusClass can no longer be casted to u16, use ToPrimitive methods now.
For example `status.to_u16().unwrap()` to get the status code number.

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -1,1048 +1,235 @@
-//! Status Codes
+//! HTTP status codes
 use std::fmt;
 use std::num::{FromPrimitive, ToPrimitive};
-use std::mem::transmute;
-
-use std::cmp::Ordering::{self, Less, Equal, Greater};
+use std::cmp::Ordering;
 
 // shamelessly lifted from Teepee. I tried a few schemes, this really
-// does seem like the best.
+// does seem like the best. Improved scheme to support arbitary status codes.
 
-/// An HTTP status code (`Status-Code` in RFC 2616).
+/// An HTTP status code (`status-code` in RFC 7230 et al.).
 ///
-/// This enum is absolutely exhaustive, covering all 500 possible values (100–599).
+/// This enum contains all common status codes and an Unregistered
+/// extension variant. It allows status codes in the range [0, 65535], as any
+/// `u16` integer may be used as a status code for XHR requests. It is
+/// recommended to only use values between [100, 599], since only these are
+/// defined as valid status codes with a status class by HTTP.
 ///
-/// For HTTP/2.0, statuses belonging to the 1xx Informational class are invalid.
-///
-/// As this is a C‐style enum with each variant having a corresponding value, you may use the likes
-/// of `Continue as u16` to retreive the value `100u16`. Normally, though, you should not need to do
-/// any such thing; just use the status code as a `StatusCode`.
-///
-/// If you encounter a status code that you do not know how to deal with, you should treat it as the
-/// `x00` status code—e.g. for code 123, treat it as 100 (Continue). This can be achieved with
+/// If you encounter a status code that you do not know how to deal with, you
+/// should treat it as the `x00` status code—e.g. for code 123, treat it as
+/// 100 (Continue). This can be achieved with
 /// `self.class().default_code()`:
 ///
 /// ```rust
-/// # use hyper::status::StatusCode::{Code123, Continue};
-/// assert_eq!(Code123.class().default_code(), Continue);
+/// # use std::num::FromPrimitive;
+/// # use hyper::status::StatusCode;
+/// let statusopt: Option<StatusCode> = FromPrimitive::from_u16(137u16);
+/// assert_eq!(statusopt.unwrap().class().default_code(), StatusCode::Continue);
 /// ```
-#[derive(Debug)]
+///
+/// IANA maintain the [Hypertext Transfer Protocol (HTTP) Status Code
+/// Registry](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) which is
+/// the source for this enum (with one exception, 418 I'm a teapot, which is
+/// inexplicably not in the register).
 pub enum StatusCode {
     /// 100 Continue
-    Continue = 100,
+    /// [[RFC7231, Section 6.2.1](https://tools.ietf.org/html/rfc7231#section-6.2.1)]
+    Continue,
     /// 101 Switching Protocols
-    SwitchingProtocols = 101,
+    /// [[RFC7231, Section 6.2.2](https://tools.ietf.org/html/rfc7231#section-6.2.2)]
+    SwitchingProtocols,
     /// 102 Processing
-    Processing = 102,
-    /// 103 (unregistered)
-    Code103 = 103,
-    /// 104 (unregistered)
-    Code104 = 104,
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: c14bffb64ba9f27b5a09e001c22b3aca4465777d

---

src/status.rs
```diff
@@ -1,1601 +1,331 @@
...
-    Code597 = 597,
-    /// 598 (unregistered)
-    Code598 = 598,
-    /// 599 (unregistered)
-    Code599 = 599,
+    /// [[RFC6585](https://tools.ietf.org/html/rfc6585)]
+    NetworkAuthenticationRequired,
+
+    /// A method not in the IANA HTTP status code registry.
+    UnregisteredStatusCode(u16),
```
seanmonstar: Since it's already got the StatusCode namespace, I'd leave this as just 'Unregistered'.

src/status.rs
```diff
@@ -1,1601 +1,331 @@
...
     }
 
     /// Determine the class of a status code, based on its first digit.
     pub fn class(&self) -> StatusClass {
-        let code = *self as u16;  // Range of possible values: 100..599.
-        // We could match 100..199 &c., but this way we avoid unreachable!() at the end.
-        if code < 200 {
-            StatusClass::Informational
+        let code = self.to_u16().unwrap();
+        // TODO rewrite
```
seanmonstar: rewrite? either remove, or explain more.

src/status.rs
```diff
@@ -1,1601 +1,331 @@
...
-            StatusClass::ClientError
+            ClientError
+        } else if code < 600{
+            ServerError
         } else {
-            StatusClass::ServerError
+            NoClass
         }
     }
-
```
seanmonstar: Why remove all these convenience methods? The allow you figure out the class without requiring an additional import.

src/status.rs
```diff
@@ -1603,36 +333,118 @@ impl Copy for StatusCode {}
...
 }
 
-// Specified manually because the codegen for derived is slow (at the time of writing on the machine
-// of writing, 1.2 seconds) and verbose (though the optimiser cuts it down to size).
+impl fmt::Debug for StatusCode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let tmp_s;
+        f.write_str(match *self {
+            UnregisteredStatusCode(ref code) => {
+                tmp_s = format!("StatusCode{}", code);
```
seanmonstar: to prevent this allocation, you could do `return write!(f, ...)`

seanmonstar: Also, this would mean a code of like 999 would be `StatusCode999`. Is that actually desirable, or should it be `Unregistered(999)`?

src/status.rs
```diff
@@ -1640,22 +452,140 @@ impl Clone for StatusCode {
...
-// Of the other common derivable traits, I didn’t measure them, but I guess they would be slow too.
-
 impl FromPrimitive for StatusCode {
     fn from_i64(n: i64) -> Option<StatusCode> {
-        if n < 100 || n > 599 {
+        if n < 0 || n > 65535 {
             None
         } else {
-            Some(unsafe { transmute::<u16, StatusCode>(n as u16) })
+            Some(match n {
```
seanmonstar: Since this is already tricky (it's possible to accidentally miss a variant in this match), it'd be best if there was only one of these lists. So here, I would defer to `from_u64`. Like: `if n < 0 { None } else { FromPrimitive::from_u64(n as u64) }`

src/status.rs
```diff
@@ -1671,106 +601,225 @@ impl Ord for StatusCode {
...
+            LoopDetected => 508,
+            NotExtended => 510,
+            NetworkAuthenticationRequired => 511,
+            UnregisteredStatusCode(n) => n,
+        } as i64)
     }
 
     fn to_u64(&self) -> Option<u64> {
-        Some(*self as u64)
+        Some(match *self {
```
seanmonstar: Same reasoning as above regarding FromPrimitive applies here: `ToPrimitive::to_i64(self).map(|n| n as u64)`

src/status.rs
```diff
@@ -1780,18 +829,41 @@ impl StatusClass {
     /// assert_eq!(status, BadRequest);
     /// // So now let’s treat it as that.
     /// ```
-    #[inline]
+    /// All status codes that do not map to an existing status class are matched
+    /// by a `NoClass`, variant that resolves to 200 (Ok) as default code.
+    /// This is a common handling for unknown status codes in major browsers.
```
seanmonstar: Have sources? I'd be curious.

@Manishearth do you agree? Or should it resolve to something else?

src/status.rs
```diff
@@ -1,1601 +1,331 @@
...
 use std::num::{FromPrimitive, ToPrimitive};
-use std::mem::transmute;
-
-use std::cmp::Ordering::{self, Less, Equal, Greater};
+use std::cmp::Ordering;
 
-// shamelessly lifted from Teepee. I tried a few schemes, this really
-// does seem like the best.
+pub use self::StatusCode::*;
+pub use self::StatusClass::*;
```
seanmonstar: I'd rather keep these namespaced by enum for now. If there's much demand, this can be an immediate follow up, but it seems like a different decision than this PR.

seanmonstar:
Test failure was in a doctest, that `StatusCode` wasn't imported.

src/status.rs
Manishearth: I haven't heard of this, but I've not really looked at the status code part of implementations

src/status.rs
reem: agreed

src/status.rs
reem: I would prefer `Unregistered(999)`, especially for a `Debug` implementation.

reem:
@pyfisch any news here? We'd like to land this to unblock servo/servo#117.

Manishearth:
You mean https://github.com/servo/servo/issues/4246 :)

reem:
Wow I have no idea what I was looking at.

src/status.rs
pyfisch: `StatusCode253` was the old formatting, I think. But I will change it.

src/status.rs
pyfisch: Remove.

src/status.rs
pyfisch: I will readd them.

pyfisch:
If anything is left to do tell me.

src/status.rs
pyfisch: I just found this out when I entered really crazy StatusCodes to Chrome and Firefox. Sometimes they used 200

pyfisch: Here: 
> Only Chromium understands 2147483647, firefox thinks it is 200
> 0 makes problems (Not displayed or understood as 200)

seanmonstar: What's the Url? Mind including it in the comment?

seanmonstar:
I fixed up a couple tiny nits, and pushed this as #311

PR #294, opened 2015-02-06
fix(readme): Make the README client example work

xymostech:
`res` needs to be unwrapped, we want to print out `body`, and `res` doesn't need to be mutable.

PR #306, opened 2015-02-10
docs(readme): Make the README client example work

ctjhoa:
`res` needs to be mutable (introduce with #294) otherwise
```
   Compiling toto v0.0.1 (file:///home/ctjhoa/rust/toto)
src/main.rs:14:16: 14:20 error: cannot borrow immutable local variable `resp` as mutable
src/main.rs:14     let body = res.read_to_string().unwrap();
                              ^~~~
Could not compile `toto`.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ fn main() {
     let mut client = Client::new();
 
     // Creating an outgoing request.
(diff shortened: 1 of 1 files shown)


reem:
Thanks!

PR #307, opened 2015-02-11
fix(net): don't stop the server when an SSL handshake fails with EOF

cyndis:
HttpAcceptor::accept()'s HTTPS logic passes IO errors from the underlying
SSL stream directly to the caller. Furthermore, the caller uses the
EndOfFile error code to detect that the server should stop accepting
connections. This means that if the TCP connection was succesfully
accepted, but an EOF condition was detected during the handshake, the
server will stop accepting connections and quit. This allows for a
trivial denial of service attack and can happen accidentally as well.

Change HttpAcceptor::accept such that if the TCP stream underlying
the SSL stream returns an EndOfFile error, change the error kind
to OtherIoError.

----

The way I fixed this is a bit ugly, but can't think of a nicer and easy way at the moment.
Suggestions welcome.

Cheers,
cyndis

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -237,9 +237,17 @@ impl NetworkAcceptor for HttpAcceptor {
             HttpAcceptor::Http(ref mut tcp, _) => HttpStream::Http(try!(tcp.accept())),
             HttpAcceptor::Https(ref mut tcp, _, ref ssl_context) => {
                 let stream = try!(tcp.accept());
-                let ssl_stream = try!(SslStream::<TcpStream>::new_server(&**ssl_context, stream).
-                                     map_err(lift_ssl_error));
-                HttpStream::Https(ssl_stream)
+                match SslStream::<TcpStream>::new_server(&**ssl_context, stream) {
+                    Ok(ssl_stream) => HttpStream::Https(ssl_stream),
+                    Err(StreamError(ref e)) => {
+                        return Err(IoError {
+                            kind: ConnectionAborted,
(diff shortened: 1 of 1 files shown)


seanmonstar:
This could also happen to the client, right? Perhaps the logic would be best in `lift_ssl_error`? Also, looking in that function, I see `ConnectionAborted` being used. Perhaps that'd be a better error here as well, instead of `OtherIoError`?

cyndis:
The client can also get an EndOfFile but there it should be handled correctly, here the issue is that the server code thinks the EndOfFile is coming from the acceptor, not the stream. So really, we probably should be turning all IoErrors from the stream in accept() into something else so that they are not confused with IoErrors from the acceptor. I guess ConnectionAborted would be a good choice. That would hide the underlying reason, but I can't think why it would be ever needed.

GitCop:

 * Commit: 37a71e8634988cc316cbcd8e86a60d2429d9050d

---

cyndis:
Here's v2 according to what I described in the previous message. I will squash these once the patch is ready.

seanmonstar:
Looks good to me.

cyndis:
Great, I'll squash.

cyndis:
Whoops, something went wrong there..

cyndis:
Fixed.

seanmonstar:
Awesome work, thank you! Just waiting on travis to go green.

cyndis:
Thanks for a great library :)

PR #310, opened 2015-02-14
fix(server): Drain requests on drop.

### LINKED ISSUE #197: `request error = HttpMethodError` logs
I'm finding the following in my server logs with some frequency:

    ERROR:hyper::server: request error = HttpMethodError

After spending some time on this, I've found a simple reproduction that seems to hinge on requests with a chunked transfer encoding, though I couldn't tell if it was the hyper client or the server that was incorrect. Wireshark hinted that it may be the client, but I believe that I've encountered this log message in the wild and not just when testing the hyper server with the hyper client.

The log message is produced when the server is trying to read the http method of a request but encounters unexpected characters. In the case of the error reproduction below, the server appears to be trying to read for a second request on the connection, and it encounters \r\n instead of the expected http method.

```rust
extern crate test;
extern crate hyper;

use std::io;
use std::rand;
use hyper::method::Method;
use hyper::client::Request;
use hyper::server;
use hyper::status::StatusCode;

#[test]
fn http_method_error_is_logged() {
    let ts = TestServer::new();
    Request::new(Method::Post, ts.get_url())
        .unwrap().start().unwrap().send().unwrap();
    io::stdio::stderr().write_line(
        "^^^^^^^     That ERROR log message there     ^^^^^^^").unwrap();

    // I couldn't work out how to capture the error logs, stdio::set_stderr()
    // only catches messages sent with panic! I think. As a result this "test" 
    // doesn't assert, you just have to look at the printed output.

    // Things which cause this error to not reproduce:
    //   * changing the request to one with no body, like a GET
    //   * providing a Content-Length on the request
}

fn hello(_: server::Request, mut res: server::Response) {
    *res.status_mut() = StatusCode::Ok;
    let mut res = res.start().unwrap();
    res.write(b"Hello World!").unwrap();
    res.end().unwrap();
}


struct TestServer {
    listen_serv : hyper::server::Listening,
    port : u16,
}

impl TestServer {
    fn new() -> TestServer {
        let host = from_str("127.0.0.1").unwrap();
        // Try to guess a free port that we can bind to:
        for _ in range(0, 100u) {
            let port = rand::random();
            let server = hyper::server::Server::http(host, port);

            match server.listen(hello) {
                Ok(listening) => {
                    return TestServer {
                        listen_serv: listening,
                        port: port
                    }
                },
                Err(_) => (),
            };
        }

        panic!("Could not find a port to bind the test server to!");
    }

    fn get_url(&self) -> hyper::Url {
        let mut url = "http://127.0.0.1:".to_string();
        url.push_str(self.port.to_string().as_slice());
        url.push_str("/");
        hyper::Url::parse(url.as_slice()).unwrap()
    }
}

impl Drop for TestServer {
    fn drop(&mut self) {
        self.listen_serv.close().unwrap();
    }
}
```

### LINKED ISSUE #309: Read body for all requests with a Content-Length header
The framing of a HTTP request doesn't depend on the method [1], so any request with a non-zero Content-Length has a corresponding body. This does not change in the case of idempotent methods like GET; what is different is that for such methods it is incorrect for the server's response to depend on the content of this body.

At the moment GET and HEAD are incorrectly special cased in hyper [2] leading to problems if you send a GET request with a body (the unread body bleeds into the following request unless the connection is closed). In these cases the body data, if any, must be read from the network but could be discarded and not made avaliable through the request API.

[1] https://tools.ietf.org/html/rfc7230#section-3.3
[2] https://github.com/hyperium/hyper/blob/master/src/server/request.rs#L42

seanmonstar:
If a client sent an illegal request (like a GET request with a message
body), or if there was a legal request with a body but the Handler
didn't read all of it, the remaining bytes would be left in the stream.
The next request to come from the same client would error, as the server
would confuse the remaining bytes, and think the request was malformed.

Fixes #197
Fixes #309

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -3,7 +3,10 @@ use std::borrow::Cow::{Borrowed, Owned};
 use std::borrow::IntoCow;
 use std::cmp::min;
 use std::old_io::{self, Reader, IoResult, BufWriter};
+use std::old_io::util as io_util;
+use std::mem;
 use std::num::from_u16;
+use std::ptr;
 use std::str;
 use std::string::CowString;
 
@@ -20,7 +23,7 @@ use HttpError::{HttpHeaderError, HttpIoError, HttpMethodError, HttpStatusError,
                 HttpUriError, HttpVersionError};
 use HttpResult;
 
-use self::HttpReader::{SizedReader, ChunkedReader, EofReader, EmptyReader};
+use self::HttpReader::{SizedReader, ChunkedReader, EofReader};
 use self::HttpWriter::{ThroughWriter, ChunkedWriter, SizedWriter, EmptyWriter};
 
 /// Readers to handle different Transfer-Encodings.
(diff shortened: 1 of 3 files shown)


seanmonstar:
@reem fixed the draining logic. I just use `copy` to a `NullWriter`.

Switching to new io will just require changing NullWriter to Sink. The EOF logic is handled inside copy, so nothing to screw up.

reem:
Looks great.

reem:
Actually, this is a horrible idea security-wise, I think. I may start reading a request body, see it's super long, and quit. I do not expect hyper to silently read the other 40 GB of that request.

seanmonstar:
Oh dang, good catch. I'll look at nodejs for inspiration.

PR #312, opened 2015-02-14
feat(server): Rewrite the accept loop into a custom thread pool.

reem:
This is a modified and specialized thread pool meant for
managing an acceptor in a multi-threaded way. A single handler
is provided which will be invoked on each stream.

Unlike the old thread pool, this returns a join guard which
will block until the acceptor closes, enabling friendly behavior
for the listening guard.

This improves performance by like 15%, all the way to 105k req/sec
on my machine, which usually gets about 90k.

### DIFF
--- a/src/server/acceptor.rs
+++ b/src/server/acceptor.rs
@@ -1,0 +1,95 @@
+use std::thread::{Thread, JoinGuard};
+use std::sync::Arc;
+use std::sync::mpsc;
+use net::NetworkAcceptor;
+
+pub struct AcceptorPool<A: NetworkAcceptor> {
+    acceptor: A
+}
+
+impl<A: NetworkAcceptor> AcceptorPool<A> {
+    /// Create a thread pool to manage the acceptor.
+    pub fn new(acceptor: A) -> AcceptorPool<A> {
+        AcceptorPool { acceptor: acceptor }
+    }
+
+    /// Runs the acceptor pool. Blocks until the acceptors are closed.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if threads == 0.
+    pub fn accept<F: Fn(A::Stream) + Send + Sync>(self,
+                                                  work: F,
+                                                  threads: usize) -> JoinGuard<'static, ()> {
+        assert!(threads != 0, "Can't accept on 0 threads.");
+
+        // Replace with &F when Send changes land.
+        let work = Arc::new(work);
+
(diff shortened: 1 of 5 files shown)


GitCop:

 * Commit: 9b9e14e4ddfe0e07eaa38555a8da017f9ad0a875

---

GitCop:

 * Commit: fc07f155b28417551a5fa212ce54c4fec202ce93

---

src/server/acceptor.rs
```diff
@@ -0,0 +1,95 @@
...
+                Err(e) => {
+                    error!("Connection failed: {}", e);
+                }
+            }
+        }
+    });
+}
+
+struct Sentinel<T: Send> {
+    value: Option<T>,
```
seanmonstar: Why Option?

reem: Look at the Drop impl.

seanmonstar: And for that matter, why the value at all? It doesn't seem to be used...

seanmonstar:
15% sounds excellent.is that due to a difference here vs TaskPool, or because of moving the acceptor into multiple threads?

src/server/acceptor.rs
reem: It's a bit more general, and I think I might use it in the future for some other features, like a timeout or something.

reem:
I'm not sure, I think it's a combination of the two.

src/server/mod.rs
```diff
@@ -146,22 +91,56 @@ S: NetworkStream + Clone + Send> Server<L> {
...
     acceptor: A,
-    guard: Option<JoinGuard<'static, ()>>,
+    _guard: JoinGuard<'static, ()>,
     /// The socket addresses that the server is bound to.
     pub socket: SocketAddr,
 }
 
 impl<A: NetworkAcceptor> Listening<A> {
-    /// Causes the current thread to wait for this listening to complete.
-    pub fn await(&mut self) {
```
seanmonstar: Because of this, can you add the breaking change text to the commit?

seanmonstar:
I see, another nice improvement is only using the channel for panics. I'm sure that's a nice boost as well.

seanmonstar:
r+ with BREAKING CHANGE added

reem:
amended with BREAKING_CHANGE warning about await. It technically doesn't need to go away since we do still have a guard, but it's actually useless in most cases (doesn't do anything drop wouldn't, can't call close while it's blocked anyway).

PR #318, opened 2015-02-14
Revert "fix(server): Drain requests on drop."

reem:
Reverts hyperium/hyper#310

This is not a good idea, since it means that the request will be read in its entirety no matter what, which is very much not a desirable property for many servers.


GitCop:

 * Commit: 3b0216865af7bf0e1bf5a35b825ca4e86a4b2f2d

---

mneumann:
@reem: How about rather closing the connection on drop unless the read was fully exhausted? That should be safe.

PR #324, opened 2015-02-19
Update hyper to latest nightly

renato-zannon:
This makes hyper compile again under the recent `Send` changes, which just landed on nightly. I also addressed a few new warnings.

I tried to address the comment on [server/acceptor.rs:26](https://github.com/hyperium/hyper/blob/3528fb9b015a0959268452d5b42d5544c7b98a6a/src/server/acceptor.rs#L26), but was unable to do so. It seems that I need to improve my lifetime-fu :)

### DIFF
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -22,7 +22,7 @@ impl<S: Scheme> DerefMut for Authorization<S> {
     }
 }
 
-impl<S: Scheme> Header for Authorization<S> {
+impl<S: Scheme + 'static> Header for Authorization<S> {
     fn header_name() -> &'static str {
         "Authorization"
     }
@@ -43,7 +43,7 @@ impl<S: Scheme> Header for Authorization<S> {
     }
 }
 
-impl<S: Scheme> HeaderFormat for Authorization<S> {
+impl<S: Scheme + 'static> HeaderFormat for Authorization<S> {
     fn fmt_header(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         match Scheme::scheme(None::<S>) {
             Some(scheme) => try!(write!(fmt, "{} ", scheme)),
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -29,7 +29,7 @@ impl<S: Scheme + 'static> Header for Authorization<S> {
 
     fn parse_header(raw: &[Vec<u8>]) -> Option<Authorization<S>> {
         if raw.len() == 1 {
-            match (from_utf8(unsafe { &raw[].get_unchecked(0)[] }), Scheme::scheme(None::<S>)) {
+            match (from_utf8(unsafe { &raw.get_unchecked(0)[..] }), Scheme::scheme(None::<S>)) {
                 (Ok(header), Some(scheme))
                     if header.starts_with(scheme) && header.len() > scheme.len() + 1 => {
                     header[scheme.len() + 1..].parse::<S>().map(|s| Authorization(s)).ok()
@@ -96,7 +96,7 @@ impl Scheme for Basic {
         let mut text = self.username.clone();
         text.push(':');
         if let Some(ref pass) = self.password {
-            text.push_str(&pass[]);
+            text.push_str(&pass[..]);
         }
         write!(f, "{}", text.as_bytes().to_base64(Config {
             char_set: Standard,
@@ -113,7 +113,7 @@ impl FromStr for Basic {
         match s.from_base64() {
             Ok(decoded) => match String::from_utf8(decoded) {
                 Ok(text) => {
-                    let mut parts = &mut text[].split(':');
+                    let mut parts = &mut text.split(':');
                     let user = match parts.next() {
                         Some(part) => part.to_string(),
                         None => return Err(())
@@ -160,7 +160,7 @@ mod tests {
     #[test]
     fn test_raw_auth_parse() {
         let headers = Headers::from_raw(&mut mem("Authorization: foo bar baz\r\n\r\n")).unwrap();
-        assert_eq!(&headers.get::<Authorization<String>>().unwrap().0[], "foo bar baz");
+        assert_eq!(&headers.get::<Authorization<String>>().unwrap().0[..], "foo bar baz");
     }
 
     #[test]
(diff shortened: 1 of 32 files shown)


renato-zannon:
Seems that travis' rust is outdated. I tested on `rustc 1.0.0-nightly (dfc5c0f1e 2015-02-18) (built 2015-02-19)`

Byron:
When using the compiler version mentioned by @renato-zannon, the build, tests and benchmarks work for me too.

Byron:
Oh, actually, the benchmarks seem to hang, like a deadlock of some sort. I am on OSX 10.10.2, in case it matters. Also it's the first time I am running hyper benchmarks, and thus can't tell if this hang is new.
I will keep going, in the hopes that this is just a bug in the benchmarks.

reem:
@renato-zannon I may have been premature with that comment. Unless we want `accept` to block and for us to run it in another thread explicitly, we need to use `Arc`.

src/server/acceptor.rs
```diff
@@ -18,9 +18,8 @@ impl<A: NetworkAcceptor> AcceptorPool<A> {
     /// ## Panics
     ///
     /// Panics if threads == 0.
-    pub fn accept<F: Fn(A::Stream) + Send + Sync>(self,
-                                                  work: F,
-                                                  threads: usize) -> JoinGuard<'static, ()> {
+    pub fn accept<F>(self, work: F, threads: usize) -> JoinGuard<'static, ()>
+        where F: Fn(A::Stream) + Send + Sync + 'static {
```
reem: Additional advantage of making this blocking: `'a` here instead of `'static`, so this can be used with non-'static data.

src/server/acceptor.rs
```diff
@@ -35,16 +34,16 @@ impl<A: NetworkAcceptor> AcceptorPool<A> {
         for _ in 0..threads { spawn() }
 
         // Spawn the supervisor
-        Thread::scoped(move || for () in supervisor_rx.iter() { spawn() })
+        thread::scoped(move || for () in supervisor_rx.iter() { spawn() })
```
reem: To make this blocking we'd just remove the `thread::scoped` call here and just run this loop.

renato-zannon:
@reem I tried to address your comments on the last commit. I refrained from changing the non-blocking behavior of `Server::listen` - even though that would be necessary to make it possible to close over non-``static` data - because that would be a breaking change. It should be relatively straightforward to do though, and I would be happy to do it.

reem:
Looks like good work, but there is still another error.

seanmonstar:
An error about IntoCow, it looks like? And all those warnings about the
changed slice syntax will be errors in 'cargo test'...

renato-zannon:
Those are from a recent breaking change. I'll download the new nightly and fix that!

renato-zannon:
@reem I've updated the PR again even though it is still WIP, because I've got an error that I wasn't able to figure out by myself - The only compile error left is from lines 252 to 256 on `src/lib.rs`:

```
#[allow(unconditional_recursion)]
fn _assert_send<T: Send>() {
    _assert_send::<client::Request<net::Fresh>>();
    _assert_send::<client::Response>();
}
```

Error:

```
src/lib.rs:254:5: 254:48 error: the trait `core::marker::Send` is not implemented for the type `<str as collections::borrow::ToOwned>::Owned` [E0277]
src/lib.rs:254     _assert_send::<client::Request<net::Fresh>>();
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/lib.rs:254:5: 254:48 note: `<str as collections::borrow::ToOwned>::Owned` cannot be sent between threads safely
src/lib.rs:254     _assert_send::<client::Request<net::Fresh>>();
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

If I comment these lines out, hyper compiles, and all tests pass.

BTW: turns out here were multiple breaking changes on the nightly. Let me know if I should squash all these `fix(rustup)` commits.

reem:
After we're done reviewing and are ready to merge you'll squash.

renato-zannon:
Boiled this down to `Cow<'static, str>`. This doesn't compile:
```rust
#[allow(unconditional_recursion)]
fn _assert_send<T: Send>() {
    use std::borrow::Cow;
    _assert_send::<Cow<'static, str>>();
}
```

Fails with:
```
src/lib.rs:256:5: 256:38 error: the trait `core::marker::Send` is not implemented for the type `<str as collections::borrow::ToOwned>::Owned` [E0277]
src/lib.rs:256     _assert_send::<Cow<'static, str>>();
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/lib.rs:256:5: 256:38 note: `<str as collections::borrow::ToOwned>::Owned` cannot be sent between threads safely
src/lib.rs:256     _assert_send::<Cow<'static, str>>();
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Could not compile `hyper`.
```

Even though this compiles fine:
```rust
#[allow(unconditional_recursion)]
fn _assert_send<T: Send>() {
    use std::borrow::ToOwned;
    _assert_send::<<str as ToOwned>::Owned>();
}
```

Which seems contradictory to me.

softprops:
I updated my cargo config to point to

```toml
[dependencies.hyper]
git = "https://github.com/renato-zannon/hyper.git"
rev = "124a31fb9b7363960bebbdcf71df9e1e1cb0e075"
```

My compiler errors/warnings went down to the following afterwards

```rust
error: the trait `core::marker::Send` is not implemented for the type `<str as collections::borrow::ToOwned>::Owned` [E0277]
 _assert_send::<client::Request<net::Fresh>>();
note: `<str as collections::borrow::ToOwned>::Owned` cannot be sent between threads safely
 _assert_send::<client::Request<net::Fresh>>();
```

src/header/common/authorization.rs
```diff
@@ -22,14 +22,14 @@ impl<S: Scheme> DerefMut for Authorization<S> {
     }
 }
 
-impl<S: Scheme> Header for Authorization<S> {
+impl<S: Scheme + 'static> Header for Authorization<S> where <S as FromStr>::Err: 'static {
```
seanmonstar: Huh, seems odd to need this bound, but I see why. Not something I would immediately think of...

seanmonstar:
@renato-zannon This is awesome, thanks! I realize this can be tedious work (updating for syntax changes).

I've fixed the last remaining thing, and will merge the new PR that includes your commits and mine. I feel like the commits are sufficiently separated that they can be merged as-is, without squashing.

seanmonstar:
Merged as #329

src/header/common/authorization.rs
renato-zannon: Yes, I found this odd too. Took me a while to figure out where to put the bound

renato-zannon:
@seanmonstar Cool :) Even though they can be a bit tedious, these changes are a great way to get early exposure to the codebase. Maybe in the future I'll be able to submit more interesting PRs :)

softprops:
Thanks for patching this up quickly. You guys are awesome

PR #325, opened 2015-02-19
fix(headers): use $crate when referring to hyper modules on macros

### LINKED ISSUE #323: Unresolved import when trying to use impl_header!
I am trying to implement a custom HTTP header using impl_header macro from hyper crate but it seems it cannot resolve the hyper::header module.

Here is my code:

    #[macro_use] extern crate hyper;

    use hyper::header;

    struct CustomHeader(String);
    impl_header!(CustomHeader, "name", String);

And here is the compiler error:

 	<hyper macros>:11:14: 11:20 error: unresolved import `header::HeaderFormat`. Maybe a missing `extern crate header`?
  	<hyper macros>:11 Result { use header:: HeaderFormat ; self . fmt_header ( f ) } } }
                               ^~~~~~
	<hyper macros>:1:1: 11:67 note: in expansion of impl_header!
	lib.rs:4:1: 4:45 note: expansion site
	Could not compile `macro_issue`.

renato-zannon:
This adds a `$crate` variable missed by de1be6726

Note that even though this PR targets master, it needs #324 (or some equivalent PR) to build on the current rust nightly. Let me know if I should rebase.

Closes #323

### DIFF
--- a/src/header/common/mod.rs
+++ b/src/header/common/mod.rs
@@ -108,7 +108,7 @@ macro_rules! impl_list_header(
 
         impl ::std::fmt::Display for $from {
             fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-                use header::HeaderFormat;
+                use $crate::header::HeaderFormat;
                 self.fmt_header(f)
             }
         }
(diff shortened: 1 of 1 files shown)


seanmonstar:
Excellent!

Issue #330, opened 2015-02-21
Change QualityItem to use a newtype Quality(u16) instead of f32


seanmonstar:
f32 gives more precission than we need, since the only valid values are `0.000 - 1.000`. Also, by their very implementation, compare floats can be tricky.

Instead, we should have a `pub struct Quality(u16)`. The value inside should be between `0 - 1000`. With a simple integer, comparison can be exact, and we also shave off half the memory size.

Formatting can be easily done, however a convenience method would be required, such as `q(value: f32) -> Quality`, that converts an `f32` into the appropriate `u16`. This allows users to still write using float syntax, without the downsides (example: `QualityItem { item: "en-US", quality: q(0.9) }`).

reem:
This is such a good idea. We should have a convention for no floats in headers.

pyfisch:
+1 I did not think this far. Using integers  is much better here.

pyfisch:
I am working on this.

pyfisch:
Close this.

PR #332, opened 2015-02-22
refactor(headers): Use u16 based newtype for quality value

pyfisch:
Using floating point numbers is problematic because comparison is inexact.
They also take more space than integral numbers in this case.

Closes: #330

BREAKING_CHANGE: Replace f32 quality values in quality items with a
Quality(u16) newtype. Valid values are from 0 to 1000.

### DIFF
--- a/src/header/common/accept.rs
+++ b/src/header/common/accept.rs
@@ -33,7 +33,8 @@ impl_list_header!(Accept,
 mod tests {
     use mime::*;
 
-    use header::{Header, QualityItem, qitem};
+    use header::{Header, Quality, QualityItem, qitem};
+    
     use super::Accept;
 
     #[test]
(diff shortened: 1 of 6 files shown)


seanmonstar:
As mentioned in #330, we should provide a method that accepts a float and creates a Quality. I'd even say that the implementation should be hidden from the user. There's no need to confuse someone that we use an integer to represent a rational.

pyfisch:
Added methods for float.

@seanmonstar: I think we should provide a method for the user to create `Quality` directly with an `u16` since converting to float costs a little time.

seanmonstar:
That's fair, but it should be the "for experts or performance needs", not
the way we show everyone in the docs. It could be confusing that Quality(1)
is not q=1, but instead q=0.001.

pyfisch:
Updated the docs to give advice how to use `Quality`.

PR #348, opened 2015-02-28
fix(cargo): set openssl to compatible v0.4.3

Byron:
In openssl >= 0.5.0, it's using `std::io`, which doesn't work with
std::old_io used in hyper.
To make it work consistently, `Cookie` was reset to the previous
version v0.1.13 as well, which depends on our compatible openssl
version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,10 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
 keywords = ["http", "hyper", "hyperium"]
 
 [dependencies]
-cookie = "*"
+cookie = "= 0.1.13"
 log = ">= 0.2.0"
 mime = "*"
-openssl = "*"
+openssl = "0.4.3"
 rustc-serialize = "*"
 time = "*"
 unicase = "*"


Byron:
Oh, and it would be great if you could make a release right away, as I am still suffering from [this bug](https://github.com/rust-lang/cargo/issues/1215).
Thank you :) - using alphas is fun !

reem:
@Byron Thanks for the PR, but I don't think we'll be taking this approach right now, but will instead merge a fix to migrate to `std::io` tomorrow, rather than backtrack to old dependencies based on a deprecated library.

You can always depend on a git fork of hyper with those dependencies changed if you need something running right this second.

Byron:
That's what I am doing, and thanks to buggy cargo it recompiles hyper *whenever* I run `cargo test`, and I run it a lot.

Would it hurt to have a working version in the meanwhile ? Travis likes it too, so will everyone just pulling hyper from crates.io today.

A working version is a merge and a publish away, which helps everyone today. Tomorrow is another day, which should upgrade to std::io to track latest.
Hope I make sense ... .

reem:
Ok, you've convinced me. I'll reopen and let @seanmonstar weigh in.

Byron:
Thank you :) ! My change shouldn't even conflict with anything you cooked up in the meanwhile. If you merge, you'd just have to remember to undo my changes to track the latest openssl. Seems doable, let's see what @seanmonstar thinks about that.

seanmonstar:
You can quite easily edit the Cargo.toml to set the openssl/cookie
versions. This is how Servo handles things before upgrading rust versions.

Byron:
Thanks for that idea, this actually works ! It's important to put these in the right order though:
```toml
openssl = "= 0.4.3"
cookie = "= 0.1.13"
# If hyper would be listed before openssl, there would be a version conflict. Makes sense.
hyper = "*"
```
Compiling feels so much better now. Besides, I learned something new about cargo today, which shall help me to remain independent and workaround all of the current issues.

seanmonstar:
Sorry, I meant Cargo.lock

Byron:
That I tried in the very beginning - cargo would keep overwriting the file, it seems it's not really supposed to be edited.
Cargo.toml works fine though.

seanmonstar:
You can edit the lock file, as I said that's what Servo does. You however
cannot use 'cargo update' with it.

PR #370, opened 2015-03-13
perf(http): changes http parsing to use httparse crate

seanmonstar:
httparse is a http1 stateless push parser. This not only speeds up
parsing right now with sync io, but will also be useful for when we get
async io, since it's push based instead of pull.

BREAKING CHANGE: Several public functions and types in the `http` module
  have been removed. They have been replaced with 2 methods that handle
  all of the http1 parsing.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -1,11 +1,15 @@
 #![deny(warnings)]
 extern crate hyper;
 
+extern crate env_logger;
+
 use std::env;
 
 use hyper::Client;
 
 fn main() {
+    env_logger::init().unwrap();
+
     let url = match env::args().nth(1) {
         Some(url) => url,
         None => {
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -1,6 +1,7 @@
 #![deny(warnings)]
 #![feature(io, net)]
 extern crate hyper;
+extern crate env_logger;
 
 use std::io::Write;
 use std::net::IpAddr;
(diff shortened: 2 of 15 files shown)


examples/client.rs
```diff
@@ -1,11 +1,15 @@
 #![deny(warnings)]
 extern crate hyper;
 
+extern crate env_logger;
```
reem: Can you do this for the other examples too?

examples/hello.rs
```diff
@@ -16,6 +16,6 @@ fn hello(_: Request, res: Response) {
 
 fn main() {
     let _listening = hyper::Server::http(hello)
-        .listen(IpAddr::new_v4(127, 0, 0, 1), 3000).unwrap();
+        .listen_threads(IpAddr::new_v4(127, 0, 0, 1), 3000, 10).unwrap();
```
reem: Either we should change defaults or go back to the original.

seanmonstar: Oh good catch. I was altering numbers when testing performance improvement for the server.

reem:
Ah I didn't actually finish reviewing this... might need to make a followup PR with some more fixes.

examples/server.rs
```diff
@@ -15,7 +15,7 @@ macro_rules! try_return(
     ($e:expr) => {{
         match $e {
             Ok(v) => v,
-            Err(e) => { error!("Error: {}", e); return; }
+            Err(e) => { println!("Error: {}", e); return; }
```
reem: stray `println!`, probably should go back to `error!`

seanmonstar: I did that on purpose, since it seemed better for a simple example to actually print errors.

PR #372, opened 2015-03-14
refactor(hyper): Split hyper into multiple crates

pyfisch:
Move various parts of hyper into separate crates:
* hypernet
* hyperprotocol

There should be no breaking changes for hyper users.


Cargo.toml
```diff
@@ -12,16 +12,16 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
...
 log = ">= 0.2.0"
-mime = "*"
-openssl = "*"
-rustc-serialize = "*"
 time = "*"
-unicase = "*"
 url = "*"
 
+[dependencies.hypernet]
+path = "hypernet"
```
reem: I'm fairly confident this will prevent uploading to crates.io. We will have to upload and depend on hypernet and hyperprotocol as normal dependencies.

pyfisch: This is probably true.

If we split hyper consequently there will be no "hyper" crate only "hyperclient" and "hyperserver". I have no idea what the current crate will be useful for.

pyfisch:
I will move server and client later out of main hyper crate. They both depend on src/http.rs and src/uri.rs. I do not know where this code should be placed. I think it would be the best thing if we could get rid of `RequestUri` and use always rust-url absolute URLs.

This depends on https://github.com/rust-lang/rust/issues/23110 being fixed. Or is there a workaround for now?

reem:
@pyfisch I actually think it would be better if we just built our own custom url type that's specialized for HTTP use cases. The absolute rust-url Url type is significantly less useful in many cases then a relative URL with a virtual root.

pyfisch:
Rust broke hyper again. The whole downcast stuff is broken.

@reem Yes we should use relative URLs.

pyfisch:
Everything works now.

Path dependencies are rewritten to normal dependencies in cargo: http://users.rust-lang.org/t/multi-crate-cargo-repositories-best-practices/692/2?u=pyfisch So it stays simple to both distribute and develop hyper.

seanmonstar:
What do you mean they are rewritten as normal dependencies?

reem:
@seanmonstar a crates.io build will use `version` and a local build will use `path`.

reem:
Closing because stale, feel free to try again.

Issue #374, opened 2015-03-16
arithmetic overflow / panic when fmt_header called on empty Cookie header


fuchsnj:
```
thread '<unnamed>' panicked at 'arithmetic operation overflowed', .... /.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.3.0/src/header/common/cookie.rs:53
```

reem:
Fixed in #375

PR #375, opened 2015-03-16
fix(headers): Fix overflow with empty cookies

fuchsnj:
fixes: https://github.com/hyperium/hyper/issues/374


seanmonstar:
Thanks!

PR #380, opened 2015-03-17
feat(client): add a default UserAgent

### LINKED ISSUE #373: Hyper doesn't automatically send the User-Agent header
A variety of APIs, including github's, mark this header as mandatory. We should set a default value of `hyper` if someone tries to send a request without a `User-Agent` header.

seanmonstar:
The user agent of a client can be changed via the `set_user_agent`
method. If not changed, the default will be `hyper/$VERSION`.

Closes #373

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -25,11 +25,11 @@ use url::UrlParser;
 use url::ParseError as UrlError;
 
 use header::{Headers, Header, HeaderFormat};
-use header::{ContentLength, Location};
+use header::{ContentLength, Location, UserAgent};
 use method::Method;
 use net::{NetworkConnector, HttpConnector, ContextVerifier};
 use status::StatusClass::Redirection;
-use {Url, HttpResult};
+use {VERSION, Url, HttpResult};
 use HttpError::HttpUriError;
 
 pub use self::request::Request;
@@ -44,6 +44,7 @@ pub mod response;
 pub struct Client<C> {
     connector: C,
     redirect_policy: RedirectPolicy,
+    user_agent: Option<String>
 }
 
 impl<'v> Client<HttpConnector<'v>> {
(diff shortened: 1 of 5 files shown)


src/net.rs
```diff
@@ -98,17 +98,20 @@ impl Clone for Box<NetworkStream + Send> {
 }
 
 impl NetworkStream {
```
reem: I don't think these methods should be public. We should provide a safe, checked API.

seanmonstar: It was actually because I couldn't use them in the tests in the `client/mod.rs` file.

There should already be a public API, using `std::any::Any`, right?

reem: Yes, of course. Why do you need to use them in tests?

seanmonstar: Oh right, I can just use the checked versions in the tests. FIxing.

seanmonstar:
@reem done

src/client/mod.rs
```diff
@@ -75,6 +77,11 @@ impl<C: NetworkConnector> Client<C> {
         self.redirect_policy = policy;
     }
 
+    /// Set the UserAgent. Pass `None` to disable.
+    pub fn set_user_agent(&mut self, agent: Option<String>) {
```
reem: What if instead of a specific method for this header, we just have a set of headers we clone into every request, so users can set any default headers really easily and the API is really simple.

seanmonstar: Hmm, I started looking at other languages, and their built-in Clients don't offer a default user-agent either. The recommended way of setting one is just to set the header. A possible solution to this is `client.get(url).header(UserAgent(someStr)).send()`...

reem: Yes, but then you have to add it for every request. If we provide a way to set default headers for every request, then you can cut all that boilerplate.

seanmonstar: Perhaps. Still, that's a different approach than this PR. I'm going to close this.

PR #382, opened 2015-03-19
Adapt to current rustc (closes #381)

adrianheine:
closes #381.

PR #383, opened 2015-03-20
fix(rustup): rustc 1.0.0-nightly (ea8b82e90)

Byron:
This PR fixes `cargo build` and `cargo test`, and includes the PR of @adrianheine .

Method lookup on traits seems to have changed to force
`impl TraitName` expressions to be more specific. That means that
`method` will not be found anymore on an object of type `&Trait+Send`,
unless you provide an `impl Trait+Send`.

Now `NetworkStream` and `HeaderFormat` trait implementations
are done against `* + Send`, which helps the compiler to find the
respective `downcast*` method implementations once again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Hello World Server:
 extern crate hyper;
 
 use std::io::Write;
-use std::net::IpAddr;
+use std::net::Ipv4Addr;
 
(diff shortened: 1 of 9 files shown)


Byron:
This PR includes #382 .

PR #384, opened 2015-03-21
fix(rustup): 1.0.0-nightly (e2fa53e59 2015-03-20)

Byron:
* replace `char_at()` calls with itertor, as suggested by compiler
* fixed comparison in test

# Important Note

**Benchmarks fail if they are run** - this is nothing travis could figure out due to the `--no-run` flag. The error is:
```bash
test header::common::date::asctime::bench_format                    ... thread '<main>' panicked at 'called `Option::unwrap()` on a `None` value', /Users/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-mac/build/src/libcore/option.rs:362
```

I don't know when this regression sneaked in, it doesn't seem to be related to this commit though.

### DIFF
--- a/src/header/common/host.rs
+++ b/src/header/common/host.rs
@@ -28,7 +28,9 @@ impl Header for Host {
             // https://github.com/servo/rust-url/issues/42
             let idx = {
                 let slice = &s[..];
-                if slice.char_at(1) == '[' {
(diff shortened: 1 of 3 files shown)


pyfisch:
They did fail before.

PR #386, opened 2015-03-21
chore(build): Enable building hyper in docker containers on travis

pyfisch:
By stating that hyper does not need sudo to build it can be compiled in
a container:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+sudo: false
 
 script:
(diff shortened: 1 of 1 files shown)


reem:
Good catch! This makes builds much faster.

seanmonstar:
Oh awesome!!

PR #390, opened 2015-03-23
Add tests for header::parsing::tm_from_str

seanmonstar:


reem:
This increases our travis time to over 5 minutes, but other than that I guess it only helps, so +1.

PR #397, opened 2015-03-25
chore(all): Remove the box_syntax feature gate.

reem:


src/client/request.rs
```diff
@@ -56,7 +56,7 @@ impl Request<Fresh> {
         let (host, port) = try!(get_host_and_port(&url));
 
         let stream = try!(connector.connect(&*host, port, &*url.scheme));
-        let stream = ThroughWriter(BufWriter::new(box stream as Box<NetworkStream + Send>));
+        let stream = ThroughWriter(BufWriter::new(Box::new(stream) as Box<NetworkStream + Send>));
```
seanmonstar: This works now? I kept the `box` syntax cause there didn't used to be a way to coerce to a Boxed TraitObject without it...

seanmonstar:
I can't merge this? Did you start from an older master?

src/client/request.rs
reem: IIRC it's worked ever since `Box::new` existed, but the silent coercions with `Box::new` aren't as ambitious as with `box`.

reem:
Possibly, let me rebase.

seanmonstar:
I've included this in #403

PR #398, opened 2015-03-25
Implement Content-Encoding Header

mattnenterprise:


src/header/common/content_encoding.rs
```diff
@@ -0,0 +1,21 @@
...
+/// The `Content-Encoding` header.
+///
+/// This header describes the encoding of the message body. It can be
+/// comma-separated, including multiple encodings.
+///
+/// ```notrust
+/// Content-Encoding: gzip
+/// ```
+///
+/// The implementation uses a vector of `Encoding` values.
```
reem: since the field is pub this is apparent, so no need to mention specifically in the docs

reem:
Everything but that nit LGTM.

seanmonstar:
Excellent, thanks!

PR #400, opened 2015-03-26
fix(rustup): static bounds required on Type definition, trivial_casts

Ryman:
Removing the remaining casts trigger inference failures... so they're... not trivial?


Ryman:
Updated to not `allow` the lint, mainly due to travis having multiple rustc versions active and that it should be easy to fix these when type ascription lands due to fixme annotations.

PR #401, opened 2015-03-27
fix(http): read more before triggering TooLargeError

### LINKED ISSUE #389: Hyper client fails if the header isn't fully received in the first read
Some HTTP servers do not send the entire HTTP header in a single chunk. Hyper assumes that the entire header is received when calling fill_buf, which isn't always the case.

Running the client example in the documentation against a [Flask test server](http://flask.pocoo.org/) causes it to panic:

thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: HttpTooLargeError'

seanmonstar:
This includes a custom BufReader, since the one in libstd doesn't allow
reading additional data into the buffer without consuming it. This is
required because some connections may send shorter packets, and so we
need to perform multiple reads. After each read, the contents of the
buffer are passed to httparse to see if have a valid message. If so, the
proper amount of bytes are consumed. The additional bytes are left in
the buffer since they are the beginning of the body.

The buffer in this BufReader also grows in size, compared to the libstd
which is sized once. This is because we start with a smaller buffer,
since the majority of messages will be able to include their head in a
packet or 2. Therefore, it's a wasteful performance hit to allocate the
maximum size for every message. However, some headers can be quite big,
and to allow for many of them to be set, we include a maximum size. Once
we've hit the maximum buffer size, and still haven't determined the end
of the headers, a HttpTooLargeError will be returned.

Closes #389

cc @reem @r-darwish

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,0 +1,95 @@
+use std::cmp;
+use std::iter;
+use std::io::{self, Read, BufRead, Cursor};
+
+pub struct BufReader<R> {
+    buf: Cursor<Vec<u8>>,
+    inner: R
+}
+
+const INIT_BUFFER_SIZE: usize = 4096;
+const MAX_BUFFER_SIZE: usize = 8192 + 4096 * 100;
+
+impl<R: Read> BufReader<R> {
+    pub fn new(rdr: R) -> BufReader<R> {
+        BufReader::with_capacity(rdr, INIT_BUFFER_SIZE)
+    }
+
+    pub fn with_capacity(rdr: R, cap: usize) -> BufReader<R> {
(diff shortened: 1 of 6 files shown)


src/buffer.rs
```diff
@@ -0,0 +1,95 @@
+use std::cmp;
+use std::iter;
+use std::io::{self, Read, BufRead, Cursor};
+
+pub struct BufReader<R> {
+    buf: Cursor<Vec<u8>>,
+    inner: R
+}
+
+const INIT_BUFFER_SIZE: usize = 4096;
```
reem: Turning this down should be a nice perf win for short connections.

PR #402, opened 2015-03-27
chore(travis): Enable caching of ./target folder.

pyfisch:
Travis supports the caching of folders between builds.
By storing the ./target folder we can cache hyper's dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: rust
 sudo: false
 
+cache:
+  directories:
+      - target
+
 script:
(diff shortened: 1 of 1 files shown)


seanmonstar:
What do you mean cache it? Travis wipes it out at some point?

pyfisch:
Travis CI calls it "caching". The [documentation](http://docs.travis-ci.com/user/caching/#Arbitrary-directories) does not say anything about automatic deletion, but you can delete the cache manually.

PR #404, opened 2015-03-29
fix: rustup

fhartwig:
* enable slice_patterns feature
* add `Reflect` trait bounds where necessary

### DIFF
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::str::{FromStr, from_utf8};
 use std::ops::{Deref, DerefMut};
+use std::marker::Reflect;
 use serialize::base64::{ToBase64, FromBase64, Standard, Config, Newline};
 use header::{Header, HeaderFormat};
 
@@ -22,7 +23,7 @@ impl<S: Scheme> DerefMut for Authorization<S> {
     }
 }
 
-impl<S: Scheme + 'static> Header for Authorization<S> where <S as FromStr>::Err: 'static {
+impl<S: Scheme + Reflect + 'static> Header for Authorization<S> where <S as FromStr>::Err: 'static {
     fn header_name() -> &'static str {
         "Authorization"
     }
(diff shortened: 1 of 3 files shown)


GitCop:

 * Commit: 6a803990afe1c58f22d0e3ed326d051ad07e8bbc

---

fhartwig:
Amended commit message to make gitcop happy.

seanmonstar:
Thanks! I wonder, if slice_patterns aren't going to be stable for beta/1.0,
we should probably not use them. I think they're only used in the
authorization header, and shouldn't be too hard to switch it to a few if
clauses.

fhartwig:
https://github.com/rust-lang/rust/issues/23121 says that they hope to have it ready for the beta release, but I'm not sure if that still applies. I can re-write the function to avoid slice patterns if you like.

seanmonstar:
With beta this week, those comments don't assure me. If you could remove
the patterns, awesome. Or I can a little later.

fhartwig:
Ok, slice patterns are gone. Tests pass for me, travis seems to have trouble with github today.

src/header/common/authorization.rs
```diff
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::str::{FromStr, from_utf8};
 use std::ops::{Deref, DerefMut};
+use std::marker::Reflect;
```
reem: Reflect shouldn't be needed in this file as far as I can tell. What error comes up if you remove these bounds?

reem: I see, since our Header traits should require Reflect.

reem:
The Header downcasting stuff all needs `Reflect` bounds similar to `Any` and `NetworkStream` to be safe.

Ryman:
According to [this](http://doc.rust-lang.org/std/marker/trait.Reflect.html) the bound should be `Any` rather than `Reflect + 'static` as `Any` implies that?

src/header/common/authorization.rs
fhartwig:
```
error: the trait `core::marker::Reflect` is not implemented for the type `S` [E0277]
```
(in lines 26 and 47)

reem:
@Ryman you're right. The existing `Any` bound should be sufficient.

PR #409, opened 2015-03-30
feat(entitytag): Add EntityTag comparison, make EntityTag safe to use

pyfisch:
Adds strong and weak comparison to EntityTag as described in the RFC, add tests for this. Make EntityTag safe to use by hiding the tag field, this prevents users from inserting malicious values for the tag. Invalid values can screw up header formatting and may allow to insert headers. Introduce EntityTag::new(), .tag() and .set_tag() methods. Fix Display trait for EntityTag. DQUOTES were missing. Remove custom formatting in ETag header. Improve docs.

BREAKING_CHANGE: EntityTag.tag is private, use EntityTag.tag() and EntityTag.set_tag("foobar") to access.

### DIFF
--- a/src/header/shared/entity.rs
+++ b/src/header/shared/entity.rs
@@ -1,29 +1,6 @@
 use std::str::FromStr;
 use std::fmt::{self, Display};
 
-/// An entity tag
-///
-/// An Etag consists of a string enclosed by two literal double quotes.
-/// Preceding the first double quote is an optional weakness indicator,
-/// which always looks like this: W/
-/// See also: https://tools.ietf.org/html/rfc7232#section-2.3
-#[derive(Clone, PartialEq, Debug)]
-pub struct EntityTag {
-    /// Weakness indicator for the tag
-    pub weak: bool,
-    /// The opaque string in between the DQUOTEs
-    pub tag: String
-}
-
-impl Display for EntityTag {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        if self.weak {
-            try!(write!(fmt, "{}", "W/"));
-        }
-        write!(fmt, "{}", self.tag)
-    }
-}
-
 // check that each char in the slice is either:
 // 1. %x21, or
 // 2. in the range %x23 to %x7E, or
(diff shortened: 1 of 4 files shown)


GitCop:

 * Commit: 85dac24238df2ea92155a7e99f8edcc4184a92be

---

src/header/shared/entity.rs
```diff
@@ -1,41 +1,114 @@
...
+
+    /// Get the tag.
+    pub fn tag(&self) -> &str {
+        self.tag.as_ref()
+    }
+
+    /// Set the tag.
+    /// # Panics
+    /// If the tag contains invalid characters.
+    pub fn set_tag(&mut self, tagslice: &str) {
```
seanmonstar: It seems odd to me for a setter to ask for a reference. I'd expect either a `String`, or `Into<String>`.

seanmonstar:
@pyfisch This is superb! Just one minor comment.

src/header/shared/entity.rs
pyfisch: Good point.

src/header/shared/entity.rs
```diff
@@ -1,41 +1,114 @@
...
 
-impl Display for EntityTag {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        if self.weak {
-            try!(write!(fmt, "{}", "W/"));
+impl EntityTag {
+    /// Constructs a new EntityTag.
+    /// # Panics
+    /// If the tag contains invalid characters.
+    pub fn new(weak: bool, tagslice: &str) -> EntityTag {
```
pyfisch: I guess the same here?

src/header/shared/entity.rs
seanmonstar: That's OK. Switching to `Into<String>` should be a non-breaking change.

src/header/shared/entity.rs
seanmonstar: Ah, yes.

Specifically, the only way to pass a `&str` is to have a static str, or they need to create a `String`. Since they have already allocated a `String` in order pass a reference, it's wasteful to allocate a new String, especially since they likely just threw away the original they made.

pyfisch:
Updated code.

PR #410, opened 2015-03-30
perf(buffer): pull in std::io::BufReader improvements

seanmonstar:
The results of benches/client.rs are below.

Before:

    running 1 test
    test bench_mock_hyper ... bench:    321609 ns/iter (+/- 18127)

After:

    running 1 test
    test bench_mock_hyper ... bench:    239319 ns/iter (+/- 29938)


reem:
wow that's a huge gain!

PR #413, opened 2015-03-31
refactor(header): Implement `HttpDate`, a wrapper for dates

pyfisch:
Using `time::Tm` directly in HTTP header fields requires special handling to parse and format
the header values., this stops us from using the header macros. By wrapping `time::Time` in a
`HttpDate`, we can use the `FromStr` and `Display` traits of `HttpDate` like for most other values.

BREAKING_CHANGE: All code using one of the `Date`, `Expires`, `If-Modified-Since`,
`If-Unmodified-Since`, `Last-Modified` header fields needs to wrap `time::Tm`
with `HttpDate`. Removed `FromStr` trait of `Date`, `If-Modified-Sice` and `If-Unmodified-Sice`,
implementing the trait here is inconsistent with other headers.

### DIFF
--- a/src/header/common/date.rs
+++ b/src/header/common/date.rs
@@ -1,45 +1,10 @@
-use std::fmt;
-use std::str::FromStr;
-use time::Tm;
-use header::{Header, HeaderFormat};
-use header::parsing::from_one_raw_str;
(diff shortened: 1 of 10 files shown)


seanmonstar:
I love it.

PR #414, opened 2015-03-31
fix(buffer): zero out new capacity when buffer grows

seanmonstar:


reem:
Can you bench this change? I'm curious how much of a difference this actually makes.

seanmonstar:
So, something happened with the last rustup. My benchmarks are all ~10x faster. I can't explain why. The changelog in cargo and rust don't easily show a reason.

Anyway, Before/After is `35524`/`37852`.

PR #419, opened 2015-04-02
refactor(headers): Introduce header!() macro, improve documentation

pyfisch:
The new macro handles single value headers, list headers, and list
headers with at least one item.
It creates the item for the header and contains its documentation. The
new macro allows handling
more header cases in the future, it will also be possible to include
tests inside the macro.

BREAKING_CHANGE: Removed impl_header!() and impl_list_header!() macros,
use new header!() macro.

### DIFF
--- a/src/header/common/accept.rs
+++ b/src/header/common/accept.rs
@@ -2,39 +2,39 @@ use mime::Mime;
 
 use header::QualityItem;
 
-/// The `Accept` header.
-///
-/// The `Accept` header is used to tell a server which content-types the client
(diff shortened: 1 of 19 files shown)


pyfisch:
By the way, the documentation syntax will get better once rust-lang/rust#23812 gets fixed.

seanmonstar:
Sweet! And also, the change in docs: wtf. (I know it's not your fault).

seanmonstar:
Oh, also, the commit syntax is actually `BREAKING CHANGE:`, without the underscore.

pyfisch:
Thanks, fixed.

PR #420, opened 2015-04-02
refactor(headers): Use header!() macro for `ETag` header

pyfisch:
Change `Etag` (only the `E` is capitalized) to `ETag` as it is written in the relevant RFC.

BREAKING_CHANGE: `Etag` header field is now `ETag` header field

Depends on #419


pyfisch:
Rebased and build passes with openssl 0.5

pyfisch:
Uses normal ssl again.

PR #421, opened 2015-04-02
 refactor(headers): Use header!() macro for 3 headers with a "*" value

pyfisch:
`If-Match`, `If-None-Match` and `Vary` headers are either a "*" value meaning that the header
matches every possible item or a list of items, one of them must be matched to fulfil the condition.

BREAKING_CHANGE: `If-Match`, `If-None-Match` and `Vary` item variant name changed to `Items`

Depends on #419


pyfisch:
Also rebased.

PR #423, opened 2015-04-02
fix(rustup): update to rust beta

seanmonstar:


src/header/common/expect.rs
```diff
@@ -16,14 +19,29 @@ pub enum Expect {
...
         "Expect"
     }
 
     fn parse_header(raw: &[Vec<u8>]) -> Option<Expect> {
-        if &[b"100-continue"] == raw {
-            Some(Expect::Continue)
+        if raw.len() == 1 {
+            let text = unsafe {
+                // safe because:
+                // 1. we jsut checked raw.len == 1
```
reem: typo in just

src/header/common/expect.rs
```diff
@@ -16,14 +19,29 @@ pub enum Expect {
...
+                // 2. we don't actually care if its utf8, we just want to
```
reem: its should be it's here

reem:
LGTM except nits

PR #425, opened 2015-04-02
chore(stability): remove core feature gate

seanmonstar:


src/header/common/authorization.rs
```diff
@@ -44,7 +44,7 @@ impl<S: Scheme + Reflect + 'static> Header for Authorization<S> where <S as From
     }
 }
 
-impl<S: Scheme + Reflect + 'static> HeaderFormat for Authorization<S> where <S as FromStr>::Err: 'static {
+impl<S: Scheme + Any + 'static> HeaderFormat for Authorization<S> where <S as FromStr>::Err: 'static {
```
reem: `'static` is implied by `Any` so we can get rid of it too.

src/header/internals/item.rs
```diff
@@ -51,7 +54,7 @@ impl Item {
         &raw[..]
     }
 
-    pub fn typed<H: Header + HeaderFormat>(&self) -> Option<&H> {
+    pub fn typed<H: Header + HeaderFormat + Any>(&self) -> Option<&H> {
```
reem: This is totally a judgement call, but do you think we should add `Any` to the `HeaderFormat` supertraits?

src/header/mod.rs
```diff
@@ -61,12 +62,12 @@ pub trait HeaderFormat: fmt::Debug + HeaderClone + Any + Send + Sync {
 
 #[doc(hidden)]
 pub trait HeaderClone {
-    fn clone_box(&self) -> Box<HeaderFormat + Sync + Send>;
+    fn clone_box(&self) -> Box<HeaderFormat + Send + Sync>;
 }
 
-impl<T: HeaderFormat + Send + Sync + Clone> HeaderClone for T {
+impl<T: HeaderFormat + Send + Clone> HeaderClone for T {
```
reem: Send is implied by HeaderFormat

src/net.rs
```diff
@@ -135,7 +134,7 @@ impl NetworkStream + Send {
     /// If the underlying type is T, get a mutable reference to the contained
     /// data.
     #[inline]
-    pub fn downcast_mut<T: Reflect + 'static>(&mut self) -> Option<&mut T> {
+    pub fn downcast_mut<T: Any + 'static>(&mut self) -> Option<&mut T> {
```
reem: 'static implied by Any here too

src/status.rs
```diff
@@ -225,6 +222,136 @@ pub enum StatusCode {
 
 impl StatusCode {
 
+    #[doc(hidden)]
```
reem: Why hidden for this and `to_u16`? That's something people are likely to want to do.

src/version.rs
```diff
@@ -7,7 +7,7 @@ use std::fmt;
 use self::HttpVersion::{Http09, Http10, Http11, Http20};
 
 /// Represents a version of the HTTP spec.
-#[derive(PartialEq, PartialOrd, Copy, Debug)]
+#[derive(PartialEq, PartialOrd, Copy, Clone, Debug)]
```
reem: Eq, Ord, and Hash too while you're at it? It seems valid to want to put these in a hashmap for routing of some kind.

src/status.rs
seanmonstar: The only reason I can see for using `from_u16` is to handle parsing, which we already do. If you know the number yourself, just write the variant instead.

PR #427, opened 2015-04-03
fix(headers): Add CowStr as a temporary hack to build on beta.

reem:
We can revert this PR when rust-lang/rust#23995 lands, but it won't
land until after beta is cut.


seanmonstar:
There was a problem?

reem:
The Unicase changes broke master

reem:
I was on my phone earlier - to be a little more clear:

Unicase changed to use `AsRef` instead of `Deref` for its impls. The impl of `AsRef` for `Cow` is overly restrictive, so this causes breakage. The linked PR fixes this, but won't be in beta. It would be nice to build on beta.

PR #428, opened 2015-04-04
feat(client): remove generic parameter for Connector

### LINKED ISSUE #379: change Client to keep connector as a Box<NetworkConnector>
This removes the generics from Client, letting people use a `Client` easily without caring what connecting and streaming business is going on underneath.

seanmonstar:
Closes #379

BREAKING CHANGE: For people using the default HttpConnector and Client,
    everything should continue to just work. If the Client has been
    used with a generic parameter, it should be removed.

However, there were some breaking changes to the internals of
    NetworkConnectors. Specifically, they no longer return a
    NetworkStream, but instead a `Into<Box<NetworkStream + Send>>`. All
    implementations of NetworkStream should continue to just work,
    however.

Possible breakages could come from the stricter usage of Send
    throughout the Client API.

cc @Manishearth @reem

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -27,7 +27,7 @@ use url::ParseError as UrlError;
 use header::{Headers, Header, HeaderFormat};
 use header::{ContentLength, Location};
 use method::Method;
-use net::{NetworkConnector, HttpConnector, ContextVerifier};
+use net::{NetworkConnector, NetworkStream, HttpConnector, ContextVerifier};
 use status::StatusClass::Redirection;
 use {Url, HttpResult};
 use HttpError::HttpUriError;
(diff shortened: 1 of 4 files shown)


seanmonstar:
Also cc @Byron

reem:
LGTM. The `Into<Box<NetworkStream + Send>>` thing is a bit weird, but it works.

seanmonstar:
Yea, I didn't want to Box in Request if the Stream was already boxed in Client. It works out in the end.

reem:
It's a pretty nice way to avoid double boxing in some cases.

Byron:
Dito ! 
Something similar I do with `BorrowMut`, in case I want people to either allow me to own something, or to just borrow it mutably, without them having to care. More complexity on my side translates to more flexibility and ease of use on the client side, which is a hit I am willing to take as many benefit from it.

PR #430, opened 2015-04-04
fix(README): Update to compile example against Rust beta

mfeckie:
Remove unused import
res doesn't need to be mutable in parameter list
Change .listen() to string format


seanmonstar:
Thanks!

PR #431, opened 2015-04-06
Revert "fix(headers): Add CowStr as a temporary hack to build on beta."

reem:
Reverts hyperium/hyper#427

Now that the impl in question has been fixed, this should not be necessary.

### DIFF
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -9,7 +9,6 @@ use std::borrow::{Cow, ToOwned};
 use std::collections::HashMap;
 use std::collections::hash_map::{Iter, Entry};
 use std::iter::{FromIterator, IntoIterator};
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 002727ede4f578e8f3ca18a12eb58316fc90043d

---

seanmonstar:
When will this fix be included into beta?

reem:
@seanmonstar For some reason I thought beta was rebuilt every day, but apparently not...

seanmonstar:
This failure is actually cause cookies is hardfixed to an older openssl version...

I still want to make a cookies lib that doesn't depend on crypto (that should be a lib added on top).

reem:
Looks like we'll actually need to keep this to build on beta for a bit, so closing.

PR #433, opened 2015-04-06
refactor(headers): Implement CookiePair inside of hyper, drop dependency...

pyfisch:
... on cookie-rs

Remove the dependency on alexcrichton's cookie-rs because it depends on openssl and does to much
work with the cookies that should be placed in a higher level lib and not in hyper. Remove all
methods that require a CookieJar. Copied and adapted the Cookie type from cookie-rs to hyper.

Originally proposed by @seanmonstar.

Set openssl version to 0.5 to make this compiling.

BREAKING CHANGE: All methods using CookieJar removed from headers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
 keywords = ["http", "hyper", "hyperium"]
 
 [dependencies]
-cookie = "*"
 httparse = "*"
 log = ">= 0.2.0"
 mime = "*"
@@ -27,4 +26,3 @@ typeable = "*"
 
 [dev-dependencies]
 env_logger = "*"
-
--- a/src/header/common/cookie.rs
+++ b/src/header/common/cookie.rs
@@ -2,8 +2,7 @@ use header::{Header, HeaderFormat};
 use std::fmt;
 use std::str::from_utf8;
 
-use cookie::Cookie as CookiePair;
-use cookie::CookieJar;
+use header::CookiePair;
 
 /// The `Cookie` header. Defined in [RFC6265](tools.ietf.org/html/rfc6265#section-5.4):
 ///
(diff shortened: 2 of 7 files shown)


seanmonstar:
I would like to do this, definitely. I wonder if we'll break many people with the removal of the cookie jar support. I know servo has a bunch of cookie code, but looking through it, I think they can simply change `extern crate cookie` to `use hyper::header::shared::Cookie` (@jdm right?).

Though, ultimately, I'd like to put this in the [cookies](https://crates.io/crates/cookies) crate.

jdm:
As far as I can tell all we use is the Cookie struct and Cookie::parse. Kind of sucks to duplicate that between hyper and cookie-rs, though. Maybe there could be a cookiejar crate that extracts the higher-level stuff and leaves cookie-rs with the parsing, test cases, and data definition.

reem:
I would also be happier if we could at least use a common type with the cookies crate. Unfortunately, because of the nature of static typing, there's not really a great way for us to be compatible with the cookies crate without depending on the cookies crate (and the associated ssl stuff).

seanmonstar:
Indeed. Mostly, I've just wanted the Cookie parts, and leave the cookie signing to another crate.

@alexcrichton Would you be interested in having cookie-rs broken up? If you, we could work together on that, and not do this. If you'd rather keep cookie-rs the way it is, that's fine too.

alexcrichton:
Yeah I'm up for any direction for `cookie-rs` to go in, it's gotten a lot more use than I originally though it would! I'd also be fine just making the openssl dep an optional dependency as it's only needed for signed/encrypted cookie jars, which not everyone needs.

I'm also fine if you guys want to start your own cookie library, I'll be the first to say that `cookie-rs` would probably need some love! Regardless, lemme know any way I can help, and I surely don't mind!

pyfisch:
I promised @alexcrichton to improve cookie-rs and make openssl an optional dependency, maybe I will make also CookieJar optional.

PR #435, opened 2015-04-06
Updates for rust-openssl 0.6.0

markuskobler:


GitCop:

 * Commit: 80176c61db23e34864dce7d42b426808278c149c

 * Commit: 20178be4c89c2278203d3e37f13f72709e5bb6b1

---

paulosuzart:
@markuskobler will you try again? Just hit the same error.

markuskobler:
@paulosuzart Are you hitting the error below, because apparently its [working by design](https://github.com/rust-lang/cargo/issues/886#issuecomment-63200104).

`native library openssl is being linked to by more than one package, and can only be linked to by one package`

Their might be a better way but I got round it by using a submodule to checkout rust-openssl and then had a `.cargo/config` like this.
```
paths = [
      "vendor/hyper",
      "vendor/openssl"
]
```

src/net.rs
```diff
@@ -180,12 +172,10 @@ impl HttpListener {
...
         let mut ssl_context = try!(SslContext::new(Sslv23).map_err(lift_ssl_error));
-        try_some!(ssl_context.set_cipher_list("DEFAULT").map(lift_ssl_error));
-        try_some!(ssl_context.set_certificate_file(
-                cert, X509FileType::PEM).map(lift_ssl_error));
-        try_some!(ssl_context.set_private_key_file(
-                key, X509FileType::PEM).map(lift_ssl_error));
-        ssl_context.set_verify(SslVerifyNone, None);
+        try!(ssl_context.set_cipher_list("DEFAULT").map_err(lift_ssl_error));
+        try!(ssl_context.set_certificate_file(cert, X509FileType::PEM).map_err(lift_ssl_error));
+        try!(ssl_context.set_private_key_file(key, X509FileType::PEM).map_err(lift_ssl_error));
```
reem: Maybe we want to change the signature of this function to return `HttpResult` so that we can use `From` instead of `map_err(lift_ssl_error)`.

paulosuzart:
@markuskobler your fix are pertinent and useful. I tried to fix myself and ended up fixing exactly the way you did. Before submit my PR I found yours. So if you just fix your commit messages they might be accepted.

src/net.rs
markuskobler: Happy to change that as well but that would be the first use of HttpResult net.rs

markuskobler:
@paulosuzart Sorry now I understand. So I thought my rebase https://github.com/markuskobler/hyper/commit/660a362b68ef5671a951f2c4d66093621d0e3fd3 fixed the gitcop warning.

I think it just needs someone to kick off the travis build again now that it looks like cookie-rs has been upgraded to the latest rust openssl.

src/net.rs
reem: I think we should, so that we can keep the entire error around rather than losing information like we do now.

seanmonstar: If that's what we want, it can be in a separate PR. Although, with the `io::Error` changes to holding a `Box<Error>`, the original SSL error is no longer really lost. It's boxed into a trait object.

markuskobler: So had you envisaged adding  adding a new HttpSslError(SslError) to the HttpError enum?

seanmonstar:
Just restarted travis.

src/net.rs
markuskobler: I think you also need to change the `NetworkConnector::connect` and `NetworkListener::accept` signatures?

seanmonstar:
Thanks for this!

PR #440, opened 2015-04-07
refactor(headers): Use header!() for CORS headers.

pyfisch:
This is the last bunch of headers that should use the new macro. Moved them out of
their own folder so that the macro works. Changed them, so that they are more in
line with the other headers.

BREAKING CHANGE: `AccessControlAllowHeaders` and `AccessControlRequestHeaders` values
are case insensitive now. `AccessControlAllowOrigin` variants are now `Any` and
`Value` to match the other headers.


pyfisch:
@Manishearth you contributed much code to CORS, are you okay with the changes?

Manishearth:
Looks okay to me.

Issue #80, opened 2014-10-15
Change Header trait to allow headers that require multiple lines


seanmonstar:
blocks #71

seanmonstar:
This was done in 0.11, `Header::fmt_header` now receives a `header::Formatter` that has a `fmt_line` method.

PR #442, opened 2015-04-08
refactor(headers): Add encode_headers() function

pyfisch:
The HTTP headers block is not always a valid UTF-8 string, so it makes no sense to
implement the Display trait for it (additionally Display is intended for humans
and HTTP header blocks clearly are not). Instead the headers should be encoded as
a byte stream.

Removed a few tests, since they do not work without Display.

Remark: Currently formatted headers can't contain invalid chars, another commit
will enable this.

BREAKING CHANGE: fmt::Display is no longer implemented for Headers.

### DIFF
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -8,6 +8,7 @@ use std::any::Any;
 use std::borrow::{Cow, ToOwned};
 use std::collections::HashMap;
 use std::collections::hash_map::{Iter, Entry};
+use std::io::{self, Write};
 use std::iter::{FromIterator, IntoIterator};
 use std::ops::{Deref, DerefMut};
 use std::{mem, fmt};
@@ -223,13 +224,12 @@ impl Headers {
     }
 }
 
-impl fmt::Display for Headers {
-   fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        for header in self.iter() {
-            try!(write!(fmt, "{}\r\n", header));
-        }
-        Ok(())
+/// Encodes headers for use with HTTP/1.x
+pub fn encode_headers<W: Write>(headers: &Headers, writeable: &mut W) -> io::Result<()> {
+    for header in headers.iter() {
+        try!(write!(writeable, "{}\r\n", header));
     }
+    Ok(())
 }
 
 impl fmt::Debug for Headers {
@@ -383,11 +383,12 @@ impl AsRef<str> for CowStr {
 #[cfg(test)]
 mod tests {
     use std::fmt;
+    use std::str;
     use mime::Mime;
     use mime::TopLevel::Text;
     use mime::SubLevel::Plain;
     use super::{Headers, Header, HeaderFormat, ContentLength, ContentType,
-                Accept, Host, qitem};
+                Accept, Host, qitem, encode_headers};
     use httparse;
 
     #[cfg(feature = "nightly")]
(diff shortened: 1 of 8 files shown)


src/header/mod.rs
```diff
@@ -61,6 +62,14 @@ pub trait HeaderFormat: fmt::Debug + HeaderClone + Any + Typeable + Send + Sync
 
 }
 
+/// A trait for any header value to encode it and send it over the wire
+pub trait EncodeHeader: HeaderFormat + fmt::Display {
```
seanmonstar: This trait isn't used anywhere?

src/header/mod.rs
```diff
@@ -61,6 +62,14 @@ pub trait HeaderFormat: fmt::Debug + HeaderClone + Any + Typeable + Send + Sync
...
+    /// Encode a header to add it into a headers block
+    fn encode_header(&self) -> Vec<Vec<u8>> {
+        vec![format!("{}", self).into_bytes()]
```
seanmonstar: Since this trait isn't used, it's hard to envision how you expect it to be used. However, I think it'd likely be better to write the bytes directly to a `Write` than to allocate 2 `Vec`s and return them...

seanmonstar:
The main point is that to you, implementing `fmt::Display` for `Headers` seems wrong, since `Headers` doesn't always contain UTF-8? This may be true, but I'd like to see farther into the design of how to support non-UTF-8 printing before making a merge like this.

src/header/mod.rs
pyfisch: Looks like git tricked me here, I did not want to commit this :disappointed:

pyfisch:
No problem, here is the design in short form:
* `HeaderFormat` gets removed, it is replaced with `EncodeHeader`, providing the `encode_header()` method returning a `Vec<Vec<u8>>`. Allowing multiple lines fixes #80. Using `Vec<u8>` allows us to encode non-ASCII headers, this is at least useful for `Cookie`, `Set-Cookie`, `ETag`, all other headers using entity tags and maybe some more.
* Raw headers are also encoded as `Vec<Vec<u8>>`
* An `encode_headers()` function is used to encode `Headers` for HTTP/1.x, for HTTP/2 a hpack should be used
* All types representing headers stop implementing `Display` since they may contain non-ASCII data.

pyfisch:
@seanmonstar would you prefer to merge the current code or should I implement the complete design first and land one huge PR?

seanmonstar:
The idea makes a fair amount of sense. I'll recap:

- Headers from new Clients and Servers today *should* use only ASCII.
- Some existing implementations may exist that use non-ASCII headers.
- When writing Headers in hyper, we 99% of the time writing explicit Headers set by the user, and they should know better to use ASCII.
- When we receive Headers, it's not a requirement that they be in ASCII. It is possible to inspect non-ASCII headers via the raw representation.
- The **only** time we should be needing to *write* back out non-ASCII headers is in proxying situations: The proxy has received some headers that include non-ASCII, and we must write the Header exactly as we received it, not losing data.
- Because of this 1 scenario, we cannot use `fmt::Display` to do so, since that library type is required to output UTF-8 Strings; the header could be any encoding.

One thing I'd like to solve in this proposed design is encoding to `Vec<Vec<u8>>`. This means 2 allocations for every header, that will be promptly thrown away. Perhaps we could pass a `&mut std::io::Write` or something. Most Headers could continue to use `write!(wrt, "{}", self)`, since `std::io::Write` includes a `write_fmt` method. The few outliers could use `write(&[u8])`.

pyfisch:
Good summary.

There is a scenario two, where a website provides cookie values or entity tags containing non-ASCII data that should be send back to the website in another request.

We cannot simply pass `&mut std::io::Write` to an encode_headers function since we need to be able to serialize a header into multiple values at least for the `Set-Cookie` header and for better compression in HTTP/2. So we would need an iterator. The iterator would yield elements that have an encode() method taking `&mut std::io::Write`. I do not think that using iterators is any faster since you need also many allocations.

seanmonstar:
Reem and I were just talking about this last night, and I want to press on
it. Still trying to think of a good way to handle Set-Cookie.

PR #448, opened 2015-04-11
feat(client): initial support for HTTP/2

mlalic:
This PR adds initial support for HTTP/2 requests.

It introduces a new pair of traits: `HttpMessage` and `Protocol`.

The `HttpMessage` trait provides an API that abstracts the sending and receiving of HTTP messages away from the concrete implementation of the underlying protocol.

The `Protocol` trait provides an API for constructing new `HttpMessage`s.

The `Request` and `Response` structs are refactored to internally use a `Box<HttpMessage>` for their HTTP communication requirements.

The `Client` has a new public method: `with_protocol`. This provides an API for plugging in a particular `Protocol` implementation that will then be used for constructing the `HttpMessage` instances that are passed to the `Request` struct constructor.

The implementation of this API (i.e. the two traits) for HTTP/2 is provided in the `hyper::http2` module.

Therefore, now it is possible to inject an `Http2Protocol` instance when constructing a `hyper::Client` which seamlessly makes it switch to using HTTP/2.

The `Http2Protocol` can use the existing `NetworkConnector` API for constructing the network connection for the underlying communication. However, do note that the current TLS connector does not honor the requirements that the HTTP/2 spec additionally makes for TLS connections (particular cipher suits are banned, applications must use ALPN -- or at least NPN -- for protocol negotiation), so if such a connector is provided, it won't be possible to establish an HTTP/2 connection. The cleartext TCP connectors work as expected.

An example is included in `examples/client_http2.rs` showing how the client can be configured to use HTTP/2.

```
mlalic@skaro:~/src/hyper$ cargo run --example client_http2 -- http://http2bin.org/get?hi=hi
     Running `target/debug/examples/client_http2 http://http2bin.org/get?hi=hi`
Response: 200 OK
Headers:
x-clacks-overhead: GNU Terry Pratchett
access-control-allow-origin: *
:status: 200
content-type: application/json
server: h2o/1.1.1
date: Fri, 29 May 2015 12:29:14 GMT
access-control-allow-credentials: true

{
  "args": {
    "hi": "hi"
  }, 
  "headers": {
    "Connection": "keep-alive", 
    "Host": "http2bin.org,http2bin.org", 
    "Via": "2 http2bin.org"
  }, 
  "origin": "131.159.207.123", 
  "url": "http://http2bin.org,http2bin.org/get?hi=hi"
}
```

# Breaking Changes

The PR introduces only a few minor, however still breaking changes:

- The `hyper::client::Response` struct no longer requires any type parameters.
- The `hyper::Error` public enum has gained a new variant `Http2`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ unicase = "0.1"
 url = "0.2"
 traitobject = "0.0.1"
 typeable = "0.1"
+solicit = "0.2"
 
 [dev-dependencies]
 env_logger = "*"
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -54,11 +54,14 @@ pub mod pool;
 pub mod request;
 pub mod response;
 
+use message::Protocol;
+use http11::Http11Protocol;
+
 /// A Client to use additional features with Requests.
 ///
 /// Clients can handle things such as: redirect policy, connection pooling.
 pub struct Client {
-    connector: Connector,
+    protocol: Box<Protocol + Send>,
     redirect_policy: RedirectPolicy,
 }
 
@@ -77,15 +80,20 @@ impl Client {
     /// Create a new client with a specific connector.
     pub fn with_connector<C, S>(connector: C) -> Client
     where C: NetworkConnector<Stream=S> + Send + 'static, S: NetworkStream + Send {
+        Client::with_protocol(Http11Protocol::with_connector(connector))
+    }
+
+    /// Create a new client with a specific `Protocol`.
+    pub fn with_protocol<P: Protocol + Send + 'static>(protocol: P) -> Client {
         Client {
-            connector: with_connector(connector),
+            protocol: Box::new(protocol),
             redirect_policy: Default::default()
         }
     }
 
     /// Set the SSL verifier callback for use with OpenSSL.
     pub fn set_ssl_verifier(&mut self, verifier: ContextVerifier) {
-        self.connector.set_ssl_verifier(verifier);
+        self.protocol.set_ssl_verifier(verifier);
     }
 
     /// Set the RedirectPolicy.
@@ -131,42 +139,8 @@ impl Client {
     }
 }
 
-fn with_connector<C: NetworkConnector<Stream=S> + Send + 'static, S: NetworkStream + Send>(c: C) -> Connector {
-    Connector(Box::new(ConnAdapter(c)))
-}
-
 impl Default for Client {
     fn default() -> Client { Client::new() }
-}
-
-struct ConnAdapter<C: NetworkConnector + Send>(C);
-
-impl<C: NetworkConnector<Stream=S> + Send, S: NetworkStream + Send> NetworkConnector for ConnAdapter<C> {
-    type Stream = Box<NetworkStream + Send>;
-    #[inline]
-    fn connect(&self, host: &str, port: u16, scheme: &str)
-        -> ::Result<Box<NetworkStream + Send>> {
-        Ok(try!(self.0.connect(host, port, scheme)).into())
-    }
-    #[inline]
-    fn set_ssl_verifier(&mut self, verifier: ContextVerifier) {
-        self.0.set_ssl_verifier(verifier);
-    }
-}
-
-struct Connector(Box<NetworkConnector<Stream=Box<NetworkStream + Send>> + Send>);
-
-impl NetworkConnector for Connector {
-    type Stream = Box<NetworkStream + Send>;
-    #[inline]
-    fn connect(&self, host: &str, port: u16, scheme: &str)
-        -> ::Result<Box<NetworkStream + Send>> {
-        Ok(try!(self.0.connect(host, port, scheme)).into())
-    }
-    #[inline]
-    fn set_ssl_verifier(&mut self, verifier: ContextVerifier) {
-        self.0.set_ssl_verifier(verifier);
-    }
 }
 
 /// Options for an individual Request.
@@ -229,7 +203,11 @@ impl<'a, U: IntoUrl> RequestBuilder<'a, U> {
         };
 
         loop {
-            let mut req = try!(Request::with_connector(method.clone(), url.clone(), &client.connector));
+            let message = {
+                let (host, port) = try!(get_host_and_port(&url));
+                try!(client.protocol.new_message(&host, port, &*url.scheme))
+            };
+            let mut req = try!(Request::with_message(method.clone(), url.clone(), message));
             headers.as_ref().map(|headers| req.headers_mut().extend(headers.iter()));
 
             match (can_have_body, body.as_ref()) {
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,6 +6,7 @@ use std::io::Error as IoError;
 use httparse;
 use openssl::ssl::error::SslError;
 use url;
+use solicit::http::HttpError as Http2Error;
 
 use self::Error::{
     Method,
@@ -15,7 +16,8 @@ use self::Error::{
     Status,
     Io,
     Ssl,
-    TooLarge
+    TooLarge,
+    Http2,
 };
 
 
@@ -40,7 +42,9 @@ pub enum Error {
     /// An `io::Error` that occurred while trying to read or write to a network stream.
     Io(IoError),
     /// An error from the `openssl` library.
-    Ssl(SslError)
+    Ssl(SslError),
+    /// An HTTP/2-specific error, coming from the `solicit` library.
+    Http2(Http2Error),
 }
 
 impl fmt::Display for Error {
@@ -60,6 +64,7 @@ impl StdError for Error {
             Uri(ref e) => e.description(),
             Io(ref e) => e.description(),
             Ssl(ref e) => e.description(),
+            Http2(ref e) => e.description(),
         }
     }
 
@@ -68,6 +73,7 @@ impl StdError for Error {
             Io(ref error) => Some(error),
             Ssl(ref error) => Some(error),
             Uri(ref error) => Some(error),
+            Http2(ref error) => Some(error),
             _ => None,
         }
     }
@@ -108,12 +114,19 @@ impl From<httparse::Error> for Error {
     }
 }
 
+impl From<Http2Error> for Error {
+    fn from(err: Http2Error) -> Error {
+        Error::Http2(err)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::error::Error as StdError;
     use std::io;
     use httparse;
     use openssl::ssl::error::SslError;
+    use solicit::http::HttpError as Http2Error;
     use url;
     use super::Error;
     use super::Error::*;
@@ -156,6 +169,7 @@ mod tests {
         from_and_cause!(io::Error::new(io::ErrorKind::Other, "other") => Io(..));
         from_and_cause!(url::ParseError::EmptyHost => Uri(..));
         from_and_cause!(SslError::SslSessionClosed => Ssl(..));
+        from_and_cause!(Http2Error::UnknownStreamId => Http2(..));
 
         from!(SslError::StreamError(io::Error::new(io::ErrorKind::Other, "ssl negotiation")) => Io(..));
 
--- a/src/http11.rs
+++ b/src/http11.rs
@@ -5,6 +5,8 @@ use std::net::Shutdown;
 use method::{Method};
 use header::{ContentLength, TransferEncoding};
 use header::Encoding::Chunked;
+
+use net::{NetworkConnector, NetworkStream, ContextVerifier};
 use http::{HttpWriter, LINE_ENDING};
 use http::HttpReader::{SizedReader, ChunkedReader, EofReader};
 use http::HttpWriter::{ChunkedWriter, SizedWriter, EmptyWriter};
@@ -12,11 +14,11 @@ use buffer::BufReader;
 use http::{self, HttpReader};
 
 use message::{
+    Protocol,
     HttpMessage,
     RequestHead,
     ResponseHead,
 };
-use net::NetworkStream;
 use header;
 use version;
 
@@ -239,5 +241,65 @@ impl Http11Message {
         self.stream = Some(raw);
 
         Ok(())
+    }
+}
+
+/// The `Protocol` implementation provides HTTP/1.1 messages.
+pub struct Http11Protocol {
+    connector: Connector,
+}
+
+impl Protocol for Http11Protocol {
+    fn new_message(&self, host: &str, port: u16, scheme: &str) -> ::Result<Box<HttpMessage>> {
+        let stream = try!(self.connector.connect(host, port, scheme)).into();
+
+        Ok(Box::new(Http11Message::with_stream(stream)))
+    }
+
+    #[inline]
+    fn set_ssl_verifier(&mut self, verifier: ContextVerifier) {
+        self.connector.set_ssl_verifier(verifier);
+    }
+}
+
+impl Http11Protocol {
+    /// Creates a new `Http11Protocol` instance that will use the given `NetworkConnector` for
+    /// establishing HTTP connections.
+    pub fn with_connector<C, S>(c: C) -> Http11Protocol
+            where C: NetworkConnector<Stream=S> + Send + 'static,
+                  S: NetworkStream + Send {
+        Http11Protocol {
+            connector: Connector(Box::new(ConnAdapter(c))),
+        }
+    }
+}
+
+struct ConnAdapter<C: NetworkConnector + Send>(C);
+
+impl<C: NetworkConnector<Stream=S> + Send, S: NetworkStream + Send> NetworkConnector for ConnAdapter<C> {
+    type Stream = Box<NetworkStream + Send>;
+    #[inline]
+    fn connect(&self, host: &str, port: u16, scheme: &str)
+        -> ::Result<Box<NetworkStream + Send>> {
+        Ok(try!(self.0.connect(host, port, scheme)).into())
+    }
+    #[inline]
+    fn set_ssl_verifier(&mut self, verifier: ContextVerifier) {
+        self.0.set_ssl_verifier(verifier);
+    }
+}
+
+struct Connector(Box<NetworkConnector<Stream=Box<NetworkStream + Send>> + Send>);
+
+impl NetworkConnector for Connector {
+    type Stream = Box<NetworkStream + Send>;
+    #[inline]
+    fn connect(&self, host: &str, port: u16, scheme: &str)
+        -> ::Result<Box<NetworkStream + Send>> {
+        Ok(try!(self.0.connect(host, port, scheme)).into())
+    }
+    #[inline]
+    fn set_ssl_verifier(&mut self, verifier: ContextVerifier) {
+        self.0.set_ssl_verifier(verifier);
     }
 }
--- a/src/http11.rs
+++ b/src/http11.rs
@@ -1,0 +1,243 @@
+//! Adapts the HTTP/1.1 implementation into the `HttpMessage` API.
+use std::io::{self, Write, BufWriter, Read};
+use std::net::Shutdown;
+
+use method::{Method};
+use header::{ContentLength, TransferEncoding};
+use header::Encoding::Chunked;
+use http::{HttpWriter, LINE_ENDING};
+use http::HttpReader::{SizedReader, ChunkedReader, EofReader};
+use http::HttpWriter::{ChunkedWriter, SizedWriter, EmptyWriter};
+use buffer::BufReader;
+use http::{self, HttpReader};
+
+use message::{
+    HttpMessage,
+    RequestHead,
+    ResponseHead,
+};
+use net::NetworkStream;
+use header;
+use version;
+
+/// An implementation of the `HttpMessage` trait for HTTP/1.1.
+#[derive(Debug)]
+pub struct Http11Message {
+    stream: Option<Box<NetworkStream + Send>>,
+    writer: Option<HttpWriter<BufWriter<Box<NetworkStream + Send>>>>,
+    reader: Option<HttpReader<BufReader<Box<NetworkStream + Send>>>>,
+}
+
+impl Write for Http11Message {
+    #[inline]
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        match self.writer {
+            None => Err(io::Error::new(io::ErrorKind::Other,
+                                          "Not in a writable state")),
+            Some(ref mut writer) => writer.write(buf),
+        }
+    }
+    #[inline]
+    fn flush(&mut self) -> io::Result<()> {
+        match self.writer {
+            None => Err(io::Error::new(io::ErrorKind::Other,
+                                          "Not in a writable state")),
+            Some(ref mut writer) => writer.flush(),
+        }
+    }
+}
+
+impl Read for Http11Message {
+    #[inline]
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        match self.reader {
+            None => Err(io::Error::new(io::ErrorKind::Other,
+                                          "Not in a readable state")),
+            Some(ref mut reader) => reader.read(buf),
+        }
+    }
+}
+
+impl HttpMessage for Http11Message {
+    fn set_outgoing(&mut self, mut head: RequestHead) -> ::Result<RequestHead> {
+        if self.stream.is_none() {
+            return Err(From::from(io::Error::new(
+                        io::ErrorKind::Other,
+                        "Message not idle, cannot start new outgoing")));
+        }
+        let mut stream = BufWriter::new(self.stream.take().unwrap());
+
+        let mut uri = head.url.serialize_path().unwrap();
+        if let Some(ref q) = head.url.query {
+            uri.push('?');
+            uri.push_str(&q[..]);
+        }
+
+        let version = version::HttpVersion::Http11;
+        debug!("request line: {:?} {:?} {:?}", head.method, uri, version);
+        try!(write!(&mut stream, "{} {} {}{}",
+                    head.method, uri, version, LINE_ENDING));
+
+        let stream = match head.method {
+            Method::Get | Method::Head => {
+                debug!("headers={:?}", head.headers);
+                try!(write!(&mut stream, "{}{}", head.headers, LINE_ENDING));
+                EmptyWriter(stream)
+            },
+            _ => {
+                let mut chunked = true;
+                let mut len = 0;
+
+                match head.headers.get::<header::ContentLength>() {
+                    Some(cl) => {
+                        chunked = false;
+                        len = **cl;
+                    },
+                    None => ()
+                };
+
+                // can't do in match above, thanks borrowck
+                if chunked {
+                    let encodings = match head.headers.get_mut::<header::TransferEncoding>() {
+                        Some(&mut header::TransferEncoding(ref mut encodings)) => {
+                            //TODO: check if chunked is already in encodings. use HashSet?
+                            encodings.push(header::Encoding::Chunked);
+                            false
+                        },
+                        None => true
+                    };
+
+                    if encodings {
+                        head.headers.set::<header::TransferEncoding>(
+                            header::TransferEncoding(vec![header::Encoding::Chunked]))
+                    }
+                }
+
+                debug!("headers={:?}", head.headers);
+                try!(write!(&mut stream, "{}{}", head.headers, LINE_ENDING));
+
+                if chunked {
+                    ChunkedWriter(stream)
+                } else {
+                    SizedWriter(stream, len)
+                }
+            }
+        };
+
+        self.writer = Some(stream);
+
+        Ok(head)
+    }
+
+    fn get_incoming(&mut self) -> ::Result<ResponseHead> {
+        try!(self.flush_outgoing());
+        if self.stream.is_none() {
+            // The message was already in the reading state...
+            // TODO Decide what happens in case we try to get a new incoming at that point
+            return Err(From::from(
+                    io::Error::new(io::ErrorKind::Other,
+                    "Read already in progress")));
+        }
+
+        let stream = self.stream.take().unwrap();
+        let mut stream = BufReader::new(stream);
+
+        let head = try!(http::parse_response(&mut stream));
+        let raw_status = head.subject;
+        let headers = head.headers;
+
+        let body = if headers.has::<TransferEncoding>() {
+            match headers.get::<TransferEncoding>() {
+                Some(&TransferEncoding(ref codings)) => {
+                    if codings.len() > 1 {
+                        trace!("TODO: #2 handle other codings: {:?}", codings);
+                    };
+
+                    if codings.contains(&Chunked) {
+                        ChunkedReader(stream, None)
+                    } else {
+                        trace!("not chuncked. read till eof");
+                        EofReader(stream)
+                    }
+                }
+                None => unreachable!()
+            }
+        } else if headers.has::<ContentLength>() {
+            match headers.get::<ContentLength>() {
+                Some(&ContentLength(len)) => SizedReader(stream, len),
+                None => unreachable!()
+            }
+        } else {
+            trace!("neither Transfer-Encoding nor Content-Length");
+            EofReader(stream)
+        };
+
+        self.reader = Some(body);
+
+        Ok(ResponseHead {
+            headers: headers,
+            raw_status: raw_status,
+            version: head.version,
+        })
+    }
+
+    fn close_connection(&mut self) -> ::Result<()> {
+        try!(self.get_mut().close(Shutdown::Both));
+        Ok(())
+    }
+}
+
+impl Http11Message {
+    /// Consumes the `Http11Message` and returns the underlying `NetworkStream`.
+    pub fn into_inner(mut self) -> Box<NetworkStream + Send> {
+        if self.stream.is_some() {
+            self.stream.take().unwrap()
+        } else if self.writer.is_some() {
+            self.writer.take().unwrap().into_inner().into_inner().unwrap()
+        } else if self.reader.is_some() {
+            self.reader.take().unwrap().into_inner().into_inner()
+        } else {
+            panic!("Http11Message lost its underlying stream somehow");
+        }
+    }
+
+    /// Gets a mutable reference to the underlying `NetworkStream`, regardless of the state of the
+    /// `Http11Message`.
+    pub fn get_mut(&mut self) -> &mut Box<NetworkStream + Send> {
+        if self.stream.is_some() {
+            self.stream.as_mut().unwrap()
+        } else if self.writer.is_some() {
+            self.writer.as_mut().unwrap().get_mut().get_mut()
+        } else if self.reader.is_some() {
+            self.reader.as_mut().unwrap().get_mut().get_mut()
+        } else {
+            panic!("Http11Message lost its underlying stream somehow");
+        }
+    }
+
+    /// Creates a new `Http11Message` that will use the given `NetworkStream` for communicating to
+    /// the peer.
+    pub fn with_stream(stream: Box<NetworkStream + Send>) -> Http11Message {
+        Http11Message {
+            stream: Some(stream),
+            writer: None,
+            reader: None,
+        }
+    }
+
+    /// Flushes the current outgoing content and moves the stream into the `stream` property.
+    ///
+    /// TODO It might be sensible to lift this up to the `HttpMessage` trait itself...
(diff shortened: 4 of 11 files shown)


GitCop:

 * Commit: 830ebe2c7c2c02b4ee7c22c3043417cafdac5a55

 * Commit: 42d4364e4b2c45c8bd8e3d0905f92b26b9b9c727

 * Commit: d426c99ed4287d633978e3874a9d4b91465163cd

 * Commit: abf43e03ef1cfd6901026799db9be7dfb4bcdf88

 * Commit: 8e61b4351526ce468a887fc30eb0bc62f5a134a0

 * Commit: 0a3c8e13a8338190c7ce16d5bb2194cf2229ee96

 * Commit: 4eb9786a105c74c2b940954a6f6fbf6da2c44019

 * Commit: 45f509c9e9e8dd6a382a54c804e77d814488a7ba

 * Commit: acd28069931d90775605d730d23515c9945c235c

 * Commit: b3395ef688dbbb17cd8f9e3403ed5a654383abc3

 * Commit: 14c1d60a2af5885cf80bcc3be8b371490f620d66

 * Commit: 84925772d5411e71a29d4261d6b7dc1bcb6f65b1

 * Commit: ac1fe5467d2b44080c70cdcf6acea36ace4e9e3d

 * Commit: 11ae013fdb5d2f82cc208a6f6ae4bd0c73fb2842

 * Commit: 42bcf865f35d44876e618dece0355afa3afd60b4

 * Commit: 4d5f293883ce65f0f6dd4cd1f995bf5b17efc590

 * Commit: 9196bab0529033680b64afc0ac89ddeff24d2c98

 * Commit: d8c13fc07d86a13bfc80ddd0ed81df37276a15a1

 * Commit: 121e9ab84c66a2e1c0a0c70fd74503e703127545

 * Commit: 6ffc23f21d1910073c1c6a24229bb82802049161

 * Commit: 482e24f834434889feec8b72d2db6ada45072f10

 * Commit: e13ce02766cc3c0772ea6ca88d9f4a43f82107c6

 * Commit: 190101f9f0e2f3c35d38074c542c65f91d088a08

 * Commit: f553ffddb8bbc20e0aae694cbd4934fcc32eb0b9

 * Commit: 4ad58bc9753e640d81d78ffc757a4193cee98ee7

---

seanmonstar:
So excite.

Cargo.toml
```diff
@@ -25,6 +25,8 @@ url = "*"
 traitobject = "*"
 typeable = "*"
 
+[dependencies.solicit]
+git = "https://github.com/mlalic/solicit"
```
seanmonstar: Not on crates.io yet?

src/client/mod.rs
```diff
@@ -32,39 +32,120 @@ use status::StatusClass::Redirection;
...
+        self.connector = with_connector(HttpConnector(Some(verifier)));
+    }
+}
+
 /// A Client to use additional features with Requests.
 ///
 /// Clients can handle things such as: redirect policy.
-pub struct Client {
-    connector: Connector,
+pub struct Client<F=Http11RequestFactory> where F: HttpRequestFactory {
```
seanmonstar: So as to keep the Client without type parameters, what if we keep a `Box<HttpRequestFactory>` instead?

src/header/mod.rs
```diff
@@ -164,8 +164,9 @@ impl Headers {
     /// # let mut headers = Headers::new();
     /// headers.set_raw("content-length", vec![b"5".to_vec()]);
     /// ```
-    pub fn set_raw<K: Into<Cow<'static, str>>>(&mut self, name: K, value: Vec<Vec<u8>>) {
-        self.data.insert(UniCase(CowStr(name.into())), Item::new_raw(value));
+    pub fn set_raw<'a, K: Into<Cow<'a, str>>>(&mut self, name: K, value: Vec<Vec<u8>>) {
```
seanmonstar: This change means the example on L165 allocates the `"content-length"` as a `String`, but there's no need.

src/http2.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        // TODO Refactor this to a helper function, as it is reused verbatim
+        //      from the HTTP/1.x code.
+        let mut uri = self.url.serialize_path().unwrap();
+        if let Some(ref q) = self.url.query {
+            uri.push('?');
+            uri.push_str(&q[..]);
+        }
+
+        let mut path: Vec<u8> = Vec::new();
```
seanmonstar: This can be `uri.into_bytes()` I believe.

src/http2.rs
```diff
@@ -0,0 +1,271 @@
...
+    type Response = Http2Response;
+    fn send(self) -> HttpResult<Http2Response> { self.send() }
+}
+
+/// A fake implementation since writing bodies is not yet supported, but
+/// we need it to satisfy the interface.
+impl io::Write for Http2Request<Streaming> {
+    #[inline]
+    fn write(&mut self, msg: &[u8]) -> io::Result<usize> {
+        Ok(msg.len())
```
seanmonstar: I think it might be better to toss an `unimplemented!()` here instead.

src/lib.rs
```diff
@@ -1,6 +1,4 @@
 #![doc(html_root_url = "https://hyperium.github.io/hyper/hyper/index.html")]
-#![deny(missing_docs)]
-#![cfg_attr(test, deny(warnings))]
```
seanmonstar: woops?

Cargo.toml
mlalic: Yeah, not yet... I wanted to maintain some flexibility, without pushing multiple versions out to crates.io, if it turned out that I needed to do some changes there to make things work better in `hyper` (so far it turned out that I didn't).

Of course it's something to do before finally landing this to make sure `hyper` is pinned to a proper version.

src/client/mod.rs
mlalic: That was my initial idea as well, but after some thought I found it wouldn't work.

You see, the `FreshHttpRequest`, `StreamingHttpRequest`, and `HttpResponse` traits define methods that *move* (i.e. consume) the object they're called on (`start`, `send`). What that means is that you can't call them on trait objects (such as `Box<FreshHttpRequest>`). Therefore, you have to know the exact types of the request and response objects, so even if the factory were boxed up, the `Client` would end up having to be generic over the `FreshHttpRequest` trait (which is what the factory produces). Perhaps there's a different workaround for this, but since the type of the fresh request needs to be known at the point of calling the `start` method, I don't see how we could fully avoid any type parameter to the `Client`.

A different option is to make the request and response methods not consume the object, but that would be a much more invasive change to the codebase and would require a lot more thought about the implications of such a change. I, myself, am not sure whether that'd be a good approach, since at least currently, the requests' "natural" lifecycle is a transformation `Fresh -> Streaming -> Response` and so it is natural that the methods consume/move the object.

src/header/mod.rs
mlalic: Fair enough... I'll revert this. Come to think of it, this change isn't even required for the `http2` code anymore (as I found a different way to do what this had been intended to).

src/http2.rs
mlalic: Sure, might make more sense for now...

src/lib.rs
mlalic: Hehe :)

As I mentioned in the opening comment, the PR is definitely not ready to be merged just yet.

I removed this since I wanted to be able to experiment with changes without being slowed down by having to put docs everywhere and get rid of warnings.

If we want to go forward with polishing this up, then of course all those issues will be solved and the commit that removed these two lines rebased out of the tree.

GitCop:

 * Commit: 830ebe2c7c2c02b4ee7c22c3043417cafdac5a55

 * Commit: 42d4364e4b2c45c8bd8e3d0905f92b26b9b9c727

 * Commit: d426c99ed4287d633978e3874a9d4b91465163cd

 * Commit: abf43e03ef1cfd6901026799db9be7dfb4bcdf88

 * Commit: 8e61b4351526ce468a887fc30eb0bc62f5a134a0

 * Commit: 0a3c8e13a8338190c7ce16d5bb2194cf2229ee96

 * Commit: 4eb9786a105c74c2b940954a6f6fbf6da2c44019

 * Commit: 45f509c9e9e8dd6a382a54c804e77d814488a7ba

 * Commit: acd28069931d90775605d730d23515c9945c235c

 * Commit: b3395ef688dbbb17cd8f9e3403ed5a654383abc3

 * Commit: 14c1d60a2af5885cf80bcc3be8b371490f620d66

 * Commit: 84925772d5411e71a29d4261d6b7dc1bcb6f65b1

 * Commit: ac1fe5467d2b44080c70cdcf6acea36ace4e9e3d

 * Commit: 11ae013fdb5d2f82cc208a6f6ae4bd0c73fb2842

 * Commit: 42bcf865f35d44876e618dece0355afa3afd60b4

 * Commit: 4d5f293883ce65f0f6dd4cd1f995bf5b17efc590

 * Commit: 9196bab0529033680b64afc0ac89ddeff24d2c98

 * Commit: d8c13fc07d86a13bfc80ddd0ed81df37276a15a1

 * Commit: 121e9ab84c66a2e1c0a0c70fd74503e703127545

 * Commit: 6ffc23f21d1910073c1c6a24229bb82802049161

 * Commit: 482e24f834434889feec8b72d2db6ada45072f10

 * Commit: e13ce02766cc3c0772ea6ca88d9f4a43f82107c6

 * Commit: 190101f9f0e2f3c35d38074c542c65f91d088a08

 * Commit: f553ffddb8bbc20e0aae694cbd4934fcc32eb0b9

 * Commit: 4ad58bc9753e640d81d78ffc757a4193cee98ee7

 * Commit: 20edf74e973ed1c382b4dfa70296ee33343061c2

 * Commit: 43d3b3af56f259272a71409dad94d81c0bbb7029

---

src/http2.rs
mlalic: Nice one!

src/http2.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn headers(&self) -> &Headers { &self.headers }
+    fn method(&self) -> Method { self.method.clone() }
+}
+
+impl Http2Request<Fresh> {
+    fn start(mut self) -> HttpResult<Http2Request<Streaming>> {
+        // Prepare the request metadata so that it fits into `solicit`'s API.
+        // It basically just adapts the method and path into byte slices.
+        let mut method_bytes: Vec<u8> = Vec::new();
+        write!(&mut method_bytes, "{}", self.method);
```
mlalic: How about we do something similar to what you suggested with using `into_bytes` for the path.

Namely, implement `Borrow<str>` for `Method` and then just pass that (`as_bytes`) to the `solicit` API that expects a `&[u8]`. It would avoid an extra allocation here...

GitCop:

 * Commit: 830ebe2c7c2c02b4ee7c22c3043417cafdac5a55

 * Commit: 42d4364e4b2c45c8bd8e3d0905f92b26b9b9c727

 * Commit: d426c99ed4287d633978e3874a9d4b91465163cd

 * Commit: abf43e03ef1cfd6901026799db9be7dfb4bcdf88

 * Commit: 8e61b4351526ce468a887fc30eb0bc62f5a134a0

 * Commit: 0a3c8e13a8338190c7ce16d5bb2194cf2229ee96

 * Commit: 4eb9786a105c74c2b940954a6f6fbf6da2c44019

 * Commit: 45f509c9e9e8dd6a382a54c804e77d814488a7ba

 * Commit: acd28069931d90775605d730d23515c9945c235c

 * Commit: b3395ef688dbbb17cd8f9e3403ed5a654383abc3

 * Commit: 14c1d60a2af5885cf80bcc3be8b371490f620d66

 * Commit: 84925772d5411e71a29d4261d6b7dc1bcb6f65b1

 * Commit: ac1fe5467d2b44080c70cdcf6acea36ace4e9e3d

 * Commit: 11ae013fdb5d2f82cc208a6f6ae4bd0c73fb2842

 * Commit: 42bcf865f35d44876e618dece0355afa3afd60b4

 * Commit: 4d5f293883ce65f0f6dd4cd1f995bf5b17efc590

 * Commit: 9196bab0529033680b64afc0ac89ddeff24d2c98

 * Commit: d8c13fc07d86a13bfc80ddd0ed81df37276a15a1

 * Commit: 121e9ab84c66a2e1c0a0c70fd74503e703127545

 * Commit: 6ffc23f21d1910073c1c6a24229bb82802049161

 * Commit: 482e24f834434889feec8b72d2db6ada45072f10

 * Commit: e13ce02766cc3c0772ea6ca88d9f4a43f82107c6

 * Commit: 190101f9f0e2f3c35d38074c542c65f91d088a08

 * Commit: f553ffddb8bbc20e0aae694cbd4934fcc32eb0b9

 * Commit: 4ad58bc9753e640d81d78ffc757a4193cee98ee7

 * Commit: 20edf74e973ed1c382b4dfa70296ee33343061c2

 * Commit: 43d3b3af56f259272a71409dad94d81c0bbb7029

 * Commit: 625a562abb4006e893960170eeae628fb3347960

---

src/client/mod.rs
reem: There is a hack to get around this where you define another trait that takes `self: Box<Self>` instead of `self`. See `FnBox` in std for an example.

src/http2.rs
reem: Probably `AsRef<str>` would be the better thing, but sounds good.

mlalic: Yeah, you're right `AsRef` would probably be better.

GitCop:

 * Commit: 830ebe2c7c2c02b4ee7c22c3043417cafdac5a55

 * Commit: 42d4364e4b2c45c8bd8e3d0905f92b26b9b9c727

 * Commit: d426c99ed4287d633978e3874a9d4b91465163cd

 * Commit: abf43e03ef1cfd6901026799db9be7dfb4bcdf88

 * Commit: 8e61b4351526ce468a887fc30eb0bc62f5a134a0

 * Commit: 0a3c8e13a8338190c7ce16d5bb2194cf2229ee96

 * Commit: 4eb9786a105c74c2b940954a6f6fbf6da2c44019

 * Commit: 45f509c9e9e8dd6a382a54c804e77d814488a7ba

 * Commit: acd28069931d90775605d730d23515c9945c235c

 * Commit: b3395ef688dbbb17cd8f9e3403ed5a654383abc3

 * Commit: 14c1d60a2af5885cf80bcc3be8b371490f620d66

 * Commit: 84925772d5411e71a29d4261d6b7dc1bcb6f65b1

 * Commit: ac1fe5467d2b44080c70cdcf6acea36ace4e9e3d

 * Commit: 11ae013fdb5d2f82cc208a6f6ae4bd0c73fb2842

 * Commit: 42bcf865f35d44876e618dece0355afa3afd60b4

 * Commit: 4d5f293883ce65f0f6dd4cd1f995bf5b17efc590

 * Commit: 9196bab0529033680b64afc0ac89ddeff24d2c98

 * Commit: d8c13fc07d86a13bfc80ddd0ed81df37276a15a1

 * Commit: 121e9ab84c66a2e1c0a0c70fd74503e703127545

 * Commit: 6ffc23f21d1910073c1c6a24229bb82802049161

 * Commit: 482e24f834434889feec8b72d2db6ada45072f10

 * Commit: e13ce02766cc3c0772ea6ca88d9f4a43f82107c6

 * Commit: 190101f9f0e2f3c35d38074c542c65f91d088a08

 * Commit: f553ffddb8bbc20e0aae694cbd4934fcc32eb0b9

 * Commit: 4ad58bc9753e640d81d78ffc757a4193cee98ee7

 * Commit: 20edf74e973ed1c382b4dfa70296ee33343061c2

 * Commit: 43d3b3af56f259272a71409dad94d81c0bbb7029

 * Commit: 625a562abb4006e893960170eeae628fb3347960

 * Commit: 87132f8e94f08c508361c2305b9fcaed18a6ed61

 * Commit: cb39b60812b209a0b83018eff0e48802de40b06f

---

src/client/mod.rs
mlalic: Right, that's a nice hack and could be used to circumvent the problem I mentioned.

That said, I'm still not convinced of the benefit of doing it like that, instead of having a generic `Client`. Consider the `send` method on the `RequestBuilder`. In this scenario, its signature needs to change to return `HttpResult<Box<HttpResponse>>` -- since we can't know the actual type of the response. This is also a change to the public API of the client -- and in my opinion a more significant one than introducing a generic type parameter (with a default).

If the `send` method returns a trait object, all clients that previously did stuff like:

```rust
    // get res the usual way
    println!("Response: {}", res.status);
```

Would need to be changed to use the appropriate trait method, instead of directly accessing the field. This also implies having to import the trait itself.

Isn't this more invasive than adding a type annotation when declaring a client instance?

I have a branch [client-box-api](https://github.com/mlalic/hyper/tree/client-box-api) on my fork where you can see how it looks with this alternative approach. (It also includes the HTTP/2 commits)

seanmonstar: Hm, you're right, changing that is unfortunate.

------

Here's an idea I haven't fully explored, but what I was thinking about before your pull request arrived in my inbox.

Instead of Client using a RequestFactory, I was thinking that Client would continue to use `client::Request` and `client::Response`. However, those would internally be altered to contain a sort of `HttpMessage`. I imagined it being something like this:

```rust
trait HttpMessage: Read + Write {
    fn get_incoming<T>(&mut self) -> HttpResult<http::MessageHead<T>>;
    fn set_outgoing<T>(&mut self, outgoing: http::MessageHead<T>) -> HttpResult<()>;
}
```

Request and Response would be altered:

```rust
struct Request<W> {
    message: Box<HttpMessage>,
    // ...
}

impl Request<Fresh> {
    pub fn start(self) -> HttpResult<Request<Streaming>> {
        try!(self.message.set_outgoing(MessageHead {
            subject: (self.method, self.uri),
            headers: self.headers,
            // version seems redundant?
        }));
    }
}

// and so forth...
```

The `HttpMessage` implementation would know how to set up its internal Read and Write. For H1, the implementation would adjust itself upon receiving an `Outgoing` to determine which `HttpWriter` to use.

There's currently a `http::Incoming`, that could be adjusted to work for both incoming and outgoing. Possible name is `MessageHead`.

boghison:
+1, been waiting for this ever since HTTP2 came out :)

reem:
@mlalic might be worth rewriting commits now just to get gitcop to shut up.

src/client/mod.rs
```diff
@@ -32,39 +32,120 @@ use status::StatusClass::Redirection;
...
+pub trait HttpRequestFactory {
+    /// The type of the `FreshHttpRequest` that this factory will produce.
+    type RequestType: FreshHttpRequest;
+    /// Creates a new instance of the particular `HttpRequestFactory`.
+    /// This method is a convenience so that the `Client` can still be
+    /// instantiated without providing a factory instance, but with a simple
+    /// `new()` call. This, however, means that all particular
+    /// `HttpRequestFactory` implementations must have a set of default
+    /// parameters with which they can be created.
+    fn new() -> Self;
```
reem: We could get rid of this and just have `HttpRequestFactory` inherit from `Default`.

src/client/mod.rs
```diff
@@ -32,39 +32,120 @@ use status::StatusClass::Redirection;
...
+/// Each particular implementation of the trait can provide a different way
+/// for achieving this, as well as a custom type of request that it produces.
+///
+/// Therefore, by defining the type of the fresh request that it will produce,
+/// it also directly defines the type of the response, since the chain of
+/// `Fresh -> Streaming -> Response` types is always well defined given a
+/// particular `FreshHttpRequest` starting point.
+///
+/// TODO: Might need a better name?
+pub trait HttpRequestFactory {
```
reem: One possible renaming: the trait can be called `HttpProtocol` or just `Protocol` and the types can just be renamed to `Http11` and `Http2`.

mlalic: It sounds good. My feeling is that it might imply a bit more than what the trait currently does (only deals with the client side aspect), but bikeshedding over the name too much doesn't gain us anything...

src/client/mod.rs
mlalic: Nice one.

src/client/mod.rs
mlalic: I see. 

Essentially, we'd end up removing the (currently) private bits of the request and response structs (which are the ones that make them tightly coupled to an HTTP/1.x implementation) and replace them with this new trait object and have the structs delegate the protocol-specific operations to it. So the client keeps its api unchanged, as it still only exposes the same response struct (at least from the perspective of the public fields). 

How about we combine the two approaches... The HTTP response trait that I introduced could be replaced by a struct with the same public fields as the current one. Then, the streaming response trait mandates that the response is exactly that struct and not a generic implementation of the trait. Coupled with the boxed up factory (on the other branch that I mentioned), we'd end up removing the generic parameter and the change in the signature of the send method (at least from the api, if not binary perspective). 
I'm currently on mobile so I can't write any code snippets to get the point across better, but I hope it's understandable what I'm proposing.

What do you think about it? I might be missing something? 

(if you want to take some time to think through alternative designs yourselves, I'm fine with that and we can put this on hold for a while, I just opened the pr since I wanted to get some code out there as a basis for such a discussion as it's always easier when there's something to base it off)

GitCop:

 * Commit: 3ce492a13535b223636212eb59c7f08995b0a791

---

GitCop:

 * Commit: 3ce492a13535b223636212eb59c7f08995b0a791

---

mlalic:
So I've updated the PR based on the discussion so far: rebased the fixes that were made along the way to the corresponding commits and adapted the commit messages to the style guidelines (except the one that needs to be rebased out of the tree anyway!).

The status is that with this, the public API of the client seems to be *very nearly* unaffected.

The return type is now `HttpResult<HttpResponse>`, where `HttpResponse` is a new struct that has the same public API as the previous `Response` struct, *apart from* not being generic. Therefore, all the examples and tests can work without any changes, but if anyone relied on the fact that this public type is generic, that code would break. Obviously, the name of the type is also different, but we could easily call this one `Response` and rename the old one, although in that case a different type of "breakage" occurs, since `Response` was a publicly exposed type, someone might have used it directly, instead of going through the `Client` (though very unlikely).

For now I also removed the `with_connector` method, but we could keep that in and instantiate an HTTP/1.1 request factory in the implementation so as to keep that change also completely transparent.

src/client/mod.rs
reem: Since this is in the client module already, I'm not too concerned with it being overly broad, and I'd be really happy to get rid of `HttpRequestBoxFactory` :P

mlalic:
So, what do you guys (@seanmonstar, @reem) think: should we go down this route? If so, I can bang this into shape with proper test coverage, documentation, error handling, etc.

seanmonstar:
@mlalic sorry, been busy. I'll look it over again today.

seanmonstar:
@mlalic I know Servo uses `Request::with_connector`, cause the XHR spec defines some fun rules slightly different than standard HTTP. /cc @Manishearth 

I have a Pool implementation in #486. That would allow keeping the requests around, and having h1 -> h2 upgrade functionality.

What did you think of my [previous suggestion](https://github.com/hyperium/hyper/pull/448#discussion_r28260410) of keeping Client, Request, and Response as opaque structs, and having the trait be a `HttpMessage`, implemented by http1 and http2? I'd like to think this would ease implementing on the server as well...

mlalic:
@seanmonstar 

> I know Servo uses Request::with_connector, cause the XHR spec defines some fun rules slightly different than standard HTTP

Indeed, currently the establishment of the underlying TCP connection that the HTTP/2 connection would use is somewhat hidden away, but letting it use the `NetworkConnector` abstraction for that should be quite simple. The `NetworkConnector` produces an object that is both `Read` and `Write`, which is all that `solicit`'s HTTP/2 connection requires. The custom connectors that Servo may already have should then work regardless of whether the tcp connection is used for h1 or h2.

> I have a Pool implementation in #486. That would allow keeping the requests around, and having h1 -> h2 upgrade functionality.

The way the pool is implemented right now doesn't lend itself at all to reusing connections for HTTP/2. What is currently pooled are the tcp connections, not HTTP connections. This is virtually the same for h1, but it is not so for h2. This is because in h1, you can simply send the raw request bytes on one of the established tcp streams and you're good; no previously sent requests on that particular stream matter for the raw payload that you transmit over the wire, nor for the parsing of the response payload. In h2, however, the payload you send for a request does depend on previously sent requests - you need to assign a new stream id, hpack encoding/decoding the headers is stateful, flow control windows depend on previously sent data, etc. Therefore, simply pooling the TCP connections does nothing to allow "keep alive"-like usage of an HTTP/2 connection.

Of course, it's doable to pool the actual HTTP 2 connections, by doing what you've done on the connector level within (what is currently named in this pr) the `HttpRequestFactory`. In fact, if acquiring a lock before creating a new request is already deemed acceptable, I could even squeeze my async client into the same api, allowing multiple concurrently issued requests to the same host to use the same HTTP 2 connection, as long as the original factory instance is kept alive and its clones given to new hyper clients (provided that spawning an io thread is also acceptable).

> What did you think of my previous suggestion of keeping Client, Request, and Response as opaque structs, and having the trait be a HttpMessage, implemented by http1 and http2? I'd like to think this would ease implementing on the server as well...

I don’t think I fully understood what you were aiming at with the `HttpMessage` trait. From what I can gather, it should be a type that exposes an interface for queuing HTTP "messages" (`set_outgoing`), as well as for grabbing them off a connection (`get_incoming`). The `Read`/`Write` trait bounds were for reading/writing the body?

The way it looks like it would be used would to implement the request's `Fresh -> Streaming -> Response` transformation would be to first queue the `MessageHead` (i.e. `Fresh -> Streaming`), then stream out the response (calls to the underlying `Write`?) and grab the response head (i.e. `Streaming -> Response`), and finally read out the response (calls to the `Read`). All this has already become encapsulated by the `FreshHttpRequest` and `StreamingHttpRequest` traits. The implementations of these two traits, along with a corresponding factory for HTTP/1.1 and HTTP/2 fully encapsulate all relevant protocol logic.

So, I don’t see the benefit of this alternate approach. It seems it would only shift the actual protocol logic a level lower... Is it that you hope to reuse this `HttpMessage` component to support server-side HTTP/2, as well?

Originally, when you wrote that comment, I thought the suggestion was so that we avoid having type parameters on the `Client`, which I’ve achieved by making the `Response` struct fully independent from the protocol that produces it. (The original post is what nudged me into realizing it was easily doable.)

Perhaps you could flesh out your proposal more because I just might not be fully getting the idea from the previous short description.

seanmonstar:
I'll expore in a branch to show what I mean.

mlalic:
@seanmonstar I have an experiment over on [this branch](https://github.com/hyperium/hyper/compare/master...mlalic:http-message) (forked off the current master HEAD, not this PR) refactoring the HTTP/1.1-specific parts of `Request`/`Response` into an implementation of an `HttpMessage` trait, such as what you described.

Is this sort of what you had in mind for the `HttpMessage` API?

seanmonstar:
Pretty much exactly. As long as such a trait could work for http2, of
course.

mlalic:
@seanmonstar 

The trait does work out quite nicely for HTTP/2 too.

So, here's an attempt at landing the PR, this time with everything polished up with proper docs and tests for the new features. Feedback of course appreciated.

(Side note: the `HttpMessage` API would allow for some nicer tests for the `Client` itself. Right now, it seems very difficult -- or at least extremely kludgy -- to test what the client itself sends, regardless of the protocol. For example, I wanted to write a test which checks whether the client properly handles redirects, as by looking at the code it seems like it always sticks to the original request method, regardless of the redirect status code it sees, but I gave up once I saw that writing a test that checks exactly what requests got sent wasn't as easy as I'd hoped. With the `HttpMessage` API we can just test which messages got dispatched, making it all much smoother.)

seanmonstar:
Sorry, I was camping this weekend (so no Internet), but this looks exciting!

src/error.rs
```diff
@@ -108,6 +114,12 @@ impl From<httparse::Error> for Error {
     }
 }
 
+impl From<Http2Error> for Error {
+    fn from(err: Http2Error) -> Error {
+        Error::Http2(err)
```
seanmonstar: Could you add a `from_and_cause!` case to the tests below, to keep the coverage happy.

src/http11.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        let stream = self.stream.take().unwrap();
+        let mut stream = BufReader::new(stream);
+
+        let head = try!(http::parse_response(&mut stream));
+        let raw_status = head.subject;
+        let headers = head.headers;
+
+        /*
```
seanmonstar: Was this commented out originally? Seems it can be killed.

src/http11.rs
```diff
@@ -0,0 +1,311 @@
...
+
+        Ok(ResponseHead {
+            headers: headers,
+            raw_status: raw_status,
+            version: head.version,
+        })
+    }
+
+    fn close_connection(&mut self) -> ::Result<()> {
+        try!(self.get_mut().close(Shutdown::Both));
```
seanmonstar: This could just be `self.get_mut().close(...)`, right?

src/http11.rs
```diff
@@ -0,0 +1,311 @@
...
+    RequestHead,
+    ResponseHead,
+};
+use header;
+use version;
+
+/// An implementation of the `HttpMessage` trait for HTTP/1.1.
+#[derive(Debug)]
+pub struct Http11Message {
+    stream: Option<Box<NetworkStream + Send>>,
```
seanmonstar: Instead of having 3 options (and the possible panic! states below if all are `None`), how about some sort of `MessageState` enum, with the 3 variants. I think that would be clearer that these determine the state of the message, whereas reading the 3 options at first left me confused. Additionally, we could remove the `panic!`s where all 3 options are None, since the enum will enforce for us the state.

src/http2.rs
```diff
@@ -0,0 +1,563 @@
...
+}
+
+/// An implementation of the `HttpMessage` trait for HTTP/2.
+///
+/// Relies on the `solicit::http::SimpleClient` for HTTP/2 communication. Adapts both outgoing and
+/// incoming messages to the API that `hyper` expects in order to be able to use the message in
+/// the `hyper::client` module.
+pub struct Http2Message<S> where S: CloneableStream {
+    client: SimpleClient<Http2Stream<S>>,
+    request: Option<Http2Request>,
```
seanmonstar: Probably similar here: if these options are for tracking state, a `MessageState` enum would probably be clearer.

src/http2.rs
```diff
@@ -0,0 +1,563 @@
...
+        // Headers are moved into `hyper`'s `Headers` collection; pseudo-headers
+        // are ignored (status has already been extracted).
+        // TODO: Coalesce possible headers with the name key into a single vector.
+        //       `solicit` returns the raw original header list and lets the clients
+        //       figure out how they would like to transform it.
+        for (name, value) in response.headers.into_iter() {
+            // Unfortunately, `hyper` wants its header names to be strings...
+            // We just panic if it ends up being an invalid utf8 string, but no
+            // copies are made.
+            headers.set_raw(String::from_utf8(name).unwrap(), vec![value]);
```
seanmonstar: The spec says that header names are valid tokens, which means they should be valid utf8. Is that not the case in http2? Either way, I'm not sure an `unwrap` should be used.

Also, in Http11, the `Set-Cookie` header is sent multiple times if there are multiple cookies. Is that different in Http2?

src/error.rs
mlalic: Done.

src/http11.rs
mlalic: Indeed, this is still done in the `Response` so as not to duplicate it in all `HttpMessage` impls. Removed.

src/http11.rs
mlalic: If you mean without the `try!`, then no. The error types are mismatched: `close_connection` requires `hyper::error::Error` and the `close` returns an `std::io::Error`. Since the types aren't automatically converted (i.e. without a `From::from` call), to me it looked nicest to go with a `try!`.

src/http11.rs
mlalic: All right, say there's an enum:

```rust
enum MessageState {
    Idle(Box<NetworkStream + Send>),
    Writing(HttpWriter<BufWriter<Box<NetworkStream + Send>>>),
    Reading(HttpReader<BufReader<Box<NetworkStream + Send>>>),
}
```

Now in `set_outgoing(&mut self, ...)`, the internal state of the message needs to transition to `MessageState::Writing` and the writer needs to be based on the stream originally wrapped in the `Idle` variant. I'm not sure how the `Box<NetworkStream>` should be moved into a different variant here.

`set_outgoing` takes `&mut self`, so neither the full state enum nor the `Box` within the `Idle` variant can be moved. The only thing I can think of around that would be to use `mem::replace`, but it'd still require something (akin to `None`) to be put in place of the original state while the stream is juggled into the correct Writer.

This `None`-like state, though, would still imply that it's technically possible to lose the stream along the way, so it's only a slight improvement...

Is there some better approach to implementing the transition which would avoid this `None` state altogether?

src/http2.rs
mlalic: Here, on the other hand, the state enum works out just fine. I suppose because there's no state that needs to be transferred between the variants when a transition happens...

src/http2.rs
mlalic: The names should still be valid utf8 ([8.1.2 HTTP Header Fields](http://http2.github.io/http2-spec/#HttpHeaders)) on the HTTP layer, that's true. (HPACK is what treats both the name and value as an opaque octet sequence.) So, indeed, if the name cannot be decoded as utf8, the response is malformed. This is now fixed (along with a test).

As for multiple `SetCookie` headers -- the situation is unchanged: multiple cookies correspond to multiple `set-cookies` headers. That said, I was hoping that `hyper` would correctly handle this itself...

What I see now, though, is that it effectively fakes it by having the *value* of the `Set-Cookie` entry in the `header::Headers` struct contain a CRLF delimiter and continue on to the next `Set-Cookie` header.

```rust
    let cookies = header::SetCookie(vec![
        cookie::Cookie::new("foo".to_owned(), "bar".to_owned()),
        cookie::Cookie::new("baz".to_owned(), "quux".to_owned())
    ]);
    // Now the following is true:
    cookies.name() == "Set-Cookie"
    cookies.value_string() == "foo=bar; Path=/\r\nSet-Cookie: baz=quux; Path=/"
```

That hack works when sending the header over HTTP/1.1, but not HTTP/2, as the CRLF is meaningless. I've worked around it in the `prepare_headers` function to special case the `Set-Cookie` and extract the cookies that it contains into separate HTTP/2 headers (although since for now there's only client support, it shouldn't actually get triggered, as clients aren't supposed to send that header).

On the other hand, when parsing a response with multiple `set-cookie` headers returned by an HTTP/2 server, it seems that `hyper`'s `Headers::from_raw` method ends up returning a headers collection with a `SetCookie` header that correctly contains all the cookies, as verified by a new unit test... Should `from_raw` not be relied on to do the right thing?

seanmonstar: `Headers::from_raw` will do the right thing. However, `headers.set_raw()` will *override* whatever value was there before, like any setter.

The cookie formatting hack was indeed a nasty hack. We've been trying to think of a way to allow headers to indicate they output over multiple lines, but keeping that invisible from 99% of headers (since they don't need it) is difficult. I'd like to try to figure that out, such that a `Header` can return a `Encoder` that implements `Iterator`, and also be able to pass that to solicit, such that the bytes are written only once through hpack, instead of allocating a temporary vector beforehand.

seanmonstar:
Just curious, from the example: `"Host": "http2bin.org,http2bin.org",`, is it supposed to have the host doubled like that?

src/http11.rs
seanmonstar: Fair enough. It could probably be done inside a method on the state enum, but don't worry about it. I might look at if it's possible to keep state similar to the `WriteStatus` that `Request` uses.

seanmonstar:
@mlalic woo! thanks for this herculean effort!

src/http2.rs
mlalic: That would be a nice improvement to aim for, for sure.

mlalic:
Happy to do it! Glad we've worked out the initial integration; we can take it from here and tweak/improve things incrementally, of course... Should be less of a big-bang than this :)

mlalic:
> Just curious, from the example: `"Host": "http2bin.org,http2bin.org",`, is it supposed to have the host doubled like that?

`hyper` probably includes the `Host` header and the HTTP/2 client includes the `:authority` pseudo-header so that's probably where it ends up looking doubled.

`:authority` "should" be used in HTTP/2 requests, but `host` is still fine, so we can tweak it that the `Http2Message` doesn't set any of those unless they're given to it in the message head? (Like HTTP/1.1 at least one of those headers is still required for every request, so it means that the `Request` always needs to set it...)

PR #457, opened 2015-04-15
fix(server): Use thread::spawn instead of thread::scoped.

reem:
Should build on both nightly and beta.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ cache:
       - target
 
 script:
-    - cargo build
-    - cargo test
+    - cargo build --features nightly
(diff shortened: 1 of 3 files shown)


reem:
This is a breaking change of sorts, since nightly consumers will have to configure hyper to use the nightly feature.

reem:
The build needs to be run with `--features nightly` if a nightly rustc version is being used, but right now no feature is passed on travis.

seanmonstar:
I guess this needs travis.yml to be updated to pass the feature for `cargo build` and `cargo test` as well?

PR #458, opened 2015-04-15
fix(server): join on thread when Listening drops

### LINKED ISSUE #447: Main thread shouldn't panic when child threads panic
When running the following code, the panic in the child thread causes the server to crash:
```
extern crate hyper;
 
use hyper::Server;
use hyper::server::Request;
use hyper::server::Response;
use hyper::net::Fresh;
 
fn panic(_: Request, _: Response<Fresh>) {
    panic!("test")
}
 
fn main() {
    Server::http(panic).listen("127.0.0.1:3000").unwrap();
}
```
Here is the output:
```
thread '<unnamed>' panicked at 'test', src/main.rs:11
thread '<unnamed>' panicked at 'child thread None panicked', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libstd/thread/mod.rs:715
thread '<main>' panicked at 'child thread None panicked', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libstd/thread/mod.rs:715
An unknown error occurred

```
This is a fairly serious reliability issue, as a server should never go down because of its child threads.

seanmonstar:
Closes #447

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -107,7 +107,7 @@ S: NetworkStream + Clone + Send> Server<'a, H, L> {
         let guard = thread::spawn(move || pool.accept(work, threads));
 
         Ok(Listening {
-            _guard: guard,
+            _guard: Some(guard),
             socket: socket,
         })
     }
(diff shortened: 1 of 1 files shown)


src/server/mod.rs
```diff
@@ -176,13 +176,24 @@ where S: NetworkStream + Clone, H: Handler {
...
+                let _ = guard.join();
+            },
+            None => ()
+        }
+    }
+}
+
 impl Listening {
     /// Stop the server from listening to its socket address.
     pub fn close(&mut self) -> HttpResult<()> {
```
reem: This function should probably take the guard for now, so that it at least lets you close a program running a server.

seanmonstar: ah right

seanmonstar:
updated

src/server/mod.rs
```diff
@@ -176,16 +176,28 @@ where S: NetworkStream + Clone, H: Handler {
...
     #[cfg(not(feature = "nightly"))]
-    _guard: JoinHandle,
+    _guard: Option<JoinHandle>,
     /// The socket addresses that the server is bound to.
     pub socket: SocketAddr,
 }
 
+impl Drop for Listening {
+    fn drop(&mut self) {
+        match self._guard.take() {
```
reem: much shorter to just write `self._guard.take().map(|g| let _ = g.join());`

seanmonstar:
updated again

PR #460, opened 2015-04-16
fix(rustup): add type parameter to JoinHandle

fhartwig:
`JoinHandle` now has a type parameter. Setting it to `()` gives the old behaviour.


fhartwig:
This fixes the build on nightly, but is obviously going to break it on beta :disappointed:

fhartwig:
Ok, sorry, I'm an idiot.

PR #463, opened 2015-04-16
refactor(error): remove redundant parts of error names

seanmonstar:
The old names followed the old style of including the module name and
"Error" in each variant. The new style is to refer to an error from its
owning module, and variants are now scoped to their enum, so there's no
need to include the enum name in the variant name.

BREAKING CHANGE: The terms `Http` and `Error` have been removed from the Error
  type and its variants. `HttpError` should now be accessed as `hyper::Error`,
  and variants like `HttpIoError` should be accessed as `Error::Io`.

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -3,7 +3,7 @@ use std::str::FromStr;
 use url::Url;
 use url::ParseError as UrlError;
 
-use error::HttpError;
+use Error;
 
 /// The Request-URI of a Request's StartLine.
 ///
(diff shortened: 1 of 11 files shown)


src/uri.rs
```diff
@@ -3,7 +3,7 @@ use std::str::FromStr;
 use url::Url;
 use url::ParseError as UrlError;
 
-use error::HttpError;
+use error::Error;
```
reem: could be just `use Error;`

reem:
LGTM except for that nit

seanmonstar:
Just holding on merging till I've completed some other breaking changes, and they can all be part of 0.4

PR #477, opened 2015-04-24
fix(server): remove type argument

daniellandau:
Doesn't compile on latest nightly without this.


daniellandau:
My bad, I didn't have the latest nightly after all.

Issue #471, opened 2015-04-21
Exposing SslContext for flexibility (e.g. client-side certs, etc)


mikedilger:
Currently, `net::HttpListener::https()` takes a certificate and a key, and sets up an SslContext with a lot of assumptions.  For example, it sets the DEFAULT cipher list.  It also sets SSL_VERIFY_NONE.
This is called by `Server::listen_threads()`.   I'm trying to make changes that allow the caller to pass in an SslContext.

Currently the SSL configuration is declared when the Server is created, by passing a certificate and a key to `Server::https()`.  Unfortunately this scheme will not direclty work with a passed in `SslContext`, because if an `SslContext` is setup in the `Server` struct, it *moves* when passed on to HttpListener::https(), causing a "use of partially moved value" error.   This cannot be solved via `clone()` because `SslContext` does not implement `clone()`.

If `Server::listen_threads()` accepted the SSL information at that point (rather than owning it in it's struct), this would cause a lot of breaking changes.

OTOH, If `HttpListener` kept a reference to an `SslContext` (instead of owning it), it would propagate another annoying lifetime parameter over a lot of code.

I don't see a clearly preferable way of implementing this. Given a preferred direction, I'd be happy to make the changes necessary.

mikedilger:
Another option I thought of:  `HttpListener` could get a separate `listen` method (where `http` and `https` would do everything except the `bind` step),  and then a new method could be created that gets a mutable reference to the `SslContext`, to be called prior to `listen`.

Also I forgot to cc @seanmonstar

seanmonstar:
@mikedilger I'm not sure I follow. In `listen_threads`, the Server is taken by value. You can move the context out of the Server in this method.

mikedilger:
`self.with_listener(listener, threads)` requires both self (Server) and listener.  If the context is moved, then 'self' is no longer available.  See https://github.com/mikedilger/hyper/tree/ssl_context

seanmonstar:
Ohhh, I see what you mean. Ok, so the `with_listener` method takes `self`, but it only needs to access `self.handler`. What about creating a private free function `with_listener`, that the method calls as `with_listener(self.handler, listener, threads)`. Then, `listen_threads` can skip the internal method, and just move out both the `ssl` and the `handler` in the same function. Make sense?

mikedilger:
I'm glad I asked, because I hadn't thought of that.

I did this in 3 commits.  See PR #479

PR #479, opened 2015-04-26
Ssl context

mikedilger:
See #471

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -164,6 +164,11 @@ impl HttpListener {
         try!(ssl_context.set_certificate_file(cert, X509FileType::PEM).map_err(lift_ssl_error));
         try!(ssl_context.set_private_key_file(key, X509FileType::PEM).map_err(lift_ssl_error));
         ssl_context.set_verify(SSL_VERIFY_NONE, None);
+        HttpListener::https_with_context(addr, ssl_context)
+    }
+
+    /// Start listening to an address of HTTPS using the given SslContext
+    pub fn https_with_context<To: ToSocketAddrs>(addr: To, ssl_context: SslContext) -> io::Result<HttpListener> {
         Ok(HttpListener::Https(try!(TcpListener::bind(addr)), Arc::new(ssl_context)))
     }
 }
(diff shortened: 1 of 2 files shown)


src/net.rs
```diff
@@ -164,6 +164,11 @@ impl HttpListener {
         try!(ssl_context.set_certificate_file(cert, X509FileType::PEM).map_err(lift_ssl_error));
         try!(ssl_context.set_private_key_file(key, X509FileType::PEM).map_err(lift_ssl_error));
         ssl_context.set_verify(SSL_VERIFY_NONE, None);
+        HttpListener::https_using_context(addr, ssl_context)
+    }
+
+    /// Start listening to an address of HTTPS using the given SslContext
+    pub fn https_using_context<To: ToSocketAddrs>(addr: To, ssl_context: SslContext) -> io::Result<HttpListener> {
```
reem: The convention is to use "with" for these sorts of constructors, so `https_with_context` may be appropriate.

src/server/mod.rs
```diff
@@ -50,6 +51,13 @@ pub mod response;
 
 mod listener;
 
+#[derive(Debug)]
+enum SslConfig<'a> {
+    CertAndKey(&'a Path, &'a Path),
+    Context(SslContext),
+    None
```
reem: better to use `Option<SslConfig<'a>>`

src/server/mod.rs
```diff
@@ -117,23 +134,27 @@ H: Handler + 'static,
...
-        let handler = self.handler;
+    pub fn with_listener(self, listener: L, threads: usize) -> HttpResult<Listening> {
+        with_listener(self.handler, listener, threads)
+    }
+}
 
-        debug!("threads = {:?}", threads);
-        let pool = ListenerPool::new(listener.clone());
-        let work = move |mut stream| handle_connection(&mut stream, &handler);
+fn with_listener<'a, H, L>(handler: H, mut listener: L, threads: usize) -> HttpResult<Listening>
```
reem: `'a` appear unused here

mikedilger:
I've amended the commits and force-pushed the amended branch.  I think I've addressed your concerns.

seanmonstar:
Thanks!

PR #481, opened 2015-04-26
Add tests for headers

pyfisch:
Already one bug discovered, I commented out the testcase, it can be fixed later.


pyfisch:
bug number 2 found.

pyfisch:
Another bug found

pyfisch:
This PR is complete now.

seanmonstar:
Wow, this is looking awesome (viewing on my phone atm)

reem:
@pyfisch should we try to move existing tests into this infrastructure, or is it meant to only be for certain types of tests? Agreed with @seanmonstart that this looks pretty awesome.

pyfisch:
We can move (hopefully) all header tests into the new infrastructure. The macro just supports parsing and formatting, some headers have different tests, but we should be able to write them also inside the new test section. It should be also possible to add `bench_header()` to the section, but I am not sure if we need special bench testcases, I think we could just bench the normal test cases.

PR #482, opened 2015-04-27
fix(net): switch to Tlsv1_2 for default ssl method

softprops:
This change substitutes the previous insecure [default ssl method](https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L90-L91) with a [more secure default](https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L97-L99). Without this I was blocked writing a client interface for a secure server supporting the more secure tls method and there wasn't a way I could find to provide the ssl method used but the client.

A more thorough changeset would probably involve something like adding a new ssl provider interface to then network connector and network listener interfaces that allow's for user defined ssl contexts, but I think that would be a breaking change. This was a much smaller change, makes the defaults more secure, and shouldn't be a breaking change, so I went with this instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,13 +17,16 @@ httparse = "*"
 log = ">= 0.2.0"
 mime = "*"
 num_cpus = "*"
-openssl = "*"
 rustc-serialize = "*"
 time = "*"
 unicase = "*"
 url = "*"
 traitobject = "*"
 typeable = "*"
+
+[dependencies.openssl]
+version = "*"
+features = ["tlsv1_2"]
 
 [dev-dependencies]
 env_logger = "*"
--- a/src/net.rs
+++ b/src/net.rs
@@ -8,7 +8,7 @@ use std::path::Path;
 use std::sync::Arc;
 
 use openssl::ssl::{Ssl, SslStream, SslContext, SSL_VERIFY_NONE};
-use openssl::ssl::SslMethod::Sslv23;
+use openssl::ssl::SslMethod::Tlsv1_2;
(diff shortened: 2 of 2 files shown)


seanmonstar:
This was actually using `Sslv23` by default, which is [Ssl v2, v3, and Tls v1](https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L88). It seemed like by default it would make the most amount of the web work. If I'm wrong, I'd love to know!

However, I recently merged in #479, which allows you to create a `SslContext` yourself. Would that serve your purposes?

(side note: I'd consider a change like this to be `feat(server)`, and probably a breaking change, since servers upgrading would silently be dropping Ssl v2 and v3 support.)

softprops:
I guess the server communicating with rejected anything less than [https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L99]. I say the other pull but it looked like it only addressed the server side of the issue. The client side connectors still have the ssl method used [hard coded](https://github.com/hyperium/hyper/blob/master/src/net.rs#L307) :/

seanmonstar:
Oh, I see. This also affects `HttpConnector` (I misread and though this was server only).

If we make this change, does this mean the Client no longer works (by default) against servers that only use SSL instead of TLS?

softprops:
Yea thinking about making this backwards compatible would probably involve some additional interfaces that allow you to specify which ssl method to use, [this](http://serverfault.com/a/179139) may help answer your question. I'm open to help take that on. I imaging it would be similar to the change to http listeners translated for network connectors

softprops:
going to close this in favor of a more flexible solution for NetworkConnectors in the future.

Issue #468, opened 2015-04-17
Add example usage for each Header


seanmonstar:
The header types docs could show an example usage on each of their respective pages.

pyfisch:
This is solved by #503

seanmonstar:
Ah, when I meant example usage, I meant showing in a rust code fench how to
create a header to to send in a Request/Response.

PR #485, opened 2015-04-27
feat(headers): Implement Content-Language header field

### LINKED ISSUE #475: Implement Content-Language header
https://tools.ietf.org/html/rfc7231#section-3.1.3.2

This will be useful for Servo.

pyfisch:
Closes #475


src/header/common/content_language.rs
```diff
@@ -0,0 +1,22 @@
...
+    #[doc=""]
+    #[doc="The `Content-Language` header field describes the natural language(s)"]
+    #[doc="of the intended audience for the representation.  Note that this"]
+    #[doc="might not be equivalent to all the languages used within the"]
+    #[doc="representation."]
+    #[doc=""]
+    #[doc="# ABNF"]
+    #[doc="```plain"]
+    #[doc="Content-Language = 1#language-tag"]
+    #[doc="```"]
```
seanmonstar: Could you add a usage example, per #468?

Or, I wonder... would the `header!` macro be able to generate a usage example automatically? Then we'd suddenly have much better docs...

src/header/shared/language.rs
```diff
@@ -0,0 +1,40 @@
+use std::str::FromStr;
+use std::fmt;
+
+/// A language tag.
+/// See http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.10
+///
+/// Note: This is no complete language tag, it should be replaced with
+/// github.com/pyfisch/rust-language-tag once it is ready.
+#[derive(Clone, PartialEq, Debug)]
+pub struct Language{
```
seanmonstar: nit: space between the name and bracket

src/header/shared/language.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl FromStr for Language {
+    type Err = ();
+    fn from_str(s: &str) -> Result<Language, ()> {
+        let mut i = s.split("-");
+        let p = i.next();
+        let s = i.next();
+        match (p, s) {
+            (Some(p),Some(s)) => Ok(Language{primary: p.to_string(),
```
seanmonstar: Mind cleaning this up also? Spaces don't cost use anything. Such as:

```rust
(Some(p), Some(s)) => Ok(Language {
    primary: p.to_string(),
    sub: Some(s.to_string())
}),
```

seanmonstar: (I know it was just copied from the original AcceptLanguage header.)

src/header/common/content_language.rs
pyfisch: Generating it automatically is not possible with macros currently. But it should be basically a copy & paste job: copy the values from the tests to the docs. Or do you want a program example? (We could create samples for classes of headers)

pyfisch:
fixed nits

PR #489, opened 2015-04-30
fix(header): exporting test_header! macro

Byron:
That way, third-parties can keep using the `header!` macro, which
now requires the `test_header!` macro as well.

Considering the `header!` macro is exported and part of the public
API, it was not desired to prevent it's usage in any way.


pyfisch:
+1

(The `test_header!` macro is not required, you can also write your own tests inside the test section)

PR #491, opened 2015-04-30
feat(headers): Add If-Range header

### LINKED ISSUE #388: Implement If-Range header parsing
Spec: `http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.27`

pyfisch:
Closes #388 

cc @jdm


src/header/common/mod.rs
```diff
@@ -286,6 +289,38 @@ macro_rules! header {
             $($tf)*
         }
     };
+
+    // Enum header
```
seanmonstar: Is this supposed to be part of this PR?

pyfisch: No.

pyfisch:
Updated pr, enum headers do not work as macro, so removed the code.

PR #492, opened 2015-04-30
fix(header): make test_module of header! optional

### LINKED ISSUE #490: Breaking change in the header! macro not documented
I updated to 0.3.15 and my code is now failing due to changes in `header!` macro:

```
src/lib.rs:250:43: 250:44 error: unexpected end of macro invocation
src/lib.rs:250 header!{ (TenantId, "TenantId") => [String] }
```

This is most likely due to commits https://github.com/hyperium/hyper/commit/efd6c96a3cf152b49b672ac87e225e45ab87a477 and https://github.com/hyperium/hyper/commit/a27e6812b9db98fd8f375f2b4b497e4bafc1871e which are not listed as breaking changes. 

With the new `header!` implementation It seems I must also add a dummy test function when implementing the header which is, IMHO, an unnecessary complication and makes the `header!` macro harder to use. E.g. for my example above: 

```
header!{ (TenantId, "TenantId") => [String] test1 {} }
```

Is this the intended behaviour?

seanmonstar:
Closes #490

/cc @pyfisch @maximih

### DIFF
--- a/examples/headers.rs
+++ b/examples/headers.rs
@@ -1,0 +1,13 @@
+#![deny(warnings)]
+
+#[macro_use]
+// TODO: only import header!, blocked by https://github.com/rust-lang/rust/issues/25003
+extern crate hyper;
+
+// A header in the form of `X-Foo: some random string`
+header! {
+    (Foo, "X-Foo") => [String]
+}
+
+fn main() {
+}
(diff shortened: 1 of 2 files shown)


pyfisch:
:+1: but you should not make `deref!` private.

seanmonstar:
Oh Woops, that was me experimenting.

PR #493, opened 2015-05-01
refactor(hyper): Fix a few nits

pyfisch:


src/header/common/pragma.rs
```diff
@@ -33,7 +33,6 @@ impl Header for Pragma {
         parsing::from_one_raw_str(raw).and_then(|s: String| {
             let slice = &s.to_ascii_lowercase()[..];
             match slice {
-                "" => None,
```
seanmonstar: an empty string is not `None`?

src/header/shared/httpdate.rs
```diff
@@ -46,12 +46,7 @@ impl FromStr for HttpDate {
 
 impl Display for HttpDate {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let tm = self.0;
-        let tm = match tm.tm_utcoff {
```
seanmonstar: Unless this was fixed in the time crate, `to_utc` is slow even if the time is already in UTC.

pyfisch: This was fixed in time crate: https://github.com/rust-lang/time/blob/fba17ed10a4774bb4666c79d58f7cec3ee858865/src/lib.rs#L770

src/header/common/pragma.rs
pyfisch: Fixed in another commit. It should always be a `None` for single value headers, so I added it to `from_one_raw_str()`. List headers already check for empty values in a similiar way. This variant is now obsolete.

PR #494, opened 2015-05-01
chore(travis): Enable test coverage generation

### LINKED ISSUE #44: generate test coverage
it'd be awesome if there was a Rust code coverage tool already. I think there's work around being compatible with kcov, but something that modifies the source and inserts counters would be awesomer.

pyfisch:
Generates a test coverage using kcov on travis. The coverage is only
generated for beta builds. kcov requires using sudo, so beta builds
can't run on container infrastructure anymore.

Closes #44 
[Coverage Status](https://coveralls.io/r/pyfisch/hyper?branch=kcov)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,12 @@ matrix:
     include:
         - rust: nightly
           env: FEATURES="--features nightly"
+          sudo: false
         - rust: nightly
           env: FEATURES="--features nightly" BENCH=true
+          sudo: false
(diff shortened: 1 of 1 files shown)


seanmonstar:
Oh excellent!

There should be a way to install locally to the user, instead needing sudo, right?

pyfisch:
I guess it will be complicated. We have to use `apt-get` to install some packages.

PR #499, opened 2015-05-02
feat(headers): Parse Upgrade header protocols further

### LINKED ISSUE #480: Reminder: Upgrade header Protocols should support version
As defined in http://tools.ietf.org/html/rfc7230#section-6.7 a protocol consists of a name and an optional version token.

pyfisch:
Parses protocols into a name and a value part matching the RFC.
An enum contains all registered or known protocols, but contains
an extension variant.

Closes #480

BREAKING CHANGE: Upgrade header Protocol changed.

### DIFF
--- a/src/header/common/upgrade.rs
+++ b/src/header/common/upgrade.rs
@@ -1,8 +1,6 @@
-use std::fmt;
+use std::fmt::{self, Display};
 use std::str::FromStr;
 use unicase::UniCase;
-
-use self::Protocol::{WebSocket, ProtocolExt};
 
 header! {
     #[doc="`Upgrade` header, defined in [RFC7230](http://tools.ietf.org/html/rfc7230#section-6.7)"]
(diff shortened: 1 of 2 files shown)


src/header/common/upgrade.rs
```diff
@@ -34,43 +32,93 @@ header! {
...
-impl FromStr for Protocol {
+impl FromStr for ProtocolName {
     type Err = ();
+    fn from_str(s: &str) -> Result<ProtocolName, ()> {
+        Ok(match s {
+            "HTTP" => ProtocolName::Http,
+            "TLS" => ProtocolName::Tls,
+            "h2c" => ProtocolName::H2c,
+            _ => {
+                if let UniCase("websocket") = UniCase(s) {
```
seanmonstar: This does pattern matching, which will skip UniCase::eq.

src/header/common/upgrade.rs
```diff
@@ -34,43 +32,93 @@ header! {
...
 }
 
-impl FromStr for Protocol {
+impl FromStr for ProtocolName {
     type Err = ();
+    fn from_str(s: &str) -> Result<ProtocolName, ()> {
+        Ok(match s {
+            "HTTP" => ProtocolName::Http,
+            "TLS" => ProtocolName::Tls,
+            "h2c" => ProtocolName::H2c,
```
seanmonstar: These name are case-sensitive?

src/header/common/upgrade.rs
```diff
@@ -34,43 +32,93 @@ header! {
...
+                    ProtocolName::Unregistered(s.to_string())
+                }
+            }
+        })
+    }
+}
+
+impl Display for ProtocolName {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Display::fmt(match *self {
```
seanmonstar: Technically could be f.write_str... But might not be consistent with elsewhere in hyper. 

Might be worth a separate issue to change those, as it is a little faster.

src/header/common/upgrade.rs
```diff
@@ -34,43 +32,93 @@ header! {
...
+    /// The protocol identifier
+    pub name: ProtocolName,
+    /// The optional version of the protocol, often in the format "DIGIT.DIGIT" (e.g.. "1.2")
+    pub version: Option<String>,
+}
+
+impl Protocol {
+    /// Creates a new Protocol with the given name and version
+    pub fn new(name: ProtocolName, version: Option<String>) -> Protocol {
+        Protocol {name: name, version: version}
```
seanmonstar: Nit: Add spaces surrounding brackets.

src/header/common/upgrade.rs
pyfisch: That's true, `if let` is not needed anyway here.

src/header/common/upgrade.rs
pyfisch: I think so. The [RFC](http://tools.ietf.org/html/rfc7230#section-6.7) does not say that they are case-insensitive. WebSocket spec says, that `websocket` must be matched case-insensitive.

pyfisch:
Updated.

PR #506, opened 2015-05-05
feat(error): add Ssl variant to hyper::Error

### LINKED ISSUE #483: Add an Ssl variant to Error
And change `NetworkConnector::connect` to return a `HttpError` instead of an `IoError`. The SSL errors are being boxed into an `ErrorKind::Other`, currently, which is losing information.

seanmonstar:
The errors from openssl were previously boxed into a
Box<std::error::Error>, which lost some specifics and made it difficult
to match against. To solve this, an `Ssl` variant is added to the
`Error` enum of hyper, and is returned when openssl returns specific
errors.

Closes #483

BREAKING CHANGE: Adds a variant to `hyper::Error`, which may break any
exhaustive matches.

cc @jdm

### DIFF
--- a/benches/client.rs
+++ b/benches/client.rs
@@ -79,7 +79,7 @@ struct MockConnector;
 
 impl net::NetworkConnector for MockConnector {
     type Stream = MockStream;
-    fn connect(&mut self, _: &str, _: u16, _: &str) -> io::Result<MockStream> {
+    fn connect(&mut self, _: &str, _: u16, _: &str) -> hyper::Result<MockStream> {
         Ok(MockStream::new())
     }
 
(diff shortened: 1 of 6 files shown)


jdm:
Oh, this sounds like it could be much more useful than https://github.com/servo/servo/blob/master/components/net/http_loader.rs#L140 !

reem:
I wish we could pull the `io::ErrorKind` trick of having an unstable error variant to make this not a breaking change, but there's no good analogue for stability markers.

seanmonstar:
@reem yea, i'd like that too, but I assume there will be a solution by the time hyper hits 1.0.

PR #511, opened 2015-05-06
feat(server): dropping a Response will write out to the underlying stream

seanmonstar:


Ryman:
The `drop` tests don't seem to actually test the body itself is flushed?

The branching on `PhantomData` looks like a neat technique!

src/server/response.rs
```diff
@@ -113,46 +107,206 @@ impl<'a> Response<'a, Fresh> {
...
+enum Body {
+    Chunked,
+    Sized(u64),
+}
+
+type StatusFresh = PhantomData<Fresh>;
+
+impl<'a, T: Any> Drop for Response<'a, T> {
+    fn drop(&mut self) {
+        if (&self._writing as &Any).is::<StatusFresh>() {
```
reem: This is inspired.

reem: That said we can get better perf by just doing `TypeId::of::<W>() == TypeId::of::<StatusFresh>()`, which will get the branch optimized away because monomorphization.

seanmonstar:
@Ryman what do you mean? That `flush` was called on the MockStream?

src/server/response.rs
```diff
@@ -26,10 +30,10 @@ pub struct Response<'a, W = Fresh> {
     // The outgoing headers on this response.
     headers: header::Headers,
 
-    _marker: PhantomData<W>
+    _writing: PhantomData<W>
```
reem: should lose the `_` now that it is used

Ryman:
@seanmonstar I mean that the tests seem to only check with a zero byte body in the response, I wonder if it's a benefit to include a test that actually ensures a body is dealt with appropriately?

seanmonstar:
@Ryman ah sure. Though, since it's a chunked the response, that last chunk written before flushing **must** be the `0\r\n\r\n` chunk, which the tests *are* checking for.

PR #515, opened 2015-05-08
fix(header): export `tm!` macro

Byron:
It is used in the `header!` macro, which is exported as well.
If `tm!` is not exported, external clients cannot user otherwise
exported macros anymore who are dependent on `tm!`.

### DIFF
--- a/src/header/common/mod.rs
+++ b/src/header/common/mod.rs
@@ -98,6 +98,7 @@ macro_rules! deref(
     }
 );
 
+#[macro_export]
 macro_rules! tm {
     ($id:ident, $tm:ident{$($tf:item)*}) => {
         #[allow(unused_imports)]


pyfisch:
Probably this should have a better name if it is exported.

Byron:
Indeed, I just did a '*I need this real quick*' and ignored the fact that I have no clue what it does. `tm` seems like too short for a rather global namespace without any hygiene.
I will leave the PR up anyway though, until the topic has received the attention it (probably) deserves.

seanmonstar:
I'm thinking along the lines of what I did for the [mime!](https://github.com/hyperium/mime.rs/blob/master/src/lib.rs#L98) macro, until https://github.com/rust-lang/rust/issues/25003 is fixed.

seanmonstar:
See #523

Byron:
Thanks. The fix works for me.

PR #518, opened 2015-05-09
fix(client): keep the underlying connector when setting an SSL verifier

### LINKED ISSUE #495: Client: Setting a custom SSL verifier discards previous Connector
When the `Client::set_ssl_verifier` method is called, it will discard any previously existing `Connector` instance that the `Client` was instantiated to use. This wasn't a problem before, when the connectors were largely stateless, but with the introduction of a `Pool` connector, this means that the pool's config will be dropped (replaced with the default one) and the client will use a brand new set of connections, instead of the originally intended ones.

mlalic:
The client no longer drops the old `Connector` instance, along with its
entire context (such as the settings and pooled connections in the case
of a `Pool` connector); rather, it passes the SSL verifier on to the
`Connector` so that it uses it for any future connections that it needs
to establish.
    
A regression test is included.
    
Closes #495

### DIFF
--- a/benches/client.rs
+++ b/benches/client.rs
@@ -8,7 +8,7 @@ use std::fmt;
 use std::io::{self, Read, Write, Cursor};
 use std::net::SocketAddr;
 
-use hyper::net;
+use hyper::net::{self, ContextVerifier};
 
 static README: &'static [u8] = include_bytes!("../README.md");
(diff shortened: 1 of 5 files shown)


seanmonstar:
Excellent, thank you!

PR #527, opened 2015-05-12
feat(server): check Response headers for Connection: close in keep_alive loop

seanmonstar:
BREAKING CHANGE: Usage of Response.deconstruct() and construct() now use
  a &mut Headers, instead of the struct proper.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -33,7 +33,7 @@ pub use net::{Fresh, Streaming};
 
 use Error;
 use buffer::BufReader;
-use header::{Headers, Expect};
+use header::{Headers, Expect, Connection};
 use http;
 use method::Method;
 use net::{NetworkListener, NetworkStream, HttpListener};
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -204,11 +204,27 @@ where S: NetworkStream + Clone, H: Handler {
         }
 
         keep_alive = http::should_keep_alive(req.version, &req.headers);
-        let mut res = Response::new(&mut wrt);
-        res.version = req.version;
-        handler.handle(req, res);
-        debug!("keep_alive = {:?}", keep_alive);
+        let version = req.version;
```
reem: This function is starting to get *really* long, can you split this section into another fn?

seanmonstar:
@reem ok, refactored that function...

reem:
LGTM - do we care about the coveralls regression as a blocker?

seanmonstar:
I don't, as I still find the coverage report confusing and inaccurate.

PR #528, opened 2015-05-13
Cleanup found by clippy lints

seanmonstar:


reem:
Why not `into` instead of `to_owned`?

seanmonstar:
@reem they do the same thing, but `ToOwned` uses an associated type, so it doesn't require inference to figure out it's output.

PR #534, opened 2015-05-16
feat(headers): return hyper::Error instead of () from header components

pyfisch:
This allows more precise errors in the future and makes it easier to use
the try!() macro in some cases.

BREAKING CHANGE: Error enum extended. Return type of header/shared/
types changed.

### DIFF
--- a/src/header/shared/charset.rs
+++ b/src/header/shared/charset.rs
@@ -103,8 +103,8 @@ impl Display for Charset {
 }
 
 impl FromStr for Charset {
-    type Err = ();
-    fn from_str(s: &str) -> Result<Charset, ()> {
+    type Err = ::Error;
+    fn from_str(s: &str) -> ::Result<Charset> {
         Ok(match s.to_ascii_uppercase().as_ref() {
             "US-ASCII" => Us_Ascii,
             "ISO-8859-1" => Iso_8859_1,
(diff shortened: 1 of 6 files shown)


src/error.rs
```diff
@@ -37,6 +38,8 @@ pub enum Error {
     TooLarge,
     /// An invalid `Status`, such as `1337 ELITE`.
     Status,
+    /// An ivalid header value or header value fragment.
```
seanmonstar: typo `ivalid`

src/error.rs
```diff
@@ -37,6 +38,8 @@ pub enum Error {
...
+    Invalid,
```
seanmonstar: Since most of the errors are about "invalid" values, I think this name should be specific to what kind of value was invalid. Perhaps `Header`.

pyfisch: `Header` is already taken and used when header blocks fail to parse. I rename it to `Value`.

seanmonstar:
I don't get why you want to use `Error::Value` here, instead of `Error::Header`. A `Value` could be from just about anything, no?

pyfisch:
Ok, I will suse `Error::Header` but I think it is a bad thing to reuse a header code. `Error::Header` is already used by httpparse. The relativly precise Errors from the parser are merged into a few codes, I think this is bad because you loose useful information, but this is not the topic here.

seanmonstar:
Optionally, a specific header could have it's own error types. Such as if there are multiple specific things that can go wrong during parsing it...

pyfisch:
Yes it could habve one. Any problems remaining?

PR #542, opened 2015-05-23
feat(ssl): make OpenSSL an optional dependency

### LINKED ISSUE #541: Make OpenSSL an optional dependency
It can be tricky to install OpenSSL, especially on Windows. It would be nice if it were an optional dependency.

nstoddard:
Compiling with default_features=false will now omit SSL support.

This introduces a ton of `cfg` annotations, but I don't know how to avoid them :(. It's quite ugly, but it seems to work.

BTW, I hope I got the title format correct.

Fixes #541

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,12 +12,10 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
 keywords = ["http", "hyper", "hyperium"]
 
 [dependencies]
-cookie = "0.1"
 httparse = "0.1"
 log = "0.3"
 mime = "0.0.11"
 num_cpus = "0.2"
-openssl = "0.6"
 rustc-serialize = "0.3"
 time = "0.1"
 unicase = "0.1"
@@ -29,4 +27,17 @@ typeable = "0.1"
 env_logger = "*"
 
 [features]
+default = ["ssl"]
+# These are grouped together b/c it doesn't really make sense to
+# enable one w/o the other
+ssl = ["openssl", "cookie"]
 nightly = []
+
+
+[dependencies.cookie]
+version = "0.1"
+optional = true
+
+[dependencies.openssl]
+version = "0.6"
+optional = true
--- a/src/error.rs
+++ b/src/error.rs
@@ -4,19 +4,11 @@ use std::fmt;
 use std::io::Error as IoError;
 
 use httparse;
+#[cfg(feature = "openssl")]
 use openssl::ssl::error::SslError;
 use url;
 
-use self::Error::{
-    Method,
-    Uri,
-    Version,
-    Header,
-    Status,
-    Io,
-    Ssl,
-    TooLarge
-};
+use self::Error::*;
 
 
 /// Result type often returned from methods that can have hyper `Error`s.
@@ -40,6 +32,7 @@ pub enum Error {
     /// An `io::Error` that occurred while trying to read or write to a network stream.
     Io(IoError),
     /// An error from the `openssl` library.
+    #[cfg(feature = "openssl")]
     Ssl(SslError)
 }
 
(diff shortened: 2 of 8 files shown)


pyfisch:
openssl is optional for cookies: https://github.com/alexcrichton/cookie-rs/blob/595a9e095bef8d98691bd1eed135d9e540f279e0/Cargo.toml, so you should not kick out cookie support completly.

src/error.rs
```diff
@@ -4,19 +4,11 @@ use std::fmt;
...
-use self::Error::{
-    Method,
-    Uri,
-    Version,
-    Header,
-    Status,
-    Io,
-    Ssl,
-    TooLarge
-};
```
pyfisch: * imports are evil, so do not change the normal import.

src/server/mod.rs
```diff
@@ -112,8 +119,12 @@ impl<'a, H: Handler + 'static> Server<'a, H, HttpListener> {
     /// Binds to a socket, and starts handling connections using a task pool.
     pub fn listen_threads<T: ToSocketAddrs>(self, addr: T, threads: usize) -> ::Result<Listening> {
         let listener = try!(match self.ssl {
+            #[cfg(feature = "openssl")]
             Some(SslConfig::CertAndKey(cert, key)) => HttpListener::https(addr, cert, key),
+            #[cfg(feature = "openssl")]
             Some(SslConfig::Context(ssl_context)) => HttpListener::https_with_context(addr, ssl_context),
+            #[cfg(not(feature = "openssl"))]
+            Some(_) => panic!("Internal error - we have an SSL config even though SSL is disabled"),
```
seanmonstar: Instead of this panic, can the cfg just be put on the `ssl: Option` property of the `Server`?

src/net.rs
```diff
@@ -341,12 +367,13 @@ impl NetworkConnector for HttpConnector {
                 let stream = try!(SslStream::new(&context, stream));
                 Ok(HttpStream::Https(stream))
             },
-            _ => {
+            x => {
                 Err(io::Error::new(io::ErrorKind::InvalidInput,
-                                "Invalid scheme for Http"))
+                                format!("Invalid scheme for Http: {}", x)))
```
seanmonstar: Whatcha think of adding a `debug!` (or info?) here explaining that enabling the feature will let you make https requests?

src/net.rs
```diff
@@ -307,16 +326,22 @@ impl NetworkStream for HttpStream {
...
 
     }
 }
 
+#[cfg(feature = "openssl")]
 /// A connector that will produce HttpStreams.
 pub struct HttpConnector(pub Option<ContextVerifier>);
+#[cfg(not(feature = "openssl"))]
+/// A connector that will produce HttpStreams.
+pub struct HttpConnector(pub Option<()>);
```
seanmonstar: Any reason to keep the `Option<()>`? Without the verifier, it used to be a `struct HttpConnector;`.

Cargo.toml
```diff
@@ -29,4 +27,17 @@ typeable = "0.1"
 env_logger = "*"
 
 [features]
+default = ["ssl"]
+# These are grouped together b/c it doesn't really make sense to
+# enable one w/o the other
+ssl = ["openssl", "cookie"]
```
seanmonstar: `ssl = ["openssl", "cookie/default"]`

seanmonstar:
I agree, let's not remove cookie entirely. We can turn just turn on or off cookie's usage of openssl.

I seem room for improvement in cookie's feature usage though: the jar could still be provided, and just have the signing methods and types conditional inside jar.rs, but oh well.

seanmonstar:
Besides those comments, this looks great! I'd been thinking of providing this at some point, but never got around to it. Thanks for the effort!

mlalic:
Since the HTTP/2 stuff got merged, openssl found its way back in :) I've published a new version of solicit (0.3) to crates.io, that relegates it to an optional dependency too, behind the "tls" feature. Hyper doesn't use anything from there yet anyway, so it should be fine to just go with the default set in all cases.

seanmonstar:
This was actually merged in #577.

nstoddard:
Oh, I totally forgot about this PR. Sorry about that!

PR #548, opened 2015-05-25
test(connection): increasing test coverage of connection module

winding-lines:
Work in progress :)


winding-lines:
@seanmonstar, I am done with mychanges once c1a8e3b gets built. We may have some overlap between tests and bench_header!(), but I think that's ok since they have different goals. Let me know if you have any feedback.

PR #549, opened 2015-05-27
test(error): increasing test coverage of error module

winding-lines:


winding-lines:
hm, this PR is not very successful in increasing coverage, I may just drop it.

seanmonstar:
I'll merge this, since it also tests the `cause()` part, which is a good idea. If there was a way to pass to kcov to ignore all `fn`s inside a `#[cfg(test)]` mod, that would greatly help reduce inflated "lines missed" values.

winding-lines:
Great :)

Marius
---

PR #551, opened 2015-05-28
refactor(headers): expose RangeUnit in hyper::header

clatour:
Allow RangeUnit to be used from user crates.


seanmonstar:
Nice catch.

PR #555, opened 2015-06-02
feat(error): add private __Nonexhaustive variant to Error

seanmonstar:
BREAKING CHANGE: Adds a new variant to public Error enum. The proper fix
  is to stop matching exhaustively on `hyper::Error`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -45,6 +45,18 @@ pub enum Error {
     Ssl(SslError),
     /// An HTTP/2-specific error, coming from the `solicit` library.
     Http2(Http2Error),
+
+    #[doc(hidden)]
+    __Nonexhaustive(Void)
+}
+
+#[doc(hidden)]
+pub enum Void {}
+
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -45,6 +45,18 @@ pub enum Error {
     Ssl(SslError),
     /// An HTTP/2-specific error, coming from the `solicit` library.
     Http2(Http2Error),
+
+    #[doc(hidden)]
+    __Nonexhaustive(Void)
```
reem: This should probably be `__NonExhaustive`.

reem:
Maybe we should pull in the `void` crate for this? I just updated it to work on stable.

src/error.rs
seanmonstar: I didn't even think about it, I just copied it from `std::io::ErrorKind`

seanmonstar:
rust-void looks nice, but I don't think I need any more that just an empty enum and an empty match.

PR #557, opened 2015-06-03
refactor(http): move h1 and h2 into http module

seanmonstar:
@mlalic This organization felt cleaner, now that there's 2 versions of http here.


mlalic:
Looks good to me.

PR #560, opened 2015-06-09
feat(headers): add `Range` header

lame-nickname:
There are 2 TODOs left in the code. I wasn't sure what is the best way to handle these cases and I'd like to hear your feedback on that.

What I also wasn't sure about, are the names of values inside the RangeSpec enum, as well as the name  of enum itself. Maybe you'll be able to come up with better ones ;)


seanmonstar:
Woo, thanks! I'll look this over. Also pinging @pyfisch, since he's done a ton of work on the headers.

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(PartialEq, Clone, Debug)]
+pub struct Range {
+    unit: RangeUnit,
+    ranges: Vec<RangeSpec>,
+}
+
+/// Each 'Range' header can contain one or more RangeSpecs.
+/// Each RangeSpec defines a range of units to fetch
+#[derive(PartialEq, Clone, Debug)]
+pub enum RangeSpec {
```
seanmonstar: I wonder, would it work out nicely to use `std::ops::{Range, RangeFrom, RangeTo}`? This could allow usage like `let ranges = vec![ 0..1, 30..40];`

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+    FromTo(u64, u64),
+    /// Get all bytes starting from x ("x-")
+    AllFrom(u64),
+    /// Get last x bytes ("-x")
+    Last(u64)
+}
+
+impl Range {
+    /// Get the most common byte range header ("bytes=from-to")
+    pub fn simple(from: u64, to: u64) -> Range {
```
seanmonstar: What about if the name for `bytes`? `Range::bytes(10..40)`?

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+            RangeSpec::Last(pos) => write!(fmt, "-{}", pos),
+            RangeSpec::AllFrom(pos) => write!(fmt, "{}-", pos),
+        }
+    }
+}
+
+
+impl fmt::Display for Range {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let as_strings : Vec<String> = self.ranges.iter().map(ToString::to_string).collect();
```
seanmonstar: This will make several heap allocations, before being pushed into the `fmt` buffer and then being tossed. It'd be faster to skip those. For instance:

```rust
try!(write!(fmt, "{}=", self.unit));

let mut iter = self.ranges.iter();
match iter.next() {
    Some(range) => try!(write!("{}", range)),
    None => return Ok(())
}
for range in iter {
    try!(write!(",{}", range));
}
Ok(())
```

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        if as_strings.is_empty() {
+            // TODO: There are few ways to handle it:
```
seanmonstar: Hm, what does the spec say? Anything?

src/header/common/range.rs
lame-nickname: Yeah, I wasn't convinced with `simple` neither, but that was the best I could come up with at the time. I'll change it to bytes()

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+            write!(fmt, "{}={}", self.unit, as_strings.connect(","))
+        }
+    }
+}
+
+impl FromStr for Range {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Range, ()> {
+        let parts: Vec<&str> = s.split("=").collect();
```
seanmonstar: To prevent allocating for a `Vec` here, the iterator from `s.split` could just be used.

```rust
let parts = match (iter.next(), iter.next()) {
    (Some(left), Some(right)) => (left, right),
    _ => return Err(())
};
```

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        Err(())
+    }
+}
+
+impl FromStr for RangeSpec {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<RangeSpec, ()> {
+        let parts : Vec<&str> = s.split("-").collect();
```
seanmonstar: Similar idea here regarding using the iterator.

src/header/common/range.rs
```diff
@@ -0,0 +1,185 @@
...
+            },
+            (start, end) => match (start.parse::<u64>(), end.parse::<u64>()) {
+                (Ok(start), Ok(end)) => Ok(RangeSpec::FromTo(start, end)),
+                _ => Err(())
+            }
+        }
+    }
+}
+
+impl Header for Range {
```
seanmonstar: This may want to use the `header!` macro, though I'm not quite sure how it would be used. @pyfisch has spent more time with it.

lame-nickname: I'm not sure how I would use the `header!` macro here. From what I've seen it autogenerates a single field struct and for `Range` I need to have both unit and list or ranges/rangeSpecs.

seanmonstar: Yea, it may that's what the macro is for, in which case nevermind me. :)

lame-nickname:
Ok. So in the new commit, I've included all your suggestions regarding iterators and unnecessary allocations. 
I've also made some decisions regarding the TODOs.
1). I think it makes sense to allow Range without any `ranges` (empty vector). In that case it will be transmitted as "bytes="  and it will be up to a server do decide what to do with it.
2) Now instead of skipping bad `ranges` during parsing, we'll return an Err, which basically means that the whole Range header is invalid.

src/header/common/range.rs
lame-nickname: I'm not sure if that makes sense. The thing is that `Range` and `RangeFrom` have their equivalent in the `RangeSpec` struct, but `RangeTo` doesn't. So we could implement `Into<RangeSpec> for RangeTo` and for `Range`, but that would leave out `RangeFrom` (and `RangeSpec::Last`), which IMO isn't very clean.

seanmonstar: You are right, I hadn't read the Range spec before suggesting that. I didn't realize that `bytes=-500` means the last 500 bytes, I assumed it means 0-500.

src/header/common/range.rs
```diff
@@ -0,0 +1,196 @@
...
+    let mut headers = Headers::new();
+    headers.set(range_header);
+
+    assert_eq!(&headers.to_string(), "Range: bytes=0-1000,2000-\r\n");
+
+    headers.clear();
+    headers.set(Range {unit: RangeUnit::Bytes, ranges: vec![]});
+
+    assert_eq!(&headers.to_string(), "Range: bytes=\r\n");
+}
```
seanmonstar: mind adding the `bench_header!` macro usage down here?

src/header/common/range.rs
lame-nickname: I've made the same assumption when I've started implementing it ;)

src/header/common/range.rs
lame-nickname: Done.

src/header/common/range.rs
pyfisch: You cannot use the macro here.

seanmonstar:
This looks good to go. Thanks! Would you mind squashing into a single commit? The changelog only really needs to report 1 new feature: added the Range header.

src/header/common/range.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        Err(())
+    }
+}
+
+impl FromStr for RangeSpec {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<RangeSpec, ()> {
+        let mut parts = s.split("-");
```
pyfisch: You should probably use [`splitn(2)`](https://doc.rust-lang.org/std/primitive.str.html#method.splitn) here. Imagine a malformed `RangeSpec` like `200-300-foobar`. It would pass as valid.

src/header/common/range.rs
```diff
@@ -0,0 +1,199 @@
...
+            (Some(""), Some(end)) => match end.parse::<u64>() {
+                Ok(end) => Ok(RangeSpec::Last(end)),
+                _ => Err(())
+            },
+            (Some(start), Some("")) => match start.parse::<u64>() {
+                Ok(start) => Ok(RangeSpec::AllFrom(start)),
+                _ => Err(())
+            },
+            (Some(start), Some(end)) => match (start.parse::<u64>(), end.parse::<u64>()) {
+                (Ok(start), Ok(end)) => Ok(RangeSpec::FromTo(start, end)),
```
pyfisch: You could check if `start` < `end`, so application code does not have to.

src/header/common/range.rs
```diff
@@ -0,0 +1,199 @@
+use std::fmt::{self, Display};
+use std::str::FromStr;
+
+use header::{Header, HeaderFormat, RangeUnit};
+use header::parsing;
+
+/// The `Range` header defined by HTTP/1.0.
```
pyfisch: Really HTTP/1.0? [RFC 1945](https://tools.ietf.org/html/rfc1945) does not mention it.

pyfisch: And could you please adjust the doccomment format like in the [Accept](https://github.com/hyperium/hyper/blob/master/src/header/common/accept.rs) header.

src/header/common/range.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+}
+
+
+impl fmt::Display for Range {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        try!(write!(fmt, "{}=", self.unit));
+
+        let mut ranges = self.ranges.iter();
```
pyfisch: This solves the same problem as https://github.com/hyperium/hyper/blob/b2eb4e5171d92d07f1ccc6b84396671f492c5445/src/header/parsing.rs#L48 (but without whitespace between the values). The algorithm should probably be the same. Most other code uses `f` for the formatter parameter. You can use `Display::fmt(range, f)` instead of `write!(fmt, "{}", range)`.

src/header/common/range.rs
```diff
@@ -0,0 +1,199 @@
...
+        let mut iter = s.split("=");
+
+        let (unit, ranges) = match (iter.next(), iter.next()) {
+            (Some(unit), Some(ranges)) => (unit, ranges),
+            _ => return Err(())
+        };
+
+        if let Ok(unit) = RangeUnit::from_str(unit) {
+            let mut valid_ranges: Vec<RangeSpec> = vec![];
+            for range in ranges.split(",") {
```
pyfisch: There can be optional whitespace between the values and empty values like `bytes=100-200, , , 300-400, -900`.

The ABNF is:
```
     byte-ranges-specifier = bytes-unit "=" byte-range-set
     byte-range-set  = 1#( byte-range-spec / suffix-byte-range-spec )
     byte-range-spec = first-byte-pos "-" [ last-byte-pos ]
     first-byte-pos  = 1*DIGIT
     last-byte-pos   = 1*DIGIT
```

The definition for `1#` is given in https://tools.ietf.org/html/rfc7230#section-7 optional whitespce is allowed but not required. You can probably use `from_one_comma_delimited()` here.

pyfisch:
Sorry for the late comments, some are a bit nitpicky but they should help to archieve a higher conformance with the spec :wink:

src/header/common/range.rs
lame-nickname: Yeah, the HTTP/1.0 was a mistake on my side. 

I've used `Pragma` header as a template and it uses `///` for the comments. 
Can you tell me why `#doc` is preferred? IMO, it's a lot noisier and harder to read.

lame-nickname:
No worries. I'm glad you've reviewed the commits, as there were several things I've clearly missed reading the HTTP spec. Will do better next time! ;)

src/header/common/range.rs
seanmonstar: I imagine he means the text. No reason to use `#[doc]` here.

pyfisch: Yes I mean the text, `Accept` just uses `#[doc]` to work around compiler bugs.

GitCop:

 * Commit: 287dc391638a99ee8604f14926d4d81344fa8a4d

 * Commit: 808ccdda443c6e6963124226185fea69b5a8a485

---

lame-nickname:
Ok guys, I've made the requested changes. I wouldn't be surprised if they were unidiomatic, so let me know if that's the case.

There are 2 open issues:
* i.e.  `bytes=-1,-1,-1` or `bytes=1-,1-`. According to the spec, there should be only one range spec of `{RangeSpec::AllFrom, RangeSpec::Last}` and it should always be at the end of the range set. Currently, it's not enforced and it's up to the application code to check that.
* i.e. `bytes=1-100,abc,200-`. Should we just drop the invalid range specs and mark the whole range as valid (as long as there is at least one valid range spec) or mark the range as invalid, if there is at least one invalid range spec? Currently, we do the former.

src/header/common/range.rs
```diff
@@ -0,0 +1,231 @@
+use std::fmt::{self, Display};
+use std::str::FromStr;
+
+use header::{Header, HeaderFormat, RangeUnit};
+use header::parsing::{from_one_raw_str, from_one_comma_delimited};
+
+/// `Range` header, defined in [RFC7233](https://tools.ietf.org/html/rfc7233#section-3.1)
```
pyfisch: Nice docs

src/header/common/range.rs
```diff
@@ -0,0 +1,231 @@
...
+        Range {
+            unit: RangeUnit::Bytes,
+            ranges: vec![RangeSpec::FromTo(from, to)],
+        }
+    }
+}
+
+
+impl fmt::Display for RangeSpec {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
```
pyfisch: f instead of fmt

src/header/common/range.rs
```diff
@@ -0,0 +1,231 @@
...
+                try!(f.write_str(","));
+            }
+            try!(Display::fmt(range, f));
+        }
+        Ok(())
+    }
+}
+
+impl FromStr for Range {
+    type Err = ();
```
pyfisch: This should use an Hyper Error as return type.

src/header/common/range.rs
```diff
@@ -0,0 +1,231 @@
...
+                    _ => Err(())
+                }
+            }
+            _ => Err(())
+        }
+    }
+}
+
+impl FromStr for RangeSpec {
+    type Err = ();
```
pyfisch: Also use an Error here

src/header/common/range.rs
lame-nickname: What should be the return value in this case? Before it was `Err(())` and now `Err(::Error::Header)`?

GitCop:

 * Commit: 287dc391638a99ee8604f14926d4d81344fa8a4d

 * Commit: 808ccdda443c6e6963124226185fea69b5a8a485

---

src/header/common/range.rs
lame-nickname: I'm not sure if `hyper::Error` makes sense here. At least not `hyper::Error::Header`. The `RangeSpec` struct isn't really a header, so `::Error::Header` doesn't make much sense for parsing failure. Plus I don't see any other `::Error` enum value, that would make sense in this context.

GitCop:

 * Commit: 287dc391638a99ee8604f14926d4d81344fa8a4d

 * Commit: 808ccdda443c6e6963124226185fea69b5a8a485

---

pyfisch:
+1

seanmonstar:
Excellent work! Thanks for sticking with it.

PR #563, opened 2015-06-12
feat(client): impl Sync for Client

### LINKED ISSUE #254: "Send"able request builder
How difficult would it be to make the request builder sendable, so a user of the lib could create a request builder and then send it to a helper task while it blocks on network io?

seanmonstar:
Connector::connect already used &self, and so would require
synchronization to be handled per connector anyway. Adding Sync to the
Client allows users to setup config for a Client once, such as using a
single connection Pool, and then making requests across multiple
threads.

Closes #254

BREAKING CHANGE: Connectors and Protocols passed to the `Client` must
  now also have a `Sync` bounds, but this shouldn't break default usage.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -20,7 +20,7 @@ fn main() {
         }
     };
 
-    let mut client = Client::new();
+    let client = Client::new();
 
     let mut res = client.get(&*url)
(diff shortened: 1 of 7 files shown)


derekdreery:
:D

PR #567, opened 2015-06-15
docs(headers): add examples

### LINKED ISSUE #468: Add example usage for each Header

clatour:
Added examples for all common header types.

Further work could be done to add descriptions around the examples, or the equivalent string representation of the headers, but this is a start.

If there is an `Extension` for the header enum type, I've tried to specifically include an example of it.

Fixes #468


seanmonstar:
:dancers:

PR #571, opened 2015-06-17
feat(headers): add bearer token support

yjv:
this allows servers/clients using bearer tokens to work out of the box without having to implement their own bearer scheme. while this would be pretty easy seems like a more general thing that is useful for a lib like this


GitCop:

 * Commit: 50068d161da4b4b3e4395e5680f77d93f82718d8

---

PR #572, opened 2015-06-17
feat(headers): add bearer token support

yjv:
this allows servers/clients using bearer tokens
 to work out of the box without having to implement
  their own bearer scheme. while this would be pretty
   easy seems like a more general thing that is useful
    for a lib like this


seanmonstar:
Good idea. Excellent on the implementation, thanks!

yjv:
Thanks!

Issue #575, opened 2015-06-17
OS X error: undefined symbols for architecture x86_64:   "_TLSv1_2_method"


winding-lines:
I am getting an undefined symbol related to openssl when I try to compile the latest on OS X. More detail below.

git log
```
commit e5dfb8233f25a4c57f2c6a68e8a7a854e42295b6
Merge: 41823ca edf6ac2
Author: Sean McArthur <sean.monstar@gmail.com>
Date:   Tue Jun 16 22:51:13 2015 -0700

    Merge pull request #572 from giftcards/master

    feat(headers): add bearer token support
```

rustc version
```
rustc -V
rustc 1.0.0 (a59de37e9 2015-05-13) (built 2015-05-14)
```

OS X version
```
sw_vers
ProductName:	Mac OS X
ProductVersion:	10.10.3
BuildVersion:	14D136
````

Libssl

```
ls -l /usr/lib/libssl.*
-rwxr-xr-x  1 root  wheel  400608 Sep  9  2014 /usr/lib/libssl.0.9.7.dylib
-rwxr-xr-x  1 root  wheel  616512 Mar 19 19:16 /usr/lib/libssl.0.9.8.dylib
lrwxr-xr-x  1 root  wheel      18 Feb 22 00:56 /usr/lib/libssl.dylib -> libssl.0.9.8.dylib
```

Actual error.

```
➜  cargo clean
➜  cargo test
   Compiling traitobject v0.0.1
   Compiling gcc v0.3.8
   Compiling pkg-config v0.3.5
   Compiling lazy_static v0.1.11
   Compiling rustc-serialize v0.3.15
   Compiling httparse v0.1.4
   Compiling unicase v0.1.0
   Compiling regex-syntax v0.1.2
   Compiling libc v0.1.8
   Compiling matches v0.1.2
   Compiling typeable v0.1.2
   Compiling bitflags v0.1.1
   Compiling time v0.1.26
   Compiling openssl-sys v0.6.2
   Compiling memchr v0.1.3
   Compiling log v0.3.1
   Compiling num_cpus v0.2.6
   Compiling aho-corasick v0.1.3
   Compiling mime v0.0.11
   Compiling hpack v0.2.0
   Compiling regex v0.1.35
   Compiling openssl v0.6.2
   Compiling solicit v0.2.0
   Compiling env_logger v0.3.1
   Compiling url v0.2.35
   Compiling cookie v0.1.21
[...]

Issue #573, opened 2015-06-17
Use WinAPI for SSL on Windows


retep998:
As an alternative to OpenSSL (which is generally quite annoying to build on Windows), it would be nice if Hyper had the option to use methods that WinAPI provides for SSL instead.

Bindings to the relevant WinAPI methods will be provided by me https://github.com/retep998/winapi-rs

Will use the abstraction provided by https://github.com/hyperium/hyper/pull/577

Related repos:
https://github.com/steffengy/rust-schannel

Relevant links:
[Authentication Functions](https://msdn.microsoft.com/en-us/library/windows/desktop/aa374731%28v=vs.85%29.aspx)
[Using SSPI with a Windows Socket Client](https://msdn.microsoft.com/en-us/library/windows/desktop/aa380536%28v=vs.85%29.aspx)
[Creating a Secure Connection Using Schannel](https://msdn.microsoft.com/en-us/library/windows/desktop/aa374782%28v=vs.85%29.aspx)
[Secure Channel](https://msdn.microsoft.com/en-us/library/windows/desktop/aa380123%28v=vs.85%29.aspx)
[SSPI with Schannel example](http://www.coastrd.com/c-schannel-smtp)

steffengy:
Now also available through https://crates.io/crates/schannel

luser:
In https://github.com/hyperium/hyper/issues/755, someone linked to https://github.com/sfackler/rust-native-tls, which is a crate wrapping openssl/security-framework/schannel, which might be a good way to go in the future.

p-jackson:
Created a gist that shows how to get schannel working with hyper, just incase someone encounters this thread while looking for that info.
https://gist.github.com/p-jackson/198dd5de05b53d14b533d272f2eaefce

schannel needs an update before it'll work with hyper 0.9

steffengy:
@p-jackson 
Be aware that it very likely won't work that way anymore for any future version.

In the future the usage of https://github.com/sfackler/rust-native-tls
or https://github.com/sfackler/schannel-rs (currently escpecially the rewrite branch).
will likely allow a similar usage.

nshtg:
Are there any plans to integrate `rust-native-tls` using a special feature flag? Easy usage of hyper without openssl would be really exciting!

sinhpham:
What is blocking this issue? I may find some time to work on this, however I'm not familiar with the code base. Can someone give me an overview of what needs to be done and what potential obstacles there are?

[...]

PR #577, opened 2015-06-19
feat(ssl): redesign SSL usage

seanmonstar:
- `Server::https` was changed to allow any implementation of `Ssl`.
- `Server` in general was also changed. This was to help the design of `Ssl`, and will also allow a future Pull Request to allow the `Server` to support an `Iterator` design, and accepting single requests; generally just not requiring that a `Handler` (and its task pool) be used.
- `HttpConnector` no longer uses SSL; use `HttpsConnector` instead. **NOTE: Servo**
- `Client` will use `DefaultConnector`, which will be `HttpsConnector` by default. If the openssl feature is disabled, it will be `HttpConnector`. Those using `Client::new()` should be unaffected.

cc #573 #131 #541 #575 @reem @retep998 @Manishearth

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,22 +12,33 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
 keywords = ["http", "hyper", "hyperium"]
 
 [dependencies]
-cookie = "0.1"
 httparse = "0.1"
 log = "0.3"
-mime = "0.0.11"
+mime = "0.0.12"
 num_cpus = "0.2"
-openssl = "0.6"
 rustc-serialize = "0.3"
 time = "0.1"
-unicase = "0.1"
-url = "0.2"
 traitobject = "0.0.1"
 typeable = "0.1"
-solicit = "0.2"
+unicase = "0.1"
+url = "0.2"
+
+[dependencies.cookie]
+version = "0.1"
+default-features = false
+
+[dependencies.openssl]
+version = "0.6"
+optional = true
+
+[dependencies.solicit]
+version = "0.3"
+default-features = false
 
 [dev-dependencies]
 env_logger = "*"
 
 [features]
+default = ["ssl"]
+ssl = ["openssl", "cookie/secure", "solicit/openssl"]
 nightly = []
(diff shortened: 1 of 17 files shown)


src/net.rs
```diff
@@ -239,123 +224,268 @@ impl Write for CloneTcpStream {
...
+                                "Invalid scheme for Http"))
+            }
+        }))
     }
-  }
 }
 
-impl Read for HttpStream {
+
+/// An abstraction to allow any SSL implementation to be used wit HttpsStreams.
```
reem: typo in with as wit

Cargo.toml
```diff
@@ -26,8 +25,13 @@ traitobject = "0.0.1"
 typeable = "0.1"
 solicit = "0.2"
```
mlalic: As I mentioned in the comments over at #542, you should bump up this version to "0.3" to make sure that hyper doesn't end up requiring openssl transitively through solicit.

src/net.rs
```diff
@@ -143,29 +136,29 @@ impl NetworkStream + Send {
...
 /// A `NetworkListener` for `HttpStream`s.
-pub enum HttpListener {
-    /// Http variant.
-    Http(TcpListener),
-    /// Https variant. The two paths point to the certificate and key PEM files, in that order.
-    Https(TcpListener, Arc<SslContext>)
-}
+pub struct HttpListener(TcpListener);
+
+/// Https variant. The two paths point to the certificate and key PEM files, in that order.
```
reem: Stray commented code related to this type. As far as I can see it can just be deleted.

seanmonstar:
Nice catches, fixed.

reem:
I like the new Server API, seems more easily extensible.

reem:
@seanmonstar give me a heads up before publishing a new version including these changes please.

seanmonstar:
Sure. I'll bump the version to 0.6 as well, since it has breaking changes.

softprops:
I'm trying to update a project to the latest versions of its dependencies including `hyper` but am running into an issue that I think was a breaking change in this pull. There used to exist a client.set_ssl_verifier which I [used here](https://github.com/softprops/shiplift/blob/master/src/lib.rs#L309-L325). This method seems to have went away, but I can't find the current alternative. Does one exists in the latest release?

softprops:
I think I figured out out

```rust
let client = Client::with_connector(
  HttpsConnector::new(
    Openssl {
      context: Arc::new(
        ssl_ctx // <- I create this now
       )
    }
  )
)
```

[this factory method](https://github.com/hyperium/hyper/blob/master/src/net.rs#L427-L428) want's quiet what I needed but enough to steer me in the right direction.

Great work on pushing this library forward.

PR #583, opened 2015-06-26
fix(cargo): fix linking on OSX 10.10

Byron:
On OSX 10.10, attempting to run `cargo test` will yield a
[linker error](https://gist.github.com/Byron/b7a50cdcf7c60234335d).

Just removing the feature fixes the issue, even though I am not
sure if it is breaking something further down the line.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,5 +41,5 @@ env_logger = "*"
 
 [features]
 default = ["ssl"]
-ssl = ["openssl", "cookie/secure", "solicit/openssl"]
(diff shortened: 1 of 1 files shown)


mlalic:
This doesn't break anything, for two reasons:

- `solicit` has no `"openssl"` feature :)
- Even the `"tls"` feature, which is what was meant here, I'm sure, isn't used anywhere

I [mentioned before](https://github.com/hyperium/hyper/pull/542#issuecomment-108216216) that we don't need it here for anything... :)

PR #595, opened 2015-07-04
refactor(headers): use String in Access-Control-Allow-Origin header

pyfisch:
Access-Control-Allow-Origin origins are URLs but they do not need to
be valid, they should just be compared as strings. So to support
invalid URLs hyper should use a string instead.

solves #562

BREAKING CHANGE: Access-Control-Allow-Origin does no longer use Url

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,6 +3,7 @@ use std::error::Error as StdError;
 use std::fmt;
 use std::io::Error as IoError;
 use std::str::Utf8Error;
+use std::string::FromUtf8Error;
 
 use httparse;
 use url;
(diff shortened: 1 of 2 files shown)


pyfisch:
Can we increase the "Coverage Decrease Threshold for failure" so that changes like this one do not fail?

seanmonstar:
Sure, I went and bumped the threshold decrease to 2% (seems like a large enough to mean that new functions forgot tests). In this case, I think the decrease was in introducing `From<FromUtf8Error>` without adding to tests. However, the error.rs tests don't seem to report correctly, which is quite frustrating.

PR #603, opened 2015-07-10
Add optional serialization of common types via `serde`.

pcwalton:
This is behind a Cargo feature to avoid forcing downstream users to
depend on `serde`. It is needed for Servo IPC to work.

### DIFF
--- a/src/header/common/mod.rs
+++ b/src/header/common/mod.rs
@@ -159,6 +159,31 @@ macro_rules! test_header {
 }
 
 #[macro_export]
+macro_rules! __hyper_generate_header_serialization {
+    ($id:ident) => {
+        #[cfg(feature = "serde-serialization")]
+        impl ::serde::Serialize for $id {
+            fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
+                            where S: ::serde::Serializer {
+                format!("{}", self).serialize(serializer)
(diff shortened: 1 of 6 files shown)


GitCop:

 * Commit: 76f48ffe05bd5b91d14f735836e3c8549efd3dd2

---

src/header/common/content_type.rs
```diff
@@ -94,4 +99,20 @@ impl ContentType {
     }
 }
 
+#[cfg(feature = "serde-serialization")]
```
reem: Everything else looks great, but can we move this into the definition of the `header` macro, so that we automatically get serialization for most headers?

pcwalton:
@reem Done.

src/header/common/mod.rs
```diff
@@ -190,6 +214,8 @@ macro_rules! header {
                 self.fmt_header(f)
             }
         }
+
+        generate_header_serialization!($id);
```
reem: Unfortunately this doesn't work without exporting `generate_header_serialization` too, since the expansion is in another namespace. 

We can just export it with a `__` prefix like `__hyper_deref`.

pcwalton:
@reem Exported and added the namespace.

reem:
Awesome, LGTM.

PR #609, opened 2015-07-13
feat(headers): Content-Range header

ealasu:


pyfisch:
@lame-nickname is also working on this. Your implementation onlyb supports byte  values but it must support all types of ranges.

ealasu:
I must have been looking at RFC 2616 instead of 7233 (the former only allows the bytes unit). Will fix.

ealasu:
@pyfisch fixed.

src/header/common/content_range.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl FromStr for ContentRangeSpec {
+    type Err = ::Error;
+
+    fn from_str(s: &str) -> ::Result<Self> {
+        let mut iter = s.splitn(2, ' ');
+        let res = match (iter.next(), iter.next()) {
+            (Some("bytes"), Some(resp)) => {
+                let mut iter = resp.split('/');
```
seanmonstar: I think this needs to use `splitn` as well, since this will currently miss something like `"123/567/89"`.

ealasu:
@seanmonstar fixed.

src/header/common/content_range.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+impl Display for ContentRangeSpec {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ContentRangeSpec::Bytes { range, instance_length } => {
+	        try!(f.write_str("bytes "));
+                match range {
+                    Some((first_byte, last_byte)) => {
+                        try!(f.write_fmt(format_args!("{}-{}", first_byte, last_byte)));
```
seanmonstar: This can be re-written as `write!(f, "{}-{}", first_byte, last_byte)`.

src/header/common/content_range.rs
```diff
@@ -0,0 +1,189 @@
...
+                    },
+                    None => {
+                        try!(f.write_str("*"));
+                    }
+                };
+	        try!(f.write_str("/"));
+	        if let Some(v) = instance_length {
+	            f.write_fmt(format_args!("{}", v))
```
seanmonstar: `write!(f, "{}", v)`

src/header/common/content_range.rs
```diff
@@ -0,0 +1,189 @@
...
+        Ok(res)
+    }
+}
+
+impl Display for ContentRangeSpec {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ContentRangeSpec::Bytes { range, instance_length } => {
+	        try!(f.write_str("bytes "));
+                match range {
```
seanmonstar: Seems to be 1 extra space of indent with this match block...

seanmonstar:
Those last few bits should also trigger a rebuild on travis, and once it's green, and assuming @pyfisch doesn't disagree, it'd be good to go!

seanmonstar:
Woops, I don't know how long ago you address this. I don't get notifications when a force push happens. Thanks sticking it out!

seanmonstar:
I squashed the commits into 1, and merged manually into master: https://github.com/hyperium/hyper/commit/af062ac954d5b90275138880ce2f5013d6664b5a

Issue #614, opened 2015-07-21
Implement Hash for StatusCode


havarnov:
Implementing Hash for StatusCode will make it possible to use the status code as key in a HashMap.

I.e.:
 ```rust
let handlers: HashMap<StatusCode, Handler> = HashMap::new();
```

jdm:
It's also possible to do `struct HashableStatusCode(StatusCode)` and implement Hash on that.

havarnov:
Yes, but you'd also had to implement `Deref` for `HashableStatusCode`. Not a big deal, but a little smoother IMO.

I've submitted a pull request, so it's up to you if you think it's useful.

PR #615, opened 2015-07-21
feat(status): implement `Hash` for `StatusCode`

havarnov:
Implment the `Hash` trait for the `StatusCode` enum.

Resolves issue #614


seanmonstar:
This could just use the derive attribute, couldn't it?

havarnov:
Of course, that's much better.

seanmonstar:
Thanks!

PR #636, opened 2015-08-27
fix(headers): case insensitive values for Connection header

### LINKED ISSUE #635: Connection Header case insensitivity
http://tools.ietf.org/html/rfc7230#section-6.1:

    ... "Connection options are case-insensitive." ...

I noticed that Apache Bench sends "Keep-Alive" and hyper misinterprets it

    DEBUG:hyper::server::request: Headers { Connection: Keep-Alive, Host: 192.168.66.62:8881, Accept: */*, User-Agent: ApacheBench/2.3, }
    TRACE:hyper::http: should_keep_alive( Http10, Some(Connection([ConnectionHeader(UniCase("Keep-Alive"))])) )

I think this may also apply to most other headers

mikedilger:
Closes #635

### DIFF
--- a/src/header/common/connection.rs
+++ b/src/header/common/connection.rs
@@ -4,6 +4,9 @@ use unicase::UniCase;
 
 pub use self::ConnectionOption::{KeepAlive, Close, ConnectionHeader};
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
:+1:

PR #639, opened 2015-08-31
feat(proxy): client proxy for http and https

winding-lines:
This is a proof of concept for supporting client side proxies. The tests
are not passing yet, I am just soliciting feedback on the architecture.

The use cases that I see are the following

- when proxy is configured by default all connections go to the proxy

- additionally there is a small list of local wan/localhost which bypass
the proxy

So once a connection to a host is established it will either use the
proxy
all the time or it will be a direct connection all the time. The Pool
keys streams on host-port-scheme so a stream is either proxy or direct.

The logic is the following:

1. connect to the proxy instead of the desired host

2. for http use the appropriate RequestUri (pre-existing class)

3. for https we need to issue a CONNECT first

Let me know what you think.


GitCop:

 * Commit: 5109c29956e3f8cb22618fde014fc0090223c831

---

GitCop:

 * Commit: 5109c29956e3f8cb22618fde014fc0090223c831

 * Commit: 3ba3bcc2c7f3a1e91db2acab7b3629ee3fdb1ab2

---

GitCop:

 * Commit: ca133ad4f3a986533b5960dbc6d933b7f3cb7266

---

GitCop:

 * Commit: d0d3f29dd0b4cee8f2ef4bdf5d695adcf3a4c4fc

---

winding-lines:
Good morning Hyper community,

I've rebased my client proxy changes on master and made all the tests pass. I need to write some tests but I would like to get some advice first on the architecture. I tried a couple of alternatives like removing the concept of the Route or not have it closely tied to the HttpStream. The challenge is that the proxying is a pretty low level concept so any change tends to propagate through the codebase.

Where would you like me to take next? We have a stable commit right now and we can compare with other approaches.

shaleh:
At the moment there is an assumption that both HTTP and HTTPS will be handled by the same proxy. This could cause problems. While I do not need it I know others do require authentication support.

As for design concerns, is there a reason you made a new constructor instead of allowing the user to call something like ``add_proxy``? Common use code will look something like:

    let client = match read_proxy_env() {
        Some(proxy_url) => {
             Client::with_proxy_config(proxy::Config {
                 proxy_host: proxy_url.some_struct_value,
                 proxy_port: proxy_url.some_other_struct_value,
                 proxy_version: "1.1".to_string()
             })
        }
        None => Client::new(),
    }

Except as I noted above this only works when one proxy config works for both settings. If there is a separate HTTP and HTTPS proxy things get more complex. This complexity needs to be duplicated in each and every client project using hyper.

Not directly this patch's problem but reading env::var("http_proxy") is not trivial. I need to use something like ``into_url`` on the string, then find the bits in the ``Url`` struct. Similarly how should one handle ``no_proxy``? It would be nice if there was convenience help provided for this set of common env variables.

shaleh:
The code works as expected for HTTP requests with the caveat that all requests go to the proxy because there is no support for ``no_proxy`` or the equivalent.

Any attempt to access a SSL url generates a SIGPIPE for me.

winding-lines:
@shaleh thanks for the feedback. I will add authentication and fix the https crash. I also need to add tests, I will do that after I refactor the internals based on other feedback.

I do not know how frequent it is for people to have multiple proxies. If we assume that most people have one proxy for all requests then the current API is the simplest. For your use case you can create 3 Hyper clients in the three configurations that you describe:
- wi proxy
- http proxy
- https proxy

This keeps the library simple and allows flexibility in the consuming application, with the downside that there is more code in the application.

Let me know what you think.

shaleh:
The point of a library is to make the application author's life easier _AND_ to have one place to put redundant code.

Almost every user of the hyper crate will need to implement proxy support at some point. The crate should make it easy for the application author to get it right. Ideally this means stuffing as much proxy knowledge into the crate as possible.

I'd like to see a baked in way to create a hyper based Client that read its settings from the environment. A similar bit of code that took a list of proxies instead of reading the environment would be another useful interface for those that want the proxy to come from their own configs.

Sorry, I would have written code instead of asking but I have been busy on other projects.

winding-lines:
Got it, I have been working on multiple proxies but I have only 6-8
hrs/week available myself. I will put out what ever code I have ready at
the end of this weekend.

seanmonstar:
I think the task of *getting the proxy settings* should be outside of hyper. The configuration could come from anywhere (environment variables, command line arguments, config files, hard coded, etc). Instead, hyper should accept the parameters in the Proxy constructor.

For instance: 

```rust
client.set_proxy(Proxy::Http {
    host: env::var("HTTP_PROXY_HOST").unwrap(),
    port: env::var("HTTP_PROXY_PORT").unwrap(),
})
```
-----

Also, it *is* possible to build a Proxy entirely outside of this crate, and use a hyper Client internally.  That's why I've mentioned trying not to adjust too much of the internal http protocol modules. The Client (and Request) just need to be able to use all the parts of RFC7230, such as the CONNECT method and alternate RequestUri formats. I'd rather if the logic to use those was contained in the `Proxy` type.

winding-lines:
Yeah, I have been trying to be less invasive. I think that some internal
refactoring is needed and it may be outside my current understanding of the
code. Let me try this weekend and see where I end up, you are certainly
welcome to take over as soon as you have the cycles :-)

GitCop:

 * Commit: 65777aed06e69e2ce48d436bdb915b5d0fffe496

---

winding-lines:
@seanmonstar I got rid of the Route and instead added a request_uri in Request (and RequestHead) as you had suggested on the initial PR. I've added a list of proxies each with their own proxy policies. All of this is only in place for the Http 1.1 code.

For the SSL NetworkConnector there is still one architecture issue because all the SSL types are defined only in the h2.rs file. So the NetworkConnector's connect still defers back to the Proxy, and this code is not ported to use the new list of proxies as per the desired APIs.

Where do you want me to take this code next?

src/client/mod.rs
```diff
@@ -107,18 +110,28 @@ impl Client {
         Client::with_connector(Pool::new(config))
     }
 
+    pub fn with_proxy_config(config: proxy::Config) -> Client {
```
seanmonstar: What's the difference between using a Connector with_proxy, and calling `client.add_proxy`?

src/header/common/proxy_authorization.rs
```diff
@@ -0,0 +1,22 @@
+use header:
+
```
seanmonstar: Is this header exported in `common/mod.rs`?

src/net.rs
```diff
@@ -292,6 +296,7 @@ impl Read for HttpStream {
 impl Write for HttpStream {
     #[inline]
     fn write(&mut self, msg: &[u8]) -> io::Result<usize> {
+        info!("write http: {}", str::from_utf8(msg).unwrap_or("-"));
```
seanmonstar: This will be far too noisy, can this be removed?

src/net.rs
```diff
@@ -310,7 +315,7 @@ impl ::std::os::windows::io::AsRawSocket for HttpStream {
 #[cfg(windows)]
 impl ::std::os::windows::io::FromRawSocket for HttpStream {
     unsafe fn from_raw_socket(sock: ::std::os::windows::io::RawSocket) -> HttpStream {
-        HttpStream(TcpStream::from_raw_socket(sock))
+        HttpStream(TcpStream::from_raw_socket(sock), Route::DirectRoute)
```
seanmonstar: `Route` no longer exists, right?

src/proxy.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Proxy only http requests.
+    ProxyHttp,
+    /// Proxy only https requestss.
+    ProxyHttps,
+    /// Proxy if the contained function returns true.
+    ProxyIf(fn(&Url) -> bool),
+}
+
+// Need to implement Clone for the time being.
+impl Clone for ProxyPolicy {
```
seanmonstar: This is due to a bug in rustc, because of the function pointer in `ProxyIf`? Is there a issue number for rustc?

src/net.rs
```diff
@@ -359,7 +364,16 @@ impl NetworkStream for HttpStream {
 
 /// A connector that will produce HttpStreams.
 #[derive(Debug, Clone, Default)]
-pub struct HttpConnector;
+pub struct HttpConnector {
+    pub proxy: Option<Proxy>
```
seanmonstar: what if a `ProxyConnector` existed in `src/proxy.rs`, that could wrap any `C: NetworkConnector`? This would allow all custom connectors to be wrapped, as well as let the logic stay in 1 module. It would also remove the `if proxy.is_some()` branch from code that never used a proxy.

GitCop:

 * Commit: 65777aed06e69e2ce48d436bdb915b5d0fffe496

---

GitCop:

 * Commit: c30780b56de9f05bcc70c8b8a1e5a87b23064f3d

---

winding-lines:
@seanmonstar, thanks for your feedback I've addressed some of it. 

I love your suggestion to have an implementation of NetworkConnector for Proxy, not yet implemented. What needs to happen is the following:

1. the Proxy implementation will connect to the proxied port
2. the socket must be passed down to the Http11 or SSL NetworkConnectors

With the current API this is not possible. This is the reason why the proxy is passed down.

Note that the Pool NetworkConnector has a different use case, the lower level NetworkConnectors are allowed to open their own sockets.

Please feel free to take over this PR and play with some approaches :)

mitsuhiko:
As mentioned in the ticket I would love to help drive this PR forward. Proxy support is very important to me.

seanmonstar:
Ideally some of this work can be brought over to the async world, as I hope to merge async io soon. It may even be that since the async version of hyper is a little lower level, using a proxy may be easier.

PR #642, opened 2015-09-01
feat(server): Add Handler per-connection hooks

mikedilger:
I'm adding a feature to my webserver:  whenever the thread pool is getting too busy (say 4 out of 8 threads are occupied, possibly idly with keep-alive requests), we force Connection close on subsequent requests until the thread pool isn't so busy anymore.

To do this, I need hooks into the beginning and end of the connection loop.  The Handler handle function is per-request, not per-connection.

A very simple addition, but please consider the function names, comments, etc., as I have no preference for these.


seanmonstar:
This seems like an acceptable situation while async support is worked out. Thanks!

mikedilger:
Go-go async support! That's gonna be awesome.  Thanks

PR #648, opened 2015-09-05
[WIP] Add a stream state enum that makes it impossible to lose a stream

mlalic:
@seanmonstar 

When the `HttpMessage` API was added and implemented for HTTP/1.1, we commented on the fact that the stream was being manually juggled in and out of `Option`s, leading to the possibility of losing the stream prematurely. At the time, I mentioned that I couldn't see how this could be avoided without resorting to some `unsafe` code. This was due to the fact that the `HttpMessage` trait methods require a `&mut self`, while the implementation needs to _move_ the stream object around.

I still think this is the case, but I think that with a minimal amount of unsafe code (3 lines which could have easily been a single one), a safe API can be exposed that guarantees that the stream will never be dropped (and eliminating the panics that this could lead to).

It hinges on the `replace_with` function that has extensive inline comments explaining why it should be safe.

The newly added `Stream` enum then represents the state of the message stream, which can only ever be idle, writing, or reading. This means that there is no chance of the underlying stream being dropped as long as the HttpMessage that references this enum is alive. Transitions to different stream states, while allowing those transitions to _move_ out of the original state variant in the &mut self methods, is performed via the `map_in_place` method of this enum. The method requires the caller to provide a closure which gets an owned `Stream` enum (meaning it can move stuff out of the variant), but requires them to return a new `Stream` that will take the place of the original one.

I deliberately left the closures passed to this method unindented for now so that you can see that a minimal amount of code really changed. Apart from minor things to fit with the new API, most of the changes actually fix spots where the current code would lead to prematurely dropping the stream (and an eventual panic).

I just wanted to get some feedback whether you think refactoring this code in this direction would be a good idea and whether my reasoning of the unsafe section is sound.

Some improvements could be made if we decide to go with this, such as allowing the closure passed to the `map_in_place` method not only a new stream state, but also a final result (referencing the `res` in the method scope is a bit awkward).

### DIFF
--- a/src/http/h1.rs
+++ b/src/http/h1.rs
@@ -35,19 +35,70 @@ use version;
 
 const MAX_INVALID_RESPONSE_BYTES: usize = 1024 * 128;
 
+#[derive(Debug)]
+struct Wrapper<T> {
+    obj: Option<T>,
+}
+
+impl<T> Wrapper<T> {
+    pub fn new(obj: T) -> Wrapper<T> {
+        Wrapper { obj: Some(obj) }
+    }
+
+    pub fn map_in_place<F>(&mut self, f: F) where F: FnOnce(T) -> T {
+        let obj = self.obj.take().unwrap();
+        let res = f(obj);
+        self.obj = Some(res);
+    }
+
+    pub fn into_inner(self) -> T { self.obj.unwrap() }
+    pub fn as_mut(&mut self) -> &mut T { self.obj.as_mut().unwrap() }
+    pub fn as_ref(&self) -> &T { self.obj.as_ref().unwrap() }
+}
+
+#[derive(Debug)]
+enum Stream {
+    Idle(Box<NetworkStream + Send>),
+    Writing(HttpWriter<BufWriter<Box<NetworkStream + Send>>>),
+    Reading(HttpReader<BufReader<Box<NetworkStream + Send>>>),
+}
+
+impl Stream {
+    fn writer_mut(&mut self) -> Option<&mut HttpWriter<BufWriter<Box<NetworkStream + Send>>>> {
+        match *self {
+            Stream::Writing(ref mut writer) => Some(writer),
+            _ => None,
+        }
+    }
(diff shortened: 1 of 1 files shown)


seanmonstar:
@mlalic It's possible to add `?w=1` on the end of diff URLs, and it will ignore whitespace changes :D

mlalic:
TIL :)

Though I'm not really surprised as it's just `git diff -w`

seanmonstar:
I haven't determined if unsound yet, but this unsafe worries me. If the
closure panics, then there will have existed 2 streams, duplicates of each
other.

That'd mean the destructor would run on both of them.

mlalic:
You're absolutely right, I disregarded what would happen during stack unwinding if an exception is thrown... As soon as you step foot into `unsafe` land, it's back to C++ :)

Here's a different approach, with no unsafe, but the same idea, though.

Here, the `Wrapper` type keeps the stream in an `Option` so that it can internally move out of it even in a `&mut self` method (by temporarily zeroing out the option). The only way to change the value is by going through the `map_in_place` method, which `take`s the current value from the `Option`, gives it to a closure, but expects the new one to be returned from it. Then, it puts that result back into the option. If the closure panics in this case, the object will have been owned only by the closure and dropped just once during unwinding. (It's all safe Rust, so there couldn't have been any aliasing introduced.) On the other hand, when the closure correctly returns, we're guaranteed that the `Wrapper` always has an object in the `Option`.

The last possibility is that someone catches the unwinding before the `Wrapper` instance itself is dropped. That would mean that the `Wrapper` would be left in a sort of poisoned state where the internal `Option` is unexpectedly `None`. This would cause another panic when any of its methods are subsequently called.

I think this is an improvement over the current version where we can get a panic by dropping the stream early, due to not being careful enough how the stream is juggled around... The only way to lose a stream here is by panicking within one of the methods of `Http11Message`, catching the panic before the message is dropped, and then trying something else with the message.

What do you say? :) (As you can tell, I'd really like to find a way to clean this up :))

seanmonstar:
I like it!

PR #649, opened 2015-09-07
fix(server): use EmptyWriter for status codes that have no body

mikedilger:
Previously, hyper was defaulting to Chunked which adds a Transfer-Encoding
header, whenever there was no Content-Length header. RFC 7230 section 3.3.1
reads:

   ...
   A server MUST NOT send a Transfer-Encoding header field in any
   response with a status code of 1xx (Informational) or 204 (No
   Content). A server MUST NOT send a Transfer-Encoding header field in
   any 2xx (Successful) response to a CONNECT request
   ...

This commit fixes the cases of 1xx (Informational), 204 (No Content) by
using the EmptyWriter. It also uses EmptyWriter for 304 (NotModified) which
should not have a body.

It does NOT address the case of responses to CONNECT requests, or to HEAD
requests which do not send a body.  These cases cannot be determined using
the data available in the response, and are left for future work.

### DIFF
--- a/src/server/response.rs
+++ b/src/server/response.rs
@@ -12,7 +12,7 @@ use time::now_utc;
 
 use header;
 use http::h1::{CR, LF, LINE_ENDING, HttpWriter};
-use http::h1::HttpWriter::{ThroughWriter, ChunkedWriter, SizedWriter};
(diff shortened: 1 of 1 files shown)


seanmonstar:
Nice!

PR #661, opened 2015-10-08
fix(server): use a timeout for Server keep-alive

### LINKED ISSUE #368: Keep-alive connections keep threads occupied infinitely
As discussed in rustless/rustless#35, suggested by @reem, Hyper seems to be too "literal" about keep-alive, keeping threads in a keep-alive loop, essentially DDoSing a Hyper server. My temporary workaround is to use many threads to delay running out of threads for the time being.

seanmonstar:
Server keep-alive is now **off** by default. In order to turn it on, the
`keep_alive` method must be called on the `Server` object.

Closes #368

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -111,8 +111,6 @@ use std::fmt;
 use std::io::{self, ErrorKind, BufWriter, Write};
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::thread::{self, JoinHandle};
-
-#[cfg(feature = "timeouts")]
 use std::time::Duration;
 
 use num_cpus;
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -273,73 +279,97 @@ impl<H: Handler + 'static> Worker<H> {
...
+                break;
+            }
+        }
 
         self.handler.on_connection_end();
 
         debug!("keep_alive loop ending for {}", addr);
     }
 
+    fn set_timeouts(&self, s: & &mut NetworkStream) -> io::Result<()> {
```
seanmonstar: The double indirection is crappy, but using a generic wasn't really possible, since the type is erased into a trait object inside the keep alive loop. And since `NetworkStream` requires `'static`, I couldn't implement the trait for `&mut T`. Fun.

PR #664, opened 2015-10-14
refactor(client): make RequestBuilder non-generic

huonw:
Improve the compile-time of downstream crates that use RequestBuilder,
by not forcing them to remonomorphise and recompile its non-generic
methods when they use it: with this change, they can just call the
precompiled versions in the `hyper` object file(s). The `send` method is
the major culprit here, since it is quite large and complicated.

For an extreme example,

    extern crate hyper;
    fn main() {
        hyper::Client::new().get("x").send().unwrap();
    }

takes ~4s to compile before this patch (i.e. generic RequestBuilder) and
~2s after. (The time spent interacting with LLVM goes from 2.2s to
0.3s.)

BREAKING CHANGE: `RequestBuilder<U>` should be replaced by `RequestBuilder`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ cache:
 script:
   - cargo build --verbose $FEATURES
   - cargo test --verbose $FEATURES
-  - '[ $TRAVIS_RUST_VERSION = nightly ] && cargo bench --no-run || :'
+  - 'if [ $TRAVIS_RUST_VERSION = nightly ]; then cargo bench --no-run; fi'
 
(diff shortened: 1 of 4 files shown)


huonw:
I suppose another approach would be to break `send` into two parts: a small generic part that calls `self.url.into_url()` which then calls a non-generic function that does all the work, but I personally feel like this version is a little nicer.

huonw:
(Based on top of #665 so tests pass on nightly.)

PR #669, opened 2015-10-18
feat(headers): better MIME types and more

pyfisch:
Introduces the `media-types` crate to hyper. It replaces the old `mime`
crate. It has supoort for complex tags and supports charsets and boundary
parameters directly and handles wildcards well. Move `Charset` to the new
crate `charsets`. Update `language-tags` crate.

BREAKING CHANGE: new media types, charsets and updated language tags


pyfisch:
Maybe I should I extend my `MediaType` to also include the shortcut method present for [`Content-Type`](https://github.com/pyfisch/hyper/blob/newmime/src/header/common/content_type.rs#L56-L96) header to create simple media types?

seanmonstar:
Interesting idea! What does the new crate improve on over `mime`?

pyfisch:
It solves many problems `mime` has. While the `mime` crate only knows `TopLevel` and `SubLevel`, my crate splits the type into four parts: type, tree, subtype and suffix. This is closer to the relevant RFCs and allows things like "check if this type is xml based". The crate uses strings instead of incomplete enums because there are so many mime types and every selection of them is somewhat random and later extension is difficult. Both crates support parameters. The `mime` crate uses incomplete enums with very few variants for both keys and values. This means most parameters will use the extension variant and makes the enums no longer useful. The `media-types` crate uses a `HashMap<String, String>` for parameters. It provides methods to access some parameters like `Charset` as strong types. These methods can be extended in the future without harming compatibility.

It supports extended MIME types (https://github.com/hyperium/mime.rs/issues/22), wildcards work (https://github.com/hyperium/mime.rs/issues/28).

seanmonstar:
Hm, I still prefer having specific Rust types for the common cases, as I find Strings to be too fragile...

pyfisch:
Some parts of the type could be enums, the rest has to many variants.

Current:
```rust
/// A Media Type commonly used to describe the contents of a resource.
#[derive(Clone, Debug, Default, Eq, PartialEq)]
pub struct MediaType {
    /// The top-level type or `None` to match all types.
    pub type_: Option<Type>,
    /// A registration tree, `None` matches all trees.
    pub tree: Option<Tree>,
    /// A subtype giving describing a concrete file format.
    pub subtype: Option<String>,
    /// Some types use a suffix to refer to the base format like XML or JSON.
    pub suffix: Option<String>,
    /// Media types can contain optional parameters for example for charsets or video codes.
    pub parameters: HashMap<String, String>
}
```

Proposal:
```rust
/// A Media Type commonly used to describe the contents of a resource.
#[derive(Clone, Debug, Default, Eq, PartialEq)]
pub struct MediaType {
    /// The top-level type or `None` to match all types.
    pub type_: Option<Type>,
    /// The subtype describes specific resource types.
    pub subtype: Option<Subtype>,
    /// Media types can contain optional parameters for example for charsets or video codes.
    pub parameters: HashMap<String, String>
}

/// A subtype, part of a media type.
pub struct Subtype {
     /// A registration tree, the standards tree uses `None`.
    pub tree: Tree,
    /// A subtype giving describing a concrete file format.
    pub subtype: String,
    /// Some types use a suffix to refer to the base format like XML or JSON.
    pub suffix: Option<String>,
}

/// Provides the five discrete and the two composite top-level media types.
pub enum Type {
    /// The "text" top-level type is intended for sending material that is
    /// principally textual in form.
    Text,
    /// A top-level type of "image" indicates that the content specifies one
    /// or more individual images.
    Image,
    /// A top-level type of "audio" indicates that the content contains audio data.
    Audio,
    /// A top-level type of "video" indicates that the content specifies a
    /// time-varying-picture image, possibly with color and coordinated sound.
    Video,
    /// The "application" top-level type is to be used for discrete data that
    /// do not fit under any of the other type names, and particularly for
    /// data to be processed by some type of application program.
    Application,
    /// The "multipart" top-level type is to be used for data consisting of multiple
    /// entities of independent data types.
    Multipart,
    /// A body of media type "message" is itself all or a portion of some
    /// kind of message object.
    Message,
    // TODO: Maybe include `model` type and `example` type?
    /// Less common top-level types.
    Unregistered(String),
}

/// Provides the four registration trees.
pub enum Tree {
    /// The standards tree is intended for types of general interest to the Internet community.
    Standards,
    /// The vendor tree is used for media types associated with publicly available products.
    Vendor,
    /// Registrations for media types created experimentally or as part of
    /// products that are not distributed commercially may be registered in
    /// the personal or vanity tree.
    Personal,
    /// Subtype names with "x." as the first facet may be used for types intended exclusively for
    /// use in private, local environments.
    Private,
    /// Other unofficial trees.
    Unregistered(String),
}
```

Examples:

string           | typed
----------------|-----------
`*/*`             | `MediaType { type_: None, subtype: None, parameters: HashMap::new() }`
`image/png` | `MediaType { type_: Some(Image), subtype: Some(Subtype { tree: Standards, subtype: Some("png".to_owned()), suffix: None }), parameters: HashMap::new() }`

Of course I will add helper methods for easier creation of `MediaTypes`.

seanmonstar:
I'm hoping to edit the mime crate to better support stars, suffixes, etc, and also reducing allocations to none (or 1 if mutable), similar to the Url crate rewrite.

PR #680, opened 2015-11-11
feat(headers): Add Content-Disposition header

### LINKED ISSUE #561: Content-Disposition header
Not strictly part of HTTP/1.1 but comes up in common usage.
See:
* RFC 6266
* Browser conformance tests at: http://greenbytes.de/tech/tc2231/
* IANA assignment at: http://www.iana.org/assignments/cont-disp/cont-disp.xhtml

(I'm just documenting this.  I'm creating a heavily stripped down variant for my multipart/form-data parsing, and don't need a proper one at present).

mikedilger:
fixes #561

This is fairly extensive and would benefit from some review.
It does not try to interpret language-tags (rfc 5646) but rather accepts any string (for now).


src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+                disposition: match &*disposition {
+                    "inline" => DispositionType::Inline,
+                    "attachment" => DispositionType::Attachment,
+                    _ => DispositionType::Ext(disposition),
+                },
+                parameters: Vec::new(),
+            };
+
+            for section in sections {
+                let mut parts = section.split('=');
```
seanmonstar: This should be `sections.splitn(2, '=')`, in case any values are `foo=bar=baz`.

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+///
+/// attr-char     = ALPHA / DIGIT
+///               / "!" / "#" / "$" / "&" / "+" / "-" / "."
+///               / "^" / "_" / "`" / "|" / "~"
+///               ; token except ( "*" / "'" / "%" )
+/// ```
+fn parse_ext_value(val: &str) -> ::Result<(Charset, Option<LanguageTag>, Vec<u8>)> {
+
+    // Break into three pieces separated by the single-quote character
+    let parts: Vec<&str> = val.split('\'').collect();
```
seanmonstar: Could the `val.splitn(3, '\'')` iterator be used, instead of allocating a temporary vec?

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+impl Header for ContentDisposition {
+    fn header_name() -> &'static str {
+        "Content-Disposition"
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> ::Result<ContentDisposition> {
+        parsing::from_one_raw_str(raw).and_then(|s: String| {
+            let mut sections = s.split(';');
+            let disposition = match sections.next() {
+                Some(s) => s.trim().to_ascii_lowercase(),
```
seanmonstar: To save on allocating a new string in lowercase, comparisons can be done using `UniCase`. Such as `if UniCase(s.trim()) == UniCase("inline")`

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+                    _ => DispositionType::Ext(disposition),
+                },
+                parameters: Vec::new(),
+            };
+
+            for section in sections {
+                let mut parts = section.split('=');
+
+                let key = if let Some(key) = parts.next() {
+                    key.trim().to_ascii_lowercase()
```
seanmonstar: Same with `UniCase` here.

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+                let val = if let Some(val) = parts.next() {
+                    val.trim()
+                } else {
+                    return Err(::Error::Header);
+                };
+
+                cd.parameters.push(
+                    match &*key {
+                        "filename" => DispositionParam::Filename(
+                            Charset::Ext("UTF-8".to_owned()), None, unquote_value(val).into_bytes()),
```
seanmonstar: I'm kinda surprised that `Utf8` isn't part of the `Charset` enum...

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+            DispositionType::Attachment => try!(write!(f, "attachment")),
+            DispositionType::Ext(ref s) => try!(write!(f, "{}", s)),
+        }
+        for param in self.parameters.iter() {
+            match param {
+                &DispositionParam::Filename(ref charset, ref opt_lang, ref bytes) => {
+                    let mut use_simple_format: bool = false;
+                    if opt_lang.is_none() {
+                        if let Charset::Ext(ref ext) = *charset {
+                            if ext.to_ascii_lowercase() == "utf-8" {
```
seanmonstar: `UniCase`

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+                    if use_simple_format {
+                        try!(write!(f, "; filename=\"{}\"",
+                                    match String::from_utf8(bytes.clone()) {
+                                        Ok(s) => s,
+                                        Err(_) => return Err(fmt::Error),
+                                    }));
+                    } else {
+                        let langstr = match *opt_lang {
+                            Some(ref lang) => format!("{}",lang),
+                            None => "".to_owned(),
```
seanmonstar: These conditions could write directly to the Formatter, instead of temporary strings.

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+                },
+                &DispositionParam::Ext(ref k, ref v) => try!(write!(f, "; {}=\"{}\"", k, v)),
+            }
+        }
+        Ok(())
+    }
+}
+
+fn unquote_value(val: &str) -> String {
+    if val.chars().next() == Some('"') && val.chars().rev().next() == Some('"') {
```
seanmonstar: This could probably be `val.trim_matches('"')`

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+    };
+
+    // Interpret the third piece as a sequence of value characters
+    let value: Vec<u8> = try!(value_chars_to_bytes(parts[2]));
+
+    Ok( (charset, lang, value) )
+}
+
+/// Parsing of `value-chars`
+fn value_chars_to_bytes(value_chars: &str) -> ::Result<Vec<u8>> {
```
seanmonstar: This function and the one below can utilize [url::percent_encoding](http://servo.github.io/rust-url/url/percent_encoding/index.html).

src/header/common/mod.rs
```diff
@@ -22,6 +22,7 @@ pub use self::allow::Allow;
 pub use self::authorization::{Authorization, Scheme, Basic, Bearer};
 pub use self::cache_control::{CacheControl, CacheDirective};
 pub use self::connection::{Connection, ConnectionOption};
+pub use self::content_disposition::ContentDisposition;
```
seanmonstar: This should also export `DispositionType` and `DispositionParam`, or else no can do much with this header.

src/header/common/content_disposition.rs
```diff
@@ -0,0 +1,384 @@
...
+/// filename-parm       = "filename" "=" value
+///                     | "filename*" "=" ext-value
+///
+/// disp-ext-parm       = token "=" value
+///                     | ext-token "=" ext-value
+///
+/// ext-token           = <the characters in token, followed by "*">
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+pub struct ContentDisposition {
```
seanmonstar: A doc test example here would 1) help people understand how to construct or use this header, and 2) properly test the header is usable from outside the crate.

seanmonstar:
@mikedilger indeed, this is a lot of work! awesome!

I've reviewed the code, though I must say, I'm not too familiar with this header, so I may have missed something.

src/header/common/content_disposition.rs
mikedilger: It is in http://www.iana.org/assignments/character-sets/character-sets.xhtml, maybe we should change the charset enum...

src/header/common/content_disposition.rs
mikedilger: I'll have to push a PR to add another EncodeSet because it's a slightly different set from the 6 or so defined in that crate already.

mikedilger: https://github.com/servo/rust-url/pull/134

mikedilger:
I think I have addressed everything above except using url::percent_encoding, which depends on PR servo/rust-url#134

seanmonstar:
Seems a new version is needed of the url crate?

mikedilger:
Precisely.   I'm in no hurry to get this PR merged as I can use raw headers in the meantime.  But if you like things tidy, @SimonSapin would have to make a release.

SimonSapin:
I just published `url` v0.3.0 on crates.io which includes https://github.com/servo/rust-url/pull/134. Publishing is easy but I sometimes forget to do it. Feel free to ask for it in PRs!

seanmonstar:
@mikedilger I see it's passing now. Awesome!

Could you rebase this into a single commit, `feat(headers): add Content-Disposition ...`? The current commits will make the changelog confusing (fixing bugs that were never released, etc).

mikedilger:
Ok, rebased onto master and collapsed into a single commit

seanmonstar:
Seems another PR updated the Cargo.toml to bump the `url` version already, making a conflict here.

mikedilger:
Ok, rebased again.

seanmonstar:
\o/

PR #685, opened 2015-11-20
chore(cargo): Update url crate to 0.5

untitaker:
~~A recent update in url triggered a version split in my application and
broke my build.~~

~~Since hyper is still compatible with 0.2, I think it might make sense to
still allow that version.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ time = "0.1"
 traitobject = "0.0.1"
 typeable = "0.1"
 unicase = "1.0"
-url = "0.2"
+url = "0.5"
 
 [dependencies.cookie]
 version = "0.1"


untitaker:
Build failure doesn't seem to be related. I'm not sure.

untitaker:
Perhaps wait for https://github.com/servo/rust-url/pull/138

untitaker:
This is now review-ready, turns out pinning only by version-range is a bad idea.

Anyone?

Ryman:
@untitaker for my own reference, what went wrong with a version range?

untitaker:
@Ryman Just problems in principle: You might expose different behavior depending on which url version is installed. Some of that behavior differences are catched by the typesystem and the resulting combination just doesn't compile, but others may be more subtle.

untitaker:
BTW: You'd probably have to bump the minor version after this, since `url` is part of your public API

PR #687, opened 2015-11-20
chore(dependencies): update openssl to 0.7 and cookie to 0.2

### LINKED ISSUE #686: Can we bump the version of the cookie crate from 0.1 to 0.2?
I'm trying to build https://github.com/cyderize/rust-websocket.
Now that fails (see https://github.com/cyderize/rust-websocket/issues/54) because websocket wants a newer version of openssl than hyper/cookie does, which causes a conflict.

Websocket could of course require an older version of ssl, to match hyper. However, in the long run, I'm guessing that hyper will move to new version of cookie though (with already uses the newer version) sooner or later, so I'm suggesting we bump it now.

I tried to do it locally, unfortunately 3 tests failed for me then, so some work is required. If I can work out what that is, I'll put up a PR, but I'm guessing that it's above my level of knowledge.

seanmonstar:
Closes #686

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,11 +25,11 @@ unicase = "1.0"
 url = "0.2"
 
 [dependencies.cookie]
-version = "0.1"
+version = "0.2"
 default-features = false
 
 [dependencies.openssl]
-version = "0.6.4"
+version = "0.7"
 optional = true
 
 [dependencies.solicit]
--- a/src/header/common/cookie.rs
+++ b/src/header/common/cookie.rs
@@ -1,9 +1,6 @@
-use header::{Header, HeaderFormat};
+use header::{Header, HeaderFormat, CookiePair, CookieJar};
 use std::fmt::{self, Display};
 use std::str::from_utf8;
-
-use cookie::Cookie as CookiePair;
-use cookie::CookieJar;
 
 /// `Cookie` header, defined in [RFC6265](http://tools.ietf.org/html/rfc6265#section-5.4)
 ///
--- a/src/header/common/set_cookie.rs
+++ b/src/header/common/set_cookie.rs
@@ -161,7 +161,7 @@ fn test_fmt() {
 
     assert_eq!(
         &headers.to_string()[..],
-        "Set-Cookie: foo=bar; HttpOnly; Path=/p\r\nSet-Cookie: baz=quux; Path=/\r\n");
+        "Set-Cookie: foo=bar; HttpOnly; Path=/p\r\nSet-Cookie: baz=quux\r\n");
 }
 
 #[test]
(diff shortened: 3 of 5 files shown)


Ryman:
Does this not warrant a version bump since it will break downstream crates which depend on `cookies = "0.1"` which interop with hyper?

seanmonstar:
@Ryman how so?

Ryman:
If lib foo depends on cookie 0.1 to build a `CookieJar` and passes it to `hyper::Cookie::from_cookie_jar` then it'll be a version mismatch I think, as cargo will not use the same version between foo and hyper as they are semver incompatible? 

i.e. `error: expected CookieJar, found CookieJar (with different crate versions)`

Ryman:
I've confirmed that when I pull this branch and use it as a local override, then [nickel-cookies](https://github.com/nickel-org/cookies/blob/master/Cargo.toml) fails to build:

```bash
src/cookies.rs:55:52: 55:55 error: mismatched types:
 expected `&cookie::jar::CookieJar<'_>`,
    found `&cookie::jar::CookieJar<'_>`
(expected struct `cookie::jar::CookieJar`,
    found a different struct `cookie::jar::CookieJar`) [E0308]
src/cookies.rs:55                 header::SetCookie::from_cookie_jar(jar)
                                                                     ^~~
src/cookies.rs:55:52: 55:55 help: run `rustc --explain E0308` to see a detailed explanation
src/cookies.rs:55:52: 55:55 note: Perhaps two different versions of crate `cookie` are being used?
src/cookies.rs:55                 header::SetCookie::from_cookie_jar(jar)
```

I don't mind fixing it in this case, but this kind of thing is a breaking change which in some cases may be non-trivial to fix :(

seanmonstar:
Ah, I see. That is irritating. I think the cookie type should probably be exported somewhere in hyper, so that it can be used without also having to depend explicitly on cookie. Such as in `header::shared::Cookie`.

Ryman:
Yeah, it can be a pain. The re-exports should help in future, I think, but (at least at the time) there were suggestions that it might not be a good practice to follow in general though on an issue I logged about this: https://github.com/rust-lang/cargo/issues/1636#issuecomment-105704163

seanmonstar:
Hm, I had read the opposite just recently, that a crate should probably
reexport types that it depends on. I can't recall where I read that,
though.

Hyper has reexported Url since forever, and I haven't heard of issues about
that.

Ryman:
I think it's only an issue when there's a diamond of dependencies, where two crates re-export something from another, and a downstream crate has to use both of these together. I think that's pretty rare though and warrants a more stern look at deciding what upstream versions to support in that case.

So my opinion is that re-exporting cookie (or whatever else) is worth doing. :+1:

Issue #651, opened 2015-09-09
Origin Header


fuchsnj:
It seems like the "Origin" header is missing  [here](http://hyper.rs/hyper/hyper/header/index.html)

http://tools.ietf.org/id/draft-abarth-origin-03.html

seanmonstar:
Is this in common use? I've never used it myself, and when checking the dev consoles, I notice neither Firefox nor Chrome set that header while browsing.

untitaker:
The Origin header is set for CORS requests.

On 15 October 2015 01:36:27 CEST, Sean McArthur <notifications@github.com> wrote:
>Is this in common use? I've never used it myself, and when checking the
>dev consoles, I notice neither Firefox nor Chrome set that header while
>browsing.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/hyperium/hyper/issues/651#issuecomment-148233353

-- 
Sent from my phone. Please excuse my brevity.

seanmonstar:
Ah yes, quite true. I had forgotten.

mfeckie:
@seanmonstar Is anyone working on this?  If not I'd be interested in giving it a go

seanmonstar:
Go for it!

untitaker:
I don't think anybody is working on this.

puhrez:
Would love to get newly-rusty hands wet; what's going on here? 

A variety of PRs have been closed, and one (to another project?)

untitaker:
I think the main discussion is under https://github.com/hyperium/hyper/pull/691, basically we waited until rust-url was 1.0 (where its Origin type was rewritten) and then forgot about it.

On 5 July 2016 01:21:00 EEST, Michael Perez <notifications@github.com> wrote:
>Would love to get newly-rusty hands wet; what's going on here? 
>
>A variety of PRs have been closed, and one (to another project?) 
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/hyperium/hyper/issues/651#issuecomment-230359643

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

seanmonstar:
The current type in rust-url is an enum with an `OpaqueOrigin` variant, which is not needed in the header. So, I'm thinking we do something like this:

```rust
pub struct Origin(String, String, Option<u16>);

// support default ports
impl PartialEq for Origin {
    fn eq(&self, other: &Origin) -> bool {
[...]

PR #690, opened 2015-11-21
Add Origin header for CORS requests

mfeckie:
In response to #651 

I've followed the pattern of using the header! macro as this header seems relatively simple. Happy to adjust if you think it's not the correct approach.


pyfisch:
You have forgotten to include the actual file containing the header.

mfeckie:
@pyfisch Whoops, was fixing up GitCop issue and missed it.  will resbumit

PR #691, opened 2015-11-21
[WIP] feat(headers): Add Origin header for CORS requests

mfeckie:
In response to #651

I've followed the pattern of using the header! macro as this header seems relatively simple. Happy to adjust if you think it's not the correct approach.


seanmonstar:
Thanks for the pull request! Just so you know, you can edit commits with `git rebase -i master`, and then force push to this branch to update the PR.

-----

Looking into the spec, I found [here](http://tools.ietf.org/html/rfc6454#section-7) a more in-depth ABNF, specifically allowing `null` or a space separated list of origins, being `scheme://host[:port]`. Does that match what is used in the wild?

@Manishearth you've done a bunch of the CORS stuff in Servo, is there anything here that should be considered? Or anyone else that should be pinged?

mfeckie:
@seanmonstar Thanks!  I work on so many different things with different policies on rebasing, squashing, merging and particularly *force pushing!*  Good to know what works for you folks.

Happy to adjust to include the additional possibilities. Will await the feedback from the folks you've pinged.

Manishearth:
Origin isn't really a string. It's a parsed URL, except that it doesn't have a path.

https://fetch.spec.whatwg.org/#origin-header

```bnf
Origin                           = origin-or-null

origin-or-null                   = origin / %x6E.75.6C.6C ; "null", case-sensitive
origin                           = scheme "://" host [ ":" port ]
```

An opaque newtype around `Url` with equality and a constructor implemented is what we want here, really.

Manishearth:
Also, fetch and the original Origin RFC seem to differ on whether or not Origin should be an array.

Fetch says that it supplants the original RFC but I'm not sure what that means for hyper.

seanmonstar:
@Manishearth Ah, so the fetch spec doesn't allow for a list of origins, just a single one.

My understanding of your suggestion would look something like this:

```rust
pub struct Origin(Url);

pub enum OriginError {
    Url(url::ParseError),
    PathNotAllowed
}

impl Origin {
    fn new<T: IntoUrl>(t: T) -> Result<Origin, OriginError> {
        let url = try!(t.into_url());
        if url.path().is_some() {
            return Err(OriginError::PathNotAllowed)
        }
        Ok(Origin(url))
    }
}
```

Manishearth:
Well, I was going to suggest something along the lines of

```rust
pub struct Origin(Url); // field is private

impl Origin {
    fn new<T: IntoUrl>(t: T) -> Result<Origin, url::ParseError> {
        let url = try!(t.into_url());
        Ok(Origin(url))
    }
}

impl PartialEq for Origin {
 /// only compare the scheme/host/port (Origin is supposed to be an opaque identifier with equality)
}
```

We can also add a stringifier which strips out the path.

src/header/common/origin.rs
```diff
@@ -0,0 +1,29 @@
+header! {
+    /// 'Origin' request header,
+    /// part of [CORS](http://www.w3.org/TR/cors/#origin-request-header)
```
jdm: We should reference http://tools.ietf.org/html/rfc6454#section-7 instead and ensure that this header implements that specification.

jdm: Sorry, wrote this without seeing the existing conversation.

GitCop:

 * Commit: d48b9ce83ff6b7e1c000ee9933a43df917cd995c

 * Commit: 1e2928b789e5fc3bd7e3f3fe4a21f52a93567d96

---

untitaker:
The url crate already has a Origin type. Perhaps use that one?

mfeckie:
cool! Will .take a look!

seanmonstar:
@untitaker I don't believe it does. At least, browsing through the docs, I couldn't find one.

Manishearth:
https://github.com/servo/rust-url/blob/66545ce19dae6d1dd7fcbce56c471e15a43a0fc3/src/lib.rs#L203

This is the correct way of handling origins (doesn't have the `null` thing though)

seanmonstar:
@Manishearth looks like port is required, but the spec doesn't require it.

Manishearth:
cc @SimonSapin

untitaker:
@seanmonstar No port implies default port, much like URLs.

Better reference: https://tools.ietf.org/html/rfc6454

In response to your question, in particular: https://tools.ietf.org/html/rfc6454#section-4

SimonSapin:
Yes, the port number has a default based on the scheme:

https://url.spec.whatwg.org/#origin
> Return a tuple consisting of URL’s scheme, its host, and its [default port](https://url.spec.whatwg.org/#default-port) if its port is null, and its port otherwise.

https://tools.ietf.org/html/rfc6454#section-4
>  If there is no port component of the URI:
>  Let uri-port be the default port for the protocol given by uri-scheme.

seanmonstar:
I realize there's a default port by scheme. I should have said more. The spec for the `Origin` header says the port is optional. By internally using a structure that doesn't record if the port was present, re-serializing will have different output from the input.

```
let s = "https://rust-lang.org"
assert_eq!(s, Origin::from_str(s).unwrap().to_string())
// "https://rust-lang.org" != "https://rust-lang.org:80"
```

If we chose to always remove the port if it is the default, we still could change the value, since the default port may have been part of the original header.

SimonSapin:
I think the `Origin` type in rust-url was designed for equality testing, possibly without considering serialization or the `Origin` HTTP header. I’d review a rust-url PR to make a change here if you think it’s appropriate.

untitaker:
I don't think a change would be appropriate to Origin. This is about semantics. The representation *could* be nicer in that ports that equal to the default one are omitted, but it really doesn't matter. Also:

```rust
let x = Url::parse("http://example.com:80/").unwrap();
x.serialize()  // http://example.com/
```

seanmonstar:
Unsure if this should be addressed in rust-url. For hyper, it could be enough to define our struct like this:

```rust
struct Origin {
    origin: url::Origin,
    show_port: bool,
}
```

untitaker:
I still don't see why we need to make a distinction between implicit and explicit default ports. The flag is not needed IMO

seanmonstar:
@untitaker I'd like to agree, but I worry humans don't actually implement what the spec says (all too common).

Example: if using `url::Origin`, then in hyper a `Origin::new("http://foo.example")` would return in `Origin(url::Origin("http:", "foo.example", 80))`. Serializing this into a request would look like this:

```
GET / HTTP/1.1
Origin: http://foo.example:80
```

Looking in the CORS code of 2 popular nodejs libraries, I *believe* this would be rejected.

- expressjs: https://github.com/expressjs/cors/blob/ddc7d03e12db8e74d44b73aa6136b90b2bd25640/lib/index.js#L26
- hapijs: https://github.com/hapijs/hapi/blob/492e59496abcbc3d73642e5b1391bd3b8b6b1e1d/lib/cors.js#L183

They both do exact String matches, forgetting that default ports should matter as well.

untitaker:
I would just omit default ports during serialization. The implementations you mentioned already wrongly detect inequalities, and I think making a distinction in hyper as well would make the situation worse.

On 17 December 2015 19:49:13 CET, Sean McArthur <notifications@github.com> wrote:
>@untitaker I'd like to agree, but I worry humans don't actually
>implement what the spec says (all too common).
>
>Example: if using `url::Origin`, then in hyper a
>`Origin::new("http://foo.example")` would return in
>`Origin(url::Origin("http:", "foo.example", 80))`. Serializing this
>into a request would look like this:
>
>```
>GET / HTTP/1.1
>Origin: http://foo.example:80
>```
>
>Looking in the CORS code of 2 popular nodejs libraries, I *believe*
>this would be rejected.
>
>- expressjs:
>https://github.com/expressjs/cors/blob/ddc7d03e12db8e74d44b73aa6136b90b2bd25640/lib/index.js#L26
>- hapijs:
>https://github.com/hapijs/hapi/blob/492e59496abcbc3d73642e5b1391bd3b8b6b1e1d/lib/cors.js#L183
>
>They both do exact String matches, forgetting that default ports should
>matter as well.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/hyperium/hyper/pull/691#issuecomment-165545652

-- 
Sent from my phone. Please excuse my brevity.

seanmonstar:
Closing as inactive.

untitaker:
That's surprising. Which concrete change would you like to see?

seanmonstar:
@untitaker this PR has `struct Origin(pub String);`. The discussion in the PR suggests having an Origin that knows the scheme, host, and port (perhaps from the url crate).

untitaker:
Right, but it wasn't clear at all if you wanted the author to actually use
url's struct (or if you think that their impl was flawed).

seanmonstar:
It may be sufficient to use url's Origin, and in the `fmt_header` method, leave off the port if it is the default port.

untitaker:
Preparing new PR. I'll probably submit the special formatting in rust-url though.

seanmonstar:
I think the `Origin` type in url looks a bit different in the proposed 1.0 branch.

untitaker:
Ah, thanks for the hint, I was about to change some things to master...

@SimonSapin I want to to add a simple `FromStr` impl for `Origin` that just does `url::Url::parse(...).map(|u| u.origin())`, for the purposes of this PR. Should I wait post-1.0?

SimonSapin:
Re earlier comments, the rust-url 1.0 branch (which I hope to merge soon-ish, I still many tests and some bug fixes in setters) now implements the "ASCII serialization of an origin" and "Unicode serialization of an origin" algorithms as defined in the HTML spec. https://github.com/servo/rust-url/blob/1.0/src/origin.rs

@untitaker In what cases do you parse an origin from a string? Does a relevant specification define a parsing algorithm? Parsing as you describe would accept and ignore URL components that are not relevant to origins, e.g. `http://me:secret@example.net:8080/foo?bar#baz` -> `https://example.net:8080`. Maybe the presence of these components should be an error?

untitaker:
@SimonSapin I want to use rust-url's Origin type in Hyper's Origin header, so I need a way to parse it from a raw header value. You're right about extra components, this restriction could probably be applied after parsing the input as URL.

SimonSapin:
https://fetch.spec.whatwg.org/#http-origin and https://tools.ietf.org/html/rfc6454#section-7.1 both define the syntax for the `Origin` header. The former says it supplants the latter. They disagree in the header containing a single origin vs a space-separated list of them but they agree that one origin is `scheme "://" host [ ":" port ]` with each component as in URLs and without extra components like a path.

I’d recommend not using the URL parser. Instead, use this regex `^([a-zA-Z][a-zA-Z0-9.+-]*)://(.+?)(:[0-9]+)?$` (where `+?` is non-greedy repetition) or an equivalent hand-written parser, plus rust-url’s `Host::parse` for the host component.

PR #696, opened 2015-11-24
feat(all): add socket timeouts

seanmonstar:
Methods added to `Client` and `Server` to control read and write
timeouts of the underlying socket.

Keep-Alive is re-enabled by default on the server, with a default
timeout of 5 seconds.

BREAKING CHANGE: This adds 2 required methods to the `NetworkStream`
  trait, `set_read_timeout` and `set_write_timeout`. Any local
  implementations will need to add them.

-----

Adding this in anticipation of bumping the minor version to `0.7`, so including other breaking changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,5 +47,4 @@ env_logger = "*"
 default = ["ssl"]
 ssl = ["openssl", "cookie/secure"]
 serde-serialization = ["serde"]
-timeouts = []
-nightly = ["timeouts"]
+nightly = []
--- a/benches/client.rs
+++ b/benches/client.rs
@@ -7,7 +7,6 @@ extern crate test;
 use std::fmt;
 use std::io::{self, Read, Write, Cursor};
 use std::net::SocketAddr;
-#[cfg(feature = "timeouts")]
 use std::time::Duration;
 
 use hyper::net;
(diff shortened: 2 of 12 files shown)


opensourcegeek:
@seanmonstar I'm trying to use `set_write_timeout` on client, but the client doesn't disconnect after the timeout. I'd like to run bunch of POST requests with big payloads to check the throughput of the connection. When the connectivity is really poor I don't want it to run all the tests. I was wondering if I can set the timeout using this method to close/drop connection after it pushed for X number of seconds?

seanmonstar:
@opensourcegeek No, this makes use of [TcpStream::set_write_timeout](https://doc.rust-lang.org/stable/std/net/struct.TcpStream.html#method.set_write_timeout). That sets a socket option to allow blocking on the write (and read for the other timeout) up to the amount of time specified. The time is not tracked between writes.

opensourcegeek:
@seanmonstar Thanks - I got around by using custom type implementing Read trait and passing it to ChunkedBody. Now I track the time in my custom type which seems to work `Ok` so far.

opensourcegeek:
@seanmonstar I'd written a small [post](https://medium.com/@opensourcegeekz/posting-chunked-body-requests-with-hyper-in-rust-lang-cf2b75b901ed) on how I'm doing chunked body POSTs using hyper, if you could look into it when you get a chance and give me any feedbacks it would be really useful for me. Thanks.

seanmonstar:
@opensourcegeek that looks correct to me!

opensourcegeek:
@seanmonstar great - thanks.

PR #697, opened 2015-11-27
feat(server): Publicize the Timeouts struct for setting all timeouts at once.

reem:


reem:
I am planning to use this in iron to expose keep alive and other timeout settings.

src/server/mod.rs
```diff
@@ -147,11 +147,14 @@ pub struct Server<L = HttpListener> {
...
 
+/// A set of durations representing the various timeouts to be set on a server.
+///
+/// Note: more fields may be added in the future.
 #[derive(Clone, Copy, Debug)]
-struct Timeouts {
-    read: Option<Duration>,
-    write: Option<Duration>,
-    keep_alive: Option<Duration>,
+pub struct Timeouts {
```
untitaker: formatting nitpick

reem: what is the nitpick?

untitaker: Yikes, misread the diff. Sorry for the noise.

seanmonstar:
I originally made that struct as a convenience when passing it around internally. I'm not so sure I'd want it part if a public API... Can you have a struct in Iron to do it?

reem:
I could just duplicate this in iron, but it seemed better to just expose it.

seanmonstar:
I'm not yet sure how timeouts will be set in the mio branch. Could you dupe
the struct in Iron for now? If this struct is kept in hyper, it can be made
public eventually, and you could swap the def for a reexport and it should
continue to work.

reem:
Yeah, that's a good reason, I'll redefine it in iron.



Thanks,
         Jonathan Reem

seanmonstar:
Closing this for now. Can revisit once hyper uses async io, and the timeout strategy is decided upon.

PR #699, opened 2015-11-29
fix(server): Removed check for GET/HEAD request when parsing body 

### LINKED ISSUE #698: Request body should be allowed with GET/HEAD requests
Currently, the body is ignored when the request method is either GET or HEAD.

I had a look through the HTTP 1.1 spec and didn't find any mention of restricting request bodies to certian methods (http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3).

I have also seen this behaviour used in the wild. For example, Elasticsearch uses it for passing the search query when performing a search request (https://www.elastic.co/guide/en/elasticsearch/reference/1.4/search-request-body.html).

Relevant code: https://github.com/hyperium/hyper/blob/master/src/server/request.rs#L44-L45
Introduced in: https://github.com/hyperium/hyper/commit/a60a67cd1e0a67c3348c34b4e3bd933557c07671

kaedroho:
Fixes #698

### DIFF
--- a/src/server/request.rs
+++ b/src/server/request.rs
@@ -9,7 +9,7 @@ use std::time::Duration;
 use buffer::BufReader;
 use net::NetworkStream;
 use version::{HttpVersion};
-use method::Method::{self, Get, Head};
+use method::Method;
 use header::{Headers, ContentLength, TransferEncoding};
 use http::h1::{self, Incoming, HttpReader};
 use http::h1::HttpReader::{SizedReader, ChunkedReader, EmptyReader};
@@ -41,9 +41,7 @@ impl<'a, 'b: 'a> Request<'a, 'b> {
         debug!("Request Line: {:?} {:?} {:?}", method, uri, version);
         debug!("{:?}", headers);
 
-        let body = if method == Get || method == Head {
-            EmptyReader(stream)
-        } else if headers.has::<ContentLength>() {
+        let body = if headers.has::<ContentLength>() {
             match headers.get::<ContentLength>() {
                 Some(&ContentLength(len)) => SizedReader(stream, len),
                 None => unreachable!()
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: d6f1980b1328cf7fe6247f8a2368c53bef2e60da

 * Commit: 650c466da95982595576c7b9615a82cf625275fc

---

src/server/request.rs
```diff
@@ -9,7 +9,7 @@ use std::time::Duration;
 use buffer::BufReader;
 use net::NetworkStream;
 use version::{HttpVersion};
-use method::Method::{self, Get, Head};
+use method::Method::{self};
```
bombless: So this last '::{self}' part can be removed now.

Hmm in this diff it seems that this use statement can be removed.

bombless:
I like this PR as we should do things right.

src/server/request.rs
kaedroho: Fixed + squashed into original commit. Thanks!

src/server/request.rs
```diff
@@ -160,6 +158,24 @@ mod tests {
...
+        let mut mock = MockStream::with_input(b"\
+            GET / HTTP/1.1\r\n\
+            Host: example.domain\r\n\
+            Content-Length: 19\r\n\
+            \r\n\
+            I'm a good request.\r\n\
+        ");
+
+        // FIXME: Use Type ascription
+        let mock: &mut NetworkStream = &mut mock;
```
reem: I think you should be able to write `&mut mock as &mut NetworkStream` here.

reem:
Giving the spec another close read, it seems that both the existing and proposed behaviors are legal, since the spec doesn't tie any defined semantics to GET or HEAD requests with bodies. Since reading the body is more flexible and I've seen GET requests with bodies in the wild, I think we should do this.

seanmonstar:
Indeed, it seems this is better behavior. Thanks for the patch!

PR #700, opened 2015-12-03
feat(status): add StatusCodes scoped by StatusClass

Ryman:
This enables users to limit their api to certain classes of `StatusCode`, e.g. if you only want to accept error codes you can do:

```rust
use hyper::status::StatusCode;
use hyper::status::scoped;

trait HttpError {}

impl HttpError for scoped::ClientError {}
impl HttpError for scoped::ServerError {}

fn only_accept_errors<E: HttpError + Into<StatusCode>>(error: E) {
     let _status_code: StatusCode = error.into();
     // ... use status code here ...
}

// Will not compile
// only_accept_errors(scoped::Informational::Continue);

// These work fine!
only_accept_errors(scoped::ClientError::BadRequest);
only_accept_errors(scoped::ServerError::InternalServerError);
```

This is somewhat similar to the proposal [here](http://smallcultfollowing.com/babysteps/blog/2015/08/20/virtual-structs-part-3-bringing-enums-and-structs-together/), but since we're able to generate the StatusCodes by macro, it's a relatively minor change to get the added benefit, I think it can be worth it to just do it without the language feature assisting.

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -2,441 +2,270 @@
 use std::fmt;
 use std::cmp::Ordering;
 
-// shamelessly lifted from Teepee. I tried a few schemes, this really
-// does seem like the best. Improved scheme to support arbitrary status codes.
+macro_rules! status_codes {
+    ($($(#[$doc:meta])+ ($code:expr, $name:ident, $reason:expr)),+$(,)*) => (
+        // shamelessly lifted from Teepee. I tried a few schemes, this really
+        // does seem like the best. Improved scheme to support arbitrary status codes.
 
-/// An HTTP status code (`status-code` in RFC 7230 et al.).
-///
-/// This enum contains all common status codes and an Unregistered
-/// extension variant. It allows status codes in the range [0, 65535], as any
-/// `u16` integer may be used as a status code for XHR requests. It is
-/// recommended to only use values between [100, 599], since only these are
-/// defined as valid status codes with a status class by HTTP.
-///
-/// If you encounter a status code that you do not know how to deal with, you
-/// should treat it as the `x00` status code—e.g. for code 123, treat it as
-/// 100 (Continue). This can be achieved with
-/// `self.class().default_code()`:
-///
-/// ```rust
-/// # use hyper::status::StatusCode;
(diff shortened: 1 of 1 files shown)


seanmonstar:
This seems clever, but I worry it might be too clever. Let me see. The status codes will be duplicated into the `scoped` module? Such that `StatusCode::BadRequest` and `scoped::ClientError::BadRequest` will exist? I imagine people trying to `*res.status_mut() = ClientError::BadRequest` and wondering why they are getting errors...

reem:
I have the same concerns, this is very cool but might be too much. However, I am leaning towards it since I think the ability to talk about only a class of status codes is pretty interesting.

The problem @seanmonstar raised can be mitigated by providing a `set_status` function which takes an `S: Into<StatusCode>`, though it still is a bit of an ergonomic hurdle. Fortunately even without such an API the fix for that error is just one `.into()`, which is not so bad.

Ryman:
> providing a set_status function which takes an S: Into<StatusCode>, though it still is a bit of an ergonomic hurdle

This is what I would suggest too, seems more flexible and people don't have to remember to deref the mut reference. (Personally, I see `foo.set_status(..)` as more ergonomic than `*foo.status_mut() = ..`)

fwiw, the error when you make the mistake is:
```
expected `hyper::status::StatusCode`,
    found `hyper::status::scoped::ClientError`
(expected enum `hyper::status::StatusCode`,
    found enum `hyper::status::scoped::ClientError`) [E0308]
```

Perhaps adding some suggestion of using `.into()` in the docstring could help? The `From` implementation gets listed under the docs for `StatusCode` while the blanket impl of `Into` doesn't appear on the docpages for the specialisations :(

seanmonstar:
I'm +0. If you two feel strongly that this is an overall win, then @reem can merge at his convenience.

reem:
If we could somehow hint in that error to use Into then I think I would be for it but right now it's just not very newbie friendly.

Ryman:
Okay, should I rebase to the first commit or just close the PR?

seanmonstar:
Closing for now, but I'm not against proposals. Thanks for the idea!

PR #703, opened 2015-12-14
feat(headers): add extended parameter parser to the public API

malept:
Move the extended parameter parser from the `Content-Disposition` header implementation into the common header parsing module. This allows crates that use Hyper to parse [RFC 5987](https://tools.ietf.org/html/rfc5987)-compliant header parameter values.

Add tests based on the examples given in the RFC.

### DIFF
--- a/src/header/common/content_disposition.rs
+++ b/src/header/common/content_disposition.rs
@@ -8,11 +8,11 @@
 
 use language_tags::LanguageTag;
 use std::fmt;
-use std::str::FromStr;
 use unicase::UniCase;
 use url::percent_encoding;
 
 use header::{Header, HeaderFormat, parsing};
+use header::parsing::parse_extended_value;
 use header::shared::Charset;
 
 /// The implied disposition of the content of the HTTP body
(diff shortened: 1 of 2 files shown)


seanmonstar:
Is this sort of thing common in many other headers? If so, perhaps it would make sense to define a struct in the `shared` module, instead of this returning a tuple?

Test failure was unrelated. Also, as this adds a new method (and possibly struct), could reword the commit message to be a `feat(headers)`?

malept:
> Is this sort of thing common in many other headers?

I'm using it for [implementing](https://github.com/malept/guardhaus) [RFC 7616](https://tools.ietf.org/html/rfc7616) (HTTP digest authentication) - the `username` parameter can alternatively be `username*`.

In searching the RFCs, I found references to RFC 5987 in:

* [RFC 5988](https://tools.ietf.org/html/rfc5988) (`Link` header)
* [RFC 6266](https://tools.ietf.org/html/rfc6266) (`Content-Disposition` header)
* [RFC 7616](https://tools.ietf.org/html/rfc7616) (`Authorization: Digest` header)

> If so, perhaps it would make sense to define a struct in the shared module, instead of this returning a tuple?

That sounds reasonable.

> Also, as this adds a new method (and possibly struct), could reword the commit message to be a `feat(headers)`?

Will do. I wasn't sure whether this was a refactor or a feature.

malept:
I've added some tests for the error cases, reworded the commit message per feedback, and added a struct. I wasn't sure whether to call the last member in the struct `value` or `data` (or something else).

seanmonstar:
This looks really good, thanks!

PR #713, opened 2015-12-31
Additional ExtendedValue API

malept:
Adds:
* the reverse of `parse_extended_value()`, via `fmt`
* derived traits, so that `ExtendedValue` can be used, for example, as a struct member
* docs for ExtendedValue members

With the first two features, I've implemented the (de)serialization of the [`username*` parameter for RFC 7616](https://tools.ietf.org/html/rfc7616#section-3.4).

### DIFF
--- a/src/header/parsing.rs
+++ b/src/header/parsing.rs
@@ -56,6 +56,7 @@ pub fn fmt_comma_delimited<T: Display>(f: &mut fmt::Formatter, parts: &[T]) -> f
 
 /// An extended header parameter value (i.e., tagged with a character set and optionally,
 /// a language), as defined in [RFC 5987](https://tools.ietf.org/html/rfc5987#section-3.2).
+#[derive(Clone, Debug, PartialEq)]
 pub struct ExtendedValue {
     pub charset: Charset,
     pub language_tag: Option<LanguageTag>,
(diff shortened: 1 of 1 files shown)


src/header/parsing.rs
```diff
@@ -130,10 +134,27 @@ pub fn parse_extended_value(val: &str) -> ::Result<ExtendedValue> {
...
+impl Display for ExtendedValue {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let encoded_value =
+            percent_encoding::percent_encode(&self.value[..],
+                                             percent_encoding::HTTP_VALUE_ENCODE_SET);
+        if self.language_tag.is_some() {
+            write!(f,
+                   "{}'{}'{}",
+                   self.charset,
+                   self.language_tag.clone().expect("No language tag found"),
```
seanmonstar: We don't need to clone the language tag here (and its Strings and copy them), we can just pass a reference to the write macro.

src/header/parsing.rs
```diff
@@ -130,10 +134,27 @@ pub fn parse_extended_value(val: &str) -> ::Result<ExtendedValue> {
     })
 }
 
+impl Display for ExtendedValue {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let encoded_value =
+            percent_encoding::percent_encode(&self.value[..],
+                                             percent_encoding::HTTP_VALUE_ENCODE_SET);
+        if self.language_tag.is_some() {
```
seanmonstar: Try with an `if let`:

```rust
if let Some(ref lang) = self.language_tag {
    write!(f, "{}'{}'{}", self.charset, lang, encoded_valued)
} else { ... }
```

malept: Thanks, fixed/rebased in da0abe8.

seanmonstar:
Thanks!

Issue #716, opened 2016-01-04
set() method for Headers could send back self


MichaelZapataScality:
This change would allow chaining of the method (or chaining another method), as in:

```rust
response.headers_mut().set(Date(something)).set(Expires(something_else));
```

I could submit a PR for this small change, but I wanted to be sure that this could be a welcomed change. This is non-breaking as there is no returned value at the moment.

The same behaviour could be used by `Headers::set_raw()` too.

jdm:
Would there be any loss of expressivity if the return value is ignored? If not, I'm in favour.

MichaelZapataScality:
No loss per se, but it enforces the use of `;` after a `set()`, meaning that a `match` only setting headers has to be surrounded in brackets.

MichaelZapataScality:
I went ahead and made a PR, the needed changes illustrate how this would end up looking like for anyone using it.

seanmonstar:
Oh yes of course, since it's returning a value, that'd mess with anyone who didn't have a semi-colon in place. Hrm...

MichaelZapataScality:
That also affects simple closures that did not need any brackets beforehand.

But this also avoids having to declare a temporary variable when doing multiple `set()` operations while still being explicit. I'm not familiar enough with hyper to know if that change is useful enough to warrant this breakage though.

seanmonstar:
@MichaelZapataScality I really appreciated the PR. This likely would be useful, but it seems this is indeed a breaking change, so if accepted, I'd wait to merge this for 0.8, with any other breaking changes.

PR #717, opened 2016-01-04
feat(headers): send headers back in set methods

MichaelZapataScality:
Following #716 suggestion, allow the `set()` and `set_raw()` methods to be used in chain.

BREAKING CHANGE: This enforces the use of a semicolon after `set()` and set_raw() methods.

The needed changes in hyper do reflect the possible impact this can have.


seanmonstar:
I'm not so sure of adding this right now. Thanks for the idea, though!

PR #727, opened 2016-02-14
Auto-set auth header if URL contains credentials

Manishearth:
The [URL spec](https://url.spec.whatwg.org/#concept-url-username)
has a concept of a URL's "username" and "password". Since rust-url implements
this spec, we should probably autoconvert instances of this username/password to
a `Basic` auth header.

Blocks https://github.com/servo/servo/pull/9625

Example to try out locally:

 
```rust
extern crate hyper;
extern crate url;
use std::io::Read;

use hyper::Client;
use hyper::header::Connection;
use url::Url;
fn main() {
    // Create a client.
    let mut client = Client::new();

    // Creating an outgoing request.
    let mut url = Url::parse("http://localhost:8000/").unwrap();
    if let Some(scheme_data) = url.relative_scheme_data_mut() {
        scheme_data.username = "manish".into();
        scheme_data.password= Some("hasanawesomepassword".into());
    }
    let mut res = client.get(url)
        // set a header
        .header(Connection::close())
        // let 'er go!
        .send().unwrap();

    // Read the Response.
    let mut body = String::new();
    res.read_to_string(&mut body).unwrap();

    println!("Response: {}", body);
}
```

r? @seanmonstar

### DIFF
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -4,6 +4,7 @@ use std::str::{FromStr, from_utf8};
 use std::ops::{Deref, DerefMut};
 use serialize::base64::{ToBase64, FromBase64, Standard, Config, Newline};
 use header::{Header, HeaderFormat};
+use url::Url;
 
 /// `Authorization` header, defined in [RFC7235](https://tools.ietf.org/html/rfc7235#section-4.2)
 ///
@@ -191,6 +192,20 @@ impl FromStr for Basic {
                 Err(::Error::Header)
             }
         }
+    }
+}
+
+impl Basic {
+    pub fn from_url(url: &Url) -> Option<Basic> {
+        if let Some(scheme) = url.relative_scheme_data() {
+            if scheme.username != "" {
+                return Some(Basic {
+                    username: scheme.username.clone(),
+                    password: scheme.password.clone(),
+                });
+            }
+        }
+        None
     }
 }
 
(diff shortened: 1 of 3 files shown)


GitCop:

 * Commit: 419aeee05e33e1a46b554cca0ef4139e18f885c7

 * Commit: 0e80f7445cffd790583349ac6ab98195057f579b

---

jdm:
Does wireshark show this happening in Firefox?

Manishearth:
When it's entered in the URL bar, it tells me that I'm sending authorization to a site that doesn't need it and complains (and doesn't send the authorization even if I tell it too). Stubborn.

However, if accessing a URL that actually needs basic auth, it diligently converts and sends it.

For example, http://manish:foo@manishearth.anapnea.net/sand/basic.php . This is running the first example from http://php.net/manual/en/features.http-auth.php with no modifications.

It works with XHR too.running

I guess the more important question is -- is this the job of hyper or Servo? The URL spec explicitly mentions usernames in the scheme data, but the HTTP spec doesn't. OTOH, what should hyper do otherwise when fed a url with auth data in it? Ignoring it is confusing and unexpected.

jdm:
Ah, see step 12.5 of https://fetch.spec.whatwg.org/#http-network-or-cache-fetch. Looks like this belongs in Servo.

Manishearth:
Alright. Perhaps I should leave the `from_url` commit here?

seanmonstar:
I agree with jdm, while this is certainly convenient, it's not part of the
HTTP spec (if it is, I can't find it). I'd expect a higher level 'request'
like crate to do several of these things automatically.

PR #729, opened 2016-02-15
feat(client): Implement Debug for Client

sfackler:
Protocol doesn't extend Debug so we have to leave that out of the
output unfortunately.


seanmonstar:
:+1:

PR #730, opened 2016-02-17
Add Headers::content_type_charset

SimonSapin:
This particular parameter is arguably commonly used, and accessing it through other APIs requires a number of imports.

Do you think this is worth having here?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "hyper"
-version = "0.7.2"
+version = "0.7.3"
 description = "A modern HTTP library."
 readme = "README.md"
 documentation = "http://hyperium.github.io/hyper/hyper/index.html"
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -87,6 +87,7 @@ use std::ops::{Deref, DerefMut};
 use std::{mem, fmt};
 
 use {httparse, traitobject};
+use mime;
 use typeable::Typeable;
 use unicase::UniCase;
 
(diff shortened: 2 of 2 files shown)


GitCop:

 * Commit: 268c4ea94410d16dd65f9c6f671d65982d634f55

---

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "hyper"
-version = "0.7.2"
+version = "0.7.3"
```
seanmonstar: I'll bump the version separately, no need to in a PR.

seanmonstar:
It feels weird to be on the Headers struct, but I think it'd be an excellent method for the ContentType struct. `headers.get::<ContentType>().and_then(|ct| ct.charset())`. What do you think of that? No additional imports required.

SimonSapin:
This would be ~half as useful. Doesn’t that code need to import `ContentType`? In addition to imports, the point is to make this more discoverable.

Initially I even wanted this on `Response`, but `some_response.headers` is easy enough to access.

For what it’s worth, the Requests library has a [`Response.encoding`](http://docs.python-requests.org/en/master/api/#requests.Response.encoding) and demonstrates it on their home page: http://docs.python-requests.org/en/master/

seanmonstar:
Interesting idea. I find the encoding property to be confusing, because
there is a Content Encoding header as well. Does it take that into account?

SimonSapin:
No, only content type: https://github.com/kennethreitz/requests/blob/057722af23e/requests/utils.py#L334-L351

I’m guessing that’s because `charset` is misnamed. What’s relevant is not so much the set of supported characters but how to decode them (the encoding).

seanmonstar:
I'd like to rewrite the mime crate soon, and I believe that would change how this works. As such, I'd rather not include something that will soon break, so closing until then.

PR #731, opened 2016-02-17
fix(net): fix the typo in `set_write_timeout`

OlegTsyba:


seanmonstar:
Yikes, thanks for catching this!

PR #736, opened 2016-02-27
refactor(headers): updated serde to version 0.7

debris:


reem:
Hmm something appears to be very wrong on some of the windows builds, they just quit randomly in the middle of the tests with a random-looking exit code.

seanmonstar:
Those have been failing for a couple weeks. I haven't taken the time to
investigate why only those environments crash.

PR #741, opened 2016-03-06
docs(readme): add clippy linting service

ctjhoa:


seanmonstar:
Thanks for the pull request! 

I'm not sure I want to be running all of clippy against hyper. Some of the links are quite useful, but quite a few others I don't fully agree with.

PR #743, opened 2016-03-11
Correctly handle repeated headers.

### LINKED ISSUE #683: Response with multiple `Transfer-Encoding: chunked` header fields not parsed as chunk
I've adapted this code from `client/response.rs`:

```rust
fn main() {
    let stream = MockStream::with_input(b"\
        HTTP/1.1 200 OK\r\n\
        Transfer-Encoding: chunked\r\n\
        Transfer-Encoding: chunked\r\n\
        \r\n\
        1\r\n\
        q\r\n\
        2\r\n\
        we\r\n\
        2\r\n\
        rt\r\n\
        0\r\n\
        \r\n"
    );

    let url = Url::parse("http://hyper.rs").unwrap();
    let mut res = Response::new(url, Box::new(stream)).unwrap();

    let mut s = String::new();
    res.read_to_string(&mut s).unwrap();
    println!("{:?}", s);
}
```

The result is:

`"1\r\nq\r\n2\r\nwe\r\n2\r\nrt\r\n0\r\n\r\n"`

Unless I'm misreading the HTTP spec, it seems

```
HTTP/1.1 200 OK\r\n
Transfer-Encoding: chunked\r\n
Transfer-Encoding: chunked\r\n
```

should be the same as

```
HTTP/1.1 200 OK\r\n
Transfer-Encoding: chunked, chunked\r\n
```

The latter correctly results in `"qwert"` being printed out.

Stebalien:
HeaderX: a
    HeaderX: b

MUST be interpreted as

    HeaderX: a, b

See: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

Fixes #683

### DIFF
--- a/src/header/common/content_length.rs
+++ b/src/header/common/content_length.rs
@@ -79,7 +79,16 @@ __hyper__tm!(ContentLength, tests {
     test_header!(test1, vec![b"3495"], Some(HeaderField(3495)));
 
     test_header!(test_invalid, vec![b"34v95"], None);
-    test_header!(test_duplicates, vec![b"5", b"5"], Some(HeaderField(5)));
+
+    // Can't use the test_header macro because "5, 5" gets cleaned to "5".
+    #[test]
+    fn test_duplicates() {
+        let parsed = HeaderField::parse_header(&[b"5"[..].into(),
+                                                 b"5"[..].into()]).unwrap();
+        assert_eq!(parsed, HeaderField(5));
+        assert_eq!(format!("{}", parsed), "5");
+    }
+
     test_header!(test_duplicates_vary, vec![b"5", b"6", b"5"], None);
 });
 
--- a/src/header/parsing.rs
+++ b/src/header/parsing.rs
@@ -23,24 +23,18 @@ pub fn from_raw_str<T: str::FromStr>(raw: &[u8]) -> ::Result<T> {
 
 /// Reads a comma-delimited raw header into a Vec.
 #[inline]
-pub fn from_comma_delimited<T: str::FromStr>(raw: &[Vec<u8>]) -> ::Result<Vec<T>> {
-    if raw.len() != 1 {
-        return Err(::Error::Header);
+pub fn from_comma_delimited<T: str::FromStr, S: AsRef<[u8]>>(raw: &[S]) -> ::Result<Vec<T>> {
+    let mut result = Vec::new();
+    for s in raw {
+        let s = try!(str::from_utf8(s.as_ref()));
+        result.extend(s.split(',')
+                      .filter_map(|x| match x.trim() {
+                          "" => None,
+                          y => Some(y)
+                      })
+                      .filter_map(|x| x.parse().ok()))
     }
-    // we JUST checked that raw.len() == 1, so raw[0] WILL exist.
-    from_one_comma_delimited(& unsafe { raw.get_unchecked(0) }[..])
-}
-
-/// Reads a comma-delimited raw string into a Vec.
-pub fn from_one_comma_delimited<T: str::FromStr>(raw: &[u8]) -> ::Result<Vec<T>> {
-    let s = try!(str::from_utf8(raw));
-    Ok(s.split(',')
-        .filter_map(|x| match x.trim() {
-            "" => None,
-            y => Some(y)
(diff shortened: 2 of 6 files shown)


GitCop:

 * Commit: 2ba4227d09656a5b7f2f0815d69e3c62869fdc9e

---

src/header/common/content_length.rs
```diff
@@ -79,7 +79,6 @@ __hyper__tm!(ContentLength, tests {
     test_header!(test1, vec![b"3495"], Some(HeaderField(3495)));
 
     test_header!(test_invalid, vec![b"34v95"], None);
-    test_header!(test_duplicates, vec![b"5", b"5"], Some(HeaderField(5)));
     test_header!(test_duplicates_vary, vec![b"5", b"6", b"5"], None);
```
Stebalien: Why did you test this case?

src/header/parsing.rs
```diff
@@ -23,24 +23,18 @@ pub fn from_raw_str<T: str::FromStr>(raw: &[u8]) -> ::Result<T> {
 
 /// Reads a comma-delimited raw header into a Vec.
 #[inline]
-pub fn from_comma_delimited<T: str::FromStr>(raw: &[Vec<u8>]) -> ::Result<Vec<T>> {
-    if raw.len() != 1 {
-        return Err(::Error::Header);
+pub fn from_comma_delimited<T: str::FromStr, S: AsRef<[u8]>>(raw: &[S]) -> ::Result<Vec<T>> {
+    let mut result = Vec::new();
```
Stebalien: Technically, this now allows raw to have length 0. Is that OK?

seanmonstar: I'd like to believe that `httparse` will error if there is a header name with no value, such as `Host:\r\nContent-Length: 10\r\n\r\n`. Which, if that's the case, then `raw` could never be 0.

src/header/common/content_length.rs
seanmonstar: For the `test_duplicates` test, from [RFC7230#3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2):

> If a message is received that has multiple Content-Length header
   fields with field-values consisting of the same decimal value, or a
   single Content-Length header field with a field value containing a
   list of identical decimal values (e.g., "Content-Length: 42, 42"),
   indicating that duplicate Content-Length header fields have been
   generated or combined by an upstream message processor, then the
   recipient MUST either reject the message as invalid or replace the
   duplicated field-values with a single valid Content-Length field
   containing that decimal value prior to determining the message body
   length or forwarding the message.

For the `test_duplicates_vary` test, step 4 of [RFC7230#3.3.3](https://tools.ietf.org/html/rfc7230#section-3.3.3):

> If a message is received without Transfer-Encoding and with
       either multiple Content-Length header fields having differing
       field-values or a single Content-Length header field having an
       invalid value, then the message framing is invalid and the
       recipient MUST treat it as an unrecoverable error.

The request should error out with varying Content-Lengths, which hyper doesn't do, but at the very least, it shouldn't pick any of the stated values as valid.

src/header/parsing.rs
Stebalien: So can I debug_assert this?

src/header/common/content_length.rs
Stebalien: Ok, this change never picks any of the stated values even if they are all identical.

seanmonstar: It looks like the parsing of `ContentLength` has been left alone, so this test should still pass. I'd rather keep the functionality of `ContentLength` the same for now.

seanmonstar:
This looks fantastic. If you could add the `ContentLength` test back in, this will be good to go!

src/header/common/content_length.rs
Stebalien: Good point, my diagnosis was wrong (guessing is bad). This change doesn't change the behavior of `ContentLength` as you say. However, it changes the behavior of the `test_header` macro to correctly verify that parsing and then formatting a header returns the original header. However, in this case, hyper will "clean" the header removing the second content length so this test will fail.

seanmonstar:
Thanks!

PR #744, opened 2016-03-14
feat(status): add HTTP statuses 421 and 451

jongiddy:
The official IANA list at
http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
has two new status codes that are missing from the hyper status list.


seanmonstar:
Excellent, thanks!

PR #750, opened 2016-03-24
feat(header): add prefer and preference applied headers

### LINKED ISSUE #747: Add prefer header
I'd like to use the [`Prefer`](http://tools.ietf.org/html/rfc7240) header, could types for this header be added?

calebmer:
Closes #747 

The specification can be found [here](http://tools.ietf.org/html/rfc7240).


seanmonstar:
Oh wow, this is extensive. Thanks for being thorough!

I was wondering in the issue #747 whether many people use this, because I was wondering if it should be part of hyper, or if the implementation can live in your own crate.

calebmer:
After a quick Google search, I found [Microsoft Azure](https://msdn.microsoft.com/en-us/library/azure/dn535596.aspx) uses it ([here](https://msdn.microsoft.com/en-us/library/hh537533.aspx) is there specific extensions). Of course, as I mentioned in #747 the popular (~6,500 stars) [PostgREST](https://github.com/begriffs/postgrest) REST API uses it.

And a few other links of people using it that my Google search turned up:
- https://github.com/Reading-eScience-Centre/coverage-restapi/blob/master/spec.md#61-prefer-header-method-for-embedding-data
- https://github.com/davidyack/Xrm.Tools.CRMWebAPI/commit/fd54a5a2fec221956f4b5a238751cba4794e2ba3

calebmer:
I'm not sure why the test is failing, could you shed some light on that?

seanmonstar:
@calebmer seems the error was an intermittent. I restarted the build. I've also asked on IRC for some others to put feedback, as I have no experience with this header :)

src/header/common/prefer.rs
```diff
@@ -0,0 +1,201 @@
...
+        }, f)
+    }
+}
+
+impl FromStr for Preference {
+    type Err = Option<<u32 as FromStr>::Err>;
+    fn from_str(s: &str) -> Result<Preference, Option<<u32 as FromStr>::Err>> {
+        use self::Preference::*;
+        let mut params = s.split(';').map(|p| {
+          let mut param: Vec<_> = p.splitn(2, '=').collect();
```
seanmonstar: Just use the split iterator instead of allocating this temporary vector. It'd change the match to `match (params.next(), params.next())`.

seanmonstar: Oh, and bump the indent for this closure please!

src/header/common/prefer.rs
```diff
@@ -0,0 +1,201 @@
...
+
+__hyper__deref!(Prefer => Vec<Preference>);
+
+impl Header for Prefer {
+    fn header_name() -> &'static str {
+        "Prefer"
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> ::Result<Prefer> {
+        let preferences = raw.iter()
```
seanmonstar: I believe this can use the fixed [`from_comma_delimited`](https://github.com/hyperium/hyper/blob/70c6914217a9b48880e61b7fb59acd15c6e1421e/src/header/parsing.rs#L26)

seanmonstar:
@calebmer I'll merge this after comments are addressed. Thanks again for such a thorough job.

calebmer:
Sorry about the indentation, my editor is set to two spaces and looks for an `.editorconfig` for other levels of indentation.

Comments addressed :blush::+1:

seanmonstar:
Excellent! Thank you!

PR #757, opened 2016-04-06
feat(net): Split Ssl into SslClient and SslServer

### LINKED ISSUE #756: Separate hyper::net::Ssl traits for server and client usage
Should help with https://github.com/hyperium/hyper/issues/752 and https://github.com/hyperium/hyper/issues/755

sfackler:
SSL libraries other than OpenSSL don't necessarily have the ability to
easily create server and client side connections at the same time. This
is backwards compatible due to blanket impls.

Closes #756

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -410,7 +410,9 @@ impl<F> NetworkConnector for F where F: Fn(&str, u16, &str) -> io::Result<TcpStr
     }
 }
 
-/// An abstraction to allow any SSL implementation to be used with HttpsStreams.
+/// Deprecated
+///
+/// Use `SslClient` and `SslServer` instead.
 pub trait Ssl {
     /// The protected stream.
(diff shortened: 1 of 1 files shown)


src/net.rs
```diff
@@ -493,7 +527,7 @@ impl<S: NetworkStream> NetworkStream for HttpsStream<S> {
 
 /// A Http Listener over SSL.
 #[derive(Clone)]
-pub struct HttpsListener<S: Ssl> {
+pub struct HttpsListener<S: SslServer> {
```
frewsxcv: Will this break constructing `HttpsListener` for users who don't have the `SslServer` imported?

seanmonstar: I don't believe so. You don't need to have the trait imported for a type to implement it.

frewsxcv: Yeah, that sounds right on second thought

PR #758, opened 2016-04-06
feat(client): handle errors in Client.request

jwilm:
Previously, failing to send on the client channel or receiving a wakeup
error would panic the current thread. The errors are now exposed so that
the consumer can take appropriate action instead of panicking.
Additionally, the WakeupError::Full case is handled by trying to wake up
the machine in a loop. This error was originally going to be returned
with the others (in a different variant), but the consumer of the client
has no recourse for retrying a loop wakeup other than to send another
request which could fail as well.

This might not be the right answer for error handling here, but it's a step in the right direction from panicking when the client looks at you funny. At the very least, this is hopefully enough to start a discussion about what to do with such errors.

This was motivated by some stress testing I was doing on the mio branch. Sending 200k requests to my local nginx server would panic once the loop was full. They all complete once we properly handle the loop wakeup. Although, I'm not sure we need to keep retrying in a loop. It may be that, since the loop does have items in the queue already, and it was notified at least once before, that the queue would get flushed anyway. Again, I would need to dig more into rotor to verify that. The docs didn't really spell it out.

---
This change is [Reviewable](https://reviewable.io/reviews/hyperium/hyper/758)

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -70,7 +70,7 @@ fn main() {
 
     let (tx, rx) = mpsc::channel();
     let client = Client::new().expect("Failed to create a Client");
-    client.request(url.parse().unwrap(), Dump(tx));
+    client.request(url.parse().unwrap(), Dump(tx)).unwrap();
 
     let mut i =-0;
     loop {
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 23b62cc90b01a645a14f4182e1de40327505999f

---

jwilm:
This has been resolved through a separate commit

PR #759, opened 2016-04-14
feat(client): introduce a RedirectCounter policy

bltavares:
To prevent requests getting stuck on a infinite loop, this commit
introduces a redirect count policy.

The policy will take a number of times it should follow redirects, and
then return the response on the last hop.

This commit includes tests when the count is suficient to reach the OK
and when the count is not enough.


seanmonstar:
Closing per conversation that async hyper is removing `RedirectPolicy`.

PR #762, opened 2016-04-16
Add SslClient and SslServer implementations for security-framework

sfackler:


GitCop:

 * Commit: 4bf4accc5a4b31724f2858858962bf741dde025a

 * Commit: 6cf43c1bd0b72349de23530dddef0aadca37557c

---

jdm:
💃

src/net.rs
```diff
@@ -593,14 +593,17 @@ impl<S: SslClient> NetworkConnector for HttpsConnector<S> {
...
+#[cfg(all(not(feature = "openssl"), not(feature = "security-framework")))]
 #[doc(hidden)]
 pub type DefaultConnector = HttpConnector;
 
 #[cfg(feature = "openssl")]
 #[doc(hidden)]
 pub type DefaultConnector = HttpsConnector<self::openssl::Openssl>;
 
+#[cfg(all(feature = "security-framework", not(feature = "openssl")))]
+pub type DefaultConnector = HttpsConnector<self::security_framework::ClientWrapper>;
```
seanmonstar: Openssl feature would still be enabled to provide the secure feature of cookies. So if you enable both, you'd be using the Openssl connector, right?

seanmonstar:
\o/

src/net.rs
sfackler: If you're using that feature, yeah, though that's a thing that we can fix on the cookie side. But it seems a bit unlikely that you'd be enabling that feature anyway if you're using the client side of hyper.

PR #775, opened 2016-04-27
fix(client): fix Host header when using a Proxy

### LINKED ISSUE #774: Proxy handling does not properly set destination Host header
Here is a script transcript of a working proxy session

    Script started on Wed 27 Apr 2016 09:32:58 AM PDT
    [~/repos/rust/hyper-upstream]
    ~$ telnet my-proxy 8080
    Trying x.x.x.x...
    Connected to my-proxy.
    Escape character is '^]'.
    GET http://xkcd.com http/1.1
    host: xkcd.com:80

    HTTP/1.1 200 OK
    Cache-Control: public, max-age=300
    Expires: Wed, 27 Apr 2016 16:23:15 GMT
    Content-Type: text/html; charset=utf-8
    ETag: "1703862119"
    Last-Modified: Wed, 27 Apr 2016 05:16:40 GMT
    Server: lighttpd/1.4.28
    Content-Length: 6196
    Accept-Ranges: bytes
    Date: Wed, 27 Apr 2016 16:18:15 GMT
    Via: 1.1 varnish
    X-Served-By: cache-dfw1838-DFW
    X-Cache: HIT
    X-Cache-Hits: 1
    X-Timer: S1461772234.306036,VS0,VE0
    Vary: Accept-Encoding
    Proxy-Connection: Keep-Alive
    Connection: Keep-Alive
    Age: 971

Note the destination is specified TWICE. Once on the GET and once again in the Host line. The current client code does not set the Host correctly.

Here is the diff needed to fix it:

    diff --git a/src/client/mod.rs b/src/client/mod.rs
    index 4ad5c2d..ad0b14d 100644
    --- a/src/client/mod.rs
    +++ b/src/client/mod.rs
    @@ -271,10 +271,10 @@ impl<'a> RequestBuilder<'a> {

             loop {
                 let mut req = {
    +                let hp = try!(get_host_and_port(&url));
                     let (scheme, host, port) = match client.proxy {
                         Some(ref proxy) => (proxy.0.as_ref(), proxy.1.as_ref(), proxy.2),
                         None => {
    -                        let hp = try!(get_host_and_port(&url));
                             (url.scheme(), hp.0, hp.1)
                         }
                     };
    @@ -282,11 +282,14 @@ impl<'a> RequestBuilder<'a> {
                         Some(ref headers) => headers.clone(),
                         None => Headers::new(),
                     };
    +                // Host needs to be the destination host not the proxy
                     headers.set(Host {
    -                    hostname: host.to_owned(),
    -                    port: Some(port),
    +                    hostname: hp.0.to_owned(),
    +                    port: Some(hp.1),
                     });
    +                // now connect to the proxy
                     let message = try!(client.protocol.new_message(&host, port, scheme));
    +                // and make a request for the destination
                     Request::with_headers_and_message(method.clone(), url.clone(), headers, messag\
e)
                 };

seanmonstar:
Closes #774

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -20,7 +20,20 @@ fn main() {
         }
     };
 
-    let client = Client::new();
+    let client = match env::var("HTTP_PROXY") {
+        Ok(mut proxy) => {
+            // parse the proxy, message if it doesn't make sense
+            let mut port = 80;
+            if let Some(colon) = proxy.rfind(':') {
+                port = proxy[colon + 1..].parse().unwrap_or_else(|e| {
+                    panic!("HTTP_PROXY is malformed: {:?}, port parse error: {}", proxy, e);
+                });
+                proxy.truncate(colon);
+            }
+            Client::with_http_proxy(proxy, port)
+        },
+        _ => Client::new()
+    };
 
     let mut res = client.get(&*url)
         .header(Connection::close())
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -71,10 +71,12 @@ use method::Method;
 use net::{NetworkConnector, NetworkStream};
 use Error;
 
+use self::proxy::tunnel;
 pub use self::pool::Pool;
 pub use self::request::Request;
 pub use self::response::Response;
 
+mod proxy;
 pub mod pool;
 pub mod request;
 pub mod response;
@@ -90,7 +92,7 @@ pub struct Client {
     redirect_policy: RedirectPolicy,
     read_timeout: Option<Duration>,
     write_timeout: Option<Duration>,
-    proxy: Option<(Cow<'static, str>, Cow<'static, str>, u16)>
+    proxy: Option<(Cow<'static, str>, u16)>
 }
 
 impl fmt::Debug for Client {
@@ -114,6 +116,15 @@ impl Client {
     /// Create a new Client with a configured Pool Config.
     pub fn with_pool_config(config: pool::Config) -> Client {
         Client::with_connector(Pool::new(config))
+    }
+
+    pub fn with_http_proxy<H>(host: H, port: u16) -> Client
+    where H: Into<Cow<'static, str>> {
+        let host = host.into();
+        let proxy = tunnel((host.clone(), port));
+        let mut client = Client::with_connector(Pool::with_connector(Default::default(), proxy));
+        client.proxy = Some((host, port));
+        client
     }
 
     /// Create a new client with a specific connector.
@@ -146,12 +157,6 @@ impl Client {
     /// Set the write timeout value for all requests.
     pub fn set_write_timeout(&mut self, dur: Option<Duration>) {
         self.write_timeout = dur;
-    }
-
-    /// Set a proxy for requests of this Client.
-    pub fn set_proxy<S, H>(&mut self, scheme: S, host: H, port: u16)
-    where S: Into<Cow<'static, str>>, H: Into<Cow<'static, str>> {
-        self.proxy = Some((scheme.into(), host.into(), port));
     }
 
     /// Build a Get request.
@@ -271,13 +276,12 @@ impl<'a> RequestBuilder<'a> {
 
         loop {
             let mut req = {
-                let (scheme, host, port) = match client.proxy {
-                    Some(ref proxy) => (proxy.0.as_ref(), proxy.1.as_ref(), proxy.2),
-                    None => {
-                        let hp = try!(get_host_and_port(&url));
-                        (url.scheme(), hp.0, hp.1)
-                    }
-                };
+                let (host, port) = try!(get_host_and_port(&url));
+                let mut message = try!(client.protocol.new_message(&host, port, url.scheme()));
+                if url.scheme() == "http" && client.proxy.is_some() {
+                    message.set_proxied(true);
+                }
+
                 let mut headers = match headers {
                     Some(ref headers) => headers.clone(),
                     None => Headers::new(),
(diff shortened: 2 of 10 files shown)


seanmonstar:
@shaleh do you mind testing that this works with your proxy?

shaleh:
This works for HTTP but fails for HTTPS. I get a response back from the proxy server telling me the request is invalid.

    Response: 400 Bad Request
    Headers:
    Pragma: no-cache
    Content-Length: 691
    Cache-Control: no-cache
    Content-Type: text/html; charset=utf-8
    Proxy-Connection: close
    Connection: close

seanmonstar:
To support `https` end points, I'll need to work in an initial `CONNECT example.dom HTTP/1.1` dance before assuming the connection can be wrapped in SSL... Sigh, I was hoping this would just be a minor fix. Although, I think doing the CONNECT stuff may be doable outside of hyper...

shaleh:
This is the sequence that works via telnet/nc

`
    CONNECT www.google.com:80 HTTP/1.0
    User-Agent: simple/1.1
    Host: www.google.com

    GET / HTTP/1.0
`

Note the use of a port number instead of "http://". When I tried to use http://www.example.com I would get back an invalid request response.
 
Since you have the `is_proxied` boolean now, calling the CONNECT as a preamble should be easy enough. This is what other C based http libraries are doing. curl is here https://github.com/curl/curl/blob/master/lib/http_proxy.c

shaleh:
Exchanging the 80 for 443 should be enough to get SSL working.

seanmonstar:
Yea, the issue is that by the time the code in `Http11Message` is running to write the data, the connection has already been made. However, once `CONNECT` has been passed, then the ssl handshake should take place to access the site over HTTPS.

I'm now thinking of composing `NetworkConnectors` to get this.

```rust
struct Tunnel<C> {
    connector: C,
    proxy: (String, String, u16),
}

type HttpToHttpsProxy = HttpsConnector<Openssl, Tunnel<HttpConnector>>;

let http_to_https_proxy = HttpsConnector::new(
    Openssl::default(),
    Tunnel::new(
        HttpConnector, "http", "my-proxy", 8080
    )
);
```

Steps that would occur:

1. `Client.get("https://www.google.com")`
2. `let tcp = tunnel.connector.connect(tunnel.proxy)`
3. `tcp.write_all("CONNECT www.google.com:443 HTTP/1.1\r\nHost: www.google.com:443\r\n\r\n")`
4. Check for `HTTP/1.1 20x` response on tcp
5. https_connector.wrap_client(tcp, "www.google.com")
6. Carry on like normal.

shaleh:
Are you thinking for making a ``Client`` using something like ``Client.new_with_proxy(host, port)``?

shaleh:
Or, HttpsConnector (and HttpConnector) could have a ``set_proxy()`` method that the Client could invoke when its own ``set_proxy`` is called.

seanmonstar:
It does seem that this is entirely implementable with the tools hyper provides, but I suppose this can be provided to ease usage of proxies for the easy cases. Those cases seem to be:

1. `http` to the proxy, and then `https` to the target
2. `http` to the proxy, `http` to the target
3. `https` to the proxy, and then `http` to the target
4. `https` to the proxy, `http` to the target

As before, any attempt to connect to the proxy or the target over https without the `ssl` feature would just result in an `unknown scheme` error.

Next, to set it up on the Client, it could look like this:

```rust
// if target-uri is https, tunnels over http to proxy, otherwise does absolute-uri as in current PR
// if ssl feature is disabled, requests to https would return `Err(unknown scheme)`
let client = Client::with_proxy("http", "my-proxy", 8080);
```

seanmonstar:
And thanks to a mistake in designing the `Ssl` trait, I cannot easily compose these connectors because I cannot make an `SslStream<SslStream<HttpStream>>` using the `Ssl` trait, since `HttpStream` is hard coded in.

Though just think of that sounds horrible. Double encryption yum.

seanmonstar:
Ah, seems that is a job for NPN...

shaleh:
NPN?

seanmonstar:
Next Protocol Negotiation, a recent addition to TLS. It allows converting a TLS connection into some other connection (it was originally created for SPDY/HTTP2 so clients can upgrade TLS to h2).

But that's a lot of work, and from my googling, the majority of people use HTTP proxies, not HTTPS, so I'm going to settle by only providing to HTTP out of the box (I want to get back to the async work).

```rust
Client::with_http_proxy("my-proxy", 8080)
```

It will always try the `http` scheme, being an http proxy.

shaleh:
But it still proxies for connections to HTTPS sites, right? Lots and lots of websites are HTTPS these days.

seanmonstar:
Yes, it will tunnel to the HTTP proxy, and negotiate TLS after the proxy has tunneled tcp.

shaleh:
Cool. Let me know when you need me to play QA again :-)

seanmonstar:
I'm not super pleased with the code, but it seems to work. Tests pass, and I edited the `examples/client.rs` to look for `HTTP_PROXY` env var, allowing me to to easily compare it's results with `curl -x http://my-proxy:8080`. My tests against real proxies seemed to work (or at least, if it worked with curl, it worked in hyper. Failures in hyper were also failures in curl, so... 🍌 ).

seanmonstar:
@shaleh does it work with your corporate proxy?

With this PR, you can do `HTTP_PROXY=my-proxy:80 cargo run --example client -- https://www.google.com`

shaleh:
I will check it tomorrow. I had planned today but was busy.

shaleh:
I get a failure:

    thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Error { repr: Custom(Custom { kind: Other, error: StringError("failed to lookup address information: Name or service not known") }) })', ../src/libcore/result.rs:746

wget, curl, etc. all work. I have not had the time to debug this. Sorry.

seanmonstar:
@shaleh The value provided does not include the scheme, right? The code in the example doesn't do as much parsing as it could have the `HTTP_PROXY` env var.

shaleh:
Bingo. I have had problems with other software failing if the variable left off the ``http://``.

PR #793, opened 2016-05-17
feat(net): Add OpensslClient constructor

### LINKED ISSUE #792: Openssl vs opensslclient
I recently noticed that the ssl trait was deprecated. I've been using the openssl struct as a network connector for clients. I seems that is just an impl for the deprecated form of secure network connectors. I looked at opensslclient, an impl of sslclient, but it seems it's ssl context member is private, meaning I can only use the default impl. Is this by design or should I be able to provide my own ssl context? If it's just a matter of adding a pub keyword to it I can submit that pull

softprops:
fixes #792


softprops:
Hrm. It looks like this pr failed the build checks but the error looks unrelated to the [error](https://ci.appveyor.com/project/seanmonstar/hyper/build/0.9.master.462/job/7hcpacoffdmd5qge). Does this look familiar?

seanmonstar:
The windows builds are currently expected now that master is using async :(

I'll merge this and cherry pick it into 0.9 as well, thanks!

softprops:
Woot! Thanks so much for such an awesome project @seanmonstar I can't wait for the async stuff

PR #808, opened 2016-05-27
Without this semicolon, multiple threads didn't spawn properly

ktsujister:
I'm not exactly sure of the reason, but without this semicolon, the example only used one thread.
After adding the semicolon, threads spawned up as expected.

```
$ rustc --version
rustc 1.9.0 (e4e8b6668 2016-05-18)
```
Checked the threads in `htop` command.
And I'm using ubuntu. Please let me know if you need more details.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -42,7 +42,7 @@ fn main() {
         let listener = listener.try_clone().unwrap();
         handles.push(::std::thread::spawn(move || {
             Server::new(listener)
-                .handle(|_| Hello).unwrap()
+                .handle(|_| Hello).unwrap();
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 884cf850cb48bde34dfa49a6ab8cec5891722330

---

ktsujister:
I modified the commit message to follow the style.

seanmonstar:
Ah, I see what is happening. Without the semi-colon, it's passing the ServerLoop back to the main thread, because of the call to `join()`, instead of the loop being dropped (and thus running forever).

Thanks!

PR #810, opened 2016-05-29
client: Make it clear what the `Read` impl for `Response` does

crumblingstatue:
In case someone reads the documentation for `client::Response` without
having read the documentation for the `client` module first, they might
wonder what the `Read` impl actually reads.

Let's clarify it by annotating the `Read` impl.

### DIFF
--- a/src/client/response.rs
+++ b/src/client/response.rs
@@ -64,6 +64,7 @@ impl Response {
     }
 }
 
+/// Read the response body.
 impl Read for Response {
     #[inline]
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {


GitCop:

 * Commit: 7f975b0f3c232c044b8f57a4d0572dde159829f7

---

crumblingstatue:
What the hell. How did the coverage decrease by adding a doc comment?

seanmonstar:
@crumblingstatue because coveralls changes its mind frequently XD

What is the comment clarifying? I'm not against more comments, just I wonder if this could be made clearer still.

crumblingstatue:
> What is the comment clarifying? I'm not against more comments, just I wonder if this could be made clearer still.

I explained it in the commit message.
If one just reads the doc page for `Response`, they can only see that `Response` implements `Read`.
But it might not be immediately apparent what the `Read` implementation actually reads. This comment clarifies that the `Read` implementation is for reading the response body.

seanmonstar:
Makes sense, thanks!

PR #817, opened 2016-06-04
Clippy run and derive Default for Buffer

leodasvacas:
Some refactorings, most were suggested by clippy.

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -88,7 +88,7 @@ impl Connect for HttpConnector {
         if let Some(key) = self.key(url) {
             let host = url.host_str().expect("http scheme must have a host");
             self.dns.as_ref().expect("dns workers lost").resolve(host);
-            self.resolving.entry(host.to_owned()).or_insert(Vec::new()).push(key.clone());
+            self.resolving.entry(host.to_owned()).or_insert_with(Vec::new).push(key.clone());
             Ok(key)
         } else {
             Err(io::Error::new(io::ErrorKind::InvalidInput, "scheme must be http"))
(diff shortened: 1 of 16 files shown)


GitCop:

 * Commit: e100e6093ec5bdf2394dc1b28cb636deba4a2972

---

src/status.rs
```diff
@@ -28,7 +28,7 @@ use std::cmp::Ordering;
 /// Registry](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) which is
 /// the source for this enum (with one exception, 418 I'm a teapot, which is
 /// inexplicably not in the register).
-#[derive(Debug, Hash)]
+#[derive(Debug, Hash, Copy, Clone)]
```
seanmonstar: The manual implementations were deliberate actually, because it's faster to compile than rustc generating the huge match statements for this big enum...

seanmonstar:
Thanks for the pull request! Sorry for the delay, I've been traveling and only have occasional computer access.

These all look really good. The only refactor that gives me pause is the derives for the `StatusCode` enum. If the compiler no longer takes a noticeably longer time generating them, then the fix looks fine. Last I checked, those derives took added several seconds each to the compilation.

leodasvacas:
Benchmarked on nightly with precision of seconds, no change in compile times for debug or release. However I suspect the improvement is due to rust-lang/rust#31414 which I believe is in beta. Should this wait for that change to land in stable?

seanmonstar:
Ah, glad to see that PR. Yes, I'd probably want to wait to introduce the
change when stable users have a safe way of keeping compile times better.

leodasvacas:
Very well, reverted changes to `status.rs`.

seanmonstar:
Thank you!

PR #833, opened 2016-06-17
Kill raw part when getting mutable reference to typed header

### LINKED ISSUE #821: get_mut doesn't seem to work
I'm not sure what happened here:

```rust
extern crate hyper;
use hyper::header::Headers;
use hyper::header::{ContentLength, ContentType};
use hyper::http::RawStatus;
use hyper::method::Method;
use hyper::mime::{Attr as MimeAttr, Value};

fn main() {
    let mut x = Headers::new();
    x.set_raw("Content-Type", vec!["text/plain;charset=shift-jis".bytes().collect()]);

    println!("old: {:?}", x); // charset is shift-jis
    if let Some(old_ct) = x.get_mut::<ContentType>() {
        println!("old old_ct: {:?}", old_ct); // still shift-jis
        let mut new_ct = old_ct.clone();
        for param in (new_ct.0).2.iter_mut() {
            if param.0 == MimeAttr::Charset {
                println!("old param {:?}", param); // shift-jis
                *param = (MimeAttr::Charset, Value::Ext("hi".into()));
                println!("new param {:?}", param); // hi
            }
        }
        *old_ct = new_ct;
        println!("new old_ct: {:?}", old_ct); // hi
    }

    // at this stage I'd expect the header to have changed. But it hasn't.
    println!("new get: {:?}", x.get::<ContentType>()); // hi
    println!("new get raw: {:?}", x.get_raw("content-type")); // shift-jis again??
    println!("new: {:?}", x); // shift-jis again??
}
```

andrebeat:
Fixes #821.

### DIFF
--- a/src/header/internals/item.rs
+++ b/src/header/internals/item.rs
@@ -78,6 +78,9 @@ impl Item {
                 Err(_) => ()
             }
         }
+        if self.raw.is_some() && self.typed.get_mut(tid).is_some() {
+            self.raw = OptCell::new(None);
+        }
         self.typed.get_mut(tid).map(|typed| unsafe { typed.downcast_mut_unchecked() })
     }
 }


GitCop:

 * Commit: c471ce67da0a2b144374049479db2665277b2054

---

seanmonstar:
Thanks! Just a couple comments.

- I *think* a typed could exist from parsing by getting `typed()`, and then later on a call to `typed_mut()` would not remove the `raw` representation.
- Could you add a couple tests for the multiple combinations that might cause it to be out of sync?
- Since the commit message becomes part of the changelog, could it be something tamer? :D Maybe just replacing "kill" with "remove"?

PR #835, opened 2016-06-17
feat(server): add more request information to Handler::check_continue

bozaro:
I want to pull request check_continue functionality to https://github.com/iron/iron and can't get request.remote_addr in current check_continue implementation.

Unfortunately this change is incompatible with old trait :(

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -124,10 +124,8 @@ use Error;
 use buffer::BufReader;
 use header::{Headers, Expect, Connection};
 use http;
-use method::Method;
 use net::{NetworkListener, NetworkStream, HttpListener, HttpsListener, SslServer};
 use status::StatusCode;
-use uri::RequestUri;
 use version::HttpVersion::Http11;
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
This probably would have been the better implementation, but it's a breaking change, and I don't plan to make any more of those for the 0.9.x era... :(

bozaro:
I think about `chunk_continue` method and came to the conclusion: this method is not good solution and I prefer to remove it.

Much better to send "100 Continue" status on first `Body::read()` call.

In this case no additional logic/callback needed for make desigion about "100 Continue":

 * If you call `Body::read()` method, then you needs to request body and hyper send to client "100 Continue";
 * If you have anought information to send final status without body, then hyper send status without "100 Continue".

I try to implement this logic on 0.9.8 branch and create a new PR.

bozaro:
I tried to implement the logic of sending "100 Continue" status at the beginning of reading, but I could not find a way to do this without losing compatibility :(

bozaro:
Move to issue #838

PR #841, opened 2016-06-20
fix(headers): Remove raw part when getting mutable reference to typed header

### LINKED ISSUE #821: get_mut doesn't seem to work

seanmonstar:
> If you get a mutable reference to a typed header it is possible to make
the two representations be out of sync. To avoid this, after parsing the
raw part it should be removed.
>
> Closes #821 

Rebase with a test added.
Closes #833

### DIFF
--- a/src/header/internals/item.rs
+++ b/src/header/internals/item.rs
@@ -78,6 +78,9 @@ impl Item {
                 Err(_) => ()
             }
         }
+        if self.raw.is_some() && self.typed.get_mut(tid).is_some() {
+            self.raw = OptCell::new(None);
+        }
         self.typed.get_mut(tid).map(|typed| unsafe { typed.downcast_mut_unchecked() })
     }
 }
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -670,6 +670,7 @@ mod tests {
     fn test_get_mutable() {
         let mut headers = Headers::from_raw(&raw!(b"Content-Length: 10")).unwrap();
         *headers.get_mut::<ContentLength>().unwrap() = ContentLength(20);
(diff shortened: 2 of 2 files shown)


andrebeat:
Sorry for the radio silence (and forgetting to add that test). Cheers!

Manishearth:
Could we get a crate version published for this?

PR #844, opened 2016-06-22
feat(client): add convenience method 'with_token' to RequestBuilder

mattforni:
may not be necessary, but I found the api for adding an authorization
header a bit clunky and thought this might be the first of many
convenience methods for the RequestBuilder

n/a


seanmonstar:
I won't be adding any new features to the 0.9 RequestBuilder. I do however have a new crate that provides convenience over the Client API, where something like this might make sense. I haven't published the crate yet, though...

mattforni:
@seanmonstar Would be happy to contribute to that crate when you publish. Keep me posted?

PR #850, opened 2016-07-01
feat(headers): Add Referrer-Policy header

aravind-pg:
This is required to implement [referrer policy delivery by header](https://www.w3.org/TR/referrer-policy/#referrer-policy-header) in Servo (see servo/servo#11860).


jdm:
Using an enum that contains an `Nonstandard(String)` variant would make sense here.

aravind-pg:
@jdm Actually, why not just have an enum of the six possible referrer policy types and regard anything else as an error? I'm on the fence about whether we'd want nonstandard as a variant here. In one sense it's useful since it adds flexibility, but then again it muddles the issue in a case where the spec really does specify exactly six variants (i.e. we don't really need the flexibility in this case). Maybe we should just defer to Hyper standard practice here -- cc @seanmonstar.

seanmonstar:
It does look like there are only 6 valid values. In those cases, I feel
enums fit perfectly. Is it expected to be extendable? If not, no need to
allow a String, right?

jdm:
I'm fine with pretending that unrecognized header values don't exist.

seanmonstar:
@jdm if you send `Content-Length: boogy\r\n`, then `headers.get::<ContentLength>()` will return `None`.

seanmonstar:
Also, if you need in this in Servo before switching to async io, I recommend targeting the PR against the 0.9.x branch.

aravind-pg:
Sounds good, will update to use an enum. Also, sounds like I will indeed need to target this against 0.9.x.  This header will probably be generally useful in master as well, so I can open a separate PR against master if you like (or maybe you can cherry-pick?).

aravind-pg:
Updated to use an enum, and also to handle legacy keywords `never`, `default`, and `always` as per [the spec](https://www.w3.org/TR/referrer-policy/#determine-policy-for-token).

Opened basically the same PR against 0.9.x as well: #851. Feel free to close this one if you don't want to merge into master at this point.

seanmonstar:
Hm, since the spec says those older keywords are semantically identical,
what if they just parsed to the equivalent new value. Such that there were
only 6 variants to the enum, and 3 additional strings that would parse to
it. @jdm I kinda prefer it, but are there issues I'm not thinking of?

aravind-pg:
Actually the current implementation does do exactly what you said -- there are only six variants, and the additional legacy strings just parse to their equivalent variants (e.g. both "no-referrer" and "never" parse to NoReferrer, etc.).

seanmonstar:
Oh! I just assumed from reading the comment in my inbox, without looking at
code :o)

seanmonstar:
Excellent, thanks!

PR #851, opened 2016-07-02
feat(headers): Add Referrer-Policy header

aravind-pg:
Required to implement [referrer policy delivery by header](https://www.w3.org/TR/referrer-policy/#referrer-policy-header) in Servo (see servo/servo#11860).

(See #850 for previous discussion. This is just the 0.9.x version.)


aravind-pg:
@seanmonstar Could you please bump the version?

aravind-pg:
@seanmonstar Ping, Servo needs the version bump.

seanmonstar:
Woops! Good news is that the wait means 0.9.10 also includes the `Origin` header. Published!

aravind-pg:
Great, thanks!

PR #853, opened 2016-07-06
feat(headers): Origin Header 0.9.x

### LINKED ISSUE #651: Origin Header

puhrez:
Add an Origin header so users may properly send CORS requests.

Closes #651


seanmonstar:
Could you rebase your commit against the 0.9.x branch? This is trying to merge all of master into 0.9.x.

puhrez:
@seanmonstar can u take a look  at this and let me know how I can make it better?

seanmonstar:
Oh sorry, it doesn't send me a notification that you had rebased. Will look today.

src/header/common/origin.rs
```diff
@@ -0,0 +1,125 @@
...
+///     }
+/// );
+/// ```
+
+#[derive(Clone, Debug)]
+pub struct Origin {
+    /// The scheme, such as http or https
+    pub scheme: String,
+    /// The host, such as Host{hostname: "hyper.rs".to_owned(), port: None}
+    pub host: Host,
```
seanmonstar: Clever.

A downside is that to set the `Origin`, you need to also have the `Host` type imported... Makes me think maybe the properties be private, with accessor methods?

src/header/common/origin.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl Header for Origin {
+    fn header_name() -> &'static str {
+        static NAME: &'static str = "Origin";
+        NAME
+    }
+
+    fn parse_header(raw: &[Vec<u8>]) -> ::Result<Origin> {
+        from_one_raw_str(raw).and_then(|s: String| {
+            let url = Url::parse(&s).unwrap(); // unwrap so we can let url::ParseError Propogate
```
seanmonstar: `unwrap` doesn't propagate in that way. 

This is `unwrap`:

```rust
fn unwrap(self) -> T {
    match self {
        Ok(t) => t,
        Err(e) => panic!("tried to unwrap, found err {:?}", e)
    }
}
```

You probably want the `try!` macro, which will pull out `T` if it can, otherwise `return Err(e)`.

seanmonstar: Some discussion about parsing using `Url` has pointed out that we shouldn't accept an `Origin` that has a path part.

How about using this strategy instead:

1. Find `://`.
2. Scheme is `raw[..colon]`.
3. Then `Host::from_str(&raw[colon + 3..])`

I just took a peek, and saw there is no `FromStr` impl for `Host`, it'd probably be nice to move the `parse_header` code into a FromStr impl...

src/header/common/origin.rs
```diff
@@ -0,0 +1,125 @@
...
+    use header::{Header, Host};
+
+    #[test]
+    fn test_origin() {
+        let origin = Header::parse_header([b"http://foo.com".to_vec()].as_ref());
+        assert_eq!(origin.ok(), Some(Origin {
+            scheme: "http".to_owned(),
+            host: Host {
+                hostname: "foo.com".to_owned(),
+                port: Some(80),
```
seanmonstar: This should probably be `port: None`, since the raw header didn't actually include the port. Otherwise, once writing this header again, it will appear as if the header were really `http://foo.com:80`.

src/header/common/origin.rs
```diff
@@ -0,0 +1,125 @@
...
+                host: host
+            })
+        })
+    }
+}
+
+
+impl HeaderFormat for Origin {
+    fn fmt_header(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self.host.port {
```
seanmonstar: 👍

GitCop:

 * Commit: d7f675eee3fd32df7ed27b5ba9a9e615423c2d31

---

puhrez:
@seanmonstar could you rerun both builds? One didn't start, the other failed downloading a dependency.

src/header/common/origin.rs
```diff
@@ -0,0 +1,127 @@
...
+        Ok(Origin{
+            scheme: scheme.to_owned(),
+            host: host
+        })
+    }
+}
+
+impl HeaderFormat for Origin {
+    fn fmt_header(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // I wanna the responsibilty of writing the host/port to Host
```
seanmonstar: using `write!(f, "{}://{}", self.scheme, self.host)` should do the same.

src/header/common/host.rs
```diff
@@ -97,6 +64,29 @@ impl HeaderFormat for Host {
     }
 }
 
+impl FromStr for Host {
+    type Err = ::Error;
+
+    fn from_str(s: &str) -> ::Result<Host> {
+        let (host_port, res) = domain_to_unicode(s);
```
seanmonstar: Nice! I had noticed that fixme and was going to file a separate issue, but you win!

src/header/common/host.rs
```diff
@@ -47,43 +49,8 @@ impl Header for Host {
...
-            match idx {
-                Some(idx) => s.truncate(idx),
-                None => ()
-            }
-
-            Ok(Host {
-                hostname: s,
-                port: port
-            })
+        from_one_raw_str(raw).and_then(|s: String| {
```
seanmonstar: I believe this step can be skipped (this step to a `String`). I'd expect this entire function could just be `from_one_raw_str(raw)`, since inferrence should tell it to use `<Host as FromStr>`.

src/header/common/origin.rs
puhrez: I originally did that, but then I get an unused result warning

seanmonstar:
@untitaker whatcha think? My only last hessitation is having people build the `Origin` struct literal, instead of some sort of `Origin::new("https", "www.rust-lang.org", None)`...

puhrez:
@seanmonstar btw I still plan to obfuscate the setting if host/port via your earlier suggestion.

src/header/common/origin.rs
seanmonstar: That doesn't sound right. I meant for this `write!` to be the only thing in the function. That way the `Result` is returned.

src/header/common/host.rs
puhrez: !

src/header/common/origin.rs
puhrez: So I see the reason for a constructor, by why obstruct access?

seanmonstar: It might not be important yet. The idea I had was to have `scheme()`, `hostname()`, and `port()` accessors. By keeping the fields themselves private, it doesn't matter if we keep using `Host` internally, or some other field scheme.

But all the other headers are currently wide open. I'll consider a accessor pass before 1.0.

src/header/common/origin.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(Clone, Debug)]
+pub struct Origin {
+    /// The scheme, such as http or https
+    pub scheme: String,
+    /// The host, such as Host{hostname: "hyper.rs".to_owned(), port: None}
+    pub host: Host,
+}
+
+impl Origin {
+    pub fn new(scheme: &str, hostname: &str, port: Option<u16>) -> Origin {
```
seanmonstar: I'd try `fn new<S: Into<String>, H: Into<String>(scheme: S, hostname: H, ...)`. The reason is that it's probably likely that a user had to construct a `String` *anyways* to create the hostname or scheme, it's a shame to essentially have `format!("{}.com", sld).as_slice().to_owned()`.

src/header/common/origin.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    fn parse_header(raw: &[Vec<u8>]) -> ::Result<Origin> {
+        from_one_raw_str(raw)
+    }
+}
+
+impl FromStr for Origin {
+    type Err = ::Error;
+
+    fn from_str(s: &str) -> ::Result<Origin> { // not exactly sure why this works and why im not getting "you need 2 types error"
```
seanmonstar: This is using `Result` from the crate root, which is `hyper::error::Result<T>`, which is an alias to `::std::result::Result<T, ::hyper::error::Error>`.

seanmonstar:
This is looking really good! 2 comments left, and we can merge!

puhrez:
so 1, the comment about  `new` and 2? the accessor funcs?

seanmonstar:
@puhrez sorry, 2 would be to remove the comment at `impl FromStr for Origin` about being unsure why there is no type error.

puhrez:
oh haha

puhrez:
@seanmonstar so i think we're good here,
aside: could you label other issues that I may tackle with `easy`?

PR #865, opened 2016-07-15
docs(guide): copy over html files, not markdown files

seanmonstar:


seanmonstar:
Failure is a bug in nightly rustc compiling regex crate.

PR #867, opened 2016-07-18
Let Server::new take one or more listeners

### LINKED ISSUE #859: Listen on multiple addresses in async hyper
Given that hyper is now async it would be great to be able to easily have a single server listen on multiple addresses.

It's already possible to get the local address of a connection via `Request::transport`.

It's not possible to implement your own `Transport` to do this as `Evented::register` is only passed one `Token`.

If you agree with this I'm happy to implement it. I think the easiest way is to let `Server` take multiple `Accept`s of the same type e.g (but with more traits so you could pass a bare HttpListener as before).

```rust
impl<I> Server<I> where I: IntoIterator, I::Item: Accept, <I::Item as Accept>::Output: Transport {
    ...
}
```

Then you could have something like this in a different crate.

```rust
enum HttpOrHttpsListener {
    Http(HttpListener),
    Https(HttpsListener),
}

impl Accept for HttpOrHttpsListener {
    ...
}
```

ebarnard:
Closes #859

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -38,7 +38,8 @@ impl<A: Accept, H: HandlerFactory<A::Output>> fmt::Debug for ServerLoop<A, H> {
 /// A Server that can accept incoming network requests.
 #[derive(Debug)]
 pub struct Server<A> {
-    listeners: Vec<A>,
+    lead_listener: A,
+    other_listeners: Vec<A>,
     keep_alive: bool,
     idle_timeout: Option<Duration>,
     max_sockets: usize,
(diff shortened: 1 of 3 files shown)


src/server/mod.rs
```diff
@@ -105,33 +112,49 @@ impl<S: SslServer> Server<HttpsListener<S>> {
...
         let mut loop_ = rotor::Loop::new(&config).unwrap();
+
+        let mut addrs = Vec::with_capacity(self.listeners.len());
+        let mut listeners = self.listeners.into_iter();
+
+        // Add the first listener. This one handles shutdown messages.
         let mut notifier = None;
         {
             let notifier = &mut notifier;
+            let listener = listeners.next().unwrap();
```
seanmonstar: I'd use `next().expect("Server::new requires at least 1 listener")`, or something similar.

src/server/mod.rs
```diff
@@ -299,29 +322,29 @@ where A: Accept,
...
+            .field("addrs", &self.addrs)
             .field("closed", &self.shutdown.0.load(Ordering::Relaxed))
             .finish()
     }
 }
 
 impl fmt::Display for Listening {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Display::fmt(&self.addr, f)
+        fmt::Debug::fmt(&self.addrs, f)
```
seanmonstar: Hm, before this would allow a `Listening` to be printed nicely, by doing `println!("listening on {}", listening)`. This change will make the output look like `listening on [V4(127.0.0.1:8080)]`.

I suppose if there are muliple addrs, it likely should output a comma separated list. But either way, it'd probably be best to not have the `Debug` output of the `SocketAddr`s.

src/server/mod.rs
ebarnard: I'm going to split listeners in to `lead_listener` and `other_listeners` to avoid having to unwrap here.

src/server/mod.rs
ebarnard: Damn. Didn't realise that included the V4.

seanmonstar:
This looks superb!

ebarnard:
Thanks. Nits addressed.

seanmonstar:
Thanks a bunch!

PR #869, opened 2016-07-21
feat(header): introduce header::Raw

seanmonstar:
The `Raw` type repesents the raw bytes of a header-value.

Having a special type allows a couple of benefits:

- The exact representation has become private, allowing "uglier"
internals. Specifically, since the common case is for a header to only
have 1 line of bytes, an enum is used to skip allocating a Vec for only
1 line. Additionally, a Cow<'static, [u8]> is used, so static bytes
don't require a copy. Finally, since we can use static bytes, when
parsing, we can compare the incoming bytes against a couple of the most
common header-values, and possibly remove another copy.

- As its own type, the `Headers.set_raw` method can be generic over
`Into<Raw>`, which allows for more ergnomic method calls.

BREAKING CHANGE: `Header::parse_header` now receives `&Raw`, instead of
  a `&[Vec<u8>]`. `Raw` provides several methods to ease using it, but
  may require some changes to existing code.

### DIFF
--- a/src/header/common/access_control_allow_credentials.rs
+++ b/src/header/common/access_control_allow_credentials.rs
@@ -1,7 +1,7 @@
 use std::fmt::{self, Display};
 use std::str;
 use unicase::UniCase;
-use header::{Header};
+use header::{Header, Raw};
 
 /// `Access-Control-Allow-Credentials` header, part of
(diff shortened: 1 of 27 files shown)


seanmonstar:
cc @mitsuhiko

PR #871, opened 2016-07-23
accept cert chain files as well as single cert files

kali:
Openssl::with_cert_and_key is often useless in its current form: in most setups, you need to provide the intermediate certificate chain to openssl. One way would be to change with_cert_and_key to allow passing a third file containing the authority chain. But fortunately, we have another option: openssl accepts combined certificates (your certs, then whatever certs you need to link it to the widely distrubuted ones). This is not exotic, both nginx and apache use this setup.

As far as I can tell, this should not break anything, and would fix https support for all libraries relying on with_cert_and_key...


GitCop:

 * Commit: fe8d11d254d728d051ebf1accb562f887556255b

---

seanmonstar:
Thanks!

PR #893, opened 2016-08-20
feat(client): add a Fn(Response) RedirectPolicy

mmirate:
Add a new RequestPolicy variant whose argument is a `Fn(Response) -> bool`, instead of a `Fn(Url) -> bool`.

The usual `RedirectIf` is often too limited since it only provides the redirect destination URL.

(Use case: Apache's mod_cas&mdash;an authentication component used by some websites&mdash;apparently sometimes sets vital cookies in 302 responses, and I'd like to halt those redirections so I can store those cookies ... if any were set. [n.b. ignoring these cookies causes an unwarranted redirect loop] I suppose I could handle HTTP redirections entirely in my own code ... if RequestBuilder impl'ed `Clone`, or if its  `send` method didn't consume `self`.)

Hence, this new RedirectPolicy provides a reference to the entire Response instead of just the URL.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -360,6 +360,7 @@ impl<'a> RequestBuilder<'a> {
                 // separate branches because they can't be one
                 RedirectPolicy::FollowAll => (), //continue
                 RedirectPolicy::FollowIf(cond) if cond(&url) => (), //continue
+                RedirectPolicy::FollowIfRes(cond) if cond(&res) => (), //continue
                 _ => return Ok(res),
             }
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! I agree this would perhaps be a more useful policy than just the `Url`. hyper on the master branch has moved away from providing a redirect policy, so I don't believe adding this here would be the solution. I'd imagine it would be a good fit for higher level client libraries, such as https://github.com/seanmonstar/request (wip).

PR #898, opened 2016-08-22
Make Headers::remove return the header

### LINKED ISSUE #891: Headers should offer method to remove and return header
If I take a `Headers` by value (moving it into my function), I want to avoid copying them. For that reason it would be useful if either `Headers::remove` would return the popped item (as `HashMap::remove` does), or if a new method was added.

untitaker:
Fix #891

### DIFF
--- a/src/header/internals/cell.rs
+++ b/src/header/internals/cell.rs
@@ -87,6 +87,20 @@ impl<V: ?Sized + Any + 'static> PtrMapCell<V> {
     }
 
     #[inline]
+    pub fn into_value(self, key: TypeId) -> Option<Box<V>> {
+        let map = unsafe { self.0.into_inner() };
+        match map {
+            PtrMap::Empty => None,
+            PtrMap::One(id, v) => if id == key {
+                Some(v)
+            } else {
+                None
+            },
+            PtrMap::Many(mut hm) => hm.remove(&key)
+        }
+    }
+
+    #[inline]
     pub unsafe fn insert(&self, key: TypeId, val: Box<V>) {
         let mut map = &mut *self.0.get();
         match *map {
--- a/src/header/internals/item.rs
+++ b/src/header/internals/item.rs
@@ -82,6 +82,14 @@ impl Item {
         }
         self.typed.get_mut(tid).map(|typed| unsafe { typed.downcast_mut_unchecked() })
     }
+
+    pub fn into_typed<H: Header>(self) -> Option<H> {
+        let tid = TypeId::of::<H>();
+        match self.typed.into_value(tid) {
+            Some(val) => Some(val),
+            None => parse::<H>(self.raw.as_ref().expect("item.raw must exist")).ok()
+        }.map(|typed| unsafe { typed.downcast_unchecked() })
+    }
 }
 
 #[inline]
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -182,6 +182,11 @@ impl Header + Send + Sync {
     unsafe fn downcast_mut_unchecked<T: 'static>(&mut self) -> &mut T {
         &mut *(mem::transmute::<*mut _, (*mut (), *mut ())>(self).0 as *mut T)
     }
+
+    #[inline]
+    unsafe fn downcast_unchecked<T: 'static>(self: Box<Self>) -> T {
+        *Box::from_raw(mem::transmute::<*mut _, (*mut (), *mut ())>(Box::into_raw(self)).0 as *mut T)
+    }
 }
 
 impl Clone for Box<Header + Send + Sync> {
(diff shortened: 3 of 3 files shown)


GitCop:

 * Commit: d89a5549ad41595f31662bbfc66118b838d99dea

---

src/header/internals/item.rs
```diff
@@ -82,6 +82,14 @@ impl Item {
         }
         self.typed.get_mut(tid).map(|typed| unsafe { typed.downcast_mut_unchecked() })
     }
+
+    pub fn into_typed<H: Header>(self) -> Option<H> {
```
untitaker: Because a `H` may be unsized, I probably can only return `Option<Box<H>>` here. However, then I'm unsure how I'd be casting the box to a concrete type.

src/header/internals/cell.rs
```diff
@@ -87,6 +87,20 @@ impl<V: ?Sized + Any + 'static> PtrMapCell<V> {
     }
 
     #[inline]
+    pub fn into_value(self, key: TypeId) -> Option<Box<V>> {
+        let mut map = unsafe { self.0.into_inner() };
```
untitaker: I was trying to keep this analogous to `get_mut` and `get`, but `UnsafeCell::into_inner` has [some extra caveats](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html#method.into_inner).

untitaker: Also `UnsafeCell::get` is not marked as unsafe, I don't understand why it is wrapped in an `unsafe` block in `get` and `get_mut`.

seanmonstar:
Oh woops, I  didn't realize that the internal `Item` would need a new method, making this less easy than I suspected. Still, we should be able to figure it out.

src/header/internals/cell.rs
seanmonstar: `UnsafeCell::get` is not unsafe, because it just returns a raw pointer. You can toss those around freely. However, **dereferencing** it is unsafe, which is needed to turn it into a safe pointer (the other way is to use mem::transmute, but that's also unsafe).

The extra caveats for `into_inner` are the same as using the pointer from `UnsafeCell::get`: accessing it is not synchronized over threads, but that is fine for us here, because of the mutable requirement of `fn remove(&mut self)`.

src/header/mod.rs
```diff
@@ -320,10 +320,11 @@ impl Headers {
     }
 
     /// Removes a header from the map, if one existed.
-    /// Returns true if a header has been removed.
-    pub fn remove<H: Header>(&mut self) -> bool {
+    /// Returns the header, if one has been removed.
```
untitaker: This is actually incorrect. `None` may also be returned if the removed header was not parseable.

I wonder if we should return `Option<Raw, Option<H>>` or something like that instead, to enable as many usecases as possible.

src/header/internals/item.rs
untitaker: To be clear, I get a type conflict exactly at this spot: Expected `H`, got `Box<Header + Send + Sync>`. The first idea here is to add a new `downcast_*` method to `impl Header + Send + Sync` (again analogous to the `get` methods), but I don't know how to implement such a method.

src/header/mod.rs
seanmonstar: Hm, I don't love that the return value could be more complicated, but how about this crazy idea: `Option<Result<H, Raw>>`...

I'd personally just drop it, but is there a convincing case where the raw is still needed to justify complicating the return type?

untitaker: I can't think of a reason why `Raw` is needed, but with the current return type there is no upgrade path for people who were checking the boolean value, as `.is_some()` is `false` if the header *was* removed, but could not be parsed.

untitaker: Therefore `Option<Option<H>>` or `Option<Result<H, ()>>` would be appropriate if we don't want to both with `Raw`.

src/header/internals/item.rs
seanmonstar: Ah yes, of course, because the `typed` is storing them as `Box<Header>`. I think there is a way to add a downcast taking self by value, such as something like `fn downcast(self: Box<Self>)`, but I'll toy on playground to find the right thing and report back.

seanmonstar: Adding this to the `impl Header + Send + Sync` should do it:
```rust
#[inline]
unsafe fn downcast_unchecked<T: 'static>(self: Box<Self>) -> T {
    *(mem::transmute::<*mut _, (*mut (), *mut ())>(Box::into_raw(self)).0 as *mut T)
}
```

Then you can call `val.downcast_unchecked()` after having gotten the type from the TypeId.

src/header/mod.rs
seanmonstar: Both feel yucky. I'd still prefer to just have `Option<H>`. The upgrade path could be to check `headers.has()` before calling `remove()`.

untitaker: Fair enough, I will update the docstring then.

GitCop:

 * Commit: 0cfb36544e42e1a53e2a7c57f6e4c58b88050589

---

GitCop:

 * Commit: 0cfb36544e42e1a53e2a7c57f6e4c58b88050589

 * Commit: d59552448f36e69bdaf1dd34dd61ce20da8070fa

---

src/header/internals/item.rs
untitaker: Unfortunately this doesn't work, apparently I can't deref `*mut T`.

untitaker:
Damn you @gitcop leave me alone. This will get squashed anyway!

src/header/internals/item.rs
seanmonstar: Oh right. We can convert the `*mut T` into a `Box<T>` for free, with `Box::from_raw`, and then we can deref move out of the box:

```rust
*Box::from_raw(mem::transmute::<*mut _, (*mut (), *mut ())>(Box::into_raw(self)).0 as *mut T)
```

untitaker: Amazing.

untitaker:
@seanmonstar This should be ready to merge.

seanmonstar:
Looks right to me. Seems travis-ci's OSX containers are lagging, all the linux tests pass, so I'll merge anyways.

untitaker:
Thanks!

Issue #897, opened 2016-08-22
Add `query()` method to `server::Request`


seanmonstar:
Similar to #896 

```rust
fn query(&self) -> Option<&str>;
```

jonashagstedt:
What about using a hashmap for the query string?
I imagine that any time you want to work with the querystring you want to work with the values.

e.g

```
let page = request.query["page"].unwrap();
```

seanmonstar:
You might want the query params, but not necessarily. A somewhat common usecase for a query string is as a cache buster, i.e. `/foo/bar.js?abcdef`, such as adding a timestamp or sha to the end. I'd also prefer leaving the choice of what to parse them into up to the user. Whether that is `Vec<(Name, Value)>`, a HashMap, or whatever.

PR #899, opened 2016-08-23
feat(server): add path() and query() to Request

### LINKED ISSUE #896: Add `path()` method to `server::Request`
```rust
fn path(&self) -> Option<&str>
```

It'd need to check for a query string in the `AbsolutePath` variant.

ahmedcharles:
Fixes #896 and #897.

Also improve RequestUri tests.

### DIFF
--- a/src/server/request.rs
+++ b/src/server/request.rs
@@ -66,17 +66,25 @@ impl<'a, T> Request<'a, T> {
     #[inline]
     pub fn version(&self) -> &HttpVersion { &self.version }
 
-    /*
     /// The target path of this Request.
     #[inline]
     pub fn path(&self) -> Option<&str> {
-        match *self.uri {
-            RequestUri::AbsolutePath(ref s) => Some(s),
-            RequestUri::AbsoluteUri(ref url) => Some(&url[::url::Position::BeforePath..]),
-            _ => None
+        match self.uri {
+            RequestUri::AbsolutePath { path: ref p, .. } => Some(p.as_str()),
+            RequestUri::AbsoluteUri(ref url) => Some(url.path()),
+            _ => None,
         }
     }
-    */
+
+    /// The query string of this Request.
+    #[inline]
+    pub fn query(&self) -> Option<&str> {
+        match self.uri {
+            RequestUri::AbsolutePath { query: ref q, .. } => q.as_ref().map(|x| x.as_str()),
+            RequestUri::AbsoluteUri(ref url) => url.query(),
+            _ => None,
+        }
+    }
 
     /// Deconstruct this Request into its pieces.
     ///
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -26,8 +26,13 @@ pub enum RequestUri {
     /// The most common request target, an absolute path and optional query.
     ///
     /// For example, the line `GET /where?q=now HTTP/1.1` would parse the URI
-    /// as `AbsolutePath("/where?q=now".to_string())`.
-    AbsolutePath(String),
+    /// as `AbsolutePath { path: "/where".to_string(), query: Some("q=now".to_string()) }`.
+    AbsolutePath {
+        /// The path part of the request uri.
+        path: String,
+        /// The query part of the request uri.
+        query: Option<String>,
+    },
 
     /// An absolute URI. Used in conjunction with proxies.
     ///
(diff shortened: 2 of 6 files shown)


ahmedcharles:
Are there tests for Request that I should update? I didn't find any obvious ones. Alternatively, I can write unit tests for Request.

src/server/request.rs
```diff
@@ -66,17 +68,31 @@ impl<'a, T> Request<'a, T> {
...
     #[inline]
-    pub fn path(&self) -> Option<&str> {
-        match *self.uri {
-            RequestUri::AbsolutePath(ref s) => Some(s),
-            RequestUri::AbsoluteUri(ref url) => Some(&url[::url::Position::BeforePath..]),
-            _ => None
+    pub fn path(&self) -> Option<Cow<str>> {
+        match self.uri {
+            RequestUri::AbsolutePath(ref s) => {
+                let url = Url::parse("http://example.com").ok().and_then(|u| u.join(s).ok());
```
seanmonstar: I was imagine using something more primitive:

```rust
let end = s.find('?').unwrap_or(s.len());
&s[..end]
```

Is there a reason that this is incorrect?

ahmedcharles: I was going for consistency and doing it in a way that I don't have to read the RFC and implement correct parsing and interpretation of urls. Url::parse normalizes the urls, and therefore, it makes a copy. It would be weird if normalized paths and queries were returned for AbsoluteUri but not for AbsolutePath.

seanmonstar: Oh ok, good call. I'm wondering whether it's better to run the Url parser
when constructing the RequestUri, and thus be able to return a reference,
or do it on demand, as this patch does.

ahmedcharles: I also found that appealing as an option. Though I didn't want to change the RequestUri enum. Also, I'm not sure what the appropriate design would be. It'd be nice if Url supported parsing partial urls, then it could ignore the domain.

Perhaps it would be best to implement a custom Url to return as part of AbsolutePath instead of just a String. That way we can return references to owned data while only parsing once, up front.

I'll take a look and see what I come up with.

ahmedcharles:
Did you notice the test changes and do you know if they are actually correct or is the implementation buggy? It seems like some things that are accepted should be rejected.

src/uri.rs
```diff
@@ -92,14 +92,17 @@ impl Display for RequestUri {
...
+    fn read(s: &str, result: Option<RequestUri>) {
+        assert_eq!(s.parse::<RequestUri>().ok(), result);
     }
 
-    read("*", RequestUri::Star);
-    read("http://hyper.rs/", RequestUri::AbsoluteUri(Url::parse("http://hyper.rs/").unwrap()));
-    read("hyper.rs", RequestUri::Authority("hyper.rs".to_owned()));
-    read("/", RequestUri::AbsolutePath("/".to_owned()));
+    read("*", Some(RequestUri::Star));
+    read("**", Some(RequestUri::Authority("**".to_owned())));
```
seanmonstar: This one is surprising. I would expect it to be a parse error.

src/uri.rs
```diff
@@ -92,14 +92,17 @@ impl Display for RequestUri {
...
+    read("http://hyper.rs/", Some(RequestUri::AbsoluteUri(Url::parse("http://hyper.rs/").unwrap())));
+    read("hyper.rs", Some(RequestUri::Authority("hyper.rs".to_owned())));
+    read("hyper.rs?key=value", Some(RequestUri::Authority("hyper.rs?key=value".to_owned())));
```
seanmonstar: Likewise, this one seems incorrect.

src/uri.rs
```diff
@@ -92,14 +92,17 @@ impl Display for RequestUri {
 
 #[test]
 fn test_uri_fromstr() {
```
seanmonstar: I'd be in favor of this test instead having 2 helpers, to prevent the proliferation of `Some(...)` in every test. Maybe something like:

```rust
assert_eq!(parse("*"), RequestUri::Star);
assert!(parse_err("**"));
```

src/server/request.rs
seanmonstar: The [spec claims this variant has "absolute-path [ "?" query ]"](https://tools.ietf.org/html/rfc7230#section-5.3.1). Maybe its worth adjusting the variant to `AbsolutePath { path: String, query: Option<String> }`?

ahmedcharles:
I implemented all of the suggested changes, though it doesn't fix one of the test cases, since I'm not entirely sure how to go about changing the request uri parser.

seanmonstar:
travis logs say the code examples in the guide are now failing.

ahmedcharles:
I guess I didn't run the doc tests. I'll do that. :)

seanmonstar:
Excellent, thanks for sticking through to the end!

ahmedcharles:
No problem. I really like that Rust has a tool for testing documentation, since that's something that's always out of date.

PR #909, opened 2016-09-04
Async Chunked Decoding

shanegibbs:
Async decoder implementation.

**Todo**: Extensions

_Work in progress (will squash commits eventually)_

### DIFF
--- a/src/http/h1/decode.rs
+++ b/src/http/h1/decode.rs
@@ -151,7 +151,7 @@ impl ChunkedState {
             ChunkedState::Body => try!(ChunkedState::read_body(body, size, buf, read)),
             ChunkedState::BodyCr => try!(ChunkedState::read_body_cr(body)),
             ChunkedState::BodyLf => try!(ChunkedState::read_body_lf(body)),
-            ChunkedState::End => panic!("Calling step on End state"),
+            ChunkedState::End => ChunkedState::End,
         })
     }
     fn read_size<R: Read>(rdr: &mut R, size: &mut u64) -> io::Result<ChunkedState> {
@@ -363,6 +363,28 @@ mod tests {
         assert_eq!(16, count);
         let result = String::from_utf8(buf.to_vec()).expect("decode String");
         assert_eq!("1234567890abcdef", &result);
+    }
+
+    #[test]
+    fn test_read_chunked_after_eof() {
+        let content = b"10\r\n1234567890abcdef\r\n0\r\n";
+        let mut mock_buf = io::Cursor::new(content);
+        let mut buf = [0u8; 50];
+        let mut decoder = Decoder::chunked();
+
+        // normal read
+        let count = decoder.decode(&mut mock_buf, &mut buf).expect("decode");
+        assert_eq!(16, count);
+        let result = String::from_utf8(buf[0..count].to_vec()).expect("decode String");
+        assert_eq!("1234567890abcdef", &result);
+
+        // eof read
+        let count = decoder.decode(&mut mock_buf, &mut buf).expect("decode");
+        assert_eq!(0, count);
+
+        // ensure read after eof also returns eof
+        let count = decoder.decode(&mut mock_buf, &mut buf).expect("decode");
+        assert_eq!(0, count);
     }
 
     // perform an async read using a custom buffer size and causing a blocking
--- a/src/http/h1/decode.rs
+++ b/src/http/h1/decode.rs
@@ -1,14 +1,39 @@
 use std::cmp;
 use std::io::{self, Read};
 
-use self::Decoder::{Length, Chunked, Eof};
+use self::Kind::{Length, Chunked, Eof};
 
 /// Decoders to handle different Transfer-Encodings.
 ///
 /// If a message body does not include a Transfer-Encoding, it *should*
 /// include a Content-Length header.
 #[derive(Debug, Clone)]
-pub enum Decoder {
+pub struct Decoder {
+    kind: Kind,
+}
+
+impl Decoder {
+    pub fn length(x: u64) -> Decoder {
+        Decoder {
+            kind: Kind::Length(x)
+        }
+    }
+
+    pub fn chunked() -> Decoder {
+        Decoder {
+            kind: Kind::Chunked(None)
+        }
+    }
+
+    pub fn eof() -> Decoder {
+        Decoder {
+            kind: Kind::Eof(false)
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+enum Kind {
     /// A Reader used when a Content-Length header is passed with a positive integer.
     Length(u64),
     /// A Reader used when Transfer-Encoding is `chunked`.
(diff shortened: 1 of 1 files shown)


src/http/h1/decode.rs
```diff
@@ -290,4 +292,40 @@ mod tests {
         assert_eq!(e.kind(), io::ErrorKind::Other);
         assert_eq!(e.description(), "early eof");
     }
+
+    fn read_chunked_async(content: &[u8], block_at: usize, read_buffer_size: usize) -> String {
+        let content_len = content.len();
+        let mock_buf = io::Cursor::new(content.clone());
+        let mut ins = Async::new(mock_buf, block_at);
+        let mut decoder = Decoder::chunked();
```
shanegibbs: Can probably have 'decode' as a argument here, and make this just a `read_async` function to test the other decoders as well perhaps.

src/http/h1/decode.rs
```diff
@@ -56,9 +31,37 @@ enum Kind {
...
+enum ChunkedState {
+    Init,
+    Size(u64),
+    SizeCr,
+    SizeLf,
+    Body(usize),
+    BodyCr,
+    BodyLf,
+    End,
+}
```
seanmonstar: Excellent, I was thinking something like this would be needed.

src/http/h1/decode.rs
```diff
@@ -1,39 +1,14 @@
...
-use self::Kind::{Length, Chunked, Eof};
+use self::Decoder::{Length, Chunked, Eof};
 
 /// Decoders to handle different Transfer-Encodings.
 ///
 /// If a message body does not include a Transfer-Encoding, it *should*
 /// include a Content-Length header.
 #[derive(Debug, Clone)]
-pub struct Decoder {
-    kind: Kind,
```
seanmonstar: This was done to keep the internals private. As a public enum, the internals are exposed and it'd be a breaking change later on to change them, since someone could have depended on the exposed internals.

shanegibbs: Oh interesting. I see. Will undo.

shanegibbs:
@seanmonstar I've completed a basic implementation, excluding extensions. See what you think. If okay, I'll continue this pattern to implement extensions.

seanmonstar:
@shanegibbs wow the work here is impressive! I'll be able to sit down and give a proper review in the morning.

src/http/h1/decode.rs
```diff
@@ -37,7 +37,7 @@ enum Kind {
     /// A Reader used when a Content-Length header is passed with a positive integer.
     Length(u64),
     /// A Reader used when Transfer-Encoding is `chunked`.
-    Chunked(Option<u64>),
+    Chunked(ChunkedState, usize),
```
seanmonstar: A `u64` was used before in case a chunk size claimed the incoming chunk was >4GB. On a 32-bit platform, `usize` cannot represent a number bigger than ~4 billion. Sounds absurd to send a chunk so big, but seemed prudent to be able to survive it.

src/http/h1/decode.rs
```diff
@@ -290,4 +347,75 @@ mod tests {
...
+                    }
+                    ins.block_in(content_len); // we only block once
+                    0 as usize
+                }
+            };
+        }
+        String::from_utf8(outs).expect("decode String")
+    }
+
+    fn all_async_cases(content: &str, expected: &str, decoder: Decoder) {
```
seanmonstar: Pretty clever, though it took me reading over it several times before I understood what you were trying to test specifically.

src/http/h1/decode.rs
```diff
@@ -136,113 +126,177 @@ impl Decoder {
...
+
+        let to_read = cmp::min(*rem, buf.len());
+        let count = try!(rdr.read(&mut buf[..to_read]));
 
+        trace!("to_read = {}", to_read);
+        trace!("count = {}", count);
+
+        if count == 0 {
+            *rem = 0;
+            return Err(io::Error::new(io::ErrorKind::Other, "early eof"));
```
seanmonstar: It didn't exist when this code was originally written (pre Rust 1.0), but there is now `io::ErrorKind::UnexpectedEof`, which seems more appropriate here.

src/http/h1/decode.rs
```diff
@@ -136,113 +126,177 @@ impl Decoder {
...
-        match try!(rdr.read(&mut buf)) {
-            1 if buf[0] == b => (),
+macro_rules! byte (
+    ($rdr:ident) => ({
+        let mut buf = [0];
+        match try!($rdr.read(&mut buf)) {
+            1 => buf[0],
             _ => return Err(io::Error::new(io::ErrorKind::InvalidInput,
-                                          "Invalid characters found")),
+                                                "Invalid chunk size line")),
```
seanmonstar: Since this is saying that a read of size 1 is required, seems the error is more correctly thought of "unexpected eof during chunk size line". So it could now be represented as `io::ErrorKind::UnexpectedEof`.

seanmonstar:
@shanegibbs Splendid work here. This is a great improvement, as the code should now be robust in a non-blocking world.

I'd prefer to merge only after extensions are supported, or else we suddenly no longer work on those peculiar streams sending them.

src/http/h1/decode.rs
shanegibbs: Ahhh. Makes sense.

src/http/h1/decode.rs
shanegibbs: Fair enough. I'll add some doc 😄

shanegibbs:
> I'd prefer to merge only after extensions are supported, or else we suddenly no longer work on those peculiar streams sending them.

Agreed. I'll continue working on this branch.

src/http/h1/decode.rs
```diff
@@ -136,113 +126,183 @@ impl Decoder {
...
+            buf: &mut [u8],
+            read: &mut usize) -> io::Result<ChunkedState> {
+        Ok(match *self {
+            ChunkedState::Size => try!(ChunkedState::read_size(body, size)),
+            ChunkedState::SizeLws => try!(ChunkedState::read_size_lws(body)),
+            ChunkedState::SizeLf => try!(ChunkedState::read_size_lf(body, size)),
+            ChunkedState::Body => try!(ChunkedState::read_body(body, size, buf, read)),
+            ChunkedState::BodyCr => try!(ChunkedState::read_body_cr(body)),
+            ChunkedState::BodyLf => try!(ChunkedState::read_body_lf(body)),
+            ChunkedState::End => panic!("Calling step on End state"),
```
shanegibbs: I have observed this `panic`. I need to investigate how `Decoder` is called and formulate a test for it.

seanmonstar: A user may call read() again, and it should continue to just say EOF,  or Ok(0).

src/http/h1/decode.rs
```diff
@@ -87,183 +93,248 @@ impl Decoder {
...
+                Err(io::Error::new(io::ErrorKind::InvalidInput,
+                                   "Invalid chunk size linear white space"))
+            }
+        }
+    }
+    fn read_extension<R: Read>(rdr: &mut R) -> io::Result<ChunkedState> {
+        trace!("read_extension");
+        match byte!(rdr) {
+            b'\r' => return Ok(ChunkedState::SizeLf),
+            _ => return Ok(ChunkedState::Extension), // no supported extensions
```
shanegibbs: Pretty crude for now. If we receive a `;` in the size line, we just keep reading until the next `\r`.

seanmonstar:
Thanks for seeing this through!

PR #910, opened 2016-09-05
feat(headers): add Warning header

### LINKED ISSUE #883: add Warning header
I would like to use the `Warning` header to warn users of deprecated API routes as proposed in this SO answer: http://stackoverflow.com/a/29623798.
Spec: https://tools.ietf.org/html/rfc7234#section-5.5

smithsps:
Implements Warning header from RFC7234
Closes #883


seanmonstar:
Looks excellent! Thanks!

PR #916, opened 2016-09-22
Evict idle connections when client is full

jwilm:
In the scenario where a request is started on the `Client`, the client
has a full slab, and sockets for a *different* domain are idling in
keep-alive, the new request would previously cause the client to panic!.

This patch adds a `spawn_error` handler which attempts to evict an idle
connection to make space for the new request. If space cannot be made,
the error handler is run (passed `Error::Full`) and the `Handler` is
dropped.

This is a breaking change because of the new variant of `Error`.

Some inefficient use of `Vec` in the client was replaced with `VecDeque`
to support push/pop from either end.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -3,7 +3,7 @@
 //! The HTTP `Client` uses asynchronous IO, and utilizes the `Handler` trait
 //! to convey when IO events are available for a given request.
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for digging into this!

PR #920, opened 2016-10-06
Referrer-Policy Header

### LINKED ISSUE #882: Referrer-Policy header should support multiple values
In order to support fallback policies when a subset of the full policies is supported (such as when a new policy is introduced in the specification), hyper's Referrer-Policy header should store the most recent value that is successfully parsed from the list of potential multiple values, per https://w3c.github.io/webappsec-referrer-policy/#parse-referrer-policy-from-header .

gsquire:
Instead of parsing one value in the `ReferrerPolicy` struct, allow for a list and use the last value. I believe this is what the ticket mentions, but I could have misinterpreted.

Closes #882

### DIFF
--- a/src/header/common/referrer_policy.rs
+++ b/src/header/common/referrer_policy.rs
@@ -54,19 +54,23 @@ impl Header for ReferrerPolicy {
 
     fn parse_header(raw: &Raw) -> ::Result<ReferrerPolicy> {
         use self::ReferrerPolicy::*;
-        parsing::from_one_raw_str(raw).and_then(|s: String| {
-            let slice = &s.to_ascii_lowercase()[..];
-            // See https://www.w3.org/TR/referrer-policy/#determine-policy-for-token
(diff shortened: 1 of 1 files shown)


jdm:
This isn't quite right - this still only looks at the very last comma-separated values and ignores the others. We want to return the rightmost value that is successfully recognized - this implies that we either need to iterate left to right and track the most recent successful parse, or iterate right to left and return the first value that is successful.

gsquire:
Ah I see.

I think the ideal way to do that would be to scan right to left checking the value each time and returning the first valid one. I will make that change. Thanks for your input.

seanmonstar:
@gsquire it would also help here to include a test in this module of this behavior

gsquire:
I don't think my solution is elegant so I am open to suggestions on improving it.

gsquire:
I refactored the ugly flow I had before. It should be good to go now.

jdm:
This looks good to me!

seanmonstar:
Thanks!

PR #924, opened 2016-10-07
* most certifactes has chains of certifactes, so we should

mabels:
also read them by default


seanmonstar:
Thanks for the PR! Could you clarify you comment a bit?

mabels:
yes I lost early control of this pull request. And did not found the github
function to remove the pull request.
I will start over again soon

cheers

meno

mabels:
close it to have a fresh start!
There is no way to change the subject later on!

PR #930, opened 2016-10-10
fix(conn): Handle remote hangup

jwilm:
Not handling this was an issue for keep-alive connections because
requests would get assigned to a closed connection and then immediately
error. Handling the HUP event makes this situation much less likely. It
is still possible however; consider the situation where a HUP arrives
while the event loop is busy processing new requests to add. The
connection is disconnected, but the HUP hasn't been processed, and a
request could be assigned to it. This case is, however, unlikely.

### DIFF
--- a/src/http/conn.rs
+++ b/src/http/conn.rs
@@ -553,6 +553,13 @@ impl<K: Key, T: Transport, H: MessageHandler<T>> Conn<K, T, H> {
             }
         }
 
+        if events.is_hup() {
+            trace!("Conn::ready got hangup");
+            let _ = scope.deregister(&self.0.transport);
+            self.on_remove();
+            return ReadyResult::Done(None);
+        }
+
         // if the user had an io interest, but the transport was blocked differently,
         // the event needs to be translated to what the user was actually expecting.
         //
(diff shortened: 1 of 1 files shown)


src/http/conn.rs
```diff
@@ -553,6 +553,13 @@ impl<K: Key, T: Transport, H: MessageHandler<T>> Conn<K, T, H> {
             }
         }
 
+        if events.is_hup() {
```
seanmonstar: I was certain that the `EPOLLERR` would be set if there was also `EPOLLHUP`...

src/http/conn.rs
```diff
@@ -610,6 +617,8 @@ impl<K: Key, T: Transport, H: MessageHandler<T>> Conn<K, T, H> {
             return ReadyResult::Continue(self);
         }
 
+        events = events | EventSet::hup();
```
seanmonstar: epoll docs suggest that `EPOLLHUP` is always listened for. May not be the case in kqueue/windows, I suppose...

jwilm:
Maybe just a problem of not checking for it then? Seems fine to explicitly register for it.

PR #931, opened 2016-10-12
Add 'take_buf' method to BufReader.

SergioBenitez:
Hello!

I'm using the latest 0.9.x release on a project I'm working on, and I'm quite in need of this method. As I'd like to release my project on crates.io soon, I was hoping you would merge this in and then release a new version of hyper 0.9.x on crates.io. I understand that 0.9.x has effectively been abandoned in favor of 0.10, so I hope it won't be too much trouble.

Thanks!
Sergio

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -44,6 +44,24 @@ impl<R: Read> BufReader<R> {
         }
     }
 
+    /// Extracts the buffer from this reader. Return the current cursor position
+    /// and the position of the last valid byte.
+    ///
+    /// This operation does not copy the buffer. Instead, it directly returns
+    /// the internal buffer. As a result, this reader will no longer have any
+    /// buffered contents and any subsequent read from this reader will not
+    /// include the returned buffered contents.
+    #[inline]
+    pub fn take_buf(&mut self) -> (Vec<u8>, usize, usize) {
+        let (pos, cap) = (self.pos, self.cap);
+        self.pos = 0;
+        self.cap = 0;
+
+        let mut output = vec![0; INIT_BUFFER_SIZE];
+        ::std::mem::swap(&mut self.buf, &mut output);
+        (output, pos, cap)
+    }
+
     #[inline]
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Note that the Travis-CI failure is unrelated to my PR.

SergioBenitez:
@seanmonstar ping.

seanmonstar:
This actually worries me a little. What happens if you take the buf and then try to keep on reading on the `Request`? It's internal counter of whether it can keep-alive will be all ruined, won't it?

SergioBenitez:
@seanmonstar Everything works as expected: the buffer will be refilled, and reading will continue. I explicitly tested that keep-alive continues working as expected.

seanmonstar:
Reading through the code, it looks like `HttpReader` wraps `BufReader`s, so the `SizedReader` (and `Chunked`) will not have counted how many bytes have been read off the socket, only how many have been read from the buffer.

What depends on this?

SergioBenitez:
That's right, but the original `Reader` shouldn't be used after calling `take_buf`. Even then, in the worst case, the `SizedReader` will simply say that there was an `eof`.

In a soon-to-be released project, I keep the TCP stream around for later reading. The current API forces me to keep around two lifetimes if I want to keep the TCP stream around, neither of which are necessary. With this API, I don't need to keep any lifetimes around, and I can rewrap the stream in more friendly APIs.

SergioBenitez:
Hi @seanmonstar, still waiting on this. With very high probability, I'll be publishing my crate next week, and I'd really like to not have a fork of Hyper in crates.io. Let me know. If it helps with getting this out, I'd be happy to add a `#[doc(hidden)]` tag so that others don't use this unnecessarily.

SergioBenitez:
@seanmonstar Can we get a new release with this change?

seanmonstar:
I published 0.9.14 a few days ago that included this.

PR #941, opened 2016-10-26
doc(intro): there is no guide

tshepang:


tshepang:
there is, only it doesn't render locally

PR #943, opened 2016-10-28
feat(headers): Add strict-origin and strict-origin-when-cross-origin referer policy

mrnayak:
Support for strict-origin and strict-origin-when-cross-origin in referer
policy required for imporving network security. This pull request will attempt
to add missing pieces of referrer policy.


seanmonstar:
Thanks!

Issue #907, opened 2016-08-31
Upgrade openssl dependency


zsck:
Currently, Hyper's [Cargo.toml](https://github.com/hyperium/hyper/blob/master/Cargo.toml) specifies using [Rust-OpenSSL](https://github.com/sfackler/rust-openssl) version `0.7.x`.  At the time of this writing, the [latest release](https://github.com/sfackler/rust-openssl/releases) is version `0.8.2`, which includes a number of bug fixes that, [in some applications](https://github.com/mattnenterprise/rust-ftp/pull/51) has resolved some functional failings.

I'd like to ask for Hyper to adopt these changes and upgrade to the latest version.

kwaegel:
Upgrading may also fix a build issue on Widows, as I mentioned in the linked bug for rust-rosetta: https://github.com/Hoverbear/rust-rosetta/issues/566 .

Unfortunately, `cookie` v0.3.0 also depends on `rust-openssl` v0.7, so fixing the build issue would require an upgrade in that crate as well. Such a change is already in master, but isn't part of a release yet.

Thomasdezeeuw:
`cookie` v0.3.1 now depends on v0.8. Is their anything else that is blocking an update to v0.8?

seanmonstar:
Bumping to openssl 0.8 is a breaking change, so that'd require bumping hyper as well. openssl is also addinv support for opensslv1.1, which will likely bump it to 0.9. I've been hoping hyper 0.10 would be async hyper.

My plan has been this:

- **hyper v0.10**: Be "async" hyper. Utilize tokio-tls.
- [**reqwest**](https://github.com/seanmonstar/reqwest): Be the convenient, higher level Client crate. This will release very soon, and depend on rust-native-tls. Everyone using hyper for a client should be able to easily switch the `reqwest`, getting better TLS support immediately, and a (as much as possible) not-breaking API, even as hyper v0.10 comes out with its async Client.

Thomasdezeeuw:
@seanmonstar that sounds great, looking forward to it. I would offer help, but I started learning Rust last week and I'm not sure I would be any help.

zsck:
Is there anything I could do to help speed this process along? My project at work has actually had to drop a couple of features because of the OpenSSL version mismatch between a few of our dependencies, and I'd like to be able to help resolve that if possible.

seanmonstar:
[...]

PR #945, opened 2016-10-31
Forcing OpenSSL to 0.8 -- discussion

i-e-b:
Hi.

The current master of hyper isn't building for me under Windows, due to a `fatal error C1083: Cannot open include file: 'openssl/ssl.h': No such file or directory` issue.
I have a 64-bit OpenSSL install, and the environment variables set: `OPENSSL_DIR`, `OPENSSL_LIB_DIR` and `OPENSSL_INCLUDE_DIR`.

A little bit of experimentation shows that the `openssl v0.7.14` package fails, but the `openssl v0.8.3` package builds OK.

Running the tests with OpenSSL 0.8 results in a few failures against the test server -- but I can't compare to 0.7.

Would it be possible to move from 0.7 to 0.8?

My OpenSSL install is `1.1.0b` x64 from http://slproweb.com/products/Win32OpenSSL.html
Rust version is `rustc 1.12.1 (d4f39402a 2016-10-19)` with cargo `cargo 0.13.0-nightly (109cb7c 2016-08-19)`
Running under Win 7 64-bit.

Thanks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,8 @@ authors = ["Sean McArthur <sean.monstar@gmail.com>",
 keywords = ["http", "hyper", "hyperium"]
 
 [dependencies]
+openssl-sys="*"
+openssl="*"
 httparse = "1.0"
 language-tags = "0.2"
 log = "0.3"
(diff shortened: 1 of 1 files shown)


seanmonstar:
Hi there, and thanks!

Unfortunately, using `*` means that every time openssl releases a new breaking change, hyper would try to use it immediately, even if the API has changed, and thus would no longer compile.

As for the eventual solution, this has been my plan: https://github.com/hyperium/hyper/issues/907#issuecomment-255509020

PR #946, opened 2016-10-31
fix(headers): Allow IPv6 Addresses in Host header

untitaker:


untitaker:
For reference see https://github.com/servo/rust-url/blob/master/src/host.rs#L85. Using the parser function from rust-url would be a bit wasteful in our case as we are storing the host as string.

seanmonstar:
Thanks!

untitaker:
Note that this has been merged into the maintenance branch and still needs to be merged into master

seanmonstar:
Thanks for the reminder, I've cherry picked this commit into master as well.

untitaker:
Fwiw I usually *merge* maintenance branches into master, because then git knows which commits are not yet merged into master and which are

seanmonstar:
Sure, the thing is that the 0.9.x branch contains some commits that are based on synchronous hyper, and so I cannot merge those back in to master.

PR #958, opened 2016-11-17
fix(http) move hup check after read/write check

shanegibbs:


jwilm:
Come to think of it, it probably doesn't make sense to try and write if the connection has terminated. Maybe the hup check should be between read and write?

seanmonstar:
Thanks!

shanegibbs:
@jwilm this makes sense to me. I created another PR https://github.com/hyperium/hyper/pull/962

PR #961, opened 2016-11-18
Fix examples and get tests compiling with api changes

mattwoodyard:
Fix  to work on the chunking bug and add some more tests.


seanmonstar:
Thanks!

PR #963, opened 2016-11-26
Moved futures/tokio deps to crates where possible to repair build

clemensw:
I wasn't able to build with a fresh checkout of the git-based dependencies.  Moving to the ones available as crates allowed the build to succeed


seanmonstar:
Thanks! I incorporated most of these changes, with a few modifications. The tokio branch is now up-to-date.

PR #971, opened 2016-12-14
[Tokio Branch] Create a `body::Body` struct, exposed as `hyper::Body` that wraps a `tokio_proto`-`Body`.

### LINKED ISSUE #969: [Tokio Branch] hyper::server::response::{ Body and IntoBody } are private, propose to expose publically
Currently, you can call `response.set_body()` with anything that implements `hyper::server::response::IntoBody`, and right now that is just `Body` itself, `Vec<u8>` and `&'static [u8]`.

In my application, I have my own ResponseBody struct, that I would like implement `IntoBody`, (to pass to `response.set_body()` but I cannot because `IntoBody` is private.

Additionally, I would like to be able to use the `hyper::server::response::Body` type, so that I can do the `Body::pair()` and `tx.start_send()` within my application, then pass that result to `response.set_body()`.

Is that potentially breaking any Hyper conventions? Are they private for a reason, and is there a better way of doing what I want to do? (Other than simply converting my ResponseBody to `Vec<u8>` and passing that in.)

ashleysommer:
Create a `body::Body` struct, exposed as `hyper::Body` that wraps a `tokio_proto`-`Body`.

Added `std::From` methods to convert from `tokio_proto`-`Body` to `body::Body` and back.
Made naming of `Body` declarations somewhat consistent across `Server`, `Server Request`, `Server Response`, `Client`, `Client Request`, and `Client Response`.
Impl `Stream` for `body::Body` so that applications can read from the body as if it were a `tokio_proto::streaming`-`Body`

All examples still work and tests are passing (except those that were broken before this modification).

Part 1/2 to close #969

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,0 +1,58 @@
+//! The Hyper Body, which is a wrapper around the tokio_proto Body.
+//! This is used as the Body for a Server Request, Server Response,
+//! and also for a Client Request and Client Response.
+//! It is based on a tokio_proto::streaming::Body for now, but will be
+//! changed to tokio_proto::multiplex::Body int the future for SSL support.
+
+use std::convert::From;
+
+use tokio_proto;
+use http::Chunk;
+use futures::{Poll, Stream};
+use futures::sync::mpsc;
+use futures::StartSend;
+
+pub type TokioBody = tokio_proto::streaming::Body<Chunk, ::Error>;
+
+#[derive(Debug)]
+pub struct Body(TokioBody);
+
+impl Body {
+    /// Return an empty body stream
+    pub fn empty() -> Body {
+        Body(TokioBody::empty())
+    }
+
+    /// Return a body stream with an associated sender half
+    pub fn pair() -> (mpsc::Sender<Result<Chunk, ::Error>>, Body) {
+        let (tx, rx) = TokioBody::pair();
+        let rx = Body(rx);
+        (tx, rx)
+    }
+}
+
+impl Stream for Body {
+    type Item = Chunk;
+    type Error = ::Error;
+
+    fn poll(&mut self) -> Poll<Option<Chunk>, ::Error> {
+        self.0.poll()
+    }
+}
+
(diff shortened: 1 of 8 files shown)


ashleysommer:
Added another commit.
Removed the two different `IntoBody` traits from `client::response` and `server::request`.
Replaced them both with a `Into<Body>` using the new `body::Body` (in the from of `From<>` impls).
There was one difference between the two existing implementations of `IntoBody` for `Vec<u8>`,
`client::request` used this implementation:
```
fn into(self) -> Body {
    let (mut tx, rx) = Body::pair();
    ::futures::lazy(move || tx.start_send(Ok(Chunk::from(self)))).wait().unwrap();
    rx
}
```
and `server::response` this this implementation:
```
fn into(self) -> Body {
    let (mut tx, rx) = Body::pair();
    tx.start_send(Ok(http::Chunk::from(self)));
    tx.poll_complete();
    rx
}
```
I understand the difference between the two, but unfortunately we cannot have two different impls with this blanket implementation of `hyper::Body`. I've gone with the latter implementation for the `Into<Body>` impl in this commit.
Part 2/2 Closes #969

seanmonstar:
Wow, thanks!

The client implementation for `IntoBody`, that uses `futures::lazy` was needed because calling `tx.start_send()` when not inside a `futures::task`, it will panic. Wrapping it in a lazy future, and then spawning a task via `future.wait()` makes the call no longer panic. I've spoken to @carllerche about it before, it might be that `tokio_proto::Body` gets a constructor that accepts an immediate result at some point.

ashleysommer:
Thanks for the clarification about the different implementations.

I see now that using that implementation does cause some client tests to crash (specifically the tests like client_post which include a body with the request) because it is calling `tx.start_send()` when not inside a `futures::task`;

I assumed that when creating a `Body`, we would always be inside a `futures::task`, but I will need to rethink that.

seanmonstar:
Thanks for all the work in doing this. I'll fix up that implementation myself, and update the tokio branch.

PR #972, opened 2016-12-14
[Tokio Branch] Fix examples failing in `cargo test`

ashleysommer:
`hyper::method::Method` is no longer accessible by users of hyper.
Must use `hyper::Method` instead.

All tests and examples now pass.


seanmonstar:
Thanks!

PR #975, opened 2016-12-17
feat(net): update openssl to 0.9 and cookies to 0.4

SergioBenitez:
As the title suggests. This change came about because I want to use the latest version of cookies, which requires the latest version of openssl. Furthermore, this allows for compilation with `ssl` on OS X without messing up the local libraries due to improvements in the `openssl` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,15 +25,11 @@ unicase = "1.0"
 url = "1.0"
 
 [dependencies.cookie]
-version = "0.2"
+version = "0.4"
 default-features = false
 
 [dependencies.openssl]
-version = "0.7"
-optional = true
-
-[dependencies.openssl-verify]
-version = "0.1"
+version = "0.9"
 optional = true
 
 [dependencies.security-framework]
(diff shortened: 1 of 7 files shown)


upsuper:
I guess the owner may want a patch based on `master` rather than `0.9.x` branch.

SergioBenitez:
@seanmonstar ping

steveklabnik:
Landing this and cutting a new version would be great 👍

drusellers:
Yes please

doomsplayer:
now hyper becomes the main stumbling stone of projects because of openssl issue actually...

dirk:
@seanmonstar, @reem: Anything we can do as a community to help this PR along?

seanmonstar:
I haven't been ignoring this, but instead talking out the problem in detail with some folks, and I think I have a solution. I've filed a new issue with the long write up at https://github.com/hyperium/hyper/issues/985

**tl;dr**: A proposal to release a new version of synchronous hyper that removes openssl from its dependency list.

seanmonstar:
The solution proposed in #985 was used, and 0.10 is now published: https://github.com/hyperium/hyper/releases/tag/v0.10.0

Thanks everyone for pushing on this idea, I really feel like this was the best solution and we can all move forward!

PR #976, opened 2016-12-20
Don't write a 0-length buffer in flush()

alexcrichton:
Ends up fixing some cases when the underlying writer doesn't handle a
0-length write well, and also hopefully a nice operation to skip!


seanmonstar:
This was done in https://github.com/hyperium/hyper/commit/fc5003125201b0d6cc476cd470a2070061da724e, thanks though!

PR #979, opened 2016-12-31
refactor(client): remove unused client type aliases

jamesmunns:
These type aliases cause test build to fail due to strict LINT settings

If these types should be preserved (used outside the test suite), please let me know and I can instead disable the LINT for these types.


seanmonstar:
Thanks!

PR #983, opened 2017-01-02
implemented a feature to use native-tls

tisoft:
Use https://github.com/sfackler/rust-native-tls to provide the TLS stack.

I have only tested this with hyper client, because I only need client side TLS in my application. Also it is based on th 0.9.x branch, not current master. If you think this could actually be merged, I will try to rebase it to master. Right now it is just in the state that works for me . :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+.idea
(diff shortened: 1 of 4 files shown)


seanmonstar:
A different solution was used, explained in #985, and a new version has been published: https://github.com/hyperium/hyper/releases/tag/v0.10.0

Thanks for pushing on this!

PR #989, opened 2017-01-07
docs(server): fix Response::set_status() documentation

d-unseductable:
`Response::set_status(&mut self, status: StatusCode)` sets the status of the response and does not return a mutable reference.


seanmonstar:
Thanks!

PR #990, opened 2017-01-07
Fix tests compilation

GuillaumeGomez:


seanmonstar:
Thanks! The failing test is do to an error in `MemBuf::reserve`. I'll fix that up.

PR #994, opened 2017-01-10
address tokio_service mutability changes

ashleygwilliams:
ideally eventually fixes #993

PR #996, opened 2017-01-10
fix(client): Client functions don't need &mut self

pimeys:
Hi,

Is there a reason these client functions require a &mut self? Makes
integrating this to our project much easier if they would be immutable.

Tests seem to pass.


seanmonstar:
they used to, when Service.call required mutable self, but that changed in #994, and we just forgot to update these. Thanks!

pimeys:
<3

PR #998, opened 2017-01-11
feat(headers): Add support for Retry-After header

pimeys:
This used to be an external crate, which seem to be abandoned and
stuck with Hyper 0.7. https://github.com/jwilm/retry-after

I think it would be better to include it with Hyper and I'm willing
to support this code here if needed. I took the code from that repo
and made sure it works with the tokio branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,20 +17,17 @@ httparse = "1.0"
 language-tags = "0.2"
 log = "0.3"
 mime = "0.2"
-mio = "0.6"
-native-tls = "0.1.0"
 relay = "0.1"
 rustc-serialize = "0.3"
 time = "0.1"
 tokio-core = "0.1"
-tokio-tls = {git = "https://github.com/tokio-rs/tokio-tls"}
 tokio-proto = "0.1"
 tokio-service = "0.1"
 unicase = "1.0"
 url = "1.0"
-vecio = "0.1"
 
 [dev-dependencies]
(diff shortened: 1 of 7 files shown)


markuskobler:
Rather than adding the chrono dependency is it easy enough to use the time crate?

pimeys:
You are absolutely right. Removed the dependency and relying only on what time crate has to offer.

seanmonstar:
You may find the `headers::shared::HttpDate` useful here,

pimeys:
Cool. I'll take a look tomorrow. This is my first real dip into this project's internals, so thanks for pointing nice things. Also if there's somewhere a TODO what is left to get this tokio branch to be 0.1 ready, you can point me some tasks. Our company will be one of the first ones using this in production, so I better know every single detail how hyper-tokio works...

jwilm:
This should probably include the license from the retry-after crate. Glad to see it making it into hyper!

pimeys:
@jwilm Will add it today. Thanks for a great library. We need this and having it as a separate library makes life much harder than just having it in here.

pimeys:
@jwilm @seanmonstar @markuskobler et.al. What is the preferred way of adding the license from retry-after to hyper? Retry-after is MIT and Apache, Hyper is MIT.

The latest amend utilized the `HttpDate` module allowing me to delete more code from the `retry_after` module.

PR #1005, opened 2017-01-13
Fix the server_get_fixed_response test

alexcrichton:
I found that when decoding headers the connection would erroneously move itself
to the `Reading::KeepAlive` state even though `wants_keep_alive` is false. This
commit touches up the logic in conn.rs to check this predict the next state
accordingly, fixing a failing server_get_fixed_response locally for me.


seanmonstar:
There was actually a different underlying bug that this put a bandaid on, I've fixed it up in https://github.com/hyperium/hyper/commit/722661e6ea317a0b0cd8bc4d8864a26527f93da1

Issue #950, opened 2016-11-06
Requests to hyper server eventually start failing.


jswrenn:
I wrote a basic server application using hyper 0.9.10 that prints the IP, referrer, and body of each request to stdout (which I then redirect into a file):
```
extern crate hyper;
use std::env;
use std::io::Read;
use hyper::method::Method::Post;
use hyper::server::{Server, Request, Response};
use hyper::header::{Referer, AccessControlAllowOrigin};
use hyper::net::Openssl;

fn main() {

  let mut args = env::args().skip(1);
  let addr = args.next().unwrap();
  let cert = args.next().unwrap();
  let key  = args.next().unwrap();
  
  let ssl = Openssl::with_cert_and_key(cert, key).unwrap();

  let _ = Server::https((addr.as_str(), 443), ssl).unwrap()
    .handle(|req: Request, mut res: Response| {
      res.headers_mut().set(AccessControlAllowOrigin::Any);
      let _ = res.send(b"");
      if req.method == Post {
        let (addr, _, headers, _, _, mut body) = req.deconstruct();
        headers.get::<Referer>().
          map(|&Referer(ref referer)| {
            let mut buffer = String::new();
            let _ = body.read_to_string(&mut buffer);
            println!("{:}\t{:}\t{:}", addr, referer, buffer);
          });
      }
    });
}
```

Everything works as expected initially, but after some number of hours after starting the server, all requests to it begin to fail (and do so very slowly). When this happens, the results of `time curl -k ...` look like this:
```
curl: (35) Unknown SSL protocol error in connection to logbook.pyret.org:443 

real	1m43.616s
user	0m0.013s
sys	0m0.000s
```

Restarting the server application corrects this issue and requests are handled within 100ms, tops.

I'm flummoxed.

seanmonstar:
Does this problem occur without the use of openssl?

jswrenn:
I'm not sure. I have not yet managed to reproduce the problem in a short amount of time, and running the application without openssl is not possible.

hannobraun:
I'm possibly seeing the same issue. I haven't spoken up so far because I'm not sure it's Hyper-related or caused by my own code, and I haven't had the time to look into it yet. But since this issue is open now, I figured I'd chime in.

If my issue is Hyper-related at all, it definitely only occurs in connection with OpenSSL. HTTPS requests don't get a reply, while HTTP requests are handled promptly. The issue goes away after a restart.

[...]

PR #1006, opened 2017-01-13
Set timeouts immediately after accepting a connection

hannobraun:
This is a possible solution for #950. I'm not running this in production yet, and thus could not confirm whether it really solves the problem. I thought I'd submit it anyway, so it can get some review. Whether or not this really is a solution for #950, I believe it fixes a real problem and should be merged in any case.

I've struggled and failed to come up with an automated test to verify that this does what it intends to do. Since there doesn't seem to be a test for timeout behavior currently, and this is a relatively straight-forward change, I decided to submit it without an automated test.

Please note that this only affects the server-side code. I haven't checked whether the same problem is present client-side.

Another note: This pull request sets the default read and write timeouts to `None`, which I believe preserves the current behavior. This might not be a good default for a server-side application, so some thought should be put into whether to change this.

A reminder to me, or in case I forgot, anyone else who might read this: This pull request adds default implementations of trait methods that should be removed in the next release which includes breaking changes (see comments in code). It would probably be a good idea to open an issue as a reminder, in case this pull request is merged.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -31,6 +31,22 @@ pub trait NetworkListener: Clone {
     fn incoming(&mut self) -> NetworkConnections<Self> {
         NetworkConnections(self)
     }
+
+    /// Sets the read timeout for all streams that are accepted
+    fn set_read_timeout(&mut self, _: Option<Duration>) {
+        // This default implementation is only here to prevent the addition of
+        // these methods from being a breaking change. They should be removed
+        // when the next breaking release is made.
+        warn!("Ignoring read timeout");
+    }
+
+    /// Sets the write timeout for all streams that are accepted
+    fn set_write_timeout(&mut self, _: Option<Duration>) {
+        // This default implementation is only here to prevent the addition of
+        // these methods from being a breaking change. They should be removed
+        // when the next breaking release is made.
+        warn!("Ignoring write timeout");
+    }
 }
 
 /// An iterator wrapper over a `NetworkAcceptor`.
@@ -205,6 +221,9 @@ impl NetworkStream + Send {
 /// A `NetworkListener` for `HttpStream`s.
 pub struct HttpListener {
     listener: TcpListener,
+
+    read_timeout : Option<Duration>,
+    write_timeout: Option<Duration>,
 }
 
 impl Clone for HttpListener {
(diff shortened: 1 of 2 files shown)


src/net.rs
```diff
@@ -31,6 +31,22 @@ pub trait NetworkListener: Clone {
     fn incoming(&mut self) -> NetworkConnections<Self> {
         NetworkConnections(self)
     }
+
+    /// Sets the read timeout for all streams that are accepted
+    fn set_read_timeout(&mut self, _: Option<Duration>) {
+        // This default implementation is only here to prevent the addition of
+        // these methods from being a breaking change. They should be removed
+        // when the next breaking release is made.
+        unimplemented!();
```
seanmonstar: Since these methods are called unconditionally later, anyone just upgrading hyper will find that suddenly, their custom `NetworkConnector` panics, which would be a breaking change. These should probably just do nothing, instead of calling `unimplemented!`.

If we want, we could put a `warn!` or even `error!` in here instead, so people looking at logs can see they should implement these. Or we could just do nothing.

hannobraun: You're right. I didn't think this through.

I think a `warn!` would be appropriate. Doing nothing would just mean that timeouts would be silently ignored, which could be very surprising to both users and implementers of custom `NetworkConnector`s. An `error!` seems too strong to me, as nothing is actually going wrong (yet) at this point.

hannobraun:
Pushed the fixes suggested by @seanmonstar. Also adjusted a misleading commit title (replaced "fix(net)" with "fix(server)".

seanmonstar:
I merge manually, since I wanted to change the `fix(server)` into a `refactor(server)`, but not bother you about it. Merged into the 0.10.x branch here: https://github.com/hyperium/hyper/commit/baef7abbdfe22ee4c008ce291cb96e88f5d5c8b0

Thanks a bunch for this. I hope it helps with #950.

seanmonstar:
Oh right, PS, if its at all possible to test that this does help #950 using a git dependency, that'd be great, so we can know if releasing `0.10.1` would be a notable bug fix.

hannobraun:
Great, thanks!

> Oh right, PS, if its at all possible to test that this does help #950 using a git dependency, that'd be great, so we can know if releasing 0.10.1 would be a notable bug fix.

Will do. I'll keep you updated in #950 about what I find out.

PR #1007, opened 2017-01-14
Create new Uri struct

GuillaumeGomez:
Here is the first draft of the new `Uri` struct. If this seems okay to you @seanmonstar, I'll remove the actual `UriRequest` and replace it with the new `Uri`.

PS: This is a WIP so do not merge!

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -23,12 +23,12 @@ impl Service for Echo {
 
     fn call(&mut self, req: Request) -> Self::Future {
         ::futures::finished(match (req.method(), req.path()) {
-            (&Get, Some("/")) | (&Get, Some("/echo")) => {
+            (&Get, "/") | (&Get, "/echo") => {
                 Response::new()
                     .with_header(ContentLength(INDEX.len() as u64))
                     .with_body(INDEX)
             },
-            (&Post, Some("/echo")) => {
+            (&Post, "/echo") => {
                 let mut res = Response::new();
                 if let Some(len) = req.headers().get::<ContentLength>() {
                     res.headers_mut().set(len.clone());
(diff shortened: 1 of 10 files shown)


seanmonstar:
This is absolutely the direction I had in mind. I would also remove the use
of the Url type here, since it actually does more than is wanted on the
server side, and does several allocations. Since this type is created for
every request, we need to make it faster.

I can do that as follow up if you prefer.

GuillaumeGomez:
I think I can do it in a second step. I'll just "borrow" the url method and use them directly instead of creating a type. For now I'll replace `RequestUri` everywhere and then I'll remove the url type usage.

examples/server.rs
```diff
@@ -23,12 +23,12 @@ impl Service for Echo {
 
     fn call(&mut self, req: Request) -> Self::Future {
         ::futures::finished(match (req.method(), req.path()) {
-            (&Get, Some("/")) | (&Get, Some("/echo")) => {
+            (&Get, "/") | (&Get, "/echo") => {
```
seanmonstar: This looks so much better!

src/uri.rs
```diff
@@ -1,11 +1,260 @@
...
+
+    pub fn fragment(&self) -> Option<&str> {
+        if let Some(len) = self.fragment {
+            Some(&self.source[self.source.len() - len..])
+        } else {
+            None
+        }
+    }
+
+    pub fn uri(&self) -> &str {
```
seanmonstar: I'd probably replace this with an `AsRef<str>` impl, since it sounds like stuttering when saying `req.uri().uri()`.

GuillaumeGomez: Hum, good point!

GuillaumeGomez:
I think I'll stop here for this PR. I cleaned up a bit and remove the old RequestUri.

@seanmonstar: Do you see anything I might have forgotten?

PR #1011, opened 2017-01-16
Fix heap corruption in http::buffer::grow_zerofill

kornholi:


antrik:
Hint: such issues can be significantly reduced by using a defensive programming approach based on slicing:

```
let len = buf.len();
buf.set_len(len + additional);
let to_fill = &buf[len..];
ptr::write_bytes(to_fill.as_mut_ptr(), 0, to_fill.len());
```

(In this specific case though I'm not sure why this unsafe code exists at all -- rather than just doing something like `let len = buf.len(); buf.resize(len + additional, 0)`?...)

seanmonstar:
The original code was taken from a PR on the standard library, and then broken by myself by not having tests. I've been rewriting it today with some tests and asserts, though I had somehow missed that `Vec::resize` was a thing.

PR #1012, opened 2017-01-17
Refactor the server module

alexcrichton:
This commit refactors the `server` module as we discussed a bit offline.
The main changes are:

* The entry point is how `Http`, the implementation of `ServerProto`.
  This type has a `new` constructor as well as builder methods to
  configure it.

* A high-level entry point of `Http::bind` was added which returns a
  `Server`. Binding a protocol to a port requires a socket address
  (where to bind) as well as the instance of `NewService`. Internally
  this creates a core and a TCP listener.

* The returned `Server` has a few methods to learn about itself, e.g.
  `local_addr` and `handle`, but mainly has two methods: `run` and
  `run_until`.

* The `Server::run` entry point will execute a server infinitely, never
  having it exit.

* The `Server::run_until` method is intended as a graceful shutdown
  mechanism. When the provided future resolves the server stops
  accepting connections immediately and then waits for a fixed period of
  time for all active connections to get torn down, after which the
  whole server is torn down anyway.

* Finally a `Http::bind_connection` method exists as a low-level entry
  point to spawning a server connection. This is used by `Server::run`
  as is intended for external use in other event loops if necessary or
  otherwise low-level needs.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -5,7 +5,7 @@ extern crate pretty_env_logger;
 //extern crate num_cpus;
 
 use hyper::header::{ContentLength, ContentType};
-use hyper::server::{Server, Service, Request, Response};
+use hyper::server::{Http, Service, Request, Response};
 
 static PHRASE: &'static [u8] = b"Hello World!";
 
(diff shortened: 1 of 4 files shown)


seanmonstar:
Er, I didn't realize deleting that branch would close this. I had finished merging the tokio branch into master, and it was now out of date. Also, noticed there is a compile error in the benches/end_to_end.rs.

alexcrichton:
Sure thing, I'll send the PR against master instead

PR #1017, opened 2017-01-18
feat(headers): Add support for Retry-After header

pimeys:
This used to be an external crate, which seem to be abandoned and
stuck with Hyper 0.7. https://github.com/jwilm/retry-after

I think it would be better to include it with Hyper and I'm willing
to support this code here if needed. I took the code from that repo
and made sure it works with the tokio branch.


pimeys:
@seanmonstar The only open question with this PR is, that I used parts of the retry-after library. Even though this is quite modified already, the original was Apache+MIT and I don't know how to deal with the licensing; should we include the Apache license only for this part or not.

seanmonstar:
I don't know how to have different licenses for different files. I have
been wanting to relicense hyper with a dual license, but trying to get all
past contributors to acknowledge and allow the change has so far felt like
a big hurdle.

pimeys:
@jwilm It is your call. How would you like the licensing to work here. Do want to merge it to HEAD here or should we change eetry-after library to support the modern 0.10 hyper?

jwilm:
@pimeys I think you can just paste retry-after's license into the top of the file you're adding to hyper. They're both available under the MIT license.

pimeys:
@jwilm The MIT or the Apache? I see you have a double license there. I can add both :)

jwilm:
Adding both would cover the case where hyper actually manages to re-release under a dual license. That probably makes sense. Make sure you just take the "header" portion of the Apache license, though.

Thanks for doing this work @pimeys :)

pimeys:
@jwilm Added the licenses. Hopefully all is good now :)

seanmonstar:
Thanks!

PR #1021, opened 2017-01-20
fix(uri): fix panic when uri contain default port

fmonjalet:
Hi,

Parsing `"http://host:80"` caused panic (index out of bound) because the
authority end index was computed with the original uri but the uri
stored for later used was sanitized by `Url::parse()` to `"http://host"`.

The fix computes the autority end index with the actual uri used (the
one from `Url::parse()`). Two tests have been added.

Please do not hesistate if you want me to adapt the fix differently.

Cheers,
Florent


seanmonstar:
Thanks!

Issue #1022, opened 2017-01-20
Remove usage of Url parse in Uri


seanmonstar:
There is legacy code in the `Uri` type that relies on `Url::parse`, even to parse only parts of the uri. This is unnecessary and slow.

Maybe not the easiest thing to do (regarding the easy label), since it requires making some `parse_scheme` and `parse_authority` and such kind of methods, but it's at least fairly straight forward, and contained to a single module.

M3rs:
I'm interested in giving this a try. Is it just alright to put in a pull request when I have something that passes tests?

seanmonstar:
@M3rs absolutely!

seanmonstar:
I meant to include this originally, but the `parse` function from Go is a good example: https://github.com/golang/go/blob/master/src/net/url/url.go#L469

M3rs:
Question: How do you want the authority with a default port handled? 

Should I refactor the method `pub fn authority(&self) -> Option<&str>` so if the port is a default (80, 443), it just returns the "authority" and not the port portion?

```
related test failures:
---- uri::test_uri_parse_absolute_with_default_port_http stdout ----
	thread 'uri::test_uri_parse_absolute_with_default_port_http' panicked at 'assertion failed: `(left == right)` (left: `Some("127.0.0.1:80")`, right: `Some("127.0.0.1")`)', uri.rs:324

---- uri::test_uri_parse_absolute_with_default_port_https stdout ----
	thread 'uri::test_uri_parse_absolute_with_default_port_https' panicked at 'assertion failed: `(left == right)` (left: `Some("127.0.0.1:443")`, right: `Some("127.0.0.1")`)', uri.rs:335
```

My forked/branched version: [uri.rs](https://github.com/M3rs/hyper/blob/refactor_uri/src/uri.rs)

seanmonstar:
@M3rs perhaps the best to do is include the port if it was in the original URI, and if it wasn't, do not include.

So:

```
http://foo.bar:3333/baz => foo.bar:3333
http://foo.bar:80/baz   => foo.bar:80
http://foo.bar/baz      => foo.bar
```

M3rs:
Ok - I have pushed an update to my [uri.rs](https://github.com/M3rs/hyper/blob/refactor_uri/src/uri.rs) which causes the port to be included.

What are the rules on path? The test_parse! methods give a mixed idea of what it should be expected if no path is given.

Examples:
```
address => path
---
http://127.0.0.1:80 => "/"
[...]

PR #1027, opened 2017-01-23
Refactor(uri): Remove usage of Url parse in Uri #1022

M3rs:
Remove usage of Url parse, add basic parse methods. 

Also add minimal error handling, default port/path handling (tests adjusted accordingly). 

https://github.com/hyperium/hyper/issues/1022

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -70,10 +70,13 @@ impl Uri {
                     "blob" | "file" => return Err(Error::Method),
                     _ => return Err(Error::Method),
                 }
-                if let Some(a) = auth {
-                    if (end + 3) == a {
-                        return Err(Error::Method);
-                    }
+                match auth {
+                    Some(a) => {
+                        if (end + 3) == a {
+                            return Err(Error::Method);
+                        }
+                    },
+                    None => return Err(Error::Method),
                 }
             }
             Ok(Uri {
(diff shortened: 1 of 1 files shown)


src/uri.rs
```diff
@@ -179,6 +175,39 @@ impl Uri {
     }
 }
 
+fn parse_scheme(s: &str) -> Option<usize> {
+    s.find(':')
+}
+
+fn parse_authority(s: &str) -> Option<usize> {
+    let v: Vec<&str> = s.split("://").collect();
+    let auth = v.last().unwrap()
```
michael-zapata: That straight `unwrap()` seems a bit extreme.

M3rs: It is only used in the block `else if s.contains("://")` (albeit, now in a method call).

Would you prefer the `fn parse_authority(s: &str) -> Option<usize>` method removed ([definition](https://github.com/M3rs/hyper/blob/refactor_uri/src/uri.rs#L182)) and inlined in place of `let auth = parse_authority(s);` ([relevant line](https://github.com/M3rs/hyper/blob/refactor_uri/src/uri.rs#L66)) so the `unwrap()` call is safer?

michael-zapata: Well, since this is meant to be an option, it may be a good idea to use the `?` operator there to return the `None` value instead of panicking.

seanmonstar: I planned on filing issues for a couple more passes through the file, removing the temporary `Vec`s and what-not. It can be fixed here if you want, or in another pass.

M3rs: Good point! Admittedly, that was a copy-paste from the previous code and I didn't think about it. Pushed an update.

M3rs:
Build fail (nightly) note:

error: unused `#[macro_use]` import
  --> src/lib.rs:20:1

Is that something I can address, or not my fault?

seanmonstar:
Looks like the newest nightly updated to better detect unused macros, not your fault.

seanmonstar:
Thanks so much @M3rs, this is fantastic work! I'll file follow up issues to remove the `Vec` usage as well.

Issue #1029, opened 2017-01-23
Remove allocations (Vec) in Uri parsing


seanmonstar:
Currently some `Vec`s are created for temporary purposes in `Uri` parsing. These could be removed entirely.

M3rs:
I'd be happy to work on this.

PR #1030, opened 2017-01-23
refactor(uri): Remove vec in uri parsing

M3rs:
Remove Vec allocations in Uri parsing. Additionally, change some tests
to check the port.

https://github.com/hyperium/hyper/issues/1029


seanmonstar:
Excellent, thank you!

Issue #1028, opened 2017-01-23
Replace rustc-serialize with base64 crate


seanmonstar:
The rustc-serialize crate is used by the `Basic` authorization header. It's a big crate, we don't need all of it, and the [base64](https://crates.io/crates/base64) is faster anyways.

M3rs:
This just in authorization.rs, right? I can try to swap these.

seanmonstar:
Correct.

robinst:
Looks like this was done on master and is a relatively small change. Would you mind if I opened a PR to backport it to 0.10.x?

seanmonstar:
@robinst sure, that'd be fine!

robinst:
Thanks! For anyone wondering, the above is now also released in 0.10.10.

PR #1031, opened 2017-01-24
refactor(header): Change to base64 serialize

M3rs:
Replace rustc-serialize with base64 crate.

https://github.com/hyperium/hyper/issues/1028


seanmonstar:
Excellent, thank you!

PR #1035, opened 2017-01-25
Add PUT method to exported list

itsmontoya:


itsmontoya:
I don't  understand why the commit-message-check failed, my message is less than 72 chars :/

GuillaumeGomez:
Oh nice catch!

seanmonstar:
The issue was that the commit message didn't match the format defined https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md#git-commit-guidelines

I modified the commit message to match, thanks!

seanmonstar:
They are very uncommon, and you can always access them as `hyper::Method::Patch` etc.

PR #1040, opened 2017-01-29
feat(http): add Expect 100-continue support

martell:

This is a WIP


seanmonstar:
This seems blocked on tokio-proto trying to buffer the first body frame, https://github.com/tokio-rs/tokio-proto/issues/153.

I'm going to close this until that is figured out, so that I don't keep thinking that there is a Pull Request I haven't looked at.

PR #1041, opened 2017-01-31
Make Uri use MemSlice internally

GuillaumeGomez:


GitCop:

 * Commit: 6ef21d39fdf7c7bfdd60490068c1b2dc26af9b19

---

GitCop:

 * Commit: 546cb554ca8d1d570907ab6e7756b8f91171ed88

---

src/uri.rs
```diff
@@ -30,7 +30,7 @@ use Error;
 /// scheme          authority                 path            query         fragment
 #[derive(Clone)]
 pub struct Uri {
-    source: Cow<'static, str>,
```
seanmonstar: Again, instead of using only `MemSlice`, I'd think an enum of static, owned, or shared would be the best here. This is because it's not free to convert to a `MemSlice` if we don't already have one. And we want to keep the optimization for a super common route, `/` being just a static constant.

src/http/buf.rs
```diff
@@ -226,6 +227,20 @@ impl MemSlice {
     fn get(&self) -> &[u8] {
         unsafe { &(*self.buf.get())[self.start..self.end] }
     }
+
+    pub fn as_str(&self) -> &str {
```
seanmonstar: I've been thinking we'd need something like this, but a slightly different way may be better.

```rust
pub struct MemStr(MemSlice);

impl MemStr {
    pub unsafe fn from_utf8_unchecked(slice: MemSlice) -> MemStr {
        MemStr(slice)
    }
}
```

This forces us to use an `unsafe {}` block whenever we create a string, so we don't accidentally just create a `MemStr` that might not be utf8.

src/uri.rs
GuillaumeGomez: Ok, let's update it this way then!

src/http/buf.rs
GuillaumeGomez: I'm not so sure to understand the usefulness of having a new struct for this. Just making the method unsafe should be enough, no?

GuillaumeGomez:
New version with internal enum.

src/http/buf.rs
seanmonstar: It's because we want to keep the `unsafe` near the proof. The proof that a certain `MemSlice` is valid utf8 is when we look at the `req.path` from httparse. Consider:

```rust
impl Uri {
    pub fn with_some_slice(slice: MemSlice) {
        // save the slice internally
    }

    fn something_else(&self) -> &str {
        unsafe { from_utf8_unchecked(self.mem_slice.as_ref()) }
    }
}
```

Now, what happens if future us (or some other future person), perhaps not remembering that this unsafety exists, creates a `Uri::with_some_slice(my_maybe_not_utf8_slice)` somewhere else in the code. It wasn't unsafe to create the `Uri`, so we don't realize that we did anything unsafe.

Compare:

```rust
impl Uri {
    pub fn with_some_mem_str(str: MemStr) -> Uri {

    }

    fn something_else(&self) -> &str {
        &self.mem_str.as_ref() // not unsafe!
    }
}
```

A future persons then says "I want to create a `Uri` over here, using my `MemSlice`, how do I do that? Oh, it takes a `MemStr`, how do I make that? Uh oh, only with `unsafe from_utf8_unchecked()`, I guess I need to check my `MemSlice` before trying that.

src/uri.rs
```diff
@@ -207,13 +250,19 @@ impl FromStr for Uri {
...
 impl From<Url> for Uri {
     fn from(url: Url) -> Uri {
-        Uri::new(url.as_str()).expect("Uri::From<Url> failed")
+        Uri::new(InternalUri::from(url.as_str())).expect("Uri::From<Url> failed")
+    }
+}
+
+impl From<MemSlice> for Uri {
+    fn from(m: MemSlice) -> Uri {
+        Uri::new(InternalUri::from(m)).expect("Uri::From<MemSlice> failed")
```
seanmonstar: A `MemSlice` comes from receiving bytes from the socket. It has gone through `httparse`, which made sure there weren't radically invalid bytes here (just that it is a `str` without spaces), but we don't know if it is actually a well-formed `Uri`. So, we can't really provide the `From<MemSlice>` implementation, here.

seanmonstar:
There would also need to be a change in `http/h1/parse`, instead of using `req.path.unwrap().parse()`. Instead, we want to create a `Uri` using the related `MemSlice`. It could look like this:

```rust
let path = slice.slice(path_start..path_end);
// path was found to be utf8 by httparse
let path = unsafe { MemStr::from_utf8_unchecked(path) };
let uri = try!(::uri::from_mem_str(path));
```

src/uri.rs
GuillaumeGomez: Seems fair, I remove it.

GuillaumeGomez:
Updated.

src/http/buf.rs
```diff
@@ -226,6 +240,25 @@ impl MemSlice {
     fn get(&self) -> &[u8] {
         unsafe { &(*self.buf.get())[self.start..self.end] }
     }
+
+    pub fn as_str(&self) -> MemStr {
+        unsafe { MemStr::from_utf8_unchecked(self.clone()) }
+    }
+
+    pub fn find(&self, c: char) -> Option<usize> {
```
seanmonstar: It looks like this is no longer used, or at least I cannot find it.

src/http/buf.rs
```diff
@@ -226,6 +240,25 @@ impl MemSlice {
     fn get(&self) -> &[u8] {
         unsafe { &(*self.buf.get())[self.start..self.end] }
     }
+
+    pub fn as_str(&self) -> MemStr {
```
seanmonstar: With the constructor on `MemStr`, this method is no longer needed.

src/http/buf.rs
```diff
@@ -226,6 +240,25 @@ impl MemSlice {
...
+        for (pos, x) in self.get().iter().enumerate() {
+            if *x == c {
+                return Some(pos);
+            }
+        }
+        None
+    }
+
+    #[doc(hidden)]
+    pub fn as_full_str(&self) -> &str {
```
seanmonstar: This method should not exist on `MemSlice`.

src/http/buf.rs
```diff
@@ -303,8 +336,8 @@ impl PartialEq<[u8]> for MemSlice {
     }
 }
 
-impl PartialEq<str> for MemSlice {
-    fn eq(&self, other: &str) -> bool {
+impl<'a> PartialEq<&'a str> for MemSlice {
```
seanmonstar: Why this change?

src/uri.rs
```diff
@@ -1,11 +1,48 @@
...
+use http::buf::{MemSlice, MemStr};
 use url::Url;
 use url::ParseError as UrlError;
 
 use Error;
 
+#[derive(PartialEq, Eq, Clone)]
+enum InternalUri {
+    Static(Cow<'static, str>),
+    Shared(MemSlice),
```
seanmonstar: This should hold a `MemStr` instead.

src/uri.rs
```diff
@@ -207,13 +250,19 @@ impl FromStr for Uri {
...
 impl From<Url> for Uri {
     fn from(url: Url) -> Uri {
-        Uri::new(url.as_str()).expect("Uri::From<Url> failed")
+        Uri::new(InternalUri::from(url.as_str())).expect("Uri::From<Url> failed")
+    }
+}
+
+impl From<MemStr> for Uri {
+    fn from(url: MemStr) -> Uri {
+        Uri::new(InternalUri::from(url.as_str())).expect("Uri::From<MemStr> failed")
```
seanmonstar: `From` implementations should be for conversions that cannot fail. As mentioned earlier, we receive a bunch of bytes from the socket, and after some `httparse` calls, we know the URI is a string, but we don't know if it is a valid `Uri`. It needs to go through `Uri::new` to parse it, and that could fail. It's better to not panic.

I've updated master with a test case in the parse.rs module that will trigger this case, so if you rebase you should see the panic (that should be an `Err` instead).

src/http/buf.rs
GuillaumeGomez: It is quite useful actually for some internal operation when you want to translate directly from `MemSlice` to `&str`.

src/http/buf.rs
GuillaumeGomez: To be able to perform this comparison in an easier way.

src/uri.rs
GuillaumeGomez: So what am I supposed to do in here?

GuillaumeGomez: I just removed it.

GuillaumeGomez:
@seanmonstar: Updated: anything else to change?

GuillaumeGomez:
New benches:

```
running 2 tests
test get_one_at_a_time  ... bench:     128,290 ns/iter (+/- 11,401) = 402 MB/s
test post_one_at_a_time ... bench:     127,876 ns/iter (+/- 12,776) = 403 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured


     Running target/release/deps/hyper-d8f89bd3a96bdf05

running 361 tests
test header::common::accept::bench::bench_format                    ... bench:         583 ns/iter (+/- 191)
test header::common::accept::bench::bench_parse                     ... bench:         586 ns/iter (+/- 29)
test header::common::accept_language::bench::bench_format           ... bench:         794 ns/iter (+/- 385)
test header::common::accept_language::bench::bench_parse            ... bench:       1,154 ns/iter (+/- 53)
test header::common::allow::bench::bench_format                     ... bench:         371 ns/iter (+/- 189)
test header::common::allow::bench::bench_parse                      ... bench:         633 ns/iter (+/- 53)
test header::common::authorization::basic::bench_format             ... bench:         339 ns/iter (+/- 143)
test header::common::authorization::basic::bench_parse              ... bench:         193 ns/iter (+/- 52)
test header::common::authorization::bearer::bench_format            ... bench:         183 ns/iter (+/- 42)
test header::common::authorization::bearer::bench_parse             ... bench:          45 ns/iter (+/- 6)
test header::common::authorization::raw::bench_format               ... bench:          52 ns/iter (+/- 14)
test header::common::authorization::raw::bench_parse                ... bench:          38 ns/iter (+/- 12)
test header::common::cache_control::normal::bench_format            ... bench:         198 ns/iter (+/- 18)
test header::common::cache_control::normal::bench_parse             ... bench:         275 ns/iter (+/- 47)
test header::common::connection::close::bench_format                ... bench:          60 ns/iter (+/- 9)
test header::common::connection::close::bench_parse                 ... bench:          66 ns/iter (+/- 18)
test header::common::connection::header::bench_format               ... bench:          54 ns/iter (+/- 8)
test header::common::connection::header::bench_parse                ... bench:         104 ns/iter (+/- 40)
test header::common::connection::keep_alive::bench_format           ... bench:          54 ns/iter (+/- 21)
test header::common::connection::keep_alive::bench_parse            ... bench:          80 ns/iter (+/- 30)
test header::common::content_encoding::multiple::bench_format       ... bench:         138 ns/iter (+/- 17)
test header::common::content_encoding::multiple::bench_parse        ... bench:         159 ns/iter (+/- 40)
test header::common::content_encoding::single::bench_format         ... bench:          56 ns/iter (+/- 25)
test header::common::content_encoding::single::bench_parse          ... bench:          59 ns/iter (+/- 12)
test header::common::content_length::bench::bench_format            ... bench:          60 ns/iter (+/- 21)
test header::common::content_length::bench::bench_parse             ... bench:          37 ns/iter (+/- 6)
test header::common::content_type::bench::bench_format              ... bench:         219 ns/iter (+/- 29)
test header::common::content_type::bench::bench_parse               ... bench:         269 ns/iter (+/- 35)
test header::common::cookie::bench::bench_format                    ... bench:         150 ns/iter (+/- 24)
test header::common::cookie::bench::bench_parse                     ... bench:         168 ns/iter (+/- 4)
test header::common::date::asctime::bench_format                    ... bench:         897 ns/iter (+/- 137)
test header::common::date::asctime::bench_parse                     ... bench:         288 ns/iter (+/- 197)
test header::common::date::imf_fixdate::bench_format                ... bench:         812 ns/iter (+/- 484)
test header::common::date::imf_fixdate::bench_parse                 ... bench:         214 ns/iter (+/- 67)
test header::common::date::rfc_850::bench_format                    ... bench:         815 ns/iter (+/- 144)
test header::common::date::rfc_850::bench_parse                     ... bench:         234 ns/iter (+/- 149)
test header::common::etag::bench::bench_format                      ... bench:         154 ns/iter (+/- 20)
test header::common::etag::bench::bench_parse                       ... bench:          71 ns/iter (+/- 32)
test header::common::expires::asctime::bench_format                 ... bench:         809 ns/iter (+/- 445)
test header::common::expires::asctime::bench_parse                  ... bench:         284 ns/iter (+/- 141)
test header::common::expires::imf_fixdate::bench_format             ... bench:         793 ns/iter (+/- 120)
test header::common::expires::imf_fixdate::bench_parse              ... bench:         213 ns/iter (+/- 28)
test header::common::expires::rfc_850::bench_format                 ... bench:         789 ns/iter (+/- 80)
test header::common::expires::rfc_850::bench_parse                  ... bench:         230 ns/iter (+/- 99)
test header::common::host::bench::bench_format                      ... bench:         140 ns/iter (+/- 36)
test header::common::host::bench::bench_parse                       ... bench:          64 ns/iter (+/- 13)
test header::common::if_match::multi::bench_format                  ... bench:         303 ns/iter (+/- 37)
test header::common::if_match::multi::bench_parse                   ... bench:         347 ns/iter (+/- 44)
test header::common::if_match::single::bench_format                 ... bench:         148 ns/iter (+/- 42)
test header::common::if_match::single::bench_parse                  ... bench:         108 ns/iter (+/- 34)
test header::common::if_match::star::bench_format                   ... bench:          52 ns/iter (+/- 5)
test header::common::if_match::star::bench_parse                    ... bench:           2 ns/iter (+/- 0)
test header::common::if_modified_since::asctime::bench_format       ... bench:         784 ns/iter (+/- 54)
test header::common::if_modified_since::asctime::bench_parse        ... bench:         290 ns/iter (+/- 39)
test header::common::if_modified_since::imf_fixdate::bench_format   ... bench:         790 ns/iter (+/- 127)
test header::common::if_modified_since::imf_fixdate::bench_parse    ... bench:         211 ns/iter (+/- 16)
test header::common::if_modified_since::rfc_850::bench_format       ... bench:         789 ns/iter (+/- 291)
test header::common::if_modified_since::rfc_850::bench_parse        ... bench:         227 ns/iter (+/- 43)
test header::common::if_none_match::bench::bench_format             ... bench:         154 ns/iter (+/- 20)
test header::common::if_none_match::bench::bench_parse              ... bench:         129 ns/iter (+/- 50)
test header::common::if_unmodified_since::asctime::bench_format     ... bench:         801 ns/iter (+/- 238)
test header::common::if_unmodified_since::asctime::bench_parse      ... bench:         279 ns/iter (+/- 44)
test header::common::if_unmodified_since::imf_fixdate::bench_format ... bench:         794 ns/iter (+/- 346)
test header::common::if_unmodified_since::imf_fixdate::bench_parse  ... bench:         213 ns/iter (+/- 40)
test header::common::if_unmodified_since::rfc_850::bench_format     ... bench:         791 ns/iter (+/- 287)
test header::common::if_unmodified_since::rfc_850::bench_parse      ... bench:         231 ns/iter (+/- 56)
test header::common::last_modified::asctime::bench_format           ... bench:         796 ns/iter (+/- 100)
test header::common::last_modified::asctime::bench_parse            ... bench:         280 ns/iter (+/- 35)
test header::common::last_modified::imf_fixdate::bench_format       ... bench:         800 ns/iter (+/- 273)
test header::common::last_modified::imf_fixdate::bench_parse        ... bench:         230 ns/iter (+/- 119)
test header::common::last_modified::rfc_850::bench_format           ... bench:         792 ns/iter (+/- 157)
test header::common::last_modified::rfc_850::bench_parse            ... bench:         229 ns/iter (+/- 26)
test header::common::location::bench::bench_format                  ... bench:          52 ns/iter (+/- 11)
test header::common::location::bench::bench_parse                   ... bench:          46 ns/iter (+/- 18)
test header::common::origin::bench::bench_format                    ... bench:         166 ns/iter (+/- 36)
test header::common::origin::bench::bench_parse                     ... bench:         128 ns/iter (+/- 15)
test header::common::prefer::normal::bench_format                   ... bench:         201 ns/iter (+/- 50)
test header::common::prefer::normal::bench_parse                    ... bench:         609 ns/iter (+/- 146)
test header::common::preference_applied::normal::bench_format       ... bench:         252 ns/iter (+/- 37)
test header::common::preference_applied::normal::bench_parse        ... bench:         600 ns/iter (+/- 42)
test header::common::range::bytes_multi::bench_format               ... bench:         351 ns/iter (+/- 10)
test header::common::range::bytes_multi::bench_parse                ... bench:         334 ns/iter (+/- 36)
test header::common::range::custom_unit::bench_format               ... bench:         167 ns/iter (+/- 23)
test header::common::range::custom_unit::bench_parse                ... bench:          90 ns/iter (+/- 29)
test header::common::referer::bench::bench_format                   ... bench:          52 ns/iter (+/- 10)
test header::common::referer::bench::bench_parse                    ... bench:          46 ns/iter (+/- 11)
test header::common::server::bench::bench_format                    ... bench:          52 ns/iter (+/- 2)
test header::common::server::bench::bench_parse                     ... bench:          45 ns/iter (+/- 1)
test header::common::strict_transport_security::bench::bench_format ... bench:         159 ns/iter (+/- 25)
test header::common::strict_transport_security::bench::bench_parse  ... bench:         162 ns/iter (+/- 3)
test header::common::transfer_encoding::ext::bench_format           ... bench:          53 ns/iter (+/- 3)
test header::common::transfer_encoding::ext::bench_parse            ... bench:          79 ns/iter (+/- 3)
test header::common::transfer_encoding::normal::bench_format        ... bench:         111 ns/iter (+/- 25)
test header::common::transfer_encoding::normal::bench_parse         ... bench:         150 ns/iter (+/- 32)
test header::common::upgrade::bench::bench_format                   ... bench:         242 ns/iter (+/- 59)
test header::common::upgrade::bench::bench_parse                    ... bench:         429 ns/iter (+/- 66)
test header::tests::bench_headers_fmt                               ... bench:         171 ns/iter (+/- 108)
test header::tests::bench_headers_get                               ... bench:          12 ns/iter (+/- 5)
test header::tests::bench_headers_get_miss                          ... bench:           3 ns/iter (+/- 1)
test header::tests::bench_headers_has                               ... bench:           5 ns/iter (+/- 2)
test header::tests::bench_headers_new                               ... bench:          91 ns/iter (+/- 13)
test header::tests::bench_headers_set                               ... bench:          58 ns/iter (+/- 23)
test header::tests::bench_headers_view_is                           ... bench:           1 ns/iter (+/- 0)
test http::h1::parse::tests::bench_parse_incoming                   ... bench:       3,194 ns/iter (+/- 516)

test result: ok. 0 passed; 0 failed; 257 ignored; 104 measured
```

Seems like global performances improved quite a bit! :D

src/http/h1/parse.rs
```diff
@@ -35,15 +36,24 @@ impl Http1Transaction for ServerTransaction {
                 trace!("Request.parse Complete({})", len);
                 let mut headers = Headers::with_capacity(req.headers.len());
                 let slice = buf.slice(len);
+                let path = req.path.unwrap();
+                let path_start = path.as_ptr() as usize -
+                    unsafe { MemStr::from_utf8_unchecked(slice.clone()) }.as_str()
```
seanmonstar: This doesn't need to clone the slice just to compare the pointer. It can be this:

```rust
let path_start = path.as_ptr() as usize - slice.as_ref().as_ptr() as usize;
```

src/http/h1/parse.rs
```diff
@@ -35,15 +36,24 @@ impl Http1Transaction for ServerTransaction {
...
                 });
+                // path was found to be utf8 by httparse
+                let path = unsafe { MemStr::from_utf8_unchecked(path) };
+
                 Some((MessageHead {
                     version: if req.version.unwrap() == 1 { Http11 } else { Http10 },
                     subject: RequestLine(
                         try!(req.method.unwrap().parse()),
-                        try!(req.path.unwrap().parse())
+                        try!(::uri::Uri::from_str(path.as_str())),
```
seanmonstar: This goes through `Uri::from_str` -> `InternalUri::from(&str)` -> `str.to_owned()`, which means we're still copying the bytes here. We could instead introduce a new function in the `uri` module like so:

```rust
// in uri.rs
pub fn from_mem_str(s: MemStr) -> Result<Uri, Error> {
    Uri::new(InternalUri::from(s))
}
```

And then we can use `::uri::from_mem_str(path)`. Since the `uri` module isn't public, and only the `Uri` type is re-exported outside of the crate, we can keep the implementation details private.

src/uri.rs
```diff
@@ -1,11 +1,54 @@
...
+use std::str::{self, FromStr};
+use http::buf::{MemSlice, MemStr};
 use Url;
 use url::ParseError as UrlError;
 
 use Error;
 
+#[derive(PartialEq, Eq, Clone)]
+enum InternalUri {
+    Static(Cow<'static, str>),
```
seanmonstar: I'd rename this variant to `Cow`, since it is misleading when you have `InternalUri::Static(Cow::Owned("/foo".to_string()))`, because it looks like a static string, but it is owned.

src/http/h1/parse.rs
GuillaumeGomez: Outch... Good catch!

GuillaumeGomez:
Updated.

GuillaumeGomez:
Updated.

PR #1044, opened 2017-02-02
Link header implementation.

tabac:
@seanmonstar 

Hi, this is in progress of course but I would like to know if you're OK with the direction it's taking.

The main question is if the `enum` approach to storing Link parameters is too much or not.

Also, I am not parsing `title*` and `link-extension` at the moment, it would be nice if I could use [regex](https://github.com/rust-lang/regex), if we are to do it, but it's not in `Cargo.toml`.


src/header/common/link.rs
```diff
@@ -0,0 +1,622 @@
...
+}
+
+impl Header for Link {
+    fn header_name() -> &'static str {
+        static NAME: &'static str = "Link";
+        NAME
+    }
+
+    fn parse_header(raw: &Raw) -> ::Result<Link> {
+       from_one_raw_str(raw)
```
seanmonstar: Is there only allowed to be 1 `Link` header? If there a multiple, does the spec say anything about that? This function will parse if there is 1 link, and reject if there are more than 1, which is fine for some headers, just asking if that is the case with `Link`.

src/header/common/link.rs
```diff
@@ -0,0 +1,622 @@
...
+///
+/// # Examples
+///
+/// ...
+///
+///
+#[derive(Clone, PartialEq, Debug)]
+pub struct Link {
+    /// Target IRI: `link-value`
+    pub link: Uri,
```
seanmonstar: I've started to wonder if exposing all the fields publicly on headers was a good idea. I recently changed the `Host` header to no longer have public fields, but to have method accessors instead. That'd allow changing the exact types of `Host.hostname`, for instance, without breaking people.

Does that seem like a good idea here?

src/header/common/link.rs
```diff
@@ -0,0 +1,622 @@
...
+                        media_type: None,
+                        link_extension: None,
+                    }
+                },
+            },
+        };
+
+        // Parse the Link header's parameters
+        for link_param in link_split {
+            let mut link_param_split = link_param.split('=');
```
seanmonstar: Probably should use `.splitn(2, '=')`, for weird cases of `rel="foo=bar"`.

src/header/common/link.rs
```diff
@@ -0,0 +1,622 @@
...
+
+            let link_param_name = match link_param_split.next() {
+                None => return Err(::Error::Header),
+                Some(p) => p.trim(),
+            };
+
+            match link_param_name.to_lowercase().as_ref() {
```
seanmonstar: If we wanted to prevent this allocation+copy of the string, `UniCase(link_param_name)` could be used. Doesn't work with `match`, though.

src/header/common/link.rs
```diff
@@ -0,0 +1,622 @@
...
+                    return Err(::Error::Header);
+                }
+            }
+        }
+
+        Ok(link_header)
+    }
+}
+
+fn verify_and_trim(s: &str, start: char, end: char) -> ::Result<&str> {
```
seanmonstar: It looks like `verify_and_trim` only ever is used with ASCII characters, so it can be done more efficiently with bytes. Changing `start` and `end` to `u8`, and calls to `verify_and_trim(s, b'"', b'"')`.

src/header/common/link.rs
tabac: If we make fields private we would have to provide setters and getters. I am planning to implement a builder pattern like constructor so we could skip the setters. I think it's fine to make things private.

About making some of the fields generic, it makes sense but which ones? We can do it for `title*` and `link-extension` and that way leave the parsing of those two fields to the user. Not sure if it makes sense for other fields maybe `link`, `anchor` but I like `Uri` as it is.

src/header/common/link.rs
tabac: Oh, I haven't realized that for `from_one_raw_str`, I will look into it.

src/header/common/link.rs
tabac: I have tried `UniCase` before doing the above. Maybe there's a better way I will check.

src/header/common/link.rs
seanmonstar: A builder may be a good idea!

I didn't mean to make those 2 fields generic, but rather, I was wondering if they are served well enough as `String`, and if someone wanted to, they can make a whole new `struct MyLink { ... }` header.

tabac: OK, I will leave those two as `String`s then.

tabac:
@seanmonstar I have to work on the way we `.split()`. The `Link` header's delimiters are `;` and `,` but these characters can appear inside quoted text. So we have to ignore them in that case.

Do you have anything in mind that does something similar? I did a bit of search and unfortunately haven't found something. I was thinking of an iterator over ASCII characters yielding string slices.

seanmonstar:
Yea, I suppose state would need to be recorded, to know that inside `"some text"`, a delimiter means nothing.

src/header/common/link.rs
```diff
@@ -0,0 +1,975 @@
...
+
+    /// Update LinkValue's `rev` parameter
+    pub fn push_rev(mut self, rev: RelationType) -> LinkValue {
+        let mut v = self.rev.take().unwrap_or(Vec::new());
+
+        v.push(rev);
+
+        self.rev = Some(v);
+
+        self
```
tabac: Is there a more idiomatic way to do this?

src/header/common/link.rs
```diff
@@ -0,0 +1,975 @@
...
+                    None => return Err(::Error::Header),
+                    Some(p) => p.trim(),
+                };
+
+                let mut link_header = match link.values.last_mut() {
+                    None => return Err(::Error::Header),
+                    Some(l) => l,
+                };
+
+                if "rel".eq_ignore_ascii_case(link_param_name) {
```
tabac: When we want case-insensitive comparison I am using this function since we're dealing only with ASCII.

src/header/common/link.rs
```diff
@@ -0,0 +1,975 @@
...
+        match T::from_str(i.trim()) {
+            Err(_) => return Err(::Error::Header),
+            Ok(t) => v.push(t),
+        }
+    }
+
+    Ok(v)
+}
+
+fn fmt_delimited<T: fmt::Display>(f: &mut fmt::Formatter, p: &[T], d: &str, s: &str, e: &str) -> fmt::Result {
```
tabac: This function as well as `from_str_delimited` kind of overlap with the ones you have in `header/parsing.rs`. We could move them but I think it's fine like this.

src/header/common/link.rs
```diff
@@ -0,0 +1,975 @@
...
+    fn header_name() -> &'static str {
+        static NAME: &'static str = "Link";
+        NAME
+    }
+
+    fn parse_header(raw: &Raw) -> ::Result<Link> {
+        // TODO: This should probably change to support multiple link
+        //       headers in one request although we can have one link
+        //       header with multiple values.
+        from_one_raw_str(raw)
```
tabac: I don't know what to do here. I think we should support more than one Link headers per request but I have the impression that `parse_header()` should return a `Result` with one header object not a `Result<Vec<Link>>`.

src/header/common/link.rs
```diff
@@ -0,0 +1,975 @@
...
+    /// Create Link from a `Vec<LinkValue>`
+    pub fn new(link_values: Vec<LinkValue>) -> Link {
+        Link { values: link_values }
+    }
+}
+
+#[allow(dead_code)]
+impl LinkValue {
+    /// Create LinkValue from URI-Reference
+    pub fn new(uri: &str) -> ::Result<LinkValue> {
```
tabac: Returning a `Result` from the constructor doesn't look very good. I did it like that because `uri::Uri` is not public. We could change that if you want to.

tabac:
@seanmonstar Please take a look and let me know what you think. I intend to do some more testing and add some tests. I also left some comments/questions.

tabac:
@seanmonstar The tests fail because of the change introduced in this [commit](https://github.com/hyperium/hyper/commit/70a3f76a276a429980a8fc877cf4de175364d708).

The change doesn't allow relative URIs containing something like `../` or `./` which I think that are allowed based on: [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2), [RFC1808](https://tools.ietf.org/html/rfc1808#section-2).

Does `uri::Uri` refere to absolute URIs only? If so I could just change the unit test.

seanmonstar:
Oh interesting point. Yes, at the moment, `uri::Uri` was meant specifically for the `request-target` of HTTP requests. In [RFC7230](https://tools.ietf.org/html/rfc7230#section-5.3), it seems that relative URLs are not allowed.

- One possibility is that we decide that `Uri` can be used for any URI case, in which case it could also be used in things like the `Location` header. Then the check that the `Uri` is absolute would need to be removed from the normal parsing, and only checked when parsing the `request-target`.

- The other is to leave the `Uri` the way it is, and have `Link` represent the URI as a `String`, similar to how the `Location` header current does.

I kind of like the idea of adding types to those headers, but only if supporting them in the `Uri` doesn't make parsing more complex and thus slow down parsing of the `request-target`. I also cannot remember currently if the URIs in these headers can contain additional formats that the `request-target` does not allow (besides a relative URL).

tabac:
I went with `uri::Uri` because for the headers `Content-Location`, `Location`, `Referer` we are using `String` but you have a comment to `use URL`. All three as well as `Link` should support relative URIs.

From what I understand from [here](https://tools.ietf.org/html/rfc3986#section-4.2) a relative URI can be something like: `//something`, `/something`, `./something`, `../something`, `somethine_else`, even `./some:thing`.

Making `uri::Uri` more generic touches parts of the library which I am not fully aware of and I feel it's more like a different task. I will go with `String` if you're OK with that. We could open another issue for supporting relative URIs in `uri::Uri`.

seanmonstar:
I think that sounds like a good plan, start with String, and we can open a new bug to see about using `Uri` in the each of these headers.

src/header/common/link.rs
seanmonstar: I found in the RFC https://tools.ietf.org/html/rfc5988#section-5.5 that the `Link` header can contain multiple links in a single header, by being comma-separated, and so the normal header rules apply which state that multiple headers of the same name can be joined together.

As an example, taken from a few pieces in the RFC:

```
 Link: </TheBook/chapter2>; rel="previous", </TheBook/chapter4>; rel="next"
 Link: </TheBook/chapter3>; rel="canoncical"
```


So it seems this would need be able to accept several `Link`s, both if a line has them comma-separated, and if there are multiple lines.

tabac:
@seanmonstar I want to add more tests and try it out a bit. Other than that I think you could review if you want. (Not sure what's going on with coveralls).

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+impl Link {
+    /// Create `Link` from a `Vec<LinkValue>`.
+    pub fn new(link_values: Vec<LinkValue>) -> Link {
+        Link { values: link_values }
+    }
+}
+
+#[allow(dead_code)]
```
seanmonstar: We should be able to remove this `allow`.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+        }
+    }
+
+    /// Get the `LinkValue`'s value.
+    pub fn link(&self) -> &str {
+        self.link.as_ref()
+    }
+
+    /// Get the `LinkValue`'s `rel` parameter(s).
+    pub fn rel(&self) -> Option<&Vec<RelationType>> {
```
seanmonstar: I wonder, would it be nicer to return `Option<&[RelationType]>`? Is there value in getting the `&Vec`? Probably not, but maybe. Same question for each of the getters returning `&Vec`.

If we wanted to do that, it'd be a change to `opt.as_ref().map(AsRef::as_ref)`.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+        self.href_lang.as_ref()
+    }
+
+    /// Get the `LinkValue`'s `media` parameter(s).
+    pub fn media_desc(&self) -> Option<&Vec<MediaDesc>> {
+        self.media_desc.as_ref()
+    }
+
+    /// Get the `LinkValue`'s `title` parameter.
+    pub fn title(&self) -> Option<&String> {
```
seanmonstar: Like the comment above, maybe we want to return `Option<&str>` instead. A plus to doing that is that it doesn't leak the internal type, so if we later changed `title` to hold a `Cow<'static, str>` or something, the API wouldn't need to change.

src/header/common/link.rs
seanmonstar: No, `Option` just received a method to do this, but it's only in nightly, so this must do.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+
+        v.push(rel);
+
+        self.rel = Some(v);
+
+        self
+    }
+
+    /// Set `LinkValue`'s `anchor` parameter.
+    pub fn set_anchor(mut self, anchor: &str) -> LinkValue {
```
seanmonstar: Perhaps `T: Into<String>` here, so people who allocated a `String` anyways can just give it to us, instead of us having to copy it again.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+
+    /// Set `LinkValue`'s `title*` parameter.
+    pub fn set_title_star(mut self, title_star: &str) -> LinkValue {
+        self.title_star = Some(String::from(title_star));
+
+        self
+    }
+
+    /// Set `LinkValue`'s `type` parameter.
+    pub fn set_media_type(mut self, media_type: &Mime) -> LinkValue {
```
seanmonstar: We should take the `Mime` by value, and allow the user to clone one or use the one they already have.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+        NAME
+    }
+
+    fn parse_header(raw: &Raw) -> ::Result<Link> {
+        // If more that one `Link` headers are present in a request's
+        // headers they are combined in a single `Link` header containing
+        // all the `link-value`s present in each of those `Link` headers.
+        raw.iter()
+            .map(parsing::from_raw_str::<Link>)
+            .fold(None, |p, c| {
```
seanmonstar: Awesome!

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+                        Some(Ok(p))
+                    },
+                    _ => Some(Err(::Error::Header)),
+                }
+            })
+            .unwrap_or(Err(::Error::Header))
+    }
+
+    fn fmt_header(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt_delimited(f, self.values.as_slice(), ", ", "", "")
```
seanmonstar: Visually parsing this line is difficult. I finally got what it's doing after several reads.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+            Ok(RelationType::ExtRelType(String::from(s)))
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Utilities
+////////////////////////////////////////////////////////////////////////////////
+
+struct SplitAsciiUnquoted<'a> {
```
seanmonstar: Fantastic, good work separating the logic out to make it easier to follow exactly how parsing works.

src/header/common/link.rs
```diff
@@ -0,0 +1,1027 @@
...
+///
+/// let mut headers = Headers::new();
+/// headers.set(
+///     Link::new(vec![link_value])
+/// );
+/// ```
+#[derive(Clone, PartialEq, Debug)]
+pub struct Link {
+    /// A list of the `link-value`s of the Link entity-header.
+    pub values: Vec<LinkValue>
```
seanmonstar: Should we keep the `values` field private?

src/header/common/link.rs
tabac: I agree, it doesn't look very good. I changed it to receive a tuple, I think it's a bit better. I thought about creating a tuple struct and passing it but it won't be so efficient I believe.

src/header/common/link.rs
tabac: I can't think of a use case needing a `&Vec`, I changed everything to return `Option<&[T]>`.

tabac:
I am kind of lost with coveralls. It's reporting 0 coverage for `link.rs` and for almost everything else. Am I doing something wrong?

seanmonstar:
I believe the integration with coveralls is broken at the moment, it recently just went to 0% on master too. I don't yet know why, perhaps a change that's preventing kcov from collecting the data.

seanmonstar:
Do you feel this is done, feature wise? If so, I'll look it over once more later with the goal of merging, unless there was something else you wanted to add or adjust.

tabac:
@seanmonstar feature wise, I think it's OK, I don't have something to add at the moment.

I am currently working on an example HTTPS client in order to test it further but any possible fixes could go to a different PR.

src/header/common/link.rs
```diff
@@ -0,0 +1,1141 @@
...
+        let link = Link::new(vec![link_value]);
+
+        let mut link_header = String::new();
+        write!(&mut link_header, "{}", link).unwrap();
+
+        let expected_link_header = "<http://example.com/TheBook/chapter2>; \
+            rel=\"previous\"; anchor=\"/anchor/example/\"; \
+            rev=\"next\"; hreflang=de; media=\"screen\"; \
+            title=\"previous chapter\"; title*=title* unparsed; \
+            type=\"text/plain\"; link-extension=link-extension unparsed";
```
davidcornu: This doesn't seem right. If I'm reading the ABNF from https://tools.ietf.org/html/rfc5988#section-5 correctly

> ```
>   Link           = "Link" ":" #link-value
>   link-value     = "<" URI-Reference ">" *( ";" link-param )
>   link-param     = ( ( "rel" "=" relation-types )
>                  | ( "anchor" "=" <"> URI-Reference <"> )
>                  | ( "rev" "=" relation-types )
>                  | ( "hreflang" "=" Language-Tag )
>                  | ( "media" "=" ( MediaDesc | ( <"> MediaDesc <"> ) ) )
>                  | ( "title" "=" quoted-string )
>                  | ( "title*" "=" ext-value )
>                  | ( "type" "=" ( media-type | quoted-mt ) )
>                  | ( link-extension ) )
>   link-extension = ( parmname [ "=" ( ptoken | quoted-string ) ] )
>                  | ( ext-name-star "=" ext-value )
> ```

`link-extension` isn't a literal but either a single attribute as a `parmname` (defined in https://tools.ietf.org/html/rfc5987#section-3.2.1) or a key value pair separated by `=` where the key is a `parmname` and the value is either a `ptoken` or a `quoted-string` (defined in https://tools.ietf.org/html/rfc2616#section-2.2).

With that in mind, my understanding is that headers like

```
Link: <https://example.com>; rel="previous"; foo="bar"; baz="qux" 
```

are valid. `foo` and `bar` are both `link-extension`s.

davidcornu: If that is indeed correct, `LinkValue`'s `link_extension` field should probably be named `link_extensions` and return a `Option<Vec<LinkExtension>>` (or just a `Vec<LinkExtension>`) where `LinkExtension` is something like

```rust
struct LinkExtension {
  name: String,
  value: Option<String>
}
```

seanmonstar: I believe you are correct. Would you be willing to try to fix it?

davidcornu:
@seanmonstar started https://github.com/hyperium/hyper/compare/master...davidcornu:link-extensions. Will open a PR when I've updated the tests to get some feedback if that's OK.

seanmonstar:
@davidcornu sounds good, open whenever ready. I've filed #1085 to track the bug.

Issue #1036, opened 2017-01-26
Provide `From<Message> for Request`


alexcrichton:
I've taken most of the ideas learned here about spawning servers back to tokio-proto in https://github.com/tokio-rs/tokio-proto/pull/135 and I believe the only remaining blocker for hyper being compatible with tokio-proto's `TcpServer` is the usage of `SocketAddr` in Hyper's request type. If this can be handled to implement `From<Message> for Request` then Hyper could leverage support in https://github.com/tokio-rs/tokio-proto/pull/130 to use `TcpServer` directly, especially when working with TLS connections where Hyper's conveniences won't be available.

Talking with @carllerche over lunch today one idea was to have the socket address be a member of the messages coming from the `Stream` going into tokio-proto. That way the message itself would contain the socket address. I could imagine this being implemented via a trait bound in Hyper perhaps? 

What do you think?

seanmonstar:
That definitely sounds good. This is basically for this line: https://github.com/hyperium/hyper/blob/master/src/server/mod.rs#L258

So, `impl From<Message<ProtoRequest, http::TokioBody> for Request`? I could move the `SocketAddr` into the `ProtoRequest` type, but I don't know how someone would make use of this, since they cannot create those types...

alexcrichton:
Yeah that's the impl I was thinking of. I don't think users would use this explicitly, but that way Hyper "as is" would hook into tokio-proto's `TcpServer` so that could be used off the bat.

PR #1045, opened 2017-02-03
Extend Server to use anything that produces a stream of (Io, SocketAddr)

mattwoodyard:
I've done some refactoring on Server to make it easier to integrate with and ssl listener.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -42,18 +42,18 @@ pub struct Http {
 }
 
 
-type ConnectionStream = Box<Stream<Item=(TcpStream, SocketAddr), Error=io::Error>>;
+type ConnectionStream<I: Io> = Box<Stream<Item=(Box<I>, SocketAddr), Error=io::Error>>;
 
 /// An instance of a server created through `Http::bind`.
 ///
 /// This server is intended as a convenience for creating a TCP listener on an
 /// address and then serving TCP connections accepted with the service provided.
-pub struct Server<S> {
+pub struct Server<S, I: Io + 'static> {
     protocol: Http,
     new_service: S,
     core: Core,
     local_address: SocketAddr,
-    incoming: ConnectionStream,
+    incoming: ConnectionStream<I>,
     shutdown_timeout: Duration,
 }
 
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 6cdf656077073b437db4bd1c9754d40f30a2d85c

 * Commit: 80ae945fcdf36e50c520a0cfee727ebd6c75aa0b

---

seanmonstar:
Being generic over a stream of `Io`s was actually how the server used to work, and was changed to the current situation in https://github.com/hyperium/hyper/pull/1013!

The idea is that `Http` implements `ServerProto`, and there is also the exposed `bind_connection` method, so really, any stream of connections can already be used.

Making use of TLS for the Server just involves using `tokio_tls::proto::Server::new(http, listener)`, from here: https://github.com/tokio-rs/tokio-tls/blob/master/src/proto.rs

@alexcrichton may have more thoughts on the matter.

mattwoodyard:
I tried getting that working first (https://play.rust-lang.org/?gist=d8d0d1ed14574f4085b1d663c6354ef8&version=stable&backtrace=0), but it seemed to be missing some implementation (I think: https://github.com/hyperium/hyper/issues/1036). I was going to add that, but then I noticed that tokio-proto didn't have graceful shutdown, which is something I wanted, so I did this.  

Is there something I've missed?

alexcrichton:
@mattwoodyard sounds like you went down the same rabbit hole I did! @seanmonstar is spot on in that the server Hyper bundles isn't currently intended to be used with TLS (or other protocols for that matter). The idea is that protocols are built externally and then there's generic support for taking a protocol and spinning up a server with it. 

Note, though, that support isn't quite here yet for all that. PRs like https://github.com/tokio-rs/tokio-proto/pull/135 and issues like https://github.com/hyperium/hyper/issues/1036 are key ingredients to getting that working. With https://github.com/hyperium/hyper/issues/1036 Hyper's request/response should work "as-is" with tokio-proto's `TcpServer` and with https://github.com/tokio-rs/tokio-proto/pull/135 then tokio-proto's `TcpServer` should reach feature parity with Hyper's current server in this repository.

@seanmonstar I'll leave it up to you whether you'd like to merge this for now. The intention at least is that this shouldn't be necessary (and in fact, the entire server shouldn't be necessary!), so you can at least rest easy that in the future you won't have to maintain this :)

seanmonstar:
Oh right, what I had suggested doesn't actually work yet without #1036. With #1036 fixed, would that solve your use case @mattwoodyard?

mattwoodyard:
For my immediate task, I'm just trying to get TLS to work. I'm not particularly attached to the approach in this PR. If #1036 can make the tokio_tls stuff work, thats fine. Longterm I'd like graceful shutdown with my TLS. If the plan is to generalize the hyper approach into tokio-proto, I can take a swing at that.

alexcrichton:
Yeah I believe that's the general long term plan at least, to have Hyper work well with the tokio-proto server builders to leverage all of the logic in there (which will expand over time)

mattwoodyard:
Closing in favor of #1047

PR #1048, opened 2017-02-04
chore(cargo): add travis-ci badge on crates.io

spk:


seanmonstar:
Thanks for the PR,

I've actually purposefully not put these in for two reasons. I don't think there is value in seeing these badges on the release pages, since clearly the versions would not have been published if they didn't build. And also, this shows the status of the master branch, which is not the code people would be downloading.

spk:
No problems thanks for the explanations

PR #1052, opened 2017-02-08
refactor(examples): remove use of depricated futures features

ericchiang:
The "Finished" struct and "finished" method have been deprecated[0]
and don't show up in the futures docs. Prefer non-deprecated
features for users exploring the examples.

[0] https://github.com/alexcrichton/futures-rs/blob/0.1.10/src/future/mod.rs#L25


seanmonstar:
Great, thanks!

PR #1057, opened 2017-02-10
fix(lib): export Origin enum

michael-zapata:
Since the `Url` struct implements an `origin()` method returning an `Origin`, provide the means to use it

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,7 @@ extern crate test;
 
 
 pub use uri::Uri;
-pub use url::Url;
+pub use url::{Url, Origin};
 pub use client::Client;
 pub use error::{Result, Error};
 pub use header::Headers;


seanmonstar:
Hm, I worry about exporting `hyper::Origin` and `hyper::header::Origin` as different types.

I've so far felt that if someone wanted to really use all the extras of the url crate, to depend on the url crate themselves. I also have been starting to wonder if there is value in requiring a `Url` over just using the `Uri` defined in hyper proper.

michael-zapata:
That explains a lot. While skimming through the documentation (namely [this one](https://hyper.rs/hyper/master/hyper/struct.Url.html)), there are a few non-exported types that are still exposed somehow.

Keeping the logic contained internally can only help. There's the option of keeping everything internal, or perhaps exposing it through a feature (though that seems a bit convoluted).

michael-zapata:
Closing, this was meant as a quick-fix, but doesn't solve the underlying issue (exposing `Url`).

PR #1058, opened 2017-02-10
docs(uri): fix rendering errors

michael-zapata:
Add a code block around the URI explanation to make it legible and add a missing backquote


seanmonstar:
Thanks!

PR #1068, opened 2017-02-17
feat(server): add remote_addr back to deconstruct

MJDSys:
The remote address was removed back when the port to async first happend.
It was readded to the Request struct later, but was never readded to the
deconstruct function.  Since there seems to be no reason to keep it
removed and Iron relies on it, lets add it back.

### DIFF
--- a/src/server/request.rs
+++ b/src/server/request.rs
@@ -65,8 +65,8 @@ impl Request {
     ///
     /// Modifying these pieces will have no effect on how hyper behaves.
     #[inline]
-    pub fn deconstruct(self) -> (Method, Uri, HttpVersion, Headers, Body) {
-        (self.method, self.uri, self.version, self.headers, self.body)
+    pub fn deconstruct(self) -> (SocketAddr, Method, Uri, HttpVersion, Headers, Body) {
+        (self.remote_addr, self.method, self.uri, self.version, self.headers, self.body)
     }
 }
 


seanmonstar:
The field was actually just changed to an `Option<SocketAddr>`, because of a few reasons.

Additionally, the `deconstruct` method was more designed to allow taking the main expensive pieces without cloning. The address can be easily copied before, and so we can keep less noise in the deconstruct method.

MJDSys:
I saw the PR for the Option<_> change after I'd already submitted this PR :)

If you prefer deconstruct doesn't have SocketAddr, should the method and version be removed too?  Those seem like other easily clonable fields.  I can adjust the PR to remove those as well.  Otherwise, I can update this PR to return the Option.

seanmonstar:
The `Method` could potentially be a `String` and need cloning, if it's an odd method. The version is definitely just a cheap copy, and in the end probably not important to anyone. The current signature was just thinking "what makes up the request?"

```
POST /foo HTTP/1.1
Content-Length: 11

Hello World
```

Oh, that's `(Method, Uri, Version, Headers, Body)`! I don't know if that's the best way to to do it, though :)

MJDSys:
That seems like the most sensible way to do it.  I was only adding back remote_addr, as there didn't seem to be a specific reason why it was removed at the time.

For now I'll just close this PR.  It's easy to adjust Iron to this new reality, and this can always be revisited if it turns into a problem.

Thanks for your work in all this!

PR #1070, opened 2017-02-20
feat(http): Set default type of Http to Body

GuillaumeGomez:


seanmonstar:
I was incorrect in saying it should have been `Body`, since `Http<B>` is where `B: AsRef<[u8]>`, so it should have been `Chunk`. I've done so in master.

PR #1072, opened 2017-02-21
feat(headers): impl Eq for ContentType

### LINKED ISSUE #1071: ContentType should impl Eq
when trying to compare header::ContentType values I get 

```
the trait `std::cmp::PartialEq<hyper::header::ContentType>` is not implemented for `&hyper::header::ContentType`
```

This seems like something most headers should impl, when writing tests its very convenient to do things like assert_eq!() and such.

Keats:
#1071

### DIFF
--- a/src/header/common/content_type.rs
+++ b/src/header/common/content_type.rs
@@ -98,4 +98,6 @@ impl ContentType {
     }
 }
 
+impl Eq for ContentType {}
+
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks! I think the issue in #1071 was instead that the user was comparing `&ContentType == ContentType`... Still, this code is correct, so merged!

PR #1076, opened 2017-02-25
fix(url): fix panic when questionmark is in fragment

sinkuu:

`Url::parse` panicked if a questionmark is in the fragment part:

```
panicked at 'attempt to subtract with overflow', src/uri.rs:207
```

This was found by [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz)


seanmonstar:
Excellent, thank you!

PR #1079, opened 2017-03-01
fix(http): Consume entire chunked encoding message

sfackler:
The trailing newline otherwise breaks keepalive after a chunked request.

### DIFF
--- a/src/http/h1.rs
+++ b/src/http/h1.rs
@@ -591,12 +591,17 @@ impl<R: Read> Read for HttpReader<R> {
                 trace!("Chunked read, remaining={:?}", rem);
 
                 if rem == 0 {
+                    if opt_remaining.is_none() {
+                        try!(eat(body, LINE_ENDING.as_bytes()));
+                    }
+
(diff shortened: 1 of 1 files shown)


sfackler:
I think this is also a problem on the tokio version.

seanmonstar:
I haven't built a test yet for master, but the chunked decoder got a pretty extensive rewrite to handle reads blocking mid chunk-size, so it may have been dealt with after all.

Either way, thanks for this!

sfackler:
Thanks! Could you cut an 0.10 release with this when you have a chance?

seanmonstar:
Published in 0.10.5.

jwilm:
re: tokio

I contributed a [similar fix](https://github.com/hyperium/hyper/commit/bffde8c841353e05e9aea267ca94848ccdeeb394) for this on master (rotor) at some point before tokio was released. Not sure if it was before or after the decoder rewrite or how many of those fixes actually made it into the tokio tree.

Issue #1075, opened 2017-02-24
Http::new(core: &'a Core)


scottlamb:
Currently `hyper::server::Server` owns its own `tokio::reactor::Core`. I propose it borrow one instead.

This would allow better sharing the reactor. It'd mean that a `Server` could share the `Core` with things that it can't now, such as another hyper server (http + https). Also, it'd mean that the reactor could be available to be used prior to bringing up the hyper server.

For example, I'd like to use it with tokio_signal, which (copying from [the example](https://github.com/alexcrichton/tokio-signal/blob/master/examples/log.rs)) is intended to be used like this:

```rust
let mut core = Core::new().unwrap();
let ctrlc = tokio_signal::ctrl_c(&core.handle());
let stream = core.run(ctrlc).unwrap();
```

I'd like to install the signal handler before being able to call `server.handle()` (this would allow me to do things such as flushing async logs prior to dying if a signal is received earlier in the startup process), and I'd like to advance the core myself to turn the future into a stream before starting to serve requests via `server.run_until` (just seems cleaner in terms of error handling).

Another possible benefit is that it'd make it easier to pass a `Remote` to the service. The `new_service` argument to `Server::bind` now has to be provided before hyper creates the `Core`. So to pass it the `Remote`, you have to stuff it later into a `std::sync::Mutex` that you passed into the lambda or in a `lazy_static`. That's inconvenient. I think with this way you could simply do something like this:

```rust
let core = Core::new();
let remote = core.remote();
let server = hyper::server::Http::new(&core)
    .bind(&addr, move || Ok(MyService(remote.clone())))
    .unwrap();
```

(Although if services are always intended to be created on a reactor thread, it might be even better for tokio's `NewService::new_service` to take a `Handle`, to avoid having to deal with `Remote::handle()` returning `None`. That change would make this benefit moot.)

It looks like I can use hyper with a borrowed `Core` today by avoiding `Http::bind` in favor of `Http::bind_connection`. But in doing so, I have to roll my own logic for not only accepting connections but also graceful shutdown, duplicating a lot of `hyper/src/server/mod.rs`. I'd rather not do that.

seanmonstar:
The expect evolution is that you could use `TcpServer` in tokio-proto, and hyper can remove some of it's TCP-specific handling. Half of the work for `TcpServer` has been merged, the part adding graceful shutdown is still a PR: https://github.com/tokio-rs/tokio-proto/pull/135

[...]

PR #1091, opened 2017-03-14
feat(examples): add an example server with a custom event loop

mwanner:
Not sure if this example is really worth it. Follows the approach outlined in #1075. 

- [X ] The commit messages match the guidelines in https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md#git-commit-guidelines


izolyomi:
Being new to Rust and Hyper, this example just have saved me from a lot more of investigation. It was just what I needed, so it's absolutely useful, something like this should be between the official examples.

seanmonstar:
Sorry for not noticing a while ago!

I'm trying to have any and all examples be in the form of [guides](https://hyper.rs/guides), as opposed to in the source repo. So I'm going to close this, but the guides definitely could use an "advanced" server guide showing how to use your own `Core`.

PR #1093, opened 2017-03-15
Migrate to leveraging the tokio-io crate

alexcrichton:
This commit updates to the most recent versions (released today) of the various
Tokio libraries in use. Namely the `tokio_core::io` module has now been
deprecated in favor of an external `tokio-io` crate. This commit pulls in that
crate and uses the `AsyncRead + AsyncWrite` abstraction instead of `Io` from
tokio-core.


GitCop:

 * Commit: 0dcfd18e51b432f2b20a6345638577fc4256922c

---

alexcrichton:
Ah and I should also mention that I dropped the `relay` crate while I was at it in favor of `futures::unsync`, but I can back out that change if you'd prefer.

src/http/conn.rs
```diff
@@ -155,7 +155,8 @@ where I: Io,
     }
 
     fn maybe_park_read(&mut self) {
-        if self.io.poll_read().is_ready() {
+        // TODO: this needs fixing
```
seanmonstar: This was part of trying to keep from busy looping, how broken is the "this needs fixing" part :D

alexcrichton: d'oh right, sorry I thought I had handled this elsewhere already.

I was wondering if you know how the best way might be to handle this off-hand?

seanmonstar: Hrm, so as the comment below tries to describe, this situation would occur when hyper has read the end of a body, such as described by a `Content-Length` header, and then stops, as it's waiting on the response side to determine if it should read the next request.

If the socket were to already have some of the bytes in its buffer, then our not reading to end would mean `TcpStream` would not unpark the task again to read more. That's a scenario when hyper needs to park, and then unpark later when the response finishes and keeps the connection alive.

So, to get the same as poll_read from now on, does that mean we need to store our own `read_has_blocked` flag?

alexcrichton: Yeah I believe so, I'll look into adding such a flag.

GitCop:

 * Commit: 41fc8dff3360a955693247eee416ba0c5186f5ee

---

alexcrichton:
Ok I believe that should handled the blocked read case?

seanmonstar:
Great, thanks!

PR #1100, opened 2017-03-21
feat(lib): remove extern Url type usage

### LINKED ISSUE #1089: Remove Url depedency
Now that `hyper::Uri` has sufficient parsing capability, and can represent any URI for the server or client, does it make sense to still also use the `Url` type? When looking at the rustdocs generated, it certainly looks confusing that there is both `Url` and `Uri` at the crate root. And issues like #1087 and others sometimes appear because people assume hyper controls the `Url` type.

Some options I'm considering:

- Remove any direct usage of `Url`, but implement `From<Url> for Uri`, so people can easily do `Request::new(Get, url.into())`.
- Remove the `url` crate dependency entirely. This would probably be the ideal option, so that the crate is only included in projects that explicitly ask for it. However, it does make it slightly more annoying to create client requests.

/cc @frewsxcv @nox (@alexcrichton for sccache experience)

seanmonstar:
Closes #1089 

BREAKING CHANGE: 

The `Url` type is no longer used. Any instance in the
  `Client` API has had it replaced with `hyper::Uri`.

  This also means `Error::Uri` has changed types to
  `hyper::error::UriError`.

  The type `hyper::header::parsing::HTTP_VALUE` has been made private,
  as an implementation detail. The function `http_percent_encoding`
  should be used instead.

### DIFF
--- a/benches/end_to_end.rs
+++ b/benches/end_to_end.rs
@@ -28,7 +28,7 @@ fn get_one_at_a_time(b: &mut test::Bencher) {
 
     let client = hyper::Client::new(&handle);
 
-    let url: hyper::Url = format!("http://{}/get", addr).parse().unwrap();
+    let url: hyper::Uri = format!("http://{}/get", addr).parse().unwrap();
 
     b.bytes = 160 * 2 + PHRASE.len() as u64;
     b.iter(move || {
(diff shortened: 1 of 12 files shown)


lucab:
Just as a usecase/question: I used to have a base url stored as a `hyper::Url` and then appending path components to it via `.join()`. This PR seems to have removed this kind of usage, as there are no more methods to mutate an Uri in place once built. Am I right that this is on purpose and the recommended usage is as described by @alexcrichton in https://github.com/hyperium/hyper/issues/1089#issuecomment-286252505?

seanmonstar:
Yes, this is on purpose for now. `Uri` could possibly gain mutators in the future.

hyper is going through what most libraries go through as they mature: removal of extras, providing a strong core, and allowing extras to be built on top with whatever opinions the builder has.

PR #1103, opened 2017-03-26
feat(client): add Client::handle

d12i:
This is useful for external structs which own (or have a reference to) a `Client` but have no access to the event loop the client is associated with.


seanmonstar:
Thanks, good idea!

PR #1106, opened 2017-03-29
feature(version): impl FromStr for HttpVersion

liamchristopher:
Hello. 
I'm using hyper to take apart some log requests and construct new requests from them. This is a simple change that matches a String back to the enum. The default case uses same unreachable!() macro that display uses.

Not sure if this has been addressed before, but since I needed it to go forward, figured I'd put a PR together for it.

Thanks,
liam

- [ x] The commit messages match the guidelines in https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md#git-commit-guidelines

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -61,6 +61,7 @@ impl Default for HttpVersion {
 #[cfg(test)]
 mod tests {
     use std::str::FromStr;
+    use error::Error;
     use super::HttpVersion;
     use super::HttpVersion::{Http09,Http10,Http11,H2,H2c};
 
@@ -81,7 +82,8 @@ mod tests {
     #[test]
     fn test_from_str_panic() {
         match HttpVersion::from_str("foo") {
-            Err(_) => assert!(true),
+            Err(Error::Version) => assert!(true),
+            Err(_) => assert!(false),
             Ok(_) => assert!(false),
         }
     }
--- a/src/version.rs
+++ b/src/version.rs
@@ -57,3 +57,31 @@ impl Default for HttpVersion {
         Http11
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+    use super::HttpVersion;
+    use super::HttpVersion::{Http09,Http10,Http11,H2,H2c};
+
+    #[test]
+    fn test_default() {
+        assert_eq!(Http11, HttpVersion::default());
+    }
+
+    #[test]
+    fn test_from_str() {
+        assert_eq!(Http09, HttpVersion::from_str("HTTP/0.9").unwrap());
+        assert_eq!(Http10, HttpVersion::from_str("HTTP/1.0").unwrap());
+        assert_eq!(Http11, HttpVersion::from_str("HTTP/1.1").unwrap());
+        assert_eq!(H2, HttpVersion::from_str("h2").unwrap());
+        assert_eq!(H2c, HttpVersion::from_str("h2c").unwrap());
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_from_str_panic() {
+        HttpVersion::from_str("foo").unwrap();
+    }
+        
+}
--- a/src/version.rs
+++ b/src/version.rs
@@ -41,18 +41,14 @@ impl fmt::Display for HttpVersion {
 impl FromStr for HttpVersion {
     type Err = Error;
     fn from_str(s: &str) -> Result<HttpVersion, Error> {
-        if s == "" {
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: a0c5c8cb973e794cd31c9e9b53677f20de0c7236

---

liamchristopher:
Commit message amended.

src/version.rs
```diff
@@ -36,6 +38,24 @@ impl fmt::Display for HttpVersion {
...
+        if s == "" {
+            Err(Error::Method)
+        } else {
+            Ok(match s {
+                "HTTP/0.9" => Http09,
+                "HTTP/1.0" => Http10,
+                "HTTP/1.1" => Http11,
+                "h2" => H2,
+                "h2c" => H2c,
+                _ => unreachable!(),
```
seanmonstar: This case isn't really unreachable, `HttpVersion::from_str("foo")` would hit it. It should return an error case instead.

src/version.rs
```diff
@@ -36,6 +38,24 @@ impl fmt::Display for HttpVersion {
     }
 }
 
+impl FromStr for HttpVersion {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<HttpVersion, Error> {
+        if s == "" {
+            Err(Error::Method)
```
seanmonstar: The error should probably be `Error::Version`.

Also, this if could be removed, and empty string would match the `_` arm in the match below.

liamchristopher: D'oh!

src/version.rs
liamchristopher: Thanks, I reworked this a few times, but wasn't sure what to return. 
Let me straighten this one out, and I'll update it.

liamchristopher:
@seanmonstar Thanks. For a simple function that was somehow too complicated. I'm still getting used to rust's syntax and seeing that I can return an alternate value in constructing a success response _really_ cleared things up.

I added a few tests, too. Checks the default, from_str and error on from_str.

src/version.rs
```diff
@@ -36,8 +38,50 @@ impl fmt::Display for HttpVersion {
...
+        assert_eq!(Http10, HttpVersion::from_str("HTTP/1.0").unwrap());
+        assert_eq!(Http11, HttpVersion::from_str("HTTP/1.1").unwrap());
+        assert_eq!(H2, HttpVersion::from_str("h2").unwrap());
+        assert_eq!(H2c, HttpVersion::from_str("h2c").unwrap());
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_from_str_panic() {
+        HttpVersion::from_str("foo").unwrap();
```
seanmonstar: Prefer a test against the `Error::Version` return value instead. This kind of test doesn't notice if the panic came from inside `from_str`, or from the `unwrap`.

src/version.rs
```diff
@@ -36,8 +38,50 @@ impl fmt::Display for HttpVersion {
     }
 }
 
+impl FromStr for HttpVersion {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<HttpVersion, Error> {
+        Ok(match s {
```
seanmonstar: Exactly what I was thinking!

src/version.rs
```diff
@@ -36,8 +38,50 @@ impl fmt::Display for HttpVersion {
...
+    use super::HttpVersion;
+    use super::HttpVersion::{Http09,Http10,Http11,H2,H2c};
+
+    #[test]
+    fn test_default() {
+        assert_eq!(Http11, HttpVersion::default());
+    }
+
+    #[test]
+    fn test_from_str() {
```
seanmonstar: Yay tests!

src/version.rs
liamchristopher: @seanmonstar Ok. That took me way too long to figure out the right(?) way to do that. 
The test now asserts true if an error condition is raised, and fails otherwise.
let me know what you think.

liamchristopher:
@seanmonstar I think github ate my last comment. I updated the test to do what I think is the right thing. Let me know what you think.

seanmonstar:
Looks great! Will merge once CI finishes, thanks!

liamchristopher:
@seanmonstar Actually hold up. That test should be against Error::Version specifically. I have one more change committed.

liamchristopher:
@seanmonstar Hm. I pushed the changes a while ago, but its not reflected here. I'm guessing PR's don't update post-approval. Should I open a new one?

PR #1107, opened 2017-03-30
feat(buffer): add from_parts and into_parts functions

illegalprime:
this adds the ability to deconstruct a buffer into all of its parts
and the ability to create a buffer from all of its parts, this lets
users deconstruct the buffer, edit the reader, then build it back up
again without losing any information

Currently a whole heap of changes is coming to rust-websocket, some of the changes require that we handle streams generically and therefore we need better interoperability with other buffered reader implementations. The large changes (which had to be done all at once) are in cyderize/rust-websocket#80, and include the ability to upgrade a hyper request to a websocket connection.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -18,6 +18,16 @@ impl<R: Read> BufReader<R> {
     }
 
     #[inline]
+    pub fn from_parts(rdr: R, buf: Vec<u8>, pos: usize, cap: usize) -> BufReader<R> {
+        BufReader {
+            inner: rdr,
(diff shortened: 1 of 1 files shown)


seanmonstar:
Ok!

illegalprime:
@seanmonstar I'm ready to publish the shiny new websockets crate but I can't until this change gets published in a hyper version, I was wondering what the timeline is for that? Thanks!

seanmonstar:
Published v0.10.6 with this change. Thanks!

Issue #1104, opened 2017-03-27
More allocator friendly Header definitions 


arthurprs:
I was looking at https://github.com/TechEmpower/FrameworkBenchmarks/pull/2611 the other day and I noticed there's no way of using some Header types without allocating, like the `Server` header in the linked example.

It's not a problem per se but it's probably worth discussing as setting N headers could potentially allocate N times.

seanmonstar:
Agreed. I starting with the most common, `Host`, but others should follow. Since changing the signature would be a breaking change anyways, it also makes sense to make the fields private and use constructors instead. That way we can start with `Cow` but we can switch to `Bytes` or something later in the future and not break. Some common headers that could beneift:

- Location
- Origin
- Referer
- Server
- UserAgent

mcarlin:
@seanmonstar, I'm looking to start contributing here. Mind if I take a stab at it and convert over those headers you mentioned if you're not already working on it?

seanmonstar:
@mcarlin That'd be great! I turned the list into checkboxes, ideally we can keep them up to date.

seanmonstar:
The list of headers I made have all been converted. If there's changes want to a specific header, lets open specific issues. Thanks!

PR #1119, opened 2017-04-08
refactor(header): make some headers more allocator friendly

hannesg:
Change the internal implementation of some simple headers to make them
more allocator friendly. Also add a constructor method to allow changing
the implementation in the future again.

The headers are:

- Location
- Referrer
- Server
- UserAgent

This change was suggested in #1104.

I've left out the Origin header for now because the internal implementation is different.

BREAKING CHANGES:
- Old code that creates the header structs directly will stop working.
- It's not possible to implement DerefMut for a Cow<'static,str>. Code
that needs to modify header after creation will stop working.

### DIFF
--- a/src/header/common/mod.rs
+++ b/src/header/common/mod.rs
@@ -264,6 +264,41 @@ macro_rules! header {
             }
         }
     };
+    // Single value cow header
+    ($(#[$a:meta])*($id:ident, $n:expr) => Cow[$value:ty]) => {
+        $(#[$a])*
+        #[derive(Clone, Debug, PartialEq)]
+        pub struct $id(::std::borrow::Cow<'static,$value>);
+        impl $id {
+            /// Creates a new $id
+            pub fn new<I: Into<::std::borrow::Cow<'static,$value>>>(value: I) -> Self {
+                $id(value.into())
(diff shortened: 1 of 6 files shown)


src/header/common/mod.rs
```diff
@@ -264,6 +264,41 @@ macro_rules! header {
...
+        #[derive(Clone, Debug, PartialEq)]
+        pub struct $id(::std::borrow::Cow<'static,$value>);
+        impl $id {
+            /// Creates a new $id
+            pub fn new<I: Into<::std::borrow::Cow<'static,$value>>>(value: I) -> Self {
+                $id(value.into())
+            }
+        }
+        impl ::std::ops::Deref for $id {
+            type Target = ::std::borrow::Cow<'static,$value>;
```
seanmonstar: So that we don't leak that it is actually a `Cow` inside, let's just `Deref` to the `$value` type instead. This will let us change `Cow` to a `Bytes` or something later, and it should all still just work.

hannesg:
@seanmonstar sure! I've updated that.

seanmonstar:
Fantastic, thanks!

PR #1121, opened 2017-04-10
refactor(header): make Origin header more allocator friendly

hannesg:
Change the internal representation of Origin to be more allocator
friendly. The internals of the Origin header are now private to allow
changing them in the future.

This change was suggested in #1104.

BREAKING CHANGES:
- Old code that accesses the Origin struct directly will stop working.


seanmonstar:
Thank you!

PR #1122, opened 2017-04-10
Introduce Response::get_ref

nox:
This ultimately allows end users to retrieve the current cipher from an SSL stream, which is needed for the Web.


seanmonstar:
Thanks! Let me know if after fitting all the pieces together, this solves your use case, and I'll release a new version (so that hopefully, if it doesn't, we don't have new versions fixing nothing.)

PR #1136, opened 2017-04-13
refactor(headers): use static "http"/"https" for scheme

### LINKED ISSUE #1124: Reduce to_owned of scheme in Origin header
Now that the `scheme` propertry is a `Cow<'static, str>`, and the scheme is pretty much always only ever "http" or "https", we should compare for those and then store a `Cow::Borrowed("http")` etc, instead of calling `scheme.to_owned()`.

hannesg:
This change removes one common string allocation when parsing the origin
header.

Fixes #1124 

### DIFF
--- a/src/header/common/origin.rs
+++ b/src/header/common/origin.rs
@@ -84,6 +84,9 @@ impl Header for Origin {
     }
 }
 
+static HTTP : &'static str = "http";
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks!

PR #1141, opened 2017-04-19
feat(server): add local_addr to retrieve resolved address

SergioBenitez:
As the title suggests, this adds a `local_addr` method to `Server` to retrieve the address that was resolved. This makes it possible to retrieve the resolved address when a `Server` instance is created with the `Server::http` or `Server:https` methods.


seanmonstar:
LGTM!

seanmonstar:
v0.10.9 has been published with this change.

SergioBenitez:
Thanks so much! :)

PR #1142, opened 2017-04-19
feat(method): provide HashMap from Cookie header

### LINKED ISSUE #1139: hyper::header::Cookie not hashable
From what I can see hyper::header::Cookie provides cookies as separate "key=value" element, but there is no way to access the keys directly (let me know if I'm missing something here). I think there would be value to provide a way to access Cookies directly bye keys right in hyper rather than as a separate crate.

Are there reasons for not providing this? Is it worth adding a PR for it? Should this be done as part of hyper::header::Cookie, or as a separate struct (say hyper::header::hashedCookie)?

dorfsmay:
Cookie headers are always key/value only. Apps servers typically provide
facilities to test existence and access them by name without additional
libraries.

Closes #1139


seanmonstar:
Thanks for submitting this work! As suggested in a previous issue, I'm not certain whether this fits within the scope of hyper. It's possible and quite likely that anyone dealing with cookies much would want much more that a simple map.

The scarier part, is that once hyper commits to an API, it can't easily break it. So, it should probably at least be explored and published as a separate thing, and be considered in the future. Does that make sense?

dorfsmay:
I understand your concern about long term support, it totally makes sense... I started writing a quick answer, but turned into a novel - sorry about that. I don't want to be argumentative, I am not trying to push for this PR just for its own sake, but hope the need for a better parsing of the Cookie header isn't missed.


This PR is not trying to deal with cookie life cycle, it is very specific to the http "Cookie" header. Parsing cookies from the "Cookie" header will never need to be more complicated, as it is exactly this, a succession of keys and values (separated by ; and =) - not to be confused with the Set-Cookie header and the issues of storing cookies, determining their scope, etc... In my view, the current implementation is not a good representation of the "Cookie" header, and this PR is basically trying to bring it at par with what has been done for [hyper::header::SetCookie](https://hyper.rs/hyper/v0.10.9/hyper/header/struct.SetCookie.html).

As you mentioned in #1139, dealing with cookie life cycle is a much bigger job and the [cookie crate](https://github.com/alexcrichton/cookie-rs) can be used for this. But in my experience, the need to be bale to just check specific cookies on incoming requests (so pure k/v) arise quite often.

Is there any way to get feedback form people using hyper? I wonder how many make use of the "Cookie" header without wanting to use a heavy gun such as cookie-rs, and end up having to write their own version of this...

seanmonstar:
Thinking about it, I've realized the error of my ways. The real problem here is not adding this method, it's that `Cookie` is just a `Vec<String>`! We only ever receive headers like this: `Cookie: foo=bar; session=sean; hello=world`, and having that be a vector of `["foo=bar", "session=sean", "hello=world"]` doesn't really help anyone.

We should change the `Cookie` header to be a map-like object. I've opened an issue for discussion/design: https://github.com/hyperium/hyper/issues/1145

seanmonstar:
I'm going to close this, as a different solution is discussed in #1145.

PR #1147, opened 2017-04-20
feat(headers): support Opaque origin headers

### LINKED ISSUE #1065: Support Opaque origin headers, serializing it to `null`
Support Opaque origins:

> An internal value, with no serialization it can be recreated from (it is serialized as "null" per ASCII serialization of an origin), for which the only meaningful operation is testing for equality.

https://html.spec.whatwg.org/multipage/browsers.html#concept-origin

alexeyzab:
Add support for Opaque origin header, serializing it to `null`.
https://html.spec.whatwg.org/multipage/browsers.html#concept-origin

Closes #1065

### DIFF
--- a/src/header/common/origin.rs
+++ b/src/header/common/origin.rs
@@ -31,20 +31,39 @@ use header::parsing::from_one_raw_str;
 /// );
 /// ```
 
-#[derive(Clone, Debug)]
-pub struct Origin {
-    /// The scheme, such as http or https
-    scheme: Cow<'static,str>,
-    /// The host, such as Host{hostname: "hyper.rs".to_owned(), port: None}
-    host: Host,
+#[derive(PartialEq, Clone, Debug)]
+pub struct Origin(OriginOrNull);
+
+#[derive(PartialEq, Clone, Debug)]
+enum OriginOrNull {
+    Origin {
+        /// The scheme, such as http or https
+        scheme: Cow<'static,str>,
+        /// The host, such as Host{hostname: "hyper.rs".to_owned(), port: None}
+        host: Host,
+    },
+    Null,
 }
 
 impl Origin {
     /// Creates a new `Origin` header.
     pub fn new<S: Into<Cow<'static,str>>, H: Into<Cow<'static,str>>>(scheme: S, hostname: H, port: Option<u16>) -> Origin{
-        Origin {
+        Origin(OriginOrNull::Origin {
             scheme: scheme.into(),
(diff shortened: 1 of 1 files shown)


src/header/common/origin.rs
```diff
@@ -31,41 +31,70 @@ use header::parsing::from_one_raw_str;
...
     /// ```
     /// use hyper::header::Origin;
     /// let origin = Origin::new("https", "foo.com", Some(443));
-    /// assert_eq!(origin.scheme(), "https");
+    /// assert_eq!(origin.scheme(), Some("https"));
     /// ```
-    pub fn scheme(&self) -> &str {
-        &(self.scheme)
+    #[allow(unused_variables)]
+    // Variable `host` is considered unused, although there does not seem to be a way to omit it here.
```
seanmonstar: You can just elide all the fields you don't want, like so `OriginOrNull::Origin { ref scheme, .. }`

src/header/common/origin.rs
```diff
@@ -31,41 +31,70 @@ use header::parsing::from_one_raw_str;
...
     /// use hyper::header::{Origin,Host};
     /// let origin = Origin::new("https", "foo.com", Some(443));
-    /// assert_eq!(origin.host(), &Host::new("foo.com", Some(443)));
+    /// assert_eq!(origin.host(), Some(&Host::new("foo.com", Some(443))));
     /// ```
-    pub fn host(&self) -> &Host {
-        &(self.host)
+    #[allow(unused_variables)]
+    // Variable `scheme` is considered unused, although there does not seem to be a way to omit it here.
+    pub fn host(&self) -> Option<&Host> {
```
seanmonstar: Aw, having to use `Option`s is unfortunate, that I had forgotten about. Oh well, there's no other way!

src/header/common/origin.rs
alexeyzab: Oh, this is really neat, thanks for pointing it out! I am a bit new to Rust and didn't know about this until now.

jdm: Another trick is to add `_` to the names of otherwise-unused variables, which suppresses the warning.

alexeyzab:
I went ahead and squashed the extra commit as well as force pushed it. 

Not sure if that's the right approach here, but I wanted to keep this change as a single commit. Let me know if there are other things I can add/improve here.

Thank you!

seanmonstar:
Thanks! Excellent work!

Issue #1109, opened 2017-04-01
Add struct for TE header field


sector-f:
The TE header field is similar to Accept-Encoding, but for the Transfer-Encoding instead of the Content-Encoding

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/TE
Section 14.39 https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

almielczarek:
I'm interested in helping with this!

Just to make sure I am on the right track, I think what I should do is implement the TE struct using the `header!` macro, passing the `QualityItem<Encoding>` struct as this also has a notion of quality values, and then writing the documentation based off of RFC7230. Does this make sense?

PR #1149, opened 2017-04-24
feat(headers): add `TE` header struct

almielczarek:
The `TE` header is used by a client to specify which
transfer encodings other than `chunked` it will accept.
It also specifies whether HTTP trailers are acceptable.
This commit also adds a `Trailers` variant to the
`hyper::header::shared::Encoding` enum.

### DIFF
--- a/src/header/common/te.rs
+++ b/src/header/common/te.rs
@@ -1,0 +1,67 @@
+use header::{Encoding, QualityItem};
+
+header! {
+    /// `TE` header, defined in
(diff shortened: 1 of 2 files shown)


almielczarek:
This is to address issue #1109. I can create a new pull request if you would like that in the commit message.

almielczarek:
I need to fix an error. I will also put the issue this PR is referencing in the commit message.

PR #1153, opened 2017-04-27
feat(headers): use a Vec + a map for Cookie

dorfsmay:
Use a Vec and a map for the Cookie header in order to satisfy the need
for both duplicate names (keys) from a client perspective and single
value for a given key from a server perspective .

BREAKING CHANGE: This change the format of the cookie header, any
previous implementation will have to be changed.

### DIFF
--- a/src/header/common/cookie.rs
+++ b/src/header/common/cookie.rs
@@ -1,8 +1,7 @@
-use header::{Header, Raw};
+use header::{Header, Raw, VecMap};
 use std::fmt;
 use std::str::from_utf8;
 use std::borrow::Borrow;
-use std::collections::HashMap;
 
 /// `Cookie` header, defined in [RFC6265](http://tools.ietf.org/html/rfc6265#section-5.4)
 ///
@@ -29,10 +28,11 @@ use std::collections::HashMap;
 ///    ])
 /// );
 /// ```
-#[derive(Clone, PartialEq, Debug)]
+#[allow(missing_debug_implementations)]
+#[derive(Clone)]
 pub struct Cookie {
     cookies: Vec<(String, String)>,
-    index: HashMap<String, String>,
+    index: VecMap<String, String>,
 }
 
 
@@ -49,9 +49,11 @@ impl Header for Cookie {
         for cookies_raw in raw.iter() {
             let cookies_str = try!(from_utf8(&cookies_raw[..]));
             for cookie_str in cookies_str.split(';') {
-                //cookies.push(cookie_str.trim().to_owned())
-
-                let mut kv_iterator = cookie_str.splitn(2, '=');
+                let cookie_trimmed = cookie_str.trim();
+                if cookie_trimmed.len() == 0 {
+                    continue;
+                }
+                let mut kv_iterator = cookie_trimmed.splitn(2, '=');
                 // split returns at least one element - unwrap is safe
                 let k = kv_iterator.next().unwrap().trim();
                 let v = match kv_iterator.next() { 
@@ -85,7 +87,7 @@ impl Cookie {
     pub fn with_capacity(capacity: usize) -> Cookie {
         Cookie {
             cookies: Vec::with_capacity(capacity),
-            index: HashMap::with_capacity(capacity),
+            index: VecMap::with_capacity(capacity),
         }
     }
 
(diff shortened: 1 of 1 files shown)


dorfsmay:
@seanmonstar @LegNeato

This is a first stab to make sure that's going in the right direction.

* It doesn't matter how you dice it, at the end of the day, a "Cookie header" is really a string representation of a Vec<(String, String)>, so let's use that, and use methods and/or additional elements in the structure to provide the functionalities we need, such as a get() returning the first value for a given key.
* I know HashMap, I'll re-submit a different PR with VecMap, right now, this compiles and demonstrates the concepts.
* I use a Vec<(String, String)> + a HashMap (VecMap later), in order to be able to both not lose duplicate keys, and only give the value for the first key on a et
* Another way to do the above would be to just use a Vec<(String, String)>, and make the get() method walk it. It would be cheaper in memory, but more expensive if an implementer looks up more than one key.
* Using owned value in the HasMap is a waste of memory, but I am very green to Rust (weeks), I wasted hours trying to make it use reference to the owned values inside the Vec and make it work in hyper... Again this is more to show what I think would be the right implementation for the Cookie header. Also, we could roll with this as it gets the signatures right, and implement internal references later if it is worth it (and maybe somebody more experience add that!).
* I added a set, diff from a push (append) as per @LegNeato sugestion
* Still missing: a proper iterator. I wasted another few hours on this, will work more on that later.

GitCop:

 * Commit: 6c74b5ff9d7699339efc102dd8fe40728816ca21

---

src/header/common/cookie.rs
```diff
@@ -59,18 +77,82 @@ impl Header for Cookie {
...
+    pub fn get<T: Borrow<String>>(&self, name: T) -> Option<String> {
+        match self.index.get(name.borrow()) {
+            Some(value_ref) => Some((*value_ref).clone()),
+            None => None,
+        }
+    }
+
+
+    /// Clear the current Cookie, and add one with specified name and value.
+    pub fn set<T: Into<String>>(&mut self, name_tref: T, value_tref: T) {
```
seanmonstar: My understanding of the suggestion for this method was to clear the values of only this key. This would seem confusing:

```rust
let mut cookie = Cookie::new();
cookie.set("foo", "bar");
cookie.set("baz", "quux");
// where did my foo=bar go!
assert_eq!(cookie.to_string(), "baz=quux");
```

src/header/common/cookie.rs
```diff
@@ -59,18 +77,82 @@ impl Header for Cookie {
...
+
 impl fmt::Display for Cookie {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let cookies = &self.0;
-        for (i, cookie) in cookies.iter().enumerate() {
-            if i != 0 {
-                try!(f.write_str("; "));
+        let mut first = true;
+        for pair in self.cookies.clone() {
+            let cookie = format!("{}={}", pair.0, pair.1);
```
seanmonstar: This would allocate a brand new `String`, before then writing it to the formatter. It can be done without the temporary allocated, such as `write!(f, "{}={}", key, val)`.

src/header/common/cookie.rs
```diff
@@ -27,10 +28,15 @@ use std::str::from_utf8;
 ///    ])
 /// );
 /// ```
-#[derive(Clone, PartialEq, Debug)]
-pub struct Cookie(pub Vec<String>);
+#[allow(missing_debug_implementations)]
+#[derive(Clone)]
+pub struct Cookie {
+    cookies: Vec<(String, String)>,
+    index: VecMap<String, String>,
```
seanmonstar: The `VecMap` is sufficient on it's own. There no need to double the strings here. The `VecMap` has `insert` and `append`, and `iter` gets all the key-value pairs, with duplicates.

Also, a `Cow<'static, str>` can be used, which allows someone who wants to set the same key and value on all cookies to do so without allocating and copying memory for it:

```rust
let mut cookie = Cookie::new();
// no allocation or copy required
cookie.set("foo", "bar");
// dynamic string, use a String
cookie.set(nonce_key.to_string(), nonce_val.to_string());
```

src/header/common/cookie.rs
```diff
@@ -27,10 +28,15 @@ use std::str::from_utf8;
 ///    ])
 /// );
 /// ```
-#[derive(Clone, PartialEq, Debug)]
-pub struct Cookie(pub Vec<String>);
+#[allow(missing_debug_implementations)]
```
seanmonstar: We'd still want a `Debug` implementation. It can use `formatter.debug_map()`.

dorfsmay: Agreed, but with my implementation, I would have had to add a Debug trait to VecMap. I wanted to get the code out fast for sake of discussion, rather than spend too many hours on an implementation that wasn't going to be used.

src/header/common/cookie.rs
dorfsmay: I understood the suggestion as push()/append() to add an extra cookie name/value to the header, and set() to for the entire header to this one value.

src/header/common/cookie.rs
dorfsmay: Great point. Thanks.

dorfsmay:
Thanks for taking the time to go through my code and the comments, this is really helpful.

PR #1158, opened 2017-04-29
test(header): test valid corner cases

dorfsmay:
Test corner cases that are easy to miss when coding such as equal sign
in the cookie value and empty and empty value.

The first thing I did when installing this new version was to run those manually and thought there was value in adding them, as they'd be easy to get wrong.

The extra formatting changes are due to rustfmt from my editor.


seanmonstar:
Great, thank you!

PR #1160, opened 2017-04-30
refactor(lib): use try_nb! from tokio_io, not tokio_core

bluetech:
tokio_core::io is deprecated, and the try_nb! conceptually belongs to
that module. It is also available in tokio_io, so use that instead.


seanmonstar:
Thank you!

Issue #1155, opened 2017-04-28
Unite Request and Response types


seanmonstar:
Currently, `client::Request` is a different type from `server::Request`, and `client::Response` is different from `server::Response`.

We want to unite them into the same type. That dramatically improves generic middleware that could apply on both a server and a client, and makes things like proxies far simpler as well, since a received `client::Response` can just become the `server::Response` for free.

PR #1162, opened 2017-05-01
refactor(http): merge Request and Response from server and client

nickgonzales:
I started exploring Hyper's code and saw issue #1155, so I started tinkering with the `Request` and `Response` classes. For the most part, the merge was straight-forward, but ran into a small snag: `server::Request` has a `Body`, but `client::Request` has an `Option<Body>`. I merged them by making the final `Request` use `Option<Body>` and added the `Default` trait to `Body`. Unfortunately, this makes the formerly-concise `req.body()`, as seen in `test/server.rs#117`, a lengthy `req.take_body().unwrap_or_default()`. 

I also re-exported `Request` and `Response` to minimize breaking changes, but this has the side-effect of being visible from 3 locations: `hyper`, `hyper::server`, and `hyper::client`).

Thoughts?


GitCop:

 * Commit: ff4369835b002764d62525ad46bec3702ce6bf25

---

seanmonstar:
Oh my, this is super exciting! I was slowly working on this, since I wanted to improve a reverse proxy example for the docs, but I also didn't want to slow down releasing 0.11, so I kept pausing. In my branch, I had united the inner parts into a `Message` with a generic `Head`, but the exposed result would be the same thing, and it doesn't really matter.

I'll do a proper review in a moment.

src/http/request.rs
```diff
@@ -49,7 +52,32 @@ impl<B> Request<B> {
     /// Read the Request body.
     #[inline]
     pub fn body(&self) -> Option<&B> { self.body.as_ref() }
-    
+
+    /// Take the Body of this Request
+    #[inline]
+    pub fn take_body(self) -> Option<B> { self.body }
```
seanmonstar: So as to mirror `Response`, I was thinking that the methods should perhaps be `body_ref()` and `body()`.

Maybe exactly how the body stream is taken from a Request or Response should be different, but if so, sounds like a separate issue to discuss.

src/http/response.rs
```diff
@@ -0,0 +1,152 @@
...
+impl<B> Response<B> {
+    /// Constructs a default response
+    #[inline]
+    pub fn new() -> Response<B> {
+        Response::default()
+    }
+
+    /// Get the HTTP version of this response.
+    #[inline]
+    pub fn version(&self) -> &HttpVersion { &self.version }
```
seanmonstar: Returning a `&HttpVersion` instead of `HttpVersion` might have been inconsistent on one of the types. The intent was to just always return it by value, since it's smaller than a pointer and `Copy`. Could you update that here?

src/http/request.rs
nickgonzales: With the following implementation already existing for Response:
```
impl Response<Body> {
    /// Take the `Body` of this response.
    #[inline]
    pub fn body(mut self) -> Body {
        self.body.take().unwrap_or(Body::empty())
    }
}
```

Would this make more sense?
```
impl<B> Request<B> {...
    /// Read the Request body.
    #[inline]
    pub fn body_ref(&self) -> Option<&B> { self.body.as_ref() }
...}

impl Request<Body> {
    /// Take the Request body.
    #[inline]
    pub fn body(self) -> Body { self.body.unwrap_or_default() }
}
```

My only concern with returning a default is that I think the caller should be able to identify if a body exists or not, but `body_ref()` seems to cover that use case.

Also, should I change the above impl for Response to use `self` instead of `mut self` so it consumes the response? Or should I leave it as-is? This would reduce the differences further.

seanmonstar: Yes, this looks like the best one:

```rust
pub fn body(self) -> Body { self.body.unwrap_or_default() }
```

Technically, with that it's possible for the impl to exist for `B: Default`, but I don't know if that helps anyone, really. For all incoming streams (incoming to this machine, built by hyper internall), it will always be `Body`.

seanmonstar:
Superb work, thank you!

PR #1179, opened 2017-05-17
feat(example): add an example http proxy server

### LINKED ISSUE #1096: simple proxy example
Hyper has client and server http support and can probably be used to write an HTTP proxy. @fmonjalet claims to have already written one. In Node.JS it's very simple to write such a proxy in [a few lines](http://www.catonmat.net/http-proxy-in-nodejs/).

Here is my attempt at doing this with Hyper. The return type from `fn call` is wrong and I've no idea how to fix it. Such a simple proxy would be an useful addition to `examples/`.

```rust 
//#![deny(warnings)]
extern crate futures;
extern crate hyper;
extern crate tokio_core;

use futures::Future;
use futures::future::FutureResult;
use hyper::{client, server, Client, Url};

#[derive(Clone, Copy)]
struct Proxy;

fn server_request_to_client_request(req: &server::Request) -> client::Request {
    // todo: handle parse error
    let url = Url::parse(&format!("{}", req.uri())).unwrap();
    client::Request::new(req.method().clone(), url)
}

impl server::Service for Proxy {
    type Request = server::Request;
    type Response = server::Response;
    type Error = hyper::Error;
    type Future = FutureResult<server::Response, hyper::Error>;

    fn call(&self, req: server::Request) -> Self::Future {
        let mut core = tokio_core::reactor::Core::new().unwrap();
        let handle = core.handle();
        let client = Client::new(&handle);
        let mut client_req = server_request_to_client_request(&req);
        client_req.set_body(req.body());

        client.request(client_req)
            .and_then(|res| {
                let mut resp: server::Response = server::Response::new();
                resp = resp.with_headers(res.headers().clone());
                resp.set_status(res.status().clone());
                resp.set_body(res.body());
                futures::future::ok(resp)
            })
    }
}

fn main() {
    let addr = "127.0.0.1:1337".parse().unwrap();

    let server = server::Http::new().bind(&addr, || Ok(Proxy)).unwrap();
    println!("Listening on http://{} with 1 thread.",
             server.local_addr().unwrap());
    server.run().unwrap();
}
```

The proxy can be tested with
```bash
http_proxy=http://localhost:1337 wget http://w3.org
```

meganehouser:
hyper can write an simple proxy.Such a simple proxy would be an useful addition to examples/.

Closes #1096


seanmonstar:
Hey, I'm so sorry I didn't notice this a while ago. Thanks for the effort!

I'm trying to have all examples and documentation like things be in the [guides](https://hyper.rs/guides), and I do believe that an "advanced" server guide making a naive proxy would probably be an excellent guide to add!

PR #1183, opened 2017-05-19
fix(lib): Remove unused unimplemented! macro.

MJDSys:
This macro isn't used anywhere, std now has an unimplemented macro
if we want to use it, and the nightly compiler now warns that this
unused.  This warning is a failure when compiling tests.


seanmonstar:
Thanks!

I technically had this in hyper's source so that I could add messages to the macro invocations, like `unimplemented!("Conn::poll_complete")`. I found that immensely useful, and so I've filed a PR to add that functionality to rust proper: https://github.com/rust-lang/rust/pull/42155

PR #1185, opened 2017-05-20
docs(server): refer to `bind` method, not `serve`

### LINKED ISSUE #1180: Current master docs for `hyper::server::Http::bind_connection` confusing
https://hyper.rs/hyper/master/hyper/server/struct.Http.html#method.bind_connection:
> This method is typically not invoked directly but is rather transitively used *through the `serve` helper method above*. This can be useful, however, when writing mocks or accepting sockets from a non-TCP location.

There's no `serve` helper method above. Did you mean `bind`? I'm not even sure, reading these docs quickly for the first time. As a general suggestion, is it possible to reference items using hyperlinks instead, to make it clearer?

sanmai-NL:
Fixes #1180.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -114,7 +114,7 @@ impl<B: AsRef<[u8]> + 'static> Http<B> {
     /// will be handled (and mapped to responses).
     ///
     /// This method is typically not invoked directly but is rather transitively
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thank you!

PR #1187, opened 2017-05-23
fix(header): Allow iterate cookie

mayah:
The commit dd03e723 changed Cookie implementation, but it also hided
Cookie header internal structure. So currently we don't have any way
to iterate cookie.

Add iter.

### DIFF
--- a/src/header/common/cookie.rs
+++ b/src/header/common/cookie.rs
@@ -85,6 +85,29 @@ impl Cookie {
     pub fn get(&self, key: &str) -> Option<&str> {
         self.0.get(key).map(AsRef::as_ref)
     }
+
+    /// Iterate cookies.
+    ///
+    /// Iterate cookie (key, value) in insertion order.
+    ///
+    /// ```
+    /// use hyper::header::Cookie;
+    /// let mut cookie = Cookie::new();
+    /// cookie.append("foo", "bar");
+    /// cookie.append(String::from("dyn"), String::from("amic"));
+    ///
+    /// let mut keys = Vec::new();
+    /// let mut values = Vec::new();
+    /// for (k, v) in cookie.iter() {
+    ///     keys.push(k);
+    ///     values.push(v);
+    /// }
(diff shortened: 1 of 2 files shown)


srijs:
You might want to consider wrapping the return type of `iter` in a tuple struct to prevent implementation details from leaking, like so:

```
struct CookieIter(::std::slice::Iter<(Cow<'static, str>, Cow<'static, str>)>);

impl Iter for CookieIter {
    type Item = (&str, &str);
    ...
}
```

mayah:
Thanks for the comment. You're right.
I've updated the patch, and has written a test to use that.
PTAL?

src/header/common/cookie.rs
```diff
@@ -154,6 +159,17 @@ impl fmt::Display for Cookie {
     }
 }
 
+#[allow(missing_debug_implementations)]
```
seanmonstar: I'd go ahead a slap `#[derive(Debug)]` on this instead. It shouldn't actually leak anything about cows, and instead should allow someone to view a list of strings pairs.

Also, I'm surprised the `deny(missing_docs)` lint didn't catch no documentation on this struct...

srijs: If I were to take an educated guess, I would say the lint didn't trigger because `CookieIter` is not exported all the way to the crate API.

@mayah, you might want to add a `pub use cookie::CookieIter` to [src/header/common/mod.rs](https://github.com/hyperium/hyper/blob/master/src/header/common/mod.rs#L33) to make sure `CookieIter` is exported from the crate.

GitCop:

 * Commit: 222ef07b8e8e4ae655cbc57aae05766229d32637

---

mayah:
Thanks for the review.
I've exported CookieIter, and moved test to document test.
PTAL?

seanmonstar:
@mayah Thanks so much! Sorry for being slow.

mayah:
Thank you!

PR #1200, opened 2017-06-02
Refactoring inlined AtomicWrite, and removed unneeded branch in parse

yazaddaruvala:
- [X] The commit messages match the [guidelines](https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md#git-commit-guidelines)

I'm not sure if this is the direction you're aiming for, but I'm trying as much as possible to move hyper's encoder / decoder types to match these tokio [method signatures](https://tokio.rs/docs/going-deeper-tokio/streaming/) and finding refactorings along the way

All the refactoring I've done so far should have no behavioral change, and I'm kinda scared to make behavioral changes. Instead I've just added TODOs, I hope that is cool. For example, moving io out of Encoder would allow it to impl Tokio::Encoder, but changing when io takes place is a bit scary.

### DIFF
--- a/src/http/conn.rs
+++ b/src/http/conn.rs
@@ -3,6 +3,8 @@ use std::io::{self, Write};
 use std::marker::PhantomData;
 use std::time::Instant;
 
+use bytes::BytesMut;
+
 use futures::{Poll, Async, AsyncSink, Stream, Sink, StartSend};
 use futures::task::Task;
 use tokio_io::{AsyncRead, AsyncWrite};
@@ -216,7 +218,7 @@ where I: AsyncRead + AsyncWrite,
 
     fn has_queued_body(&self) -> bool {
         match self.state.writing {
-            Writing::Body(_, Some(_)) => true,
+            Writing::Body(_, Some(_), _) => true,
             _ => false,
         }
     }
@@ -240,7 +242,8 @@ where I: AsyncRead + AsyncWrite,
         //TODO: handle when there isn't enough room to buffer the head
         assert!(self.io.buffer(buf) > 0);
         self.state.writing = if body {
-            Writing::Body(encoder, None)
+            // TODO: Figure out a better init capacity for BytesMut
+            Writing::Body(encoder, None, BytesMut::with_capacity(0))
         } else {
             Writing::KeepAlive
         };
@@ -256,13 +259,13 @@ where I: AsyncRead + AsyncWrite,
         }
 
         let state = match self.state.writing {
-            Writing::Body(ref mut encoder, ref mut queued) => {
+            Writing::Body(ref mut encoder, ref mut queued, ref mut result_buf) => {
                 if queued.is_some() {
                     return Ok(AsyncSink::NotReady(chunk));
                 }
                 if let Some(chunk) = chunk {
                     let mut cursor = Cursor::new(chunk);
-                    match encoder.encode(&mut self.io, cursor.buf()) {
+                    match encoder.encode(&mut self.io, cursor.buf(), result_buf) {
                         Ok(n) => {
                             cursor.consume(n);
 
@@ -303,9 +306,9 @@ where I: AsyncRead + AsyncWrite,
     fn write_queued(&mut self) -> Poll<(), io::Error> {
         trace!("Conn::write_queued()");
         let state = match self.state.writing {
-            Writing::Body(ref mut encoder, ref mut queued) => {
+            Writing::Body(ref mut encoder, ref mut queued, ref mut result_buf) => {
                 let complete = if let Some(chunk) = queued.as_mut() {
-                    let n = try_nb!(encoder.encode(&mut self.io, chunk.buf()));
+                    let n = try_nb!(encoder.encode(&mut self.io, chunk.buf(), result_buf));
                     chunk.consume(n);
                     chunk.is_written()
                 } else {
@@ -490,7 +493,7 @@ enum Reading {
 
 enum Writing<B> {
     Init,
-    Body(Encoder, Option<Cursor<B>>),
+    Body(Encoder, Option<Cursor<B>>, BytesMut),
     Ending(Cursor<&'static [u8]>),
     KeepAlive,
     Closed,
@@ -511,7 +514,7 @@ impl<B: AsRef<[u8]>> fmt::Debug for Writing<B> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
             Writing::Init => f.write_str("Init"),
-            Writing::Body(ref enc, ref queued) => f.debug_tuple("Body")
+            Writing::Body(ref enc, ref queued, _) => f.debug_tuple("Body")
                 .field(enc)
                 .field(queued)
                 .finish(),
@@ -672,6 +675,7 @@ impl<'a, T: fmt::Debug + 'a, B: AsRef<[u8]> + 'a> fmt::Debug for DebugFrame<'a, 
 
 #[cfg(test)]
 mod tests {
+    use bytes::BytesMut;
     use futures::{Async, Future, Stream, Sink};
     use futures::future;
     use tokio_proto::streaming::pipeline::Frame;
@@ -688,7 +692,7 @@ mod tests {
     impl<T> Writing<T> {
         fn is_queued(&self) -> bool {
             match *self {
-                Writing::Body(_, Some(_)) => true,
+                Writing::Body(_, Some(_), _) => true,
                 _ => false,
             }
         }
@@ -784,7 +788,7 @@ mod tests {
             let io = AsyncIo::new_buf(vec![], 0);
             let mut conn = Conn::<_, http::Chunk, ServerTransaction>::new(io, Default::default());
             let max = ::http::io::MAX_BUFFER_SIZE + 4096;
-            conn.state.writing = Writing::Body(Encoder::length((max * 2) as u64), None);
+            conn.state.writing = Writing::Body(Encoder::length((max * 2) as u64), None, BytesMut::with_capacity(0));
 
             assert!(conn.start_send(Frame::Body { chunk: Some(vec![b'a'; 1024 * 8].into()) }).unwrap().is_ready());
             assert!(!conn.state.writing.is_queued());
@@ -811,7 +815,7 @@ mod tests {
         let _: Result<(), ()> = future::lazy(|| {
             let io = AsyncIo::new_buf(vec![], 4096);
             let mut conn = Conn::<_, http::Chunk, ServerTransaction>::new(io, Default::default());
-            conn.state.writing = Writing::Body(Encoder::chunked(), None);
+            conn.state.writing = Writing::Body(Encoder::chunked(), None, BytesMut::with_capacity(0));
 
             assert!(conn.start_send(Frame::Body { chunk: Some("headers".into()) }).unwrap().is_ready());
             assert!(conn.start_send(Frame::Body { chunk: Some(vec![b'x'; 8192].into()) }).unwrap().is_ready());
@@ -824,7 +828,7 @@ mod tests {
         let _: Result<(), ()> = future::lazy(|| {
             let io = AsyncIo::new_buf(vec![], 1024 * 1024 * 5);
             let mut conn = Conn::<_, http::Chunk, ServerTransaction>::new(io, Default::default());
-            conn.state.writing = Writing::Body(Encoder::length(1024 * 1024), None);
+            conn.state.writing = Writing::Body(Encoder::length(1024 * 1024), None, BytesMut::with_capacity(0));
             assert!(conn.start_send(Frame::Body { chunk: Some(vec![b'a'; 1024 * 1024].into()) }).unwrap().is_ready());
             assert!(conn.state.writing.is_queued());
             assert!(conn.poll_complete().unwrap().is_ready());
@@ -887,7 +891,7 @@ mod tests {
             let mut conn = Conn::<_, http::Chunk, ServerTransaction>::new(io, Default::default());
             conn.state.reading = Reading::KeepAlive;
             assert!(conn.poll().unwrap().is_not_ready());
-            conn.state.writing = Writing::Body(Encoder::length(5_000), None);
+            conn.state.writing = Writing::Body(Encoder::length(5_000), None, BytesMut::with_capacity(0));
             assert!(conn.poll_complete().unwrap().is_ready());
             Ok::<(), ()>(())
         });
--- a/src/http/h1/encode.rs
+++ b/src/http/h1/encode.rs
@@ -48,10 +48,10 @@ impl Encoder {
         }
     }
 
-    pub fn encode<W: Write>(&mut self, w: &mut W, msg: &[u8]) -> io::Result<usize> {
+    pub fn encode<W: Write>(&mut self, w: &mut W, msg: &[u8], result_buf: &mut BytesMut) -> io::Result<usize> {
         match self.kind {
             Kind::Chunked(ref mut chunked) => {
-                chunked.encode(w, msg)
+                chunked.encode(w, msg, result_buf)
             },
             Kind::Length(ref mut remaining) => {
                 let n = {
@@ -92,7 +92,7 @@ enum Chunked {
 }
 
 impl Chunked {
-    fn encode<W: Write>(&mut self, w: &mut W, msg: &[u8]) -> io::Result<usize> {
+    fn encode<W: Write>(&mut self, w: &mut W, msg: &[u8], result_buf: &mut BytesMut) -> io::Result<usize> {
         match *self {
             Chunked::Init => {
                 let mut size = ChunkSize {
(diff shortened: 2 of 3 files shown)


src/http/conn.rs
```diff
@@ -235,12 +237,17 @@ where I: AsyncRead + AsyncWrite,
 
         let wants_keep_alive = head.should_keep_alive();
         self.state.keep_alive &= wants_keep_alive;
-        let mut buf = Vec::new();
+
+        // TODO: Can we reuse this butter?
+        // TODO: Figure out a better init capacity
+        let mut buf = BytesMut::with_capacity(0);
```
seanmonstar: I've toyed around with this a bit, actually. Using a `Vec` is actually faster here. `BytesMut` has some atomic checks inside each of its methods, and since we never share this buffer to other threads, we don't need it.

I have toyed with passing `self.io.write_buf_mut()` directly, instead, since it is likely an already-allocated vector, and would remove the copy from the intermediate vector into the write buf. I couldn't tell with my machine if it made any noticeable change, and was also worried about the `reserve`s inside `encode` making the buffer bigger than the max buf size, so, it hasn't changed. I'd welcome a change with measurements showing an improvements!

src/http/h1/date.rs
```diff
@@ -2,19 +2,21 @@ use std::cell::RefCell;
...
 
+use bytes::BytesMut;
+
 use time::{self, Duration};
 
 // "Sun, 06 Nov 1994 08:49:37 GMT".len()
 pub const DATE_VALUE_LENGTH: usize = 29;
 
-pub fn extend(dst: &mut Vec<u8>) {
+pub fn extend(dst: &mut BytesMut) {
```
seanmonstar: Same comments about extending a `Vec` being faster than a `BytesMut` here.

src/http/h1/encode.rs
```diff
@@ -107,50 +108,54 @@ impl Chunked {
             Chunked::End => return Ok(0),
             _ => {}
         }
+
         let mut n = {
-            let pieces = match *self {
+            // TODO: Find a better way to concatente
+            //    Could use write! if we reserved the required capacity upfront
+            match *self {
```
seanmonstar: The idea here is actually to try to use the `writev` syscall, which could be done with `AsyncWrite::write_bufs`. Before tokio, that's what `AtomicWrite` tried to do. `AtomicWrite` is now just a legacy artifact I haven't refactored yet.

The point is, it's better to keep all these as individual slices, and allow the `writev` to write all of them at once, without needing to copy them in user space.

src/http/conn.rs
```diff
@@ -490,7 +497,7 @@ enum Reading {
 
 enum Writing<B> {
     Init,
-    Body(Encoder, Option<Cursor<B>>),
+    Body(Encoder, Option<Cursor<B>>, BytesMut),
```
seanmonstar: There is already a buffer to write into, in `io.write_buf`.

src/http/conn.rs
yazaddaruvala: The only reason I was using `BytesMut` is because of the Tokio docs and types. I'll change them back to `Vec`.

src/http/h1/encode.rs
yazaddaruvala: Ok, I wont lie your statements are a bit above my pay grade. Mostly I was just trying to get rid of the allocations that come from `.concat()`, but can deffer that for now.

Do you mind if I just inline `AtomicWrite` since its just a simple passthrough, or would you rather keep it to be able to extend later?

src/http/conn.rs
yazaddaruvala: Hey yeah, the only reason I wasn't using `io.write_buf` is because I'm not sure when the `io.flush` method gets called.

If you're cool with `.encode` extending `io.write_buf` and then `io.flush` inside conn, that I can do.

I'm happy with it any way, just trying to remove the extra allocations inside `.encode` if possible.

src/http/h1/encode.rs
seanmonstar: The compiler can inline itself. I've left the way it is as a reminder to investigate using `AsyncWrite::write_bufs` in the future, but I know that it's only a reminder to me, no one else would be able to determine that. Probably could use a comment to that effect. If you'd rather inline the implementation and have a comment there instead, that's fine.

src/http/conn.rs
seanmonstar: I hear ya, and it's a worthy goal! The best way would be to make use of `AsyncWrite::write_bufs`, which will prevent a copy in the plain-tcp-on-unix case. Unfortunately, on Windows, and when using TLS, flattening (copying) into a single buffer is required.

In this case, since we're dealing with performance sensitive stuff, I'd *love* if any changes could include benchmarks showing that it was an improvement. I've spent a lot of time twiddling with things thinking it was an obvious improvement, and it turned out to make no difference.

For extending the `write_buf`, there's the additional constraint that it should ideally never grow beyond `MAX_BUFFER_SIZE`, which could happen with APIs taking a `&mut Vec`.

yazaddaruvala: Yeah fair enough. I was hoping that if we could move all this to the Tokio types then the Tokio foundation would magically handle all the performance stuff.

Whats a good way to run performance tests? I only really have access to a 4 year old mackbook pro

seanmonstar: I tend to try a few things when checking performance: 

1. Look into a specific `#[bench]` in the file of the function I'm optimizing (or write one if the thing I'm optimizing isn't specifically benched). For instance, there is now a `bench_server_transaction_encode` in the `h1.rs` file, when I was writing the `date.rs` stuff.
2. I run `cargo bench --bench server` (or `--bench end_to_end` if the stuff I touched would affect the `Client` also).
3. I run `cargo run --release --example hello`, and then run [wrk](https://github.com/wg/wrk) to see if requests per second changed.

I compare the results of `master` vs my current work to determine if it was useful.

src/http/h1/parse.rs
```diff
@@ -114,12 +114,7 @@ impl Http1Transaction for ServerTransaction {
         let init_cap = 30 + head.headers.len() * AVERAGE_HEADER_SIZE;
         dst.reserve(init_cap);
         debug!("writing headers = {:?}", head.headers);
-        if head.version == ::HttpVersion::Http11 && head.subject == ::StatusCode::Ok {
-            extend(dst, b"HTTP/1.1 200 OK\r\n");
-            let _ = write!(FastWrite(dst), "{}", head.headers);
-        } else {
-            let _ = write!(FastWrite(dst), "{} {}\r\n{}", head.version, head.subject, head.headers);
-        }
+        let _ = write!(FastWrite(dst), "{} {}\r\n{}", head.version, head.subject, head.headers);
```
seanmonstar: The code here is actually to be a fast path, since use `extend` is faster than going through the formatter, and `HTTP/1.1 200 OK\r\n` is the most common path, so it was made to be slightly faster.

yazaddaruvala: Lol, I never want to me in charge of performance sensitive code.

I'll add a comment instead. Similar to the comment for date, do you remember what that speedup was?

seanmonstar: I don't recall anymore. If someone is interested, it's possible run the benchmarks with and without this optimization.

src/http/conn.rs
yazaddaruvala: Any specific parameters for wrk?

seanmonstar: It depends on resources of your computer. I tend to stick it to 1 thread, cause I use a shared server, and a duration of 10 seconds. So `-t1 -d10`. Then, I bump up the connections until it starts to reduce the req/s significantly, and then use the same configuration in the feature branch, and then compare req/s.

yazaddaruvala:
1. In theory it should cut down on allocations.
2. Move `Encoder` closer to use a pre-defined Tokio trait. Just need to remove the io from Encoder.
3. It also sets hyper up for more reuse. For example, as you previously pointed out `Buffered` has its own internal buffers.
4. A better init capacity (i.e. not 0) would probably improve performance. Any thoughts on the init-capacity? Note: This buffer can never exceed: `(ChunkSize::len - ChunkSize::pos) + 2 + msg.len() + 2`. It might be a good size to initialize it to `max(msg.len()) + 4` or something.

FWIW: I did try swapping out `BytesMut` for `Vec<u8>` (to improve performance more) but the tests weren't passing, which was strange so I didn't want to modify tests and went back to `BytesMut`. I also didn't wanna play around a bunch with init-capacity, sharing more buffers or removing the io from Encoder unless you thought this was a decent first step and thats the direction you wanted to go.

Regarding benches: There seems to be a tiny improvement for `bench_server_tcp_throughput` (i.e. ~4.5%), but nothing substantial, and no change I could discern in any of the others (including wrk). I think the wrk benchmark is not a good example, as the response is just a hello world (i.e. the buffer isn't really getting any re-use).

Let me know if I did this right:
[Origin Master](https://pastebin.com/4YZ7eTb4):
```
test get_one_at_a_time  ... bench:     160,683 ns/iter (+/- 53,933) = 338 MB/s
test post_one_at_a_time ... bench:     139,204 ns/iter (+/- 14,831) = 391 MB/s
test bench_raw_tcp_throughput    ... bench:      29,455 ns/iter (+/- 1,715) = 5 MB/s
test bench_server_tcp_throughput ... bench:      64,910 ns/iter (+/- 4,129) = 2 MB/s

test get_one_at_a_time  ... bench:     133,196 ns/iter (+/- 23,932) = 408 MB/s
test post_one_at_a_time ... bench:     134,964 ns/iter (+/- 27,127) = 403 MB/s
test bench_raw_tcp_throughput    ... bench:      30,639 ns/iter (+/- 9,273) = 5 MB/s
test bench_server_tcp_throughput ... bench:      63,335 ns/iter (+/- 3,177) = 2 MB/s

test get_one_at_a_time  ... bench:     134,292 ns/iter (+/- 22,861) = 404 MB/s
test post_one_at_a_time ... bench:     138,777 ns/iter (+/- 28,843) = 392 MB/s
test bench_raw_tcp_throughput    ... bench:      29,926 ns/iter (+/- 3,009) = 5 MB/s
test bench_server_tcp_throughput ... bench:      62,860 ns/iter (+/- 7,198) = 2 MB/s


test http::h1::parse::tests::bench_parse_incoming                   ... bench:       4,359 ns/iter (+/- 990) = 201 MB/s
test http::h1::parse::tests::bench_server_transaction_encode        ... bench:         364 ns/iter (+/- 112) = 206 MB/s

test http::h1::parse::tests::bench_parse_incoming                   ... bench:       4,490 ns/iter (+/- 1,672) = 195 MB/s
test http::h1::parse::tests::bench_server_transaction_encode        ... bench:         437 ns/iter (+/- 89) = 171 MB/s

test http::h1::parse::tests::bench_parse_incoming                   ... bench:       4,768 ns/iter (+/- 813) = 184 MB/s
test http::h1::parse::tests::bench_server_transaction_encode        ... bench:         351 ns/iter (+/- 109) = 213 MB/s

Yazads-MacBook-Pro% wrk http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   305.84us   88.61us   5.21ms   89.37%
    Req/Sec    16.19k     1.37k   21.02k    84.65%
  325384 requests in 10.10s, 40.03MB read
Requests/sec:  32216.24
Transfer/sec:      3.96MB
Yazads-MacBook-Pro% wrk http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   305.93us   62.02us   2.84ms   71.35%
    Req/Sec    16.13k     1.29k   20.49k    87.00%
  320848 requests in 10.00s, 39.47MB read
Requests/sec:  32085.10
Transfer/sec:      3.95MB
Yazads-MacBook-Pro% wrk http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   316.02us  179.23us   9.93ms   99.08%
    Req/Sec    15.86k     1.14k   19.51k    86.00%
  315543 requests in 10.00s, 38.82MB read
Requests/sec:  31551.20
Transfer/sec:      3.88MB
```
[Local Changes](https://pastebin.com/xiJsErg6):
```
test get_one_at_a_time  ... bench:     131,957 ns/iter (+/- 13,539) = 412 MB/s
test post_one_at_a_time ... bench:     160,557 ns/iter (+/- 211,404) = 339 MB/s
test bench_raw_tcp_throughput    ... bench:      29,558 ns/iter (+/- 8,365) = 5 MB/s
test bench_server_tcp_throughput ... bench:      61,322 ns/iter (+/- 3,232) = 2 MB/s

test get_one_at_a_time  ... bench:     135,429 ns/iter (+/- 29,762) = 401 MB/s
test post_one_at_a_time ... bench:     141,821 ns/iter (+/- 22,682) = 383 MB/s
test bench_raw_tcp_throughput    ... bench:      29,960 ns/iter (+/- 14,258) = 5 MB/s
test bench_server_tcp_throughput ... bench:      62,109 ns/iter (+/- 3,325) = 2 MB/s

test get_one_at_a_time  ... bench:     129,452 ns/iter (+/- 34,420) = 420 MB/s
test post_one_at_a_time ... bench:     134,567 ns/iter (+/- 12,938) = 404 MB/s
test bench_raw_tcp_throughput    ... bench:      30,774 ns/iter (+/- 18,852) = 5 MB/s
test bench_server_tcp_throughput ... bench:      58,977 ns/iter (+/- 2,983) = 2 MB/s


test http::h1::parse::tests::bench_parse_incoming                   ... bench:       4,743 ns/iter (+/- 833) = 185 MB/s
test http::h1::parse::tests::bench_server_transaction_encode        ... bench:         379 ns/iter (+/- 151) = 197 MB/s

test http::h1::parse::tests::bench_parse_incoming                   ... bench:       4,321 ns/iter (+/- 1,697) = 203 MB/s
test http::h1::parse::tests::bench_server_transaction_encode        ... bench:         395 ns/iter (+/- 159) = 189 MB/s

test http::h1::parse::tests::bench_parse_incoming                   ... bench:       4,651 ns/iter (+/- 1,507) = 188 MB/s
test http::h1::parse::tests::bench_server_transaction_encode        ... bench:         357 ns/iter (+/- 118) = 210 MB/s

Yazads-MacBook-Pro% wrk http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   313.23us   62.16us   2.79ms   78.37%
    Req/Sec    15.75k     1.32k   32.25k    95.52%
  314983 requests in 10.10s, 38.75MB read
Requests/sec:  31188.50
Transfer/sec:      3.84MB
Yazads-MacBook-Pro% wrk http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   307.08us  107.77us   7.70ms   96.92%
    Req/Sec    16.15k     0.92k   19.47k    84.65%
  324517 requests in 10.10s, 39.92MB read
Requests/sec:  32127.92
Transfer/sec:      3.95MB
Yazads-MacBook-Pro% wrk http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   322.04us  210.94us  10.11ms   99.46%
    Req/Sec    15.66k   684.91    19.80k    81.19%
  314725 requests in 10.10s, 38.72MB read
Requests/sec:  31162.78
Transfer/sec:      3.83MB
```

seanmonstar:
The benchmarks run probably aren't affected at all by this change, since this is adjusting the chunked encoder, and those benchmarks are all using the length encoder.

I'm not sure I want to be adding a secondary buffer to `enum Writing`, when there is already a write buffer in `Buffered`. I find it's starting to get more complex to know which buffer still needs to be flushed.

If there's a change here, I'd much rather the change just be switching to using `AsyncWrite::write_bufs`.

yazaddaruvala:
> The benchmarks run probably aren't affected at all by this change, since this is adjusting the chunked encoder, and those benchmarks are all using the length encoder.

lol I'm good at life. I'll try and make an actual benchmark for this.

> I'm not sure I want to be adding a secondary buffer to enum Writing, when there is already a write buffer in Buffered. I find it's starting to get more complex to know which buffer still needs to be flushed.

The new buffer always has `.take` called on it, and as such is fully flushed every time. I guess, I'm curious: Do you not consider the call to `.concat()` using a secondary buffer?

seanmonstar:
In fact, I don't see it quite the same. I realize the change here is to try to hold onto the heap memory, so that it might able to be used again instead of asking jemalloc for another slot. But as the `Writing` enum is part of the connection's `State`, to me, it adds overhead to know whether there are bytes sitting in that buffer or not.

yazaddaruvala:
Ok closing this. All of these ideas suck when I bench them.

PR #1203, opened 2017-06-03
fix(http,server): Update tests to use non-deprecated futures methods.

daboross:
These seem mostly to be renames, with the exception of Unpark -> Notify, which includes a functional change. Needed so that travis will build with the `#![deny(warnings)]`.

In the test set using Notify, I'm unsure exactly what the `id` passed around should be referring to. Here I've ignored it when receiving, and just sent `0`, though that may not be the best solution.

### DIFF
--- a/src/http/body.rs
+++ b/src/http/body.rs
@@ -133,7 +133,7 @@ fn test_body_stream_concat() {
         tx.send(Ok("world".into())).wait().unwrap();
     });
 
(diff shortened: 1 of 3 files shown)


GitCop:

 * Commit: 71fb78da4ac72f3e3e2e7a7ffaec3b3b1cda3c85

---

PR #1204, opened 2017-06-03
feat(version): implement FromStr and AsRef<str> for Version

leoschwarz:

The `FromStr` trait was implemented in the master branch too, `AsRef<str>` is not but it's rather useful when serializing and deserializing types (which is what I'm currently doing with my upcoming mocking/replaying library for reqwest).

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -3,7 +3,9 @@
 //! Instead of relying on typo-prone Strings, use expected HTTP versions as
 //! the `HttpVersion` enum.
 use std::fmt;
+use std::str::FromStr;
 
+use error::Error;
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks!

killercup:
(Damn, forgot to comment here when I saw this yesterday.)

I wrote a macro a while ago that does basically this in a way where you don't have to repeat the variant names and strings multiple times: <https://play.rust-lang.org/?gist=de3b5653915d09019d96d94113d0b7a6>. I'm not sure what your stance on internal macros is, I just wanted to leave this here :)

PR #1225, opened 2017-06-20
refactor(http): return poll more

yazaddaruvala:
- [x] The commit messages match [the guidelines](https://github.com/hyperium/hyper/blob/master/CONTRIBUTING.md#git-commit-guidelines)

**Disclaimer**: Modifies `unsafe` (very slightly) in `read_from_io`

Summary: Its a pretty big CR, but I just change the types around.

More stuff:
Ideally, `io` and `decoder` would not be coupled, but that is above my pay grade.
`conn -> io, decoder`

Currently, this is what the dependencies look like:
`conn -> decoder -> io`
`conn`: Uses `Poll`
`decoder`:  Does not use `Poll`
`io`: Should use `Poll`

So I need to push `Poll` through decoder which, on the one hand, kinda sucks because there are more characters around, and it makes the tests harder to write. On the other hand, its the only good way to refactor `Buffered` to use things like `AsyncRead` and `AsyncWrite`.

Thoughts?

### DIFF
--- a/src/http/io.rs
+++ b/src/http/io.rs
@@ -69,7 +69,7 @@ impl<T: AsyncRead + AsyncWrite> Buffered<T> {
                     }
                 },
             }
-            match try_nb!(self.read_from_io()) {
+            match try_ready!(self.read_from_io()) {
                 0 => {
                     trace!("parse eof");
                     //TODO: With Rust 1.14, this can be Error::from(ErrorKind)
@@ -80,7 +80,7 @@ impl<T: AsyncRead + AsyncWrite> Buffered<T> {
         }
     }
 
-    fn read_from_io(&mut self) -> io::Result<usize> {
+    fn read_from_io(&mut self) -> Poll<usize, io::Error> {
         use bytes::BufMut;
         // TODO: Investigate if we still need these unsafe blocks
         if self.read_buf.remaining_mut() < INIT_BUFFER_SIZE {
@@ -97,13 +97,15 @@ impl<T: AsyncRead + AsyncWrite> Buffered<T> {
                 Ok(n) => n,
                 Err(e) => {
                     if e.kind() == io::ErrorKind::WouldBlock {
+                        // TODO: Push this out, ideally, into http::Conn.
                         self.read_blocked = true;
+                        return Ok(Async::NotReady);
                     }
                     return Err(e)
                 }
             };
             self.read_buf.advance_mut(n);
-            Ok(n)
+            Ok(Async::Ready(n))
         }
     }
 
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks @yazaddaruvala!

yazaddaruvala:
Lol I was just typing up a comment asking if you wanted me to make smaller CRs.

But actually, do you prefer more bite sized CRs so you can do them at a glance or larger ones like this?

seanmonstar:
Either is fine. I do prefer that a PR be focused, if possible, but size isn't an issue.

src/http/io.rs
```diff
@@ -102,13 +97,15 @@ impl<T: AsyncRead + AsyncWrite> Buffered<T> {
                 Ok(n) => n,
                 Err(e) => {
                     if e.kind() == io::ErrorKind::WouldBlock {
+                        // TODO: Push this out, ideally, into http::Conn.
                         self.read_blocked = true;
```
yazaddaruvala: @seanmonstar my next CR will probably move `read_blocked` into `http::Conn::State`. I've verified that its semantically equivalent in all other locations, help me with errors.

For example:
```
    pub fn parse<S: Http1Transaction>(&mut self) -> Poll<MessageHead<S::Incoming>, ::Error> {
          loop {
              // Initially, read_buf will be empty; read_blocked=false
              match try!(S::parse(&mut self.read_buf)) {
                  Some(head) => {
                      // Can only get here if `parse` has all of the data it needed to construct a MessageHead
                      // Which means, read_buf was just written to.
                      // Which in turn means read_blocked = false; because read_from_io would have had to return Async::Ready
                      //trace!("parsed {} bytes out of {}", len, self.read_buf.len());
                     return Ok(Async::Ready(head.0))
                  },
                  None => {
                      if self.read_buf.capacity() >= MAX_BUFFER_SIZE {
                         // This error case can in theory happen when read_blocked = true or false;
                         // However, does the state still matter if we are returning an error?
                         debug!("MAX_BUFFER_SIZE reached, closing");
                         return Err(::Error::TooLarge);
                      }
                  },
              }
              // Semantically equivalent to expand the macro:
              // read_blocked = false
              // do read_from_io
              // If read_from_io == NotReady: read_blocked = true; return NotReady
              match try_ready!(self.read_from_io()) {
                 0 => {
                      trace!("parse eof");
                      //TODO: With Rust 1.14, this can be Error::from(ErrorKind)
                      return Err(io::Error::new(io::ErrorKind::UnexpectedEof, ParseEof).into());
                  }
                 _ => {},
              }
          }
      }
```

In that example, when we return an error, the connection will be closed correct? As such the state no longer matters, and I can set `read_blocked` to whatever without worry?

seanmonstar: Returning an `Err` from `Stream::poll` signals the stream is doomed. Tokio will drop the stream.

yazaddaruvala: Perfect

PR #1226, opened 2017-06-20
refactor(uri): make ByteStr an implementation detail of uri

yazaddaruvala:

**Disclaimer**: Modifies `unsafe`


src/http/mod.rs
```diff
@@ -15,31 +15,16 @@ use version::HttpVersion::{Http10, Http11};
...
 mod conn;
 mod io;
 mod h1;
 //mod h2;
-mod str;
 pub mod request;
 pub mod response;
 
-/*
-macro_rules! nonblocking {
```
seanmonstar: Oh right, good catch!

src/uri.rs
```diff
@@ -389,6 +389,48 @@ impl StdError for UriError {
     }
 }
 
+// TODO: Is ByteStr adding value? Inline it?
```
seanmonstar: I find it adds value, because it encapsulates the safety into a separate type, so that `Uri` doesn't really need to think about safety.

yazaddaruvala: tldr: Im very happy to leave it alone and remove the comment for now.

I was playing with it last night, what I found was that more or less we do not interact with `ByteStr` or `Bytes` mostly we interact with `&str`. But at the same time we don't know the lifetime.

What would be ideal is if we could store `ByteStr {Bytes, &str /* pointing to Bytes */}`. Then if Uri was only using a `&str`, `ByteStr` would not have much use.

Thoughts?

yazaddaruvala:
Got it.

For now, I do like that it is in `uri.rs`, having it colocated is really helpful to find out if it is useful or how we can improve the interface. Basically, it allows people to play with it without worrying about it being used elsewhere.

I can put it in its own sub-module within Uri. Especially, if you want to eventually use it in other places then that definitely makes sense, and it would be good to ensure only the pub interface is used.

src/uri.rs
seanmonstar: Storing a `&str` isn't necessary. The API of `ByteStr` works just fine. The point is that we kind of want a `String`, but one that can be shared, since we receive `Bytes` when parsing new requests.

yazaddaruvala: Not a rebuttal, I'm just curious what you mean by shared. Doesn't `ByteStr` owns `Bytes`?

seanmonstar: The `Bytes` type is sort of like an `Arc<Vec<u8>>`, which some improvements. The point is so that when a `Request` is parsed and constructed, the various pieces internally the are backed by a string of bytes, all point back to only a single shared buffer. So for a `Uri`, the string `"https://github.com/hyperium/hyper/pull/1226"` isn't copied into a new part of the heap (like it would if we used `String`), but its only one slot, and `Uri`'s `Bytes` keeps the indices of what part it's interested in.

yazaddaruvala:
@seanmonstar re you happy the way it is (i.e. an internal module)?

Or would you prefer if `ByteStr` was in its own file: `src/uri/mod.rs` + `src/uri/str.rs`?

seanmonstar:
@yazaddaruvala actually, just last night as I was refactoring code in the client pool module, I wanted to replace the `Rc<String>` with a `ByteStr`... perhaps it should stay as a tiny module in `http`?

yazaddaruvala:
I'd prefer something like `use ::common::ByteStr;`, would that work for you?

seanmonstar:
That's fine, as long as `mod common` isn't public.

yazaddaruvala:
Yup agreed, I wouldn't want to expose this to anyone.

I would also be inclined to name it `use ::private_utils::ByteStr`. No real preference on the name. Let me know.

seanmonstar:
Having read through plenty of HTTP frameworks, many tend to have a common or commons module of non-HTTP-specific stuff that is needed to implement the framework, so I like `common`.

yazaddaruvala:
Updated to `common`

seanmonstar:
Thanks!

PR #1227, opened 2017-06-21
docs(headers): clarify TransferEncoding xor ContentLength

jonhoo:
Following a brief discussion with @seanmonstar on `#rust`:

Update the docs for `Content-Length` and `Transfer-Encoding` to mention that they are mutually exclusive, and that setting one will unset the other, in accordance with RFC7230.

Also expand the docs for `Request::set_body` to say that chunked encoding is used by default, and mention how to disable this feature.


seanmonstar:
Awesome, thanks!

PR #1229, opened 2017-06-23
refactor(http): consolidate state into Conn::State

yazaddaruvala:

**Disclaimer**: Modifies `unsafe` in `read_from_io`

Moving the `read_blocked` state into `http:Conn::State`. Not only allows us to refactor `http::io` more, but consolidating state usually results in easier mental models and long term code quality.

### DIFF
--- a/src/http/conn.rs
+++ b/src/http/conn.rs
@@ -38,6 +38,7 @@ where I: AsyncRead + AsyncWrite,
         Conn {
             io: Buffered::new(io),
             state: State {
+                read_blocked: false,
                 reading: Reading::Init,
                 writing: Writing::Init,
                 read_task: None,
@@ -70,6 +71,7 @@ where I: AsyncRead + AsyncWrite,
         }
     }
 
+    // TODO: This function only ever returns `Ok`. Change type?
     fn read_head(&mut self) -> Poll<Option<Frame<http::MessageHead<T::Incoming>, http::Chunk, ::Error>>, io::Error> {
         debug_assert!(self.can_read_head());
         trace!("Conn::read_head");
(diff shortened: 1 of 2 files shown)


seanmonstar:
I get what you mean by moving this 'state' field into the 'State' struct, but, I'm actually a little worried that it's easier to forget to set and unset the state, since other places might trigger a read. What I mean is, this PR is an example, there's more places that one needs to remember this field.

yazaddaruvala:
That was going to be my next PR, but I can make it this one. As much as possible, I want to only deal with State in `Stream` and `Sink` :)

Lets see what it looks like. If you don't like it, I'm happy to abandon it.

yazaddaruvala:
Thoughts on it now, i.e. only being set in `Stream`?

seanmonstar:
It does look "clean", but it took several minutes to verify that in fact every time `NotReady` is returned, it is was coming from the `read` operation, *except* when on `KeepAlive`. That worries me that it's almost too clever, and that if another case that returns `NotReady` is ever introduced, one could easily mess this state up.

Keeping the state of the read being blocked in the IO seems clearer, and keeps the relevant state closest to where it can occur. And it also follows what Tokio does: the `TcpStream` in Tokio keeps track of its own blocked state internally, too.

yazaddaruvala:
Interesting, I was actually curious if the "*except* when on `KeepAlive`" was an accidental inconsistency.

No worries, I just sorta imagined that `read_blocked` was a legacy name, and it should have been `stream_blocked`. But since that is not the case, and if future code paths could return `NotReady` but are not supposed to set `read_blocked` then I agree, this is not the correct approach.

Issue #838, opened 2016-06-20
Change 100-continue behavior to send when Body has been polled


bozaro:
Use of the 100 (Continue) Status (https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.2.3):

> The purpose of the 100 (Continue) status (see section 10.1.1) is to allow a client that is sending a request message with a request body to determine if the origin server is willing to accept the request (based on the request headers) before the client sends the request body. In some cases, it might either be inappropriate or highly inefficient for the client to send the body if the server will reject the message without looking at the body.

In my case, for example, I can upload file or reject already uploaded file in one atomic request.

Current state:
- 0.9.x - handler can make desigion and allow/reject client request by additional `check_continue` method. This method is not supported by https://github.com/iron/iron and https://github.com/nickel-org/nickel.rs. Also this method require duplicate logic (it not called on HTTP/1.0) for routing and check permissions;
- master - looks like this feature is forgotten.

I think, much better to send "100 Continue" status on first Body::read() call (similar logic implemented by https://github.com/frewsxcv/tiny-http project).

In this case no additional logic/callback needed for make desigion about "100 Continue":
- If you call Body::read() method, then you needs to request body and hyper send to client "100 Continue";
- If you have anought information to send final status without body, then hyper send status without "100 Continue".

I try to change logic on 0.9.x branch by send "100 Continue" status of first read call, but I failed :(

seanmonstar:
> master - looks like this feature is forgotten.

Not so much forgotten, I just wanted to come up with a better design. What you outline here seems like it should work on master: 

```rust
fn on_request(&mut self, req: Request) -> Next {
    match request.headers().get() {
        Some(&ContentLength(len)) if len > 10_000_000 => {
            self.status = StatusCode::PayloadTooLarge;
            Next::write()
        },
        None => {
            self.status = StatusCode::LengthRequired;
            Next::write()
        },
        Some(&ContentLength(len)) => {
            self.length = len;
            // hyper looks for Expect-100, and sends 100 status because of this read()
            Next::read()
        }
    }
}
```

bozaro:
Yes, about this behavior and I had in mind.

I'm sorry if offended. Loss (sometimes temporary) features on full code redesign is a normal and I did not mean anything bad. On the contrary, from my point of view this is a very good time to remind about corner cases :)

seanmonstar:
Not offended at all! Thanks for helping think this issue through.

martell:
[...]

PR #1232, opened 2017-06-25
feat(server): Handle 100-continue

sfackler:
cc #838

### DIFF
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -4,7 +4,7 @@ use std::fmt;
 
 use bytes::BytesMut;
 
-use header::{Connection, ConnectionOption};
+use header::{Connection, ConnectionOption, Expect};
 use header::Headers;
 use method::Method;
 use status::StatusCode;
@@ -55,6 +55,10 @@ pub type ResponseHead = MessageHead<RawStatus>;
 impl<S> MessageHead<S> {
     pub fn should_keep_alive(&self) -> bool {
         should_keep_alive(self.version, &self.headers)
+    }
+
+    pub fn expecting_continue(&self) -> bool {
+        expecting_continue(self.version, &self.headers)
     }
 }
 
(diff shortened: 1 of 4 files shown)


sfackler:
The 100 Continue is always sent immediately because tokio-proto tries to load the first bit of the body immediately, but that's better than nothing!

martell:
> The 100 Continue is always sent immediately because tokio-proto tries to load the first bit of the body, but that's better than nothing!
@sfackler that is the reason we closed #1040

sfackler:
I don't agree with that rationale. Even blindly returning `100 Continue` is way better than not. Without this change, `curl http://localhost:1337/echo -F foo=bar` takes 1.01 seconds against the example echo server since it sits around for a second waiting for a `100 Continue` that never comes.

src/http/mod.rs
```diff
@@ -119,6 +123,17 @@ pub fn should_keep_alive(version: HttpVersion, headers: &Headers) -> bool {
     ret
 }
 
+/// Checks if a connection is expecting a `100 Continue` before sending its body.
+#[inline]
+pub fn expecting_continue(version: HttpVersion, headers: &Headers) -> bool {
```
seanmonstar: Could we get a test case at the bottom, like there is for `should_keep_alive`?

sfackler:
Yeah, I was trying to figure out the right way to test this. Client support is something I'd love to have, but it's a bit more complex than the server side - you have to wait on the continue with a timeout in case the server doesn't know how to handle `Expect: 100-continue`, and resend the headers without the `Expect` if the server returns a 417. Manually driving the client side against a raw TcpStream seems like a decent option.

sfackler:
Updated with tests

seanmonstar:
Thanks!

martell:
Awesome now we have some support in tree. :)

@seanmonstar last week you said on irc that there might be something hyper can do about the buffering. Can we get an outline of your thoughts on that maybe in the discussion on #838 ?
I will also link the relevant different implementation attempts carl tried for the futures crate.

seanmonstar:
My thoughts are really short, probably flawed, and untested. Essentially, record some `state Polled { NotPolled, TokioBufferedPolled, UserPolled }`, updating in the `Conn::poll` method. Such that, on the 2nd poll, we now believe `UserPolled`, and would then respond with a 100-continue.

This might not work at all. It's possible that the buffer in Tokio will know better, and if the user polls after the buffer has, and the `Conn` hasn't notified that the task is ready for reading, it will short circuit and not poll the Conn again. I have no idea.

sfackler:
Thanks!

PR #1244, opened 2017-07-06
feat(server): Provide reference to Response body

### LINKED ISSUE #1216: Determine if Response body is populated
I was looking through the api provided for `Response` for a way to determine if `body` has been populated (or not) by some other part of the application. I can't immediately see a way to do this however.

I note that `body` is an `Option` per https://github.com/hyperium/hyper/blob/master/src/http/response.rs#L15 so simply being able to access `Option::is_some` would suit my use case.

Alternatively `Response` could be extended with a `bool` that could be used to indicate by application code that the current response is considered to be 'committed' and probably should not be further altered.

I'd be happy to provide a PR for either (or both) but I wanted to find out if this would be palatable before proceeding.

bradleybeddoes:
The Response body was previously inaccessible to application code. This
change allows applications that may wish to alter the Response based on
previous creation (or not) of the body to do so.

n.b. discussion in #1216 indicates this function may change in
0.12 or be removed in favour of some other api.

Closes #1216.

### DIFF
--- a/src/http/response.rs
+++ b/src/http/response.rs
@@ -93,6 +93,10 @@ impl<B> Response<B> {
         self.set_body(body);
         self
     }
+
+    /// Read the body.
+    #[inline]
+    pub fn body_ref(&self) -> Option<&B> { self.body.as_ref() }
 }
 
 impl Response<Body> {


seanmonstar:
Hm, I was just about to merge and figure that we'd deal with body types in a 0.12 release, and did a final look over a noticed that, it seems `Response` at least doesn't actually expose really anywhere that there's an `Option` inside, right? So, while `Request` would need new methods, `Response` could perhaps be fixed...?

bradleybeddoes:
Sure. What did you have in mind, re: fixing `Response` here?

seanmonstar:
Crap, I think the `Response::new` and `Response::default` not having a bounds on `B` means the `Option` has to stay...

seanmonstar:
Yea, it can't really be changed in 0.11, so may as well add things like this to allow inspection. In 0.12, we'll have to make some breaking changes so that the the internal `Option` can be removed, and then an optional body would be up to the user, where `Response<Option<Body>>` vs `Response<Body>`.

Sorry for futzing around so long.

PR #1249, opened 2017-07-07
refactor(http): replace unsafe read with AsyncRead::read_buf

yazaddaruvala:

Disclaimer: Removes unsafe

Differences between:
- Hyper unsafe impl
  - Does not zero out the unused bytes of the buffer
    - (zeros them earlier when reserving capacity)
  - Uses `.read`
- The default impl in [AsyncRead](https://docs.rs/tokio-io/0.1/src/tokio_io/lib.rs.html#131-156)
  - Zeros out the unused bytes of the buffer.
  - Uses `.read`
- The [`impl AsyncRead for TcpStream`](https://docs.rs/tokio-core/0.1.8/src/tokio_core/net/tcp.rs.html#588-615) (i.e. the one used by Hyper Server)
  - Does not zero out anything
  - Does not use `.read`
    - Uses `.read_bufs`

I haven't dug too deeply into the difference between `.read` and `.read_bufs` but the benches are within their usual variance (at least on my computer).

Thoughts?

### DIFF
--- a/src/http/io.rs
+++ b/src/http/io.rs
@@ -89,6 +89,7 @@ impl<T: AsyncRead + AsyncWrite> Buffered<T> {
     fn read_from_io(&mut self) -> Poll<usize, io::Error> {
         use bytes::BufMut;
         // TODO: Investigate if we still need these unsafe blocks
+        //    Note: `AsyncRead::read_buf` also zeros memory if needed
         if self.read_buf.remaining_mut() < INIT_BUFFER_SIZE {
             self.read_buf.reserve(INIT_BUFFER_SIZE);
             unsafe { // Zero out unused memory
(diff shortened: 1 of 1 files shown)


yazaddaruvala:
For more context I understand that [`.read_bufs`](https://carllerche.github.io/mio/mio/net/struct.TcpStream.html#method.read_bufs) corresponds to a `readv` syscall, but I have no idea what the performance characteristics of using that syscall vs `.read` would be.

seanmonstar:
It using `readv` in this case should be nearly identical. The change hasn't been made I guess because `read_buf` didn't exist when this part of hyper was written. I also didn't want bad implementations of `AsyncRead` to zero the memory over and over. I guess we could just say "that's a bad transport", but I can still think people could blame hyper...

yazaddaruvala:
I have a love hate relationship with generics. At lease as far as I can tell everywhere in Hyper. I.e. Server and HttpConnector we use TcpStream.

There is a possibility that Config currently allows people to add a connector which is not HttpConnector. There is also a possibility someone doesn't use hyper::Server but then they are using the __Proto proxies. 


Couple options: 1. "Plug those holes" by making Conn explicitly use a TcpStream. 2. Make the creation of a Connector unsafe. 3. don't use slow transports - add documentation.

I prefer option 1. but this does slightly increase the commitment that hyper will only use tokio transports.

Option 1 does also deprecate what I imagine was a use case for having custom Connectors.

seanmonstar:
The reason for being generic is that HTTP can be over any transport. That could be TCP, TCP protected by TLS, Unix Domain Sockets, UDP, or anything else. It's a good thing!

The answer is probably to just do what in this PR...

yazaddaruvala:
Cool, let me know if there is anything you'd like me to tweak.

seanmonstar:
I tried out bench this branch against master also, and saw a small (30,000 ns vs 31,000 ns, 5x on each branch) slow down in this branch. I've looked into the impls of `AsyncRead` and `BufMut` and things, and I don't have a good guess as why it seems slower.

yazaddaruvala:
Yeah I mean.. the benches are weird sometimes. I've found they run faster or slower if I switch tabs in the same terminal. But its all good, if its slower its slower.

PR #1250, opened 2017-07-10
refactor(header): remove GetType trait

dmizuk:
Commit 7d122bd15fb354c427c40986c28e292f9c65f54e removed usage of `header::sealed::GetType`, so the trait can safely be removed.


seanmonstar:
Oh, good catch! I hadn't even noticed.

PR #1251, opened 2017-07-10
refactor(http): remove custom Cursor in favor of std::io::Cursor

yazaddaruvala:

Replacing the custom Cursor with the std Cursor was pretty straight forward. The only ergonomic loss was around `DebugTruncate` usage, and now we have the foundation to use get more ergonomic wins by using `Buf(Mut)` traits more (like in `Encoder::encode` and `Buffered::buffer`).

### DIFF
--- a/src/http/conn.rs
+++ b/src/http/conn.rs
@@ -1,7 +1,8 @@
 use std::fmt;
-use std::io::{self, Write};
+use std::io::{self, Cursor, Write};
 use std::marker::PhantomData;
 
+use bytes::Buf;
 use futures::{Poll, Async, AsyncSink, Stream, Sink, StartSend};
 use futures::task::Task;
 use tokio_io::{AsyncRead, AsyncWrite};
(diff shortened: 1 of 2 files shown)


seanmonstar:
Well, that's surprising. It turns out using the `std::io::Cursor` does actually make things slower. The `bench_server_tcp_thoughput_fixedsize_small_payload` is what I ran, 5 times on master and 5 times with this branch.

- master: ~30,000 ns/iter
- branch: ~32,000 ns/iter

I didn't grab a profiler to try to determine why, but some possibilities: several of the methods on `std::io::Cursor` are not `#[inline]`, such as `position`, `set_position`, `read`, `get_ref`, `get_mut`, so that requires a function call for many things that really are just accessing a field. The in-crate `Cursor` of hyper is able to have all those function calls optimized away...

seanmonstar:
Due to the performance regressions in the previous comment, I'm going to close this. I'm sorry! It seemed like a clear win, in terms of cleaner code. Maybe with some `#[inline]`s added in to std, this change could be landed in the future?

yazaddaruvala:
Lol nothing to be sorry about, the impl in std clearly needs some love. Might be a good first commit.

PR #1255, opened 2017-07-12
feat(client): add connect, read, and write timeouts to client

### LINKED ISSUE #1234: Add connect, read, and write timeouts to client

hjr3:
- Add a `TimeoutConnector` implememtation that can be used to wrap any
  implemention of `Connect`.
- Add read and write timeout support to `Client` and client `Config`.
  The internal client implementation now wraps I/O with the
  tokio-io-timeout crate. Due to the way tokio-proto works, the user
  will see a "broken pipe" error instead of a "timed out" error when
  a read or write timeout occurs.

Closes #1234

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ tokio-core = "0.1.6"
 tokio-proto = "0.1"
 tokio-service = "0.1"
 tokio-io = "0.1"
+tokio-io-timeout = "0.1.0"
 unicase = "2.0"
 
 [dev-dependencies]
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -3,10 +3,12 @@ use std::fmt;
 use std::io;
 use std::mem;
 //use std::net::SocketAddr;
+use std::time::Duration;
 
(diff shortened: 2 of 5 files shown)


seanmonstar:
It looks like this can work entirely out-of-crate? I kind of lean towards keeping it that way. Is there something that having this in crate allows that wouldn't be possible otherwise?

As a side note, looking at the tokio-io-timeout crate, the implementations of `AsyncRead` and `AsyncWrite` don't pass forward the `read_buf` and `write_buf` methods, which would mean readv and writev are no longer used for `TcpStream`.

hjr3:
@seanmonstar `TimeoutConnector` can be out of crate. I don't see how the read and write timeouts would be out of crate. That being said, I was looking at other issues where you were talking about the `ClientProto` onion. Maybe that is a better way to introduce some of this functionality.

seanmonstar:
Can't the read and write timeouts be set in any connector by returning a `TimeoutStream`?

hjr3:
Of course! Not sure why I did not think of that earlier. Ok, I will make a separate crate and add all of the logic there.

PR #1256, opened 2017-07-13
feat(headers): Add missing headers

brendanzab:
These seem to have been missed! I renamed `LastEventID` and `TE` to `LastEventId` and `Te` respoectively to follow Rust's capitalisation conventions.


brendanzab:
I get the following test error locally:

```
thread 'header::common::last_event_id::test_last_event_id::test1' panicked at 'called `Result::unwrap()` on an `Err` value: Header', src/libcore/result.rs:860:4
```

Any ideas?

seanmonstar:
Oh, wow, woops :D

For the failing test, it is because the `header!` implementation for a single value doesn't allow for an empty header. It'd probably need a manual `impl Header` instead of using the macro (it's not common for a header to be allowed to have an empty value).

It seems the tests were never running, since the module was never used...

brendanzab:
Thanks! Hopefully it's fixed now!

src/header/common/last_event_id.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// # Example
+/// ```
+/// use hyper::header::{Headers, LastEventId};
+///
+/// let mut headers = Headers::new();
+/// headers.set(LastEventId("1".to_owned()));
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+pub struct LastEventId(pub String);
```
brendanzab: I wasn't quite sure if you wanted this to be a `Cow`, but what I decided on based on the other things here...

seanmonstar: This would likely be dynamic right? As in, at compile time, how often you know what the last event that was sent on a connection? Probably never?

So then `String` seems fine... Although, is the ID frequently a string? Would people usually just generate some random hex bytes for the ID? Or is often a number?

brendanzab: No idea. Looked on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers), [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields), and [IANA](https://www.iana.org/assignments/message-headers/message-headers.xhtml) but couldn't find a reference to this header. What list did you use to figure out what to include?

seanmonstar:
I looked in HTML spec, and its referred to as a 'last-event-id string', so a `String` seems fine. Thanks for fixing these!

Issue #1235, opened 2017-06-26
Handle wildcard character for AccessControlAllowMethods


KiChjang:
The [fetch specification](https://fetch.spec.whatwg.org/#http-access-control-allow-methods) has defined a new ABNF form for the header as follows:
```
Access-Control-Allow-Methods     = #method / wildcard
wildcard                         = "*"
```

whereas hyper currently supports the following ABNF:
```
Access-Control-Allow-Methods: "Access-Control-Allow-Methods" ":" #Method
```

seanmonstar:
For now, the new version of hyper won't be including the typed headers directly, so I'm closing this here as a won't fix.

Issue #189, opened 2014-12-10
Migrate hyper::header to its own crate


DiamondLovesYou:
`rust-ppapi` maintainer here. `rust-ppapi` needs to be able to handle header formats correctly, but doesn't need all of Hyper (ie doesn't need the server + client parts). I'd like to have `rust-ppapi` just depend on the header portions of hyper, because the Pepper API has it's own http client API.

pyfisch:
It looks to me like Cargo got support for sub-crates: http://doc.crates.io/guide.html#path-dependencies

I propose to split Hyper into `hyperprotocol`, `hyperclient`, `hyperserver` and maybe `hypernet`.

jimmycuadra:
Would there be any disadvantages to extracting the status and header modules into their own crates? What about giving them generic (non-"hyper") names? It'd be great to have generic types for these concepts in the Rust ecosystem (like we have the url crate) that could be used by downstream crates without having to tie their APIs to anything specific to Hyper.

mikedilger:
Case in point: my email/mime creation crate depends on hyper, and it has nothing to do with HTTP.

I don't think splitting it up into lots of tiny crates makes sense, but headers/mime should really be separate.

brendanzab:
Yeah, looking at the [IANA list of headers](https://www.iana.org/assignments/message-headers/message-headers.xhtml), it seems that headers span more than just http. I can also see that we are missing many of them! It would be nice to see if we could split them out and do some more rgourous testing of our implementation coverage, perhaps even testing to see how many we implement based on their supplied CSV file.

seanmonstar:
This will be done as part of the 0.12 release. When the rest of the types are replaced with `http` types, that means things by default will be using `http::HeaderMap`.

This will mean a change to the API is likely needed, perhaps as just a way to get and set typed values by taking `&/&mut HeaderMap`.

Also needs a name. Some ideas:

- headers (currently [exists](https://crates.io/crates/headers), but looks like it may have just been an experiment. Perhaps Ashley would be interested in a transfer?)
- helmet (protecting your head(ers)?)
- visor (enhancing your head(ers)?)

mehcode:
My vote would be `http_headers` and then `http/headers` if/when Cargo supports sub-crates.

seanmonstar:
`http_headers` is also [currently claimed](https://crates.io/crates/http_headers) (perhaps abandoned?).

tshepang:
[...]

PR #1268, opened 2017-07-18
fix(headers): Handle wildcard character for Access-Control-Allow-Methods

LegNeato:
Change the header `Access-Control-Allow-Methods` to handle the wildcard character (`*`).
In the process, make a value required to match the spec.

Closes https://github.com/hyperium/hyper/issues/1235.
BREAKING CHANGE: The `Access-Control-Allow-Methods` header now requires a value.
Accordingly, `AccessControlAllowMethods` changed to an enum from a function.

### DIFF
--- a/src/header/common/access_control_allow_methods.rs
+++ b/src/header/common/access_control_allow_methods.rs
@@ -10,7 +10,8 @@ header! {
     ///
     /// # ABNF
     /// ```plain
-    /// Access-Control-Allow-Methods: "Access-Control-Allow-Methods" ":" #Method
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for this! As you've noticed it being a breaking change, I'd like to have this change occur with any other breaking changes. So I'd probably set this as a part of 0.12, and not merge until we know all the changes we want to make.

LegNeato:
Sounds good, no rush...I basically just did this to play around with Rust a little more and get familiar with the codebase.

seanmonstar:

PR #1270, opened 2017-07-19
feat(client): add client example with gzip and https support

king6cong:
this example supports https and gzip.
If gzip is used, information about compression ratio will be displayed.

I think this will be a helpful demo for real world usages nowadays.


seanmonstar:
I actually feel like doing all this is out of scope for examples in the hyper repo. There is an HTTPS guide on the hyper.rs website, though.

king6cong:
Yeah. Better leave for libraries like reqwest.

PR #1276, opened 2017-07-27
feat(http): add map_body to Request and Response

spinda:
Add Request::map_body and Response::map_body to allow for swapping out the body
with a new version based on the old one -- wrapping a stream combinator around
it, for example.

Motivating scenario: I have a `Service` combinator which wraps around another `Service`. The inner `Service` returns a `Response` whose body type is a stream of messages. The outer `Service` needs to wrap a serializer around this body type to serialize its output to binary chunks, before turning the resulting `Response` over to hyper.

### DIFF
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -98,6 +98,20 @@ impl<B> Request<B> {
     #[inline]
     pub fn set_body<T: Into<B>>(&mut self, body: T) { self.body = Some(body.into()); }
 
+    /// Set the body and move the Request, using the previous body as context.
+    #[inline]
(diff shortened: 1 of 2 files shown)


seanmonstar:
We've been working on a new implementation of `Request` and `Response`, which includes a way to separate the head and body parts, and then construct a new one. That'd allow you to achieve the same result, right? We found that preferable to a `map`, since the naming sounds off.

spinda:
That sounds great, and I also wrestled with the naming of this function. Feel free to close this then.

PR #1279, opened 2017-07-31
docs(headers): clarify Basic auth

jethrogb:
See https://tools.ietf.org/html/rfc7617

### DIFF
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -140,7 +140,8 @@ pub struct Basic {
     /// The username as a possibly empty string
     pub username: String,
     /// The password. `None` if the `:` delimiter character was not
-    /// part of the parsed input.
+    /// part of the parsed input. Note: A compliant client MUST
+    /// always send a password (which may be the empty string).
     pub password: Option<String>
 }
 


seanmonstar:
Is this actually true? I looked around in RFC 7617, and couldn't find text the stated this. In fact, I suspect it's common for clients to convert a request to `http://user@example.org` into the a basic auth header of `base64("user")`.

jethrogb:
The RFC doesn't leave room for not including the password:
```
   To receive authorization, the client

   1.  obtains the user-id and password from the user,

   2.  constructs the user-pass by concatenating the user-id, a single
       colon (":") character, and the password,

   3.  encodes the user-pass into an octet sequence (see below for a
       discussion of character encoding schemes),

   4.  and obtains the basic-credentials by encoding this octet sequence
       using Base64 ([RFC4648], Section 4) into a sequence of US-ASCII
       characters ([RFC0020]).
```

seanmonstar:
Ah, I see that the code in this type also does that. So then the point is that when decoding, it might not include a colon. Does it make sense to have an `Option` at all? Or should it actually fail parsing if there is no semi colon?

Probably depends on real world usage, instead of exaclty what the spec does.

jethrogb:
> Does it make sense to have an Option at all? Or should it actually fail parsing if there is no semi colon?

What is the behavior for other headers? I thought it'd be ok to be lenient in parsing and just document proper behavior.

PR #1282, opened 2017-08-06
feat(headers): add `rel=prefetch`

kpcyrd:
Add `rel=prefetch` support for [link headers][1].

This is needed for server hints.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ

### DIFF
--- a/src/header/common/link.rs
+++ b/src/header/common/link.rs
@@ -85,6 +85,9 @@ pub struct LinkValue {
     /// Context IRI: `anchor`.
     anchor: Option<String>,
 
+    /// Specify the target context that will consume the response
+    as_: Option<String>,
+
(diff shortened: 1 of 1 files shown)


kpcyrd:
Since extending an enum is a breaking change, I've added a couple more things that are related to preloading. Most of it is taken from here: https://www.w3.org/TR/preload/

The `as` target might need some more work, it's currently a string that could be replaced with an enum as well. Valid values are :
 
`audio`, `video`, `track`, `script`, `style`, `font`, `image`, `fetch`, `worker`, `embed`, `object`, `document`.

Feedback welcome. :)

seanmonstar:

PR #1287, opened 2017-08-12
feat(server): support upgrading to other protocols 

spinda:
The Http type learns a new method, bind_upgradable_connection, which returns a Future. This method behaves like bind_connection, except the Service supplied must return an UpgradableResponse containing either an HTTP Response or some information describing a protocol switch. If an Upgrade response is returned, the server task is shut down (without closing the underlying io) and this information is forwarded to the Future, along with the underlying io and any remaining buffered data that has already been read from it.

If the server task shuts down normally, the Future completes with a None output value. In the case of an error handling the connection, the error is also forwarded to the Future.

This allows for implementing servers with hyper that can upgrade their connections to different protocols, such as WebSockets, in response to client requests. Such protocols can be implemented outside of hyper itself. The API is designed to allow for potentially multiple upgrade paths out of an HTTP exchange, with the upgrade information attached to the UpgradableResponse::Upgrade signal determining which is used.

A companion crate implementing WebSocket upgrades can be found [here](https://github.com/spinda/hyper-websocket), with sample code [here](https://github.com/spinda/hyper-websocket/blob/416ca8588931aa92a737c9c1a4afef2bef5d24b6/tests/test.rs).

Importantly, the code path for normal, non-upgradable connections is entirely untouched by this addition.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -354,9 +354,9 @@ impl<B> Into<Message<__ProtoResponse, B>> for Response<B> {
     }
 }
 
-impl<B, P> Into<Message<Result<__ProtoResponse, P>, B>> for UpgradableResponse<P, B> {
+impl<B, P> Into<Message<UpgradableResponseHead<P>, B>> for UpgradableResponse<P, B> {
     #[inline]
-    fn into(self) -> Message<Result<__ProtoResponse, P>, B> {
+    fn into(self) -> Message<UpgradableResponseHead<P>, B> {
         match self {
             UpgradableResponse::Response(res) => {
                 let (head, body) = response::split(res);
@@ -366,7 +366,11 @@ impl<B, P> Into<Message<Result<__ProtoResponse, P>, B>> for UpgradableResponse<P
                     Message::WithoutBody(Ok(__ProtoResponse(head)))
                 }
             }
-            UpgradableResponse::Upgrade(upgrade_info) => Message::WithoutBody(Err(upgrade_info)),
+            UpgradableResponse::Upgrade(upgrade_info, maybe_head) => {
+                Message::WithoutBody(Err((upgrade_info, maybe_head.map(|head| {
+                    __ProtoResponse(response::split(head).0)
+                }))))
+            }
         }
     }
 }
@@ -408,6 +412,8 @@ type UpgradablePayload<T, B, P> = Result<(__ProtoTransport<T, B>, P), ::Error>;
 type UpgradableReceiver<T, B, P> = Receiver<UpgradablePayload<T, B, P>>;
 type UpgradableSender<T, B, P> = Sender<UpgradablePayload<T, B, P>>;
 
+type UpgradableResponseHead<P> = Result<__ProtoResponse, (P, Option<__ProtoResponse>)>;
+
 impl<T, B, P> ServerProto<(T, UpgradableSender<T, B, P>)> for UpgradableHttp<P, B>
     where T: AsyncRead + AsyncWrite + 'static,
           B: AsRef<[u8]> + 'static,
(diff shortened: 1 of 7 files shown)


seanmonstar:
This looks really impressive! I apologize I haven't had time to review it, I've been trying to wrap up changes to the `http` crate before RustConf. I'll take a look soon, I promise!

spinda:
No problem, and thanks!

spinda:
Hi again, have you had a chance to look at this since RustConf?

seanmonstar:
@spinda I'm so sorry, I don't have particularly good reasons other than forgetting. I'd like to look at this this week. I think upgrading protocols is necessary, and so hyper should definitely support it.

cc @carllerche

src/server/mod.rs
```diff
@@ -336,6 +405,381 @@ impl<T, B> Service for HttpService<T>
...
+    _marker: PhantomData<P>,
+}
+
+type UpgradablePayload<T, B, P> = Result<(__ProtoTransport<T, B>, P), ::Error>;
+type UpgradableReceiver<T, B, P> = Receiver<UpgradablePayload<T, B, P>>;
+type UpgradableSender<T, B, P> = Sender<UpgradablePayload<T, B, P>>;
+
+type UpgradableResponseHead<P> = Result<__ProtoResponse, (P, Option<__ProtoResponse>)>;
+
+impl<T, B, P> ServerProto<(T, UpgradableSender<T, B, P>)> for UpgradableHttp<P, B>
```
seanmonstar: That all of these boilerplate impls are needed is a bummer. Sorry to have to write that XD

tests/upgradable_server.rs
```diff
@@ -0,0 +1,242 @@
...
+        .and_then(|body| {
+            let body_str = str::from_utf8(body.as_ref()).expect("client body decode error");
+            assert_eq!(body_str, "Hello World");
+            Ok(())
+        });
+    core.run(test).expect("client body read error");
+}
+
+#[test]
+fn test_upgrade() {
```
seanmonstar: ❤️

src/server/mod.rs
```diff
@@ -336,6 +405,381 @@ impl<T, B> Service for HttpService<T>
...
+/// protocol.
+///
+/// If the latter, the `Upgrade` variant should contain information pertaining
+/// to this protocol switch.
+pub enum UpgradableResponse<P = (), B = ::Body> {
+    /// An HTTP response.
+    Response(Response<B>),
+    /// A protocol upgrade signal with accompanying information and optional
+    /// "switching protocols" HTTP response head.
+    Upgrade(P, Option<Response<()>>),
```
seanmonstar: So, what exactly is the `P`? From what I can tell, is it just some arbitrary thing that might be created when inspected a `Request`, and useful once upgraded?

spinda:
Thanks for taking a look! Not sure how much time I'm going to have over the next few days (ending an internship and moving back home). I'll address these comments as soon as I can.

seanmonstar:
I'd be especially interested in discussing the design before spending time writing code :D

seanmonstar:
I put more details of the proposal I suggested in #1323.

blueridanus:
How's progress on this? Can I help?

nayato:
is this now blocked on moving from tokio-proto?

seanmonstar:
The internal change to not use tokio-proto is already merged, just not on by default while bugs are still found. It should probably be easier to make upgrades work, however, because of this work.

I'd say the blocker for this feature is settling on an API to expose to a user.

seanmonstar:
I'm closing this due to inactivity. I recommend discussing the feature more in #1323.

PR #1295, opened 2017-08-23
fix(client): don't drop connection pools with exactly one connection left

jethrogb:
Previously, the pool cleanup logic would erroneously clean up
pools that had exactly one connection left in them. This bug
was introduced in 30fb0e777815adcaa6b287ac7720570dedd77942.


seanmonstar:
Woops, thanks for fixing this!

jethrogb:
Could you do a release?

seanmonstar:
Yep, packaging now.

PR #1299, opened 2017-08-28
feat(headers): add 'Token' token support

flokli:
This adds support for static Token authorization used in some APIs.
Code is pretty similar to [Bearer token support](https://github.com/hyperium/hyper/pull/572/files).

### DIFF
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -21,6 +21,7 @@ use header::{Header, Raw};
 /// # Example values
 /// * `Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==`
 /// * `Bearer fpKL54jvWmEGVoRdCNjG`
+/// * `Token Ithoo6daph2yekoz3aen`
 ///
 /// # Examples
 /// ```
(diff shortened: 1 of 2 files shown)


flokli:
Travis complains about an unused `extern crate pretty_env_logger` in rust nighly.

Shouldn't be related to this PR.

seanmonstar:
Hey there, sorry for delay!

The Token scheme for the Authorization header seems quite uncommon. Anyone using just a simple token usually uses `Bearer` scheme instead. There's an abandoned [draft](https://tools.ietf.org/html/draft-hammer-http-token-auth-01) for a `Token` scheme, defining a way to provide HMAC request signing. The author of that draft has since made the `Hawk` scheme instead.

Would it make sense to keep a scheme (and implementation) like this outside of hyper, since it seems uncommon, and the formats seem to differ?

flokli:
@seanmonstar No problem :-)

Sure, its kind of uncommon, nevertheless I stumbled over a (private) HTTP API using this `Token` scheme for authentication, added support for this outside of hyper, and thought it might be useful for others, too (as it doesn't add that much to the codebase).

Maybe @ernestas-poskus, who  +1'd here knows a real-world example, which could justify this more, even though it might not be an official spec?

seanmonstar:
I'm going to suggest that this stay out of hyper for now, because it seems very uncommon, and implementations may differ on how servers are using this scheme anyways.

Thanks for the effort!

flokli:
Sure, I can understand that. No worries :-)

andir:
@seanmonstar while you are right that there is no standard for `Token` there are a few applications using it. For example `django-rest-framework` is using that scheme (per default, http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication). You can change the `keyword` that is being used to `Bearer` or whatever. 

I fully support NOT merging this right now, just wanted to state that there are indeed people using `Token` (as a default).

Efforts should go towards changing the default in `django-rest-framework` and other frameworks so they just stick to the standard.

PR #1318, opened 2017-09-11
refactor(uri): Change Uri to http::Uri

Aaronepower:
This change is start of moving to using the http crate's types over
hypers.

BREAKING CHANGE: Currently http::Uri doesn't impl all the same traits as
Uri, and is missing Uri::is_absolute


Aaronepower:
I can't seem to divine what is causing these tests to fail. The only difference I've found is that hyper's `Uri` doesn't have a trailing `/` with empty path and `http::Uri` does.

seanmonstar:
@Aaronepower yes, the failing tests are that the the client is sending requests with blank URIs. Bytes like `GET  HTTP/1.1\r\nHost: localhost:42805\r\n\r\n` (note there is 2 spaces between the method and version, but no URI).

I've pushed an update to master to make debugging those tests hopefully easier.

Aaronepower:
Closing since the compat feature was implemented.

Issue #1313, opened 2017-09-09
compatibility with http crate


srijs:
Now that the `http` crate is officially published, are there plans to provide a compatibility layer for hyper, or will this have to wait for hyper 0.12?

If you're open to the former, I have some glue code I'd be willing to contribute.

jimmycuadra:
I was thinking about starting on this work too. My understanding was that the plan was to replace all the existing hyper types with ones from the http crate. Is there a more specific plan or is there any code underway already? Or can anyone from the community just pick this up?

carllerche:
I believe that is the plan, but doing so would require a breaking change (thus a 0.x bump).

srijs:
I think the question might be if there is a way to introduce parts of the http crate w/o breaking the API just yet? Two possible approaches I can think of:

- Introduce conversion/compatibility methods, e.g. to be able to pass a `Service` using http types to `Http::bind` (or an equivalent compat. method).

- Replace some of the internal logic with pieces from `http` (url parsing comes to mind as an example), making existing `hyper` types a lightweight facade on top of `http` types so that the next step of replacing them for the 0.12 release would be easier (all the while reaping the benefits of less code in hyper, and getting validation on the `http` logic).

jimmycuadra:
Personally, I'd much rather see hyper's (now redundant) types disappear completely in 0.12. My impression is that people using Rust and hyper at this point understand that the HTTP stack is not stable yet, and that making breaking changes where it's clearly needed is expected in order to push us towards 1.0.

srijs:
I think we both are in agreement that that's what should happen for 0.12. Of course @seanmonstar should make that call.

Depending on what the time table around this is though, I think it might be beneficial to offer some level of compatibility before introducing 0.12.

seanmonstar:
We do expect 0.12 to remove hyper's related types, and use `http` instead. As noted, that's a breaking change, so we'd like to group that with other breaking changes that will likely need to happen, like changes to allow h2 to work.

The idea mentioned here is neat, though: providing compatibility with `http` in hyper 0.11. That could perhaps allow downstream users and frameworks to start adopting `http` sooner, and then an when hyper 0.12 releases, that part is an easier change. I like it!

seanmonstar:
See https://github.com/hyperium/hyper/issues/1317 for the breaking change issue, but we can keep discussion in here for how to provide compatibility in 0.11!

srijs:
@seanmonstar Thanks for your input!

[...]

PR #1319, opened 2017-09-12
feat(*): implement compatibility with http crate

srijs:
Design exploration for #1313. This is missing tests at the moment (working on it), but I would appreciate some feedback as to my current approach.

TODO

- [x] Implement Request/Response conversion suitable for client
- [x] Implement tests and validate round-trip-ability of conversion functions

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -30,14 +30,14 @@ use uri::{self, Uri};
 pub use http::response::Response;
 pub use http::request::Request;
 pub use self::connect::{HttpConnector, Connect};
-#[cfg(feature = "compat")]
-pub use self::compat::CompatFutureResponse;
 
 mod connect;
 mod dns;
 mod pool;
 #[cfg(feature = "compat")]
-mod compat;
+mod compat_impl;
+#[cfg(feature = "compat")]
+pub mod compat;
 
 /// A Client to make outgoing HTTP requests.
 // If the Connector is clone, then the Client can be clone easily.
@@ -117,8 +117,14 @@ where C: Connect,
     /// Send an `http::Request` using this Client.
     #[inline]
     #[cfg(feature = "compat")]
-    pub fn request_compat(&self, req: http_types::Request<B>) -> CompatFutureResponse {
-        self::compat::future(self.call(req.into()))
+    pub fn request_compat(&self, req: http_types::Request<B>) -> compat::CompatFutureResponse {
+        self::compat_impl::future(self.call(req.into()))
+    }
+
+    /// Convert into a client accepting `http::Request`.
+    #[cfg(feature = "compat")]
+    pub fn into_compat(self) -> compat::CompatClient<C, B> {
+        self::compat_impl::client(self)
     }
 }
 
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -9,6 +9,8 @@ use std::time::Duration;
 
 use futures::{future, Poll, Async, Future, Stream};
 use futures::unsync::oneshot;
+#[cfg(feature = "compat")]
+use http_types;
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio::reactor::Handle;
 use tokio_proto::BindClient;
@@ -28,10 +30,14 @@ use uri::{self, Uri};
 pub use http::response::Response;
 pub use http::request::Request;
 pub use self::connect::{HttpConnector, Connect};
+#[cfg(feature = "compat")]
+pub use self::compat::CompatFutureResponse;
 
 mod connect;
 mod dns;
 mod pool;
+#[cfg(feature = "compat")]
+mod compat;
 
 /// A Client to make outgoing HTTP requests.
 // If the Connector is clone, then the Client can be clone easily.
@@ -106,6 +112,13 @@ where C: Connect,
     #[inline]
     pub fn request(&self, req: Request<B>) -> FutureResponse {
         self.call(req)
+    }
+
+    /// Send an `http::Request` using this Client.
+    #[inline]
+    #[cfg(feature = "compat")]
+    pub fn request_compat(&self, req: http_types::Request<B>) -> CompatFutureResponse {
+        self::compat::future(self.call(req.into()))
     }
 }
 
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -571,7 +571,7 @@ impl From<Headers> for http_types::HeaderMap {
                 .expect("attempted to convert invalid header name");
             let value = http_types::header::HeaderValue::from_bytes(header.value_string().as_bytes())
                 .expect("attempted to convert invalid header value");
-            header_map.insert(name, value);
+            header_map.append(name, value);
         }
         header_map
     }
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -77,8 +77,13 @@
 //! }
 //! ```
 use std::borrow::{Cow, ToOwned};
+#[cfg(feature = "compat")]
+use std::convert::From;
 use std::iter::{FromIterator, IntoIterator};
 use std::{mem, fmt};
+
+#[cfg(feature = "compat")]
+use http_types;
 
 use unicase::Ascii;
 
@@ -543,6 +548,32 @@ impl fmt::Debug for Headers {
         f.debug_map()
             .entries(self.iter().map(|view| (view.0.as_ref(), ValueString(view.1))))
             .finish()
+    }
+}
+
+#[cfg(feature = "compat")]
+impl From<http_types::HeaderMap> for Headers {
+    fn from(header_map: http_types::HeaderMap) -> Headers {
+        let mut headers = Headers::new();
+        for (name, value) in header_map.iter() {
+            headers.set_raw(name.as_str().to_string(), value.as_bytes());        
+        }
+        headers
+    }
+}
+
+#[cfg(feature = "compat")]
+impl From<Headers> for http_types::HeaderMap {
+    fn from(headers: Headers) -> http_types::HeaderMap {
+        let mut header_map = http_types::HeaderMap::new();
+        for header in headers.iter() {
+            let name = http_types::header::HeaderName::from_bytes(header.name().as_bytes())
+                .expect("attempted to convert invalid header name");
+            let value = http_types::header::HeaderValue::from_bytes(header.value_string().as_bytes())
+                .expect("attempted to convert invalid header value");
+            header_map.insert(name, value);
+        }
+        header_map
     }
 }
 
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -553,10 +553,16 @@ impl fmt::Debug for Headers {
 
 #[cfg(feature = "compat")]
 impl From<http_types::HeaderMap> for Headers {
-    fn from(header_map: http_types::HeaderMap) -> Headers {
+    fn from(mut header_map: http_types::HeaderMap) -> Headers {
         let mut headers = Headers::new();
-        for (name, value) in header_map.iter() {
-            headers.set_raw(name.as_str().to_string(), value.as_bytes());        
+        for (name, mut value_drain) in header_map.drain() {
+            if let Some(first_value) = value_drain.next() {
+                let mut raw: Raw = first_value.as_bytes().into();
+                for value in value_drain {
+                    raw.push(value.as_bytes());
+                }
+                headers.append_raw(name.as_str().to_string(), raw);        
+            }
         }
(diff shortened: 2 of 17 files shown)


src/lib.rs
```diff
@@ -21,6 +21,8 @@ extern crate base64;
 extern crate bytes;
 #[macro_use] extern crate futures;
 extern crate futures_cpupool;
+#[cfg(feature = "compat")]
+extern crate http as http_types;
```
seanmonstar: I'd eventually move the `hyper::http` module to be `hyper::proto`, most likely. That'd allow using the `http` namespace for the crate. I realize fixing all the renames is tedious, so I'm not saying it's required, I can get to it eventually :)

src/method.rs
```diff
@@ -156,6 +159,24 @@ impl Default for Method {
     }
 }
 
+#[cfg(feature = "compat")]
+impl From<http_types::Method> for Method {
+    fn from(method: http_types::Method) -> Method {
+        method.as_ref().parse()
```
seanmonstar: These `Method` conversions can probably skip `parse` in most cases, and just match on the common variants, so the cost should be minuscule.

src/http/request.rs
```diff
@@ -132,6 +135,24 @@ impl<B> fmt::Debug for Request<B> {
     }
 }
 
+#[cfg(feature = "compat")]
+impl From<Request> for http_types::Request<Body> {
+    fn from(req: Request) -> http_types::Request<Body> {
+        let mut builder = http_types::request::Builder::new();
```
seanmonstar: By using `From` on all the types, we can skip the builder, and use the `http::Request` directly. I say that just because for a conversion like this, performance would be crucial, and so skipping the error checking when we know there's no errors will be noticeable.

src/status.rs
```diff
@@ -596,6 +599,22 @@ impl From<StatusCode> for u16 {
     }
 }
 
+#[cfg(feature = "compat")]
+impl From<http_types::StatusCode> for StatusCode {
+    fn from(status: http_types::StatusCode) -> StatusCode {
+        StatusCode::try_from(status.as_u16())
```
seanmonstar: Thanks to `hyper::StatusCode::Unregistered` being able to have any stupid `u16`, this is likely the best approach!

src/uri.rs
```diff
@@ -315,6 +318,22 @@ impl Display for Uri {
     }
 }
 
+#[cfg(feature = "compat")]
+impl From<http_types::Uri> for Uri {
+    fn from(uri: http_types::Uri) -> Uri {
+        uri.to_string().parse()
```
seanmonstar: The `to_string` is unfortunate, but it doesn't look like there's a good way around it, as `hyper::Uri` provides `as_ref` for `&str`, requiring the holding of 1 contiguous buffer.

... Hm, though, `http::Uri` is backed by `Bytes`, just like `hyper::Uri`. I haven't looked to remind myself, but I wonder if there's a combination of methods that can be called to try to prevent a full alloc and copy?

src/version.rs
```diff
@@ -58,6 +61,39 @@ impl Default for HttpVersion {
...
+        match v {
+            HttpVersion::Http09 =>
+                http_types::Version::HTTP_09,
+            HttpVersion::Http10 =>
+                http_types::Version::HTTP_10,
+            HttpVersion::Http11 =>
+                http_types::Version::HTTP_11,
+            HttpVersion::H2 =>
+                http_types::Version::HTTP_2,
+            _ => panic!("attempted to convert unexpected http version")
```
seanmonstar: The `HttpVersion::__DontMatchMe` variant can be used here, so we know it shouldn't ever exist, and then use `unreachable!` instead of `panic!`.

src/server/mod.rs
```diff
@@ -103,6 +113,18 @@ impl<B: AsRef<[u8]> + 'static> Http<B> {
         })
     }
 
+    /// Bind a `NewService` using types from the `http` crate.
+    ///
+    /// See `Http::bind`.
+    #[cfg(feature = "compat")]
+    pub fn bind_compat<S, Bd>(&self, addr: &SocketAddr, new_service: S) -> ::Result<Server<compat::NewCompatService<S>, Bd>>
```
seanmonstar: 👍

src/lib.rs
srijs: I'm happy todo that as a smaller PR if you don't mind!

src/uri.rs
srijs: I'll see if I can come up with something. Maybe there's a feature we can land in `http` to enable a more performant conversion here?

src/version.rs
srijs: The problem is that hyper has made `H2` and `H2c` two distinct values of `HttpVersion`, rather than conveying non-cleartext/cleartext via some other channel. `http` on the other hand does not make the distinction, which is why this match is not exhaustive and the function therefore partial.

src/http/request.rs
srijs: Done in 5eb16616a081661d3fbae716e8ce83dc6840f99c

src/version.rs
seanmonstar: That's a good point, I had forgotten about that variant. Woops.

src/uri.rs
srijs: I've explored this a bit:

- Converting from `hyper::Uri` to `http::Uri` can be done without copying the underlying data. In fact, I've included this in my changes to the branch.

- Converting  from `http::Uri` to `hyper::Uri` is not easily possible, however, since `http` "chops" the `Bytes` into different parts. If allocations here really become a problem, I guess we could consider changing `http::Uri` to keep a reference to the "top-level" `Bytes` so that we could implement `http::Uri::as_bytes` or something similar.

srijs:
@seanmonstar I've pushed a couple of commits that should address your feedback so far. Would appreciate a second review, to identify further gaps that you'd like to see addressed.

src/header/mod.rs
```diff
@@ -546,6 +551,32 @@ impl fmt::Debug for Headers {
...
+}
+
+#[cfg(feature = "compat")]
+impl From<Headers> for http_types::HeaderMap {
+    fn from(headers: Headers) -> http_types::HeaderMap {
+        let mut header_map = http_types::HeaderMap::new();
+        for header in headers.iter() {
+            let name = http_types::header::HeaderName::from_bytes(header.name().as_bytes())
+                .expect("attempted to convert invalid header name");
+            let value = http_types::header::HeaderValue::from_bytes(header.value_string().as_bytes())
```
seanmonstar: This should use `header.raw()` instead of `value_string()`. The `value_string` method can panic, even though the raw bytes were a valid header value.

Additionally, when using `raw`, that should mean a 2nd loop. The `name` can be constructed once, and used for each call to `append` for each item `raw`.

src/header/mod.rs
```diff
@@ -546,6 +551,32 @@ impl fmt::Debug for Headers {
     }
 }
 
+#[cfg(feature = "compat")]
+impl From<http_types::HeaderMap> for Headers {
+    fn from(header_map: http_types::HeaderMap) -> Headers {
+        let mut headers = Headers::new();
+        for (name, value) in header_map.iter() {
+            headers.set_raw(name.as_str().to_string(), value.as_bytes());
```
seanmonstar: This should use `append_raw`, since `set_raw` will overwrite any previous value for a header name.

src/client/mod.rs
```diff
@@ -107,6 +113,13 @@ where C: Connect,
     pub fn request(&self, req: Request<B>) -> FutureResponse {
         self.call(req)
     }
+
+    /// Send an `http::Request` using this Client.
+    #[inline]
+    #[cfg(feature = "compat")]
+    pub fn request_compat(&self, req: http_types::Request<B>) -> CompatFutureResponse {
```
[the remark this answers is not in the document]
srijs: I like that idea a lot! One concern I'd have is the impact on type inference. I'll go and explore it for a bit and get back to you.

src/client/mod.rs
```diff
@@ -107,6 +113,13 @@ where C: Connect,
...
+        self::compat::future(self.call(req.into()))
+    }
```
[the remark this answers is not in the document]
srijs: - [ ] Implement compatibility for `Client` `Service` API.

src/http/request.rs
```diff
@@ -132,6 +135,41 @@ impl<B> fmt::Debug for Request<B> {
     }
 }
 
+#[cfg(feature = "compat")]
+impl From<Request> for http_types::Request<Body> {
```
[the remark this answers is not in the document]
srijs: Unfortunately not, since the only reason Rust lets me impl this is due to the fact that part of the impl receiver is from this crate (since both the trait as well as the receiver of the impl are from other crates).

src/http/response.rs
```diff
@@ -142,6 +145,42 @@ impl fmt::Debug for Response {
...
+                to_headers.set_raw(name.as_str().to_string(), value.as_bytes());        
+            }
+        }
+        let (_, body) = from_res.into_parts();
+        to_res.with_body(body)
+    }
+}
+
+#[cfg(feature = "compat")]
+impl From<Response> for http_types::Response<Body> {
```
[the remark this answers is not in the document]
srijs: See explanation for `From<Request>` below.

src/http/request.rs
```diff
@@ -132,6 +135,41 @@ impl<B> fmt::Debug for Request<B> {
...
+
+#[cfg(feature = "compat")]
+impl<B> From<http_types::Request<B>> for Request<B> {
+    fn from(from_req: http_types::Request<B>) -> Request<B> {
+        let (from_parts, body) = from_req.into_parts();
+
+        let mut to_req = Request::new(from_parts.method.into(), from_parts.uri.into());
+        to_req.set_version(from_parts.version.into());
+        {
+            let mut to_headers = to_req.headers_mut();
```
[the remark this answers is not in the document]
srijs: I suppose it could, if I used `mem::replace` on the mutable reference returned here.

Will give it a shot!

src/header/mod.rs
srijs: You got it!

src/header/mod.rs
srijs: Makes sense.

src/client/mod.rs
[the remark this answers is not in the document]
srijs: Okay, so I did a brief exploration on this one, and the main challenge that I've encountered is that while I can make `request` work for both `hyper::Request` and `http::Request`, the `Item` of `FutureResponse` is still `hyper::Response`, which IMO is not an optimal user experience. So I'd probably need to make `FutureResponse` generic over the response type, which would be a breaking API change...

Let me know if you have any other ideas on how to solve this! I'll work on the other items in the meantime.

src/http/request.rs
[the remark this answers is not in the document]
seanmonstar: Ah dang, I see what you mean, it can't be done because of the generic + `From` + target is not a crate local.

- It can stay as it is, using the default generic of `hyper::Request`, and just not have a conversion when one has `hyper::Request<SomethingElse>`. In this case, `hyper::Request<B>` can grow an `into_compat` method to compensate...
- We can have an `Into` instead that does allow the generic, but then `http::Request::from(hyper)` doesn't work...

I'm not certain which is preferable, actually. Probably depends on which case is more common (using `From::from`, or `hyper::Request<NotDefaultBody>`).

seanmonstar:
Woo, nearly there!

src/http/request.rs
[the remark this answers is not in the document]
srijs: To get an idea where you stand on this, do you see the panic-ing of the conversion functions as a concern at all, or do you think it's a non-issue? Introducing `from` methods that panic has been bothering me more and more when thinking about it during the last couple of days, and at least personally I would feel a lot more confident about the quality of my contribution if the conversion methods could safely signal an error w/o resorting to `panic!`.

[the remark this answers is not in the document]
seanmonstar: I'm personally fine with them panicking for now. It is an opt-in compat layer, and any panics should be horribly wrong, unexpected bugs. If `TryFrom` were stable, I'd be fine with using that, but it's not.

If you'd rather, `try_into_http` inherent methods can be added, to allow non-panicking methods.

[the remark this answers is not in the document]
srijs: In that case, I'd be fine with leaving it like it is. Or did you want me to add the generic `into_compat` method before we merge? Anything else you'd like to see changed?

seanmonstar:
Yay! Great work, @srijs, thanks!

PR #1324, opened 2017-09-17
perf(client): Remove parked senders when a checkout is dropped to prevent memory leaks

### LINKED ISSUE #1315: Memory leak when using client.clone() for server (proxy)
I'm writing a reverse proxy and use the HTTP client to connect to any upstream server. Code is in https://github.com/klausi/rustnish/blob/goal-06/src/lib.rs#L150 . The server is leaking memory so I must be doing something wrong.

Steps to reproduce:

1. Make sure you have any upstream HTTP service running locally on port 80, for example the default Apache install on Ubuntu which will give you a dummy page at http://localhost/
   ```
   sudo apt install apache2
   ```

2. Run rustnish reverse proxy on port 9090:
   ```
   git clone --branch goal-06 git@github.com:klausi/rustnish.git
   cd rustnish
   cargo run --release
   ```

3. Get PID of rustnish process and memory usage:

   ```
   ps aux | grep rustnish
   ```
   The 6th column is the memory usage of the process, something like 20124 (~20MB)
   
4. Fire 1 million requests at the reverse proxy with Apache Bench:
   ```
   ab -c 4 -n 1000000 http://localhost:9090/
   ```
   (This takes ~130 seconds on my computer)
   
5. Check memory usage again:
   ```
   ps aux | grep rustnish
   ```
   The 6th column should show something like 284920, which is ~280MB!
   
So although we do not cache any requests or keep them around otherwise memory usage is growing without freeing up anymore.

One solution to the problem is this patch:

```
diff --git a/src/lib.rs b/src/lib.rs
index 520dd24..a591b99 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -170,7 +170,6 @@ pub fn start_server_background(
             let http = Http::new();
             let listener = TcpListener::bind(&address, &handle)
                 .chain_err(|| format!("Failed to bind server to address {}", address))?;
-            let client = Client::new(&handle);
 
             let server = listener.incoming().for_each(move |(sock, addr)| {
                 http.bind_connection(
@@ -180,7 +179,7 @@ pub fn start_server_background(
                     Proxy {
                         port: port,
                         upstream_port: upstream_port,
-                        client: client.clone(),
+                        client: Client::new(&handle),
                     },
                 );
                 Ok(())
```

Avoiding the client.clone() call fixes the memory leak but degrades the runtime performance. Apache Bench with 1 million requests took 220 seconds instead of 130 seconds before. So the client_clone() call seems to be the right thing to do, but the clones are not dropped correctly when one request handling is done?

klausi:
Attempt to fix #1315 .

Note that I'm not sure what I'm doing here is right because I have not groked yet what the purpose of those parked senders is.

### DIFF
--- a/src/client/pool.rs
+++ b/src/client/pool.rs
@@ -271,6 +271,16 @@ impl<T: Clone> Future for Checkout<T> {
     }
 }
 
+impl<T> Drop for Checkout<T> {
+    fn drop(&mut self) {
+        // Cleanup the parked senders in the pool to prevent a memory leak where
+        // it would grow indefinitely.
+        // @todo This clears out other parked senders as well, how should we
+        // remove just our sender of this checkout?
+        self.pool.inner.borrow_mut().parked.remove(&self.key);
+    }
+}
+
 struct Expiration(Option<Duration>);
 
 impl Expiration {


seanmonstar:
Thanks for getting it this far, it helped me understand where the leak is. Completely removing the parked list could hurt any other pending requests, making the client use idle connections less often. I've opened #1325 that will cleanup when a `Checkout` is dropped, but only if the `Sender` is dead.

klausi:
Very good, glad I could help at least some way :-)

Issue #1263, opened 2017-07-15
Http::bind does not scale to uses where a Handle is needed


brendanzab:
So, I have a [`ReverseProxy` type that requires a `Client`](https://docs.rs/hyper-reverse-proxy/0.2.1/hyper_reverse_proxy/struct.ReverseProxy.html#method.new) in order to be constructed. But in order to do that I need to [do the manual set up of the Tokio event loop](https://docs.rs/hyper-reverse-proxy/0.2.1/hyper_reverse_proxy/#example), because constructing a `Client` requires access to the `Handle`, which I can't get when calling `Http::bind`. 

```rust
fn run() -> hyper::Result<()> {
    // Set up the Tokio reactor core
    let mut core = Core::new()?;
    let handle = core.handle();

    // Set up a TCP socket to listen to
    let listen_addr = SocketAddr::new(Ipv4Addr::new(127, 0, 0, 1).into(), 8080);
    let listener = TcpListener::bind(&listen_addr, &handle)?;

    // Listen to incoming requests over TCP, and forward them to a new `ReverseProxy`
    let http = Http::new();
    let server = listener.incoming().for_each(|(socket, addr)| {
        let client = Client::new(&handle);
        let service = ReverseProxy::new(client, Some(addr.ip()));
        http.bind_connection(&handle, socket, addr, service);
        Ok(())
    });

    // Start our server on the reactor core
    core.run(server)?;

    Ok(())
}
```

This gets pretty horrific once I start needing the graceful shutdown behavior of `Server::run_until`, as shown by [hyper-reverse-proxy/examples/extended.rs](https://github.com/brendanzab/hyper-reverse-proxy/blob/2b2ef26523aca56f822622dc287b4fab925a7b42/examples/extended.rs). I pretty much need to re-implement all of that stuff myself...

Is there a better way to go about this? Or could we have a `Http::bind_with_handle` which could allow us to pass a handle to our own reactor core into Hyper?

P-E-Meunier:
Same here, exposing a handle is not enough: I'm running two HTTP servers on the same event loop (one http handling let's encrypt, and redirecting other requests to https), plus a pleingres database connection, and an SSH server.

I would be able to spawn the pleingres and SSH server using just the current handle, but not to run the two HTTP servers, because each wants its own event loop.

I'd be willing to write the two extra lines in my program to start the event loop myself, though.

seanmonstar:
[...]

Issue #1322, opened 2017-09-16
Creating Server with a Handle


seanmonstar:
I've put this off for a while claiming that it will be solved in tokio-proto, but I'm now leaning towards may as well fix it in hyper, if a generic way appears in the future, no harm done. So, on to the proposal:

------

I hinted at a possible solution [earlier in another issue](https://github.com/hyperium/hyper/issues/1075#issuecomment-297192417), and I think a solution is probably similar: Add a way to combine an `Http`, `Service`, `TcpListener`, and `Handle` into a `impl Future`. 

## Proposed API

- `Http`
  - `pub fn bind_handle<S>(&self, addr: &SocketAddr, new_service: S, handle: &Handle) -> Server`
- `Server`
  - `pub fn shutdown_signal<F>(&mut self, signal: F) -> &mut Self`
  - `impl Future for Server`

## Usage

```rust
// Spawn 2 servers on the same Core
let mut core = Core::new()?;
let handle = core.handle();

let srv1 = Http::new()
    .bind_handle(addr1, new_service1, &handle);

let srv2 = Http::new()
    .bind_handle(addr2, new_service2, &handle);

handle.spawn(srv1);
handle.spawn(srv2);
core.run(futures::future::empty())?;
```

```rust
// Spawn a Server with a Client
let mut core = Core::new()?;
let handle = core.handle();

let client = Client::new(&handle);
let server = Http::new()
    .bind_handle(addr, new_service(client), &handle);
handle.spawn(server);

core.run(futures::future::empty())?;
```

```rust
// Spawn a Server and still get graceful shutdown
let mut core = Core::new()?;
let handle = core.handle();

// send on tx to shutdown server future
let (tx, rx) = oneshot::channel();

let client = Client::new(&handle);
let mut server = Http::new()
    .bind_handle(addr, new_service(client), &handle);
server.shutdown_signal(rx);

// can even just run server directly
core.run(server)?;
```

## Implementation

This proposal is reusing the `Server` type, which already exists in a form that owns a `Core`. Internally, it can just change its field to an enum of 2 variants, 1 with a `Core`, and the other with a `Handle`. It should be able to use the majority of the same code in both cases.

It might be tricky that you can create a `Server` using `Http::bind`, which **owns a Core**, and then you can send that `Server` into another `Core`, calling `handle.spawn(server)`. If you were to do that, the `impl Future for Server` could make the outer `Core` lock up, since an implementation might decide to call `core.run()` on its inner `Core`.

[...]

PR #1326, opened 2017-09-19
feat(server): Accepting futures::Streams of connections, and turning the server into a future

P-E-Meunier:

This PR is maybe not meant to be accepted as such, but solves a number of problems for me:

- I can now run my own event loop, and configure my socket as I want before, without having to resort to low-level functions. This was already possible before, at the price of code duplication between Hyper and my crate. This solves issues #1075 and #1263.

- I can now use the latest developments of rustls (for example their recent addition of SNI): since the server now takes an arbitrary stream of connections, one example use is to negotiate a TLS connection and passing it as a stream, like so:

``` rust
let mut core = Core::new().unwrap();
let http = hyper::server::Http::new();
let listener = tokio_core::net::TcpListener::bind(addr, core.handle())
  .incoming().from_err()
  .and_then(move |(socket, addr)| {
      https_config
        .accept_async(socket)
        .map(move |socket| (socket, addr))
        .from_err()
  });
http.from_listener(listener, | | Ok(MyServer { … }).unwrap()
  .future_run(core.handle())
  .from_err()
```
Of course, it work fine on plain `TcpListener`s, and that's actually what `Http::bind` does.

Also, this PR does not change the existing API, except for:
- `local_addr`, which is now impossible to implement with the more general setting of arbitrary streams of connections instead of just Incoming.
- `handle`, but this is now useless: if you need a handle, you might as well run your own event loop.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -1,6 +1,7 @@
 #![deny(warnings)]
 extern crate hyper;
 extern crate futures;
+extern crate tokio_core;
 extern crate pretty_env_logger;
 
 use futures::future::FutureResult;
@@ -31,7 +32,5 @@ impl Service for Hello {
 fn main() {
     pretty_env_logger::init().unwrap();
     let addr = "127.0.0.1:3000".parse().unwrap();
-    let server = Http::new().bind(&addr, || Ok(Hello)).unwrap();
-    println!("Listening on http://{} with 1 thread.", server.local_addr().unwrap());
-    server.run().unwrap();
+    Http::new().run(&addr, || Ok(Hello)).unwrap();
 }
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -47,7 +47,5 @@ fn main() {
     pretty_env_logger::init().unwrap();
     let addr = "127.0.0.1:1337".parse().unwrap();
 
-    let server = Http::new().bind(&addr, || Ok(Echo)).unwrap();
-    println!("Listening on http://{} with 1 thread.", server.local_addr().unwrap());
-    server.run().unwrap();
+    Http::new().run(&addr, || Ok(Echo)).unwrap();
 }
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -14,15 +14,17 @@ use std::time::Duration;
 use futures::future;
 use futures::task::{self, Task};
 use futures::{Future, Stream, Poll, Async, Sink, StartSend, AsyncSink};
-use futures::future::Map;
+use futures::future::{Map, Empty, Select};
+use futures::stream::{FromErr};
 
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio::reactor::{Core, Handle, Timeout};
-use tokio::net::TcpListener;
+use tokio::net::{TcpListener, TcpStream, Incoming};
 use tokio_proto::BindServer;
 use tokio_proto::streaming::Message;
 use tokio_proto::streaming::pipeline::{Transport, Frame, ServerProto};
 pub use tokio_service::{NewService, Service};
+
 
 use http;
 use http::response;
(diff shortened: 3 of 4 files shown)


GitCop:

 * Commit: f350a4b5cde99b18a7d4f907121cddee63d93e61
  * Your subject line is longer than 100 characters

---

GitCop:

 * Commit: f350a4b5cde99b18a7d4f907121cddee63d93e61
  * Your subject line is longer than 100 characters

---

GitCop:

 * Commit: d631ef72cdcfc206d5a702a0954691bbe273a8ab

 * Commit: 8649d5e370c09606a49048895546b9834ad0b1a0

---

seanmonstar:
Hey there, thanks for starting this up! I do indeed want to add the ability to use any stream of connections, without having to use the tokio-proto stack, so this is definitely the right idea.

I recently outlined in #1322 that shows a proposed API for using `Http`/`Server` with a `Handle`, instead of owning the `Core`. Importantly, #1322 defines a way to keep backwards compatibility, which would be needed to merge this into 0.11. Would you like to adjust this PR to match what is proposed there?

Additionally, while we would eventually want to get rid of the `local_addr` method, since it doesn't belong in hyper, we can't remove it in 0.11. So, a `SocketAddr` would need to be provided with the `from_listener` method, so it can be set in the `Server`. We could deprecate `Server::local_addr` also (maybe now, maybe later) to hint that it would go away in 0.12.

P-E-Meunier:
Hi! As I figured out with my second commit, the tests currently in master do not pass with my proposed change, so I guess we'll have to wait for 0.12.
This is not a giant problem for me, as long as 0.12 comes in the next few weeks.

By the way, I like how the server code is structured. It's really short and clean.

seanmonstar:
> so I guess we'll have to wait for 0.12. This is not a giant problem for me, as long as 0.12 comes in the next few weeks.

Not likely, 0.12 will probably be a little while, there's still some upstream changes to figure out with futures and tokio.

However! We don't need to wait for 0.12. As I said, we can morph this into the API in #1322, and be able to release it in a 0.11.x release. What do you think of trying to do that?

P-E-Meunier:
Sure. So since the API are necessarily different (some have their own `Core`, some don't), that would mean duplicating some code for a while, marking it as deprecated (as you said before, but not just for `local_addr`).

The `core` field of the struct could be an `Option`, right?

Another question I had was, why did you require anything to be `Send + Sync`? This seem a little odd for stuff supposed to be run in an event loop.

seanmonstar:
>  So since the API are necessarily different (some have their own Core, some don't), that would mean duplicating some code for a while? The core field of the struct could be an Option, right?

I believe it can be done without duplicating too much code. For instance:

- `Server.core` could become `Server.reactor`, with an `enum Reactor { Core(Core), Handle(Handle) }`.
- `Server.run`/`Server.run_until` consume the `Server`, so the `Core` can be taken out, a `Handle` grabbed, and then the `Server` can be reconstructed the same, just with a `Reactor::Handle(handle)`, and then because it implements `Future`, `run` can just call `core.run(server)`.

> Another question I had was, why did you require anything to be Send + Sync? This seem a little odd for stuff supposed to be run in an event loop.

This was because we hoped to soon move the internal `Server::run` thing to use tokio-proto's `TcpServer`, which can run on multiple threads. I'm not sure this is going to actually happen, but I'd worry about that in a different issue (in fact, there is #1275 that removes it, and I'll be looking at merging that PR soon too).

kamalmarhubi:
An added benefit of this approach is making it possible to implement HTTP over unix domain sockets. Having TCP baked in as it is currently would make that rather annoying!

seanmonstar:
Master has seen some updates, adding a `Serve` type that may eventually be the replacement of `Server` (perhaps it takes its name in 0.12). We should add a method to `Http` that can create a `Serve` from any stream of incoming connections, which should accomplish the rest of the goal of this PR. Frankly, was just stuck on naming...

Thank you so much for exploring this! This helped figure out that `Server` both owning a `Core` and possibly being a `Future` was awkward.

sanmai-NL:
@kamalmarhubi: interesting, yet I personally have no idea how user clients could browse a website via Unix domain sockets. What would the purpose of such a feature be?

P-E-Meunier:
Here's a really direct application of allowing arbitrary streams: tests! Maybe you don't can't open ports on you machine, and you want to run the protocol inside buffers (or on unix domain sockets if you want to benchmark what gets sent to the kernel).

PR #1329, opened 2017-09-20
feat(header): add ContentType::xml() constructor

spk:

This only add new constructor ContentType::xml()
Cheers

### DIFF
--- a/src/header/common/content_type.rs
+++ b/src/header/common/content_type.rs
@@ -68,13 +68,13 @@ impl ContentType {
         ContentType(mime::TEXT_PLAIN_UTF_8)
     }
 
-    /// A constructor  to easily create a `Content-Type: text/html; charset=utf-8` header.
+    /// A constructor  to easily create a `Content-Type: text/html` header.
     #[inline]
     pub fn html() -> ContentType {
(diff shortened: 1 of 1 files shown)


src/header/common/content_type.rs
```diff
@@ -74,6 +74,12 @@ impl ContentType {
         ContentType(mime::TEXT_HTML)
     }
 
+    /// A constructor  to easily create a `Content-Type: text/xml; charset=utf-8` header.
```
seanmonstar: The docs say `text/xml; charset=utf-8`, but [`TEXT_XML`](https://docs.rs/mime/0.3.4/mime/constant.TEXT_XML.html) is `text/xml`. Either the documentation is wrong, or the implementation :D

I've typically stayed away from XML as much as possible (lifestyle choice), so I can't say from experience which is more common. I suspect including the charset is?

spk: Hehe; for me its more without charset; made a fix in https://github.com/spk/hyper/commit/4d893a7fc9cf2b187598298dc467b4db9b44d516

seanmonstar:
Great, thanks!

Issue #1327, opened 2017-09-19
Reason for Sync + Send + 'static


mersinvald:
Hello!

Can you please explain reasoning behind `Sync + Send + 'static` bound on the `NewService` here
https://github.com/hyperium/hyper/blob/77bfcd95156e6f19c0618d7493ea092b4e055655/src/server/mod.rs#L88-L92
and here
https://github.com/hyperium/hyper/blob/77bfcd95156e6f19c0618d7493ea092b4e055655/src/server/mod.rs#L339-L342

I've been writing my custom `NewService` and ran into problems with this bounds.

Are they necessary since `hyper` moved to async single-thread mode with `Tokio`?
Removing them will allow to write simpler and way more effective single-thread async code.

seanmonstar:
This has been addressed in master, and the linked PRs have reasoning around it.

PR #1331, opened 2017-09-22
 fix(server): remove Sync + Send + 'static trait bound 

mersinvald:
Remove unnecessary for tokio-based async server thait bounds according to Issue #1327

BREAKING CHANGE: server NewService and Http API

Thanks to @LooMaclin for writing this patch in the first place.


mersinvald:
Since this change relaxes the thait requirements should it be concidered as breaking?

mersinvald:
Didn't notice the existing PR on the same reason: https://github.com/hyperium/hyper/pull/1275

seanmonstar:
Hey there! Thanks for submitting this!

As you noticed, there was also #1275, and some discussion happened in there around why these bounds existed. I've since finally merged that PR. This should be in 0.11.3 soon!

mersinvald:
@seanmonstar that's awesome news! Thank you!

PR #1332, opened 2017-09-22
feat(server): Creating Server with a Handle WIP

Thomspoon:

This PR is a work-in-progress for #1322.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -41,6 +41,13 @@ use http::Body;
 pub use http::response::Response;
 pub use http::request::Request;
 
+/// Hyper::Server is able to either create and use its own core, or share
+/// one with others.
+enum ServerInternal {
+    Core(Core),
+    Handle(Handle)
+}
+
 /// An instance of the HTTP protocol, and implementation of tokio-proto's
 /// `ServerProto` trait.
 ///
@@ -62,7 +69,7 @@ where B: Stream<Error=::Error>,
 {
     protocol: Http<B::Item>,
     new_service: S,
-    core: Core,
+    internal: ServerInternal,
     listener: TcpListener,
     shutdown_timeout: Duration,
 }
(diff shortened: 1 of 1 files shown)


src/server/mod.rs
```diff
@@ -376,7 +403,7 @@ impl<T, B> Service for HttpService<T>
     }
 }
 
-impl<S, B> Server<S, B>
+impl<'a, S, B> Server<S, B>
```
Thomspoon: Oops. That wasn't supposed to be in there.

src/server/mod.rs
```diff
@@ -389,7 +416,10 @@ impl<S, B> Server<S, B>
     /// Returns a handle to the underlying event loop that this server will be
     /// running on.
     pub fn handle(&self) -> Handle {
-        self.core.handle()
+        match self.internal {
+            ServerInternal::Core(ref core) => core.handle().clone(),
```
seanmonstar: The clone here isn't need, `core.handle()` already returns `Handle`.

src/server/mod.rs
```diff
@@ -105,13 +112,33 @@ impl<B: AsRef<[u8]> + 'static> Http<B> {
...
+    /// Bind the provided `addr` and return a server that has
+    /// a shared `Core`.
+    /// 
+    /// This method, when provided a handle to a `Core`, allows
+    /// the ability to bind multiple servers to the same `Core`.
+    pub fn bind_handle<S, Bd>(&self, addr: &SocketAddr, new_service: S, handle: Handle) -> Server<S, Bd>
+        where S: NewService<Request = Request, Response = Response<Bd>, Error = ::Error> + 'static,
+            Bd: Stream<Item=B, Error=::Error>,
+    {
+        let listener = TcpListener::bind(addr, &handle).unwrap();
```
seanmonstar: This probably means `bind_handle` should return a `Result`, just like `bind` does, so that we don't panic.

seanmonstar:
Master has grown to have a `Serve` type which will be used to explore replacing `Server`. `Serve` doesn't require a `Core`, but instead is a stream of binding connections with HTTP.

Thanks for taking a stab at this!

PR #1333, opened 2017-09-23
feat(client): allow custom executors for HttpConnector

### LINKED ISSUE #1330: Allow HttpConnector to be provided an Executor
The default connector in hyper, the `HttpConnector`, uses a `CpuPool` to do blocking DNS resolution. It would be useful to be able to provide a `futures::future::Executor`, so that users can try to share a `CpuPool` with other blocking tasks as well.

srijs:
Fixes #1330

This got slightly more complex than I initially thought due to trying to make a future-proof API by:
- not exposing any `futures::sync::oneshot::*` types
- catering for the possibility of `HttpConnector` wanting to execute different types of blocking tasks

---

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -2,9 +2,13 @@ use std::error::Error as StdError;
 use std::fmt;
 use std::io;
 use std::mem;
+use std::sync::Arc;
 //use std::net::SocketAddr;
 
 use futures::{Future, Poll, Async};
+use futures::future::{Executor, ExecuteError};
+use futures::sync::oneshot;
+use futures_cpupool::{Builder as CpuPoolBuilder};
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio::reactor::Handle;
 use tokio::net::{TcpStream, TcpStreamNew};
@@ -43,22 +47,35 @@ where T: Service<Request=Uri, Error=io::Error> + 'static,
 /// A connector for the `http` scheme.
 #[derive(Clone)]
 pub struct HttpConnector {
-    dns: dns::Dns,
+    executor: HttpConnectExecutor,
     enforce_http: bool,
     handle: Handle,
 }
 
 impl HttpConnector {
-
     /// Construct a new HttpConnector.
     ///
     /// Takes number of DNS worker threads.
     #[inline]
     pub fn new(threads: usize, handle: &Handle) -> HttpConnector {
+        let pool = CpuPoolBuilder::new()
+            .name_prefix("hyper-dns")
+            .pool_size(threads)
+            .create();
+        HttpConnector::new_with_executor(pool, handle)
+    }
+
+    /// Construct a new HttpConnector.
+    ///
+    /// Takes an executor to run blocking tasks on.
+    #[inline]
+    pub fn new_with_executor<E: 'static>(executor: E, handle: &Handle) -> HttpConnector
+        where E: Executor<HttpConnectorBlockingTask>
+    {
         HttpConnector {
-            dns: dns::Dns::new(threads),
+            executor: HttpConnectExecutor(Arc::new(executor)),
             enforce_http: true,
-            handle: handle.clone(),
+            handle: handle.clone()
         }
     }
 
@@ -109,7 +126,7 @@ impl Service for HttpConnector {
         };
 
         HttpConnecting {
-            state: State::Lazy(self.dns.clone(), host.into(), port),
+            state: State::Lazy(self.executor.clone(), host.into(), port),
             handle: self.handle.clone(),
         }
     }
@@ -154,8 +171,8 @@ pub struct HttpConnecting {
 }
 
 enum State {
-    Lazy(dns::Dns, String, u16),
-    Resolving(dns::Query),
+    Lazy(HttpConnectExecutor, String, u16),
+    Resolving(oneshot::SpawnHandle<dns::IpAddrs, io::Error>),
     Connecting(ConnectingTcp),
     Error(Option<io::Error>),
 }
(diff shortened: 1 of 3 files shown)


srijs:
FYI current tip of master fails for me, so I based this off of a slightly stale version of master.

seanmonstar:
Yea sorry, I pushed a change to master that broke some tests. I've pushed a fix for them just now.

Wow, this change was more involved than I thought!

srijs:
Yeah, I'm open for suggestions how to simplify it!

It's just that every attempt to slim it down on my part has resulted in exposing implementation details on the public API, which I was assuming is not what you want.

src/client/mod.rs
```diff
@@ -30,7 +30,7 @@ use version::HttpVersion;
 
 pub use http::response::Response;
 pub use http::request::Request;
-pub use self::connect::{HttpConnector, Connect};
+pub use self::connect::{HttpConnector, Connect, HttpConnectorBlockingTask};
```
seanmonstar: Is it possible to not expose this type to outside the crate?

src/client/connect.rs
```diff
@@ -231,6 +248,48 @@ impl ConnectingTcp {
     }
 }
 
+enum HttpConnectorBlockingTaskInner {
+    Dns(oneshot::Execute<dns::Work>)
+}
```
seanmonstar: Could this inner `enum` go away, and just be a field of `HttpConnectorBlockingTask`?

src/client/connect.rs
```diff
@@ -42,23 +46,31 @@ where T: Service<Request=Uri, Error=io::Error> + 'static,
 
 /// A connector for the `http` scheme.
 #[derive(Clone)]
-pub struct HttpConnector {
-    dns: dns::Dns,
+pub struct HttpConnector<E = CpuPool> {
```
seanmonstar: I'm slightly worried about exposing the generic here. I suppose with a default set, it should still compile for everyone... The alternative is to keep a `Box<Executor>` (or perhaps `Arc<Executor>`).

srijs: Since `HttpConnectExecutor` is wrapping `E` in an `Arc` in any case, there shouldn't be an issue with hiding the generic in there as a trait object. I'll give that a shot!

- [ ] Hide `Executor` generic

src/client/connect.rs
srijs: Absolutely! I was setting this up so that we can potentially add different kinds of blocking tasks, but I suppose as long as the API supports it without forcing a breaking change, we can always expand this into an enum later.

src/client/mod.rs
srijs: I'm not 100% sure what you are saying here.

Not exposing it as in not exporting it, but with the type still being part of the signature of `HttpConnector::new_with_executor`? Or instead boxing so that `HttpConnector::new_with_executor` requires an `Executor<Box<Future<Item=(), Error=()>>>`?

Both don't seem ideal to me :/

seanmonstar: Oh OK, it's used for the generic type on `futures::future::Executor<T>`... Could the constructor `new_with_executor` just be generic over `F: Future<Item=(), Error=()>`? After that, it's turned into a trait object and put in an `Arc`, so the generic isn't really important...

srijs: Hmm, maybe I'm misunderstanding, but as far as I can see when implementing a function that is generic over `F`, it's impossible for us to produce a future of type `F` without being passed one from the outside. But since our goal is to be able to execute arbitrary futures, I'm unsure how that would help us?

seanmonstar: I checked out the branch and saw what you mean. Ideally we'd accept `E: Executor<impl Future<Item=(), Error=()>`, since there isn't really a reason to tell anyone the exact type that is being executed. If I delete from this export, it's essentially doing that. It's a type that implements the future we want, but users can't name it (as would be the case with `impl Future`), and seems best to me.

srijs:
@seanmonstar resolved conflicts.

I'd like to try and resolve the last outstanding issue, whether to export `HttpConnectorBlockingTask` or not.

Could you help me out by detailing what your concerns are with exporting it? Is it about the size of the API surface? Or something different?

At the moment, I feel rather strongly against _not_ exporting it. In my opinion, this will complicate the situation for frameworks being built on top of hyper.

---

For instance, assume for a moment that we are building a framework based on hyper, and there is some cpu-intensive task that we'd like to offload to a thread pool, and this thread-pool needs to be customisable, similar to what we want for the DNS pool here.

So now there are two types of tasks that we'd like to schedule on the executor: Hyper's http connector task, and our custom framework task. However, since `HttpConnectorBlockingTask` has not been exported from hyper, it's impossible to express an enum of these two tasks which could be used in the trait constraint for the `Executor`.

---

As an alternative, I would suggest that we require an `Executor<Box<Future>>` instead of an `Executor<HttpConnectorBlockingTask>`. This does not constrain consumers of hyper in the way that I've described above, and has the additional upside of not expanding the API surface. The downside is the future will be moved to the heap, which might result in slightly worse performance.

In summary, the three options we have:

1. Don't export any new symbols, require `Executor<Box<Future>>`
    - More flexibility for consumers 👍 
    - No increase in API surface 👍 
    - Slightly worse performance 👎 
2. Don't export any new symbols, require `Executor<HttpConnectorBlockingTask>`
    - Less flexibility for consumers 👎 
    - No increase in API surface 👍 
    - No performance impact 👍 
3. Export `HttpConnectorBlockingTask`, require `Executor<HttpConnectorBlockingTask>`
    - More flexibility for consumers 👍 
    - Slight increase in API surface 👎 
    - No performance impact 👍 

--- 

Would be great to get your thoughts on this and work towards getting this merged!

seanmonstar:
> Could you help me out by detailing what your concerns are with exporting it? Is it about the size of the API surface? Or something different?

It's just about promising backwards compatibility around naming and trait impls on that type.

> However, since HttpConnectorBlockingTask has not been exported from hyper, it's impossible to express an enum of these two tasks which could be used in the trait constraint for the Executor.

True, you could not define an enum naming the `HttpConnectorBlockingTask`. However, it *would* work similar to if we just returned an `impl Future` using the `conservative_impl_trait` unstable feature. It would still implement the trait, so it can passed to anything that is generic over `Future<Item=(), Error=()>`, and it can also be boxed.

So, in your example of wanting to use a thread pool, if we use something like `futures_cpupool`, it would still be doable, since it is already generic over any `F: Future<Item=(), Error=()>`. You just wouldn't be able to name it specially, so you could have an `enum Task { Something, Connect(HttpConnectorBlockingTask), }`.

If this API were to return `Box<Future>` instead, it actually wouldn't change much. You still couldn't name the type. However, it would force boxing at a stage that might not need it. You could still end up creating an enum, like `enum Task { Something, Connect(Box<Future<Item=(), Error=()>>), }`, and just box the future as it's returned from hyper.

----

I don't mean to be annoying, my main concern is around backwards compatibility. It can be a pain at times :)

srijs:
@seanmonstar Thanks for taking the time and explaining it in the way you did! You managed to convince me that not exporting the symbol is an acceptable solution, so I made the change to _not_ export the constructor.

Let me know if you'd like to see any other changes before merging this!

seanmonstar:
Great, thank you for the work!

PR #1336, opened 2017-09-25
feat(body): implement limited buffering

srijs:
Apologies for opening yet another PR while my previous one is still open. Please do take your time!

This is some logic that I noticed I started copying between projects, so it seemed high time to make it reusable. The basic idea is to protect against DoS attacks in cases where you do need to buffer the full body in memory, for instance JSON parsing, by enforcing a maximum size while buffering.

Initially I thought of putting it into a separate crate, but it would be nice to have this available directly in hyper. For instance, I noticed that the guide is using `Stream::concat2` [here](https://hyper.rs/guides/client/json/) in the example for JSON parsing. In the spirit of Rust being safe by default, I believe we should promote `Body::buffer` as the default way of doing this, rather than `Stream::concat2` which is vulnerable to DoS attacks through large request or response bodies.

Let me know what you think!

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -139,6 +139,22 @@ impl From<httparse::Error> for Error {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+/// Possible error when the body buffer limit is exceeded.
+pub struct BodyTooLargeError;
+
+impl fmt::Display for BodyTooLargeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str(self.description())
+    }
+}
+
+impl StdError for BodyTooLargeError {
+    fn description(&self) -> &str {
+        "Body is too large"
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::error::Error as StdError;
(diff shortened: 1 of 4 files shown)


seanmonstar:
It's true that if one doesn't think about it, `body.concat2()` could buffer a ginormous body. I've so far been fine with that, since it's quite similar to showing examples with blocking IO to use `io.read_to_string(&mut s)`. I think that a web framework should probably do more here, like checking the `Content-Length` header (and endpoints with size restrictions should probably reject requests without a `Content-Length`).

Additionally, this sort of functionality is likely useful outside of hyper, for any `Stream` over bytes. I don't know if it possibly makes sense as a thing on `Stream`, perhaps. It'd be interesting if it could be a property of the thing being extended, such that at `MaxSizedChunk` could reject being extended past a certain size, but unfortunately, the `Extend` trait doesn't really have any way of doing that other than panicking.

All this to say I'm unsure if it should be part of hyper directly, or somewhere else. It's possible @alexcrichton or @carllerche have thought of these things also.

alexcrichton:
This does indeed seem like a plausible combinator to extract! My guess now though would likely be a separate crate rather thanan existing on as it's not clear where it may go

srijs:
If this is nothing that should live in `hyper` then I agree that a separate crate would be appropriate. I feel it's probably not generic enough to live in `futures` itself...

PR #1343, opened 2017-10-01
fix(http): avoid infinite recursion when Body::from is called with Cow::Owned.

aaronriekenberg:
When cow is a Cow::Owned, cow.to_owned() returns a Cow::Owned, which leads to infinite recursion.
Extract the owned or borrowed values from the cow to ensure progress is made in either case.


seanmonstar:
Woops, thanks for catching and fixing this!

PR #1346, opened 2017-10-03
feat(client): add names to DNS threads

### LINKED ISSUE #918: Add names to DNS threads
They currently just inherit the thread name from wherever the client is created.

brogowski:
Closes #918


seanmonstar:
Great, thanks!

PR #1347, opened 2017-10-03
refactor(uri): remove parsing condition

nieksand:
Scheme check never true given preceding else if


nieksand:
On line 65 we already do s.contains("://"). 

So this check on line 93 can never be true.

seanmonstar:
Thanks!

Issue #1334, opened 2017-09-23
Strict-Transport-Security header does not support 'preload'


raphaelcohn:
To allow for HTTPS pinning, browsers can check for the value of `preload` in `Strict-Transport-Security` header. Hyper doesn't support this.

raphaelcohn:
see <https://hstspreload.org/>

seanmonstar:
Thanks to `StrictTransportSecurity` having all of its fields public, it's a breaking change to add a new one, so adding this would need to be part of 0.12. (While in there, the fields should be made private with getter and setter methods instead, so we can add more fields in the future...)

gabibbo97:
Hello, if no one is working on this I would like to start an implementation

seanmonstar:
I haven't seen a pull request, so fair game! Note that as pointed out earlier, this would be a breaking change and so couldn't be released soon...

seanmonstar:
For now, the new version of hyper won't be including the typed headers directly, so I'm closing this here as a won't fix.

PR #1355, opened 2017-10-17
feat(headers): Implement HSTS preload

gabibbo97:
Prior to this, Hyper did not implement the HSTS preload header,
this commit changes the header in order to implement a constructor that verifies the validity of max-age at compile time, a coherent string to header parser and some tests to ensure functionality.

This closes issue  #1334 
**BREAKING CHANGE:** the StrictTransportSecurity struct now has a new bool member: preload

### DIFF
--- a/src/header/common/strict_transport_security.rs
+++ b/src/header/common/strict_transport_security.rs
@@ -29,7 +29,8 @@ use header::{Header, Raw, parsing};
 /// # Example values
 /// * `max-age=31536000`
 /// * `max-age=15768000 ; includeSubDomains`
-///
+/// * `max-age=31536000 ; includeSubDomains ; preload`
+/// 
 /// # Example
 /// ```
 /// # extern crate hyper;
@@ -39,7 +40,13 @@ use header::{Header, Raw, parsing};
 /// let mut headers = Headers::new();
 ///
 /// headers.set(
-///    StrictTransportSecurity::including_subdomains(31536000u64)
+///     {
+///         let mut StrictTransport = StrictTransportSecurity::new();
+///         StrictTransport.set_max_age(StrictTransportSecurity::preload_min_age());
+///         StrictTransport.set_include_subdomains(true);
+///         StrictTransport.set_preload(true);
+///         StrictTransport
+///     }
 /// );
 /// # }
 /// ```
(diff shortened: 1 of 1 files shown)


seanmonstar:
Cool! Since you're in there, doing all the breaks, how would you feel making the fields all private, and have setters/getters instead? That'd allow any new fields to be added without breaking changes in the future...

gabibbo97:
I have made a significant breaking change to the header:

- all members are now private
- you can create the header using a builder-like pattern

I have updated also the documentation
This is my first pull-request of all time in Rust 😄

gabibbo97:
I did something wrong with my understanding of the "bench_headers" macro

gabibbo97:
Please wait, currently I am refactoring the tests

gabibbo97:
Everything should be ok now

src/header/common/strict_transport_security.rs
```diff
@@ -39,43 +40,108 @@ use header::{Header, Raw, parsing};
...
         }
     }
 
-    /// Create an STS header that excludes subdomains
-    pub fn excluding_subdomains(max_age: u64) -> StrictTransportSecurity {
+    /// Returns the inserted max age
+    pub fn max_age(&self) -> u64 { self.max_age }
+
+    /// Sets the max age
+    pub fn set_max_age(&self, max_age: u64) -> StrictTransportSecurity {
```
seanmonstar: It's more usual for setters to just mutate the type itself, than to return a new type. We should probably stick with the usual :D

src/header/common/strict_transport_security.rs
```diff
@@ -39,43 +40,108 @@ use header::{Header, Raw, parsing};
...
+    /// * Redirects all of the traffic on port 80 to port 443.
+    /// * Serves all of its subdomains via HTTPS.
+    /// * Sets in the HSTS headers the parameter includeSubDomains and has max-age
+    ///   bigger than 31536000 seconds (one year).
+    /// * Any redirect will still be served with a HSTS header
+    preload: bool
 }
 
+/// This value indicates the minimum value of the max_age
+/// that can be used togheter with preload
```
seanmonstar: Typo: "together".

src/header/common/strict_transport_security.rs
```diff
@@ -39,43 +40,108 @@ use header::{Header, Raw, parsing};
...
+
+    /// Returns if the header applies HSTS preload
+    pub fn preload(&self) -> bool { self.preload }
+
+    /// Sets if the HSTS header applies HSTS preload.
+    /// This will check if the HSTS header satisfies that:
+    /// 
+    /// * It also applies to subdomains
+    /// * It has a max_age greater than 31536000 seconds
+    pub fn set_preload(&self, preload: bool) -> Result<StrictTransportSecurity,()> {
```
seanmonstar: Hm, I see why this returns a `Result`. Google requires the max age minimum in order for a site to eligible to for its own preload list.

Though, I'm not sure if we should be enforcing that specific requirement here.

src/header/common/strict_transport_security.rs
```diff
@@ -85,35 +151,56 @@ impl FromStr for StrictTransportSecurity {
...
             .and_then(|res| match res {
-                (Some(age), sub) => Ok(StrictTransportSecurity {
+                (Some(age), sub, pre) if pre.is_none() => Ok(StrictTransportSecurity{
                     max_age: age,
-                    include_subdomains: sub.is_some()
+                    include_subdomains: sub.is_some(),
+                    preload: false
                 }),
+                (Some(age), sub, pre) if pre.is_some() => {
+                    if sub.is_none() || age < PRELOAD_MIN_AGE {
```
seanmonstar: I wouldn't enforce the limit when parsing the header either. It's a perfectly valid header, it just doesn't meet the requirements for Google's list at the time of writing (but could in the future).

gabibbo97:
Hi Sean, thanks for your review, from the last time I have edited all of the setters from builder-pattern like to proper setter methods.
I also agree with you on the preload issue, the RFC tells us that a StrictTransportSecurity header that does not comply with Google's guidelines is still valid so I have removed the check and added a method that checks if your current struct matches the guidelines, in this way I believe that we can manage to parse valid but not Google compliant headers and also check if our new header meets the requisites.
The docs have been updated according to this change

seanmonstar:

Issue #1358, opened 2017-10-21
How do you write a resilient HTTP Hyper server that does not crash with "too many open files"?


klausi:
The Hello world example at https://hyper.rs/ is vulnerable to denial of service attacks if the max number of allowed open file descriptors is not high enough. I was in contact with @seanmonstar already about this and he does not think this is a security issue, so I'm posting this publicly.

Steps to reproduce:

1. Implement the Hello world example from https://hyper.rs/
2. Set a very low file descriptor limit to 50 to provoke the crash early: `ulimit -n 50`
3. Start the Hello world server: `cargo run`
4. Open another shell and attack the server with Apache bench (100 concurrent requests): `ab -c 100 -n 10000 http://localhost:3000/`

That will crash the server with an IO Error `Io(Error { repr: Os { code: 24, message: "Too many open files" } })`.

A naive solution is to just restart the server all the time with a loop:

```rust
fn main() {
    loop {
        let addr = "127.0.0.1:3000".parse().unwrap();
        let server = Http::new().bind(&addr, || Ok(Proxy)).unwrap();
        match server.run() {
            Err(e) => println!("Error: {:?}", e),
            Ok(_) =>{},
        };
    }
}
```

Which is not at all ideal because there is a downtime for a short period of time and all connections from clients are reset.

I checked the behavior of other server software, Varnish in this case. With a low file descriptor limit it just waits until it has descriptors available before accepting connections.

Can Hyper do the same? How do you run your Hyper servers in production to prevent a server crash when file descriptors run out?

klausi:
After digging around a bit in Hyper and Tokio I think that  [tokio_core::net::Incoming](https://docs.rs/tokio-core/0.1.10/tokio_core/net/struct.Incoming.html) is flawed. It has an Error type but each TCP connection attempt it produces is not a Result but a TcpStream already. So there is no way to catch the IO error on individual incoming accept() attempts and then continue to receive new connections. Contrast that with [std::net::Incoming](https://doc.rust-lang.org/std/net/struct.Incoming.html) which produces a Result for each connection attempt.

So I think the path forward here is to try out a forked version of tokio_core and Hyper to see if it is possible to produce TCP connections in the same Result way as the standard library.

seanmonstar:
A fork shouldn't be required, you can accomplish the same thing quite easily implementing a new `Future` that works on `listener.accept()` instead of `incoming()`.

[...]

PR #1369, opened 2017-11-04
tests(server): Add failing test to demonstrate port exhausting problem

klausi:
Failing test demonstration for #1358

### DIFF
--- a/tests/server.rs
+++ b/tests/server.rs
@@ -564,8 +564,8 @@ fn no_ports_exhausted_errors() {
     let handle = core.handle();
 
     // Send 100k requests (TCP connections).
-    let nr_requests = 100_000;
-    let concurrency = 10_000;
+    let nr_requests = 10_000_000;
+    let concurrency = 100_000;
 
     let mut parallel = Vec::new();
     for _i in 0..concurrency {
--- a/tests/server.rs
+++ b/tests/server.rs
@@ -6,7 +6,7 @@ extern crate pretty_env_logger;
 extern crate tokio_core;
 
 use futures::{Future, Stream};
-use futures::future::{self, FutureResult};
+use futures::future::{self, FutureResult, join_all, loop_fn, Loop};
 use futures::sync::oneshot;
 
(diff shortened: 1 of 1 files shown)


tests/server.rs
```diff
@@ -553,6 +553,42 @@ fn no_proto_nonempty_parse_eof_returns_error() {
...
+    let nr_requests = 100_000;
+    let concurrency = 10_000;
+
+    let mut parallel = Vec::new();
+    for _i in 0..concurrency {
+        let requests_til_done = loop_fn(0, |counter| {
+            // Just establish the TCP connection, do nothing otherwise.
+            let socket = tokio_core::net::TcpStream::connect(&addr, &handle);
+
+            socket.then(move |_| -> Result<_, std::io::Error> {
```
seanmonstar: It seems that this will drop the `TcpStream` right upon connect, and so there won't actually be that many concurrent connections. I think.

klausi:
This is test is trying to exhaust available ports on the system to trigger an IO error which makes the server panic. This was failing for me locally, maybe the Travis VM is too slow so that the problem does not occur? I raised the number of requests to 10 million, let's see if that makes a difference.

tests/server.rs
klausi: Good point! Not sure what we can do to keep the connection open for a bit before dropping it? Add a future that just sleeps 1 second?

klausi:
Yay, Travis is failing now as expected. This test is not super ideal because it has to make that many requests, but on the other hand it is a nice integration test to check if Hyper can keep up when a flood of requests arrives.

The next step now is to figure out how to actually fix this and make the test pass.

klausi:
I think I have to close this because a test case like this has random fails and is not reliable. I had a similar test running in Rustnish, but removed it after failing randomly on Travis CI: https://github.com/klausi/rustnish/commit/e7fcd752dc620dedd4c589efd0f3d1f05eecaf6a

Sorry! Let me know if you have better ideas how to test such an attack.

PR #1370, opened 2017-11-06
feat(header): implement ByteRangeSpec::to_satisfiable_range

srijs:
Implemented this for work recently, and it took us a couple of tries and careful reading of the RFC to get it right. Based on this experience, I could image it to be useful as a built-in.

Let me know what you think!

### DIFF
--- a/src/header/common/range.rs
+++ b/src/header/common/range.rs
@@ -76,6 +76,77 @@ pub enum ByteRangeSpec {
     Last(u64)
 }
 
+impl ByteRangeSpec {
+    /// Given the full length of the entity, attempt to normalize the byte range
+    /// into an satisfiable end-inclusive (from, to) range.
+    ///
+    /// The resulting range is guaranteed to be a satisfiable range within the bounds
+    /// of `0 <= from <= to < full_length`.
+    ///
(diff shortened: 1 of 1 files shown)


srijs:
FYI nightly seems to fail for reasons apparently not linked to this change set

seanmonstar:
Very cool, thanks! (Nightly failure is because nightly recently gained ASCII methods on `str`, making `AsciiExt` imports trigger a warning.)

I do admit, I had a hard-ish time understand what this problem was solving. But after looking at the tests, it seems to be "give me a range up to the Nth byte"?

srijs:
Thanks for merging!

Yeah, so the problem it's solving is given the full length of the requested representation of a resource, just return the inclusive range that should be used to slice the data internally before returning.

So for instance, you implement an endpoint serving a file from disk, and you want to support range requests. It's a lot easier to just get a (start, end) offset that you can use to seek and read than having to deal with all the different ways a client can request a range.

Any suggestions for how to improve the docs to make this clearer?

seanmonstar:
I see, so its trying to turn the variants into a single `a..b`, as long as `n >= b`. Maybe an example or two in the docs might help, but I'm not really sure. I've never had to use it, so my context is very shallow.

PR #1372, opened 2017-11-11
feat(client): skip dns resolution when host is a valid ip addr

srijs:
This change allows us to skip going to the thread pool for dns lookup in cases where the host itself is already a valid ipv4 or ipv6 address.

This is useful for performance reasons, as well as to allow passing in a no-op executor in cases where it's known that only static IPs will be connected to.


seanmonstar:
Very nice!

PR #1376, opened 2017-11-15
feat(headers): support known extensions

sanmai-NL:
And only known extensions. Clean up parsing of extensions in general.

### DIFF
--- a/src/header/common/cache_control.rs
+++ b/src/header/common/cache_control.rs
@@ -31,16 +31,15 @@ use header::parsing::{from_comma_delimited, fmt_comma_delimited};
 /// );
 /// ```
 /// ```
-/// use hyper::header::{Headers, CacheControl, CacheDirective};
+/// use hyper::header::{Headers, CacheControl, CacheDirective, CacheDirectiveExtension};
 ///
 /// let mut headers = Headers::new();
 /// headers.set(
 ///     CacheControl(vec![
 ///         CacheDirective::NoCache,
 ///         CacheDirective::Private,
-///         CacheDirective::MaxAge(360u32),
-///         CacheDirective::Extension("foo".to_owned(),
-///                                   Some("bar".to_owned())),
+///         CacheDirective::MaxAge(360_u32),
+///         CacheDirective::Extension(CacheDirectiveExtension::StaleWhileRevalidate(360_u32)),
 ///     ])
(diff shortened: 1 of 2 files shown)


sanmai-NL:
@seanmonstar: could you let me know when you plan to review this?

LegNeato:
This would be a breaking change, right? If so can you add `BREAKING_CHANGE: <description>` to the commit message?

seanmonstar:
Sorry, yes, this would be a breaking change, so it couldn't be part of 0.11. I'll put it in the 0.12 milestone, but if someone needs it immediately, one could just copy the code into a local module and use it instead of importing from hyper.

sanmai-NL:
@LegNeato, @seanmonstar: done. @seanmonstar: okay great, but when is a review planned, if any?

src/header/common/cache_control.rs
```diff
@@ -108,8 +129,8 @@ pub enum CacheDirective {
     /// "s-maxage=delta"
     SMaxAge(u32),
 
-    /// Extension directives. Optionally include an argument.
-    Extension(String, Option<String>)
+    /// Extension directives.
+    Extension(CacheDirectiveExtension),
```
seanmonstar: The original goal of the `Extension` variant was to allow for new arguments even if the version of hyper doesn't support them (hence being just `String`s). Would it make sense for these new variants to just be on `CacheDirective` itself?

seanmonstar:

sanmai-NL:
@seanmonstar: would it be okay to merge this PR, for the time being? Nobody is going to look up a stale PR branch to reuse any of this work, so it’d be a waste. Even if you plan to release the redesigned `hyper` version soon, then at least I could pin `hyper` to an older version still has the `header` module, along with this enhancement.

sfackler:
Is there any particular reason you need to pull the header from the Hyper crate specifically?

sanmai-NL:
No there isn’t. I was under the impression that the new headers crate will not be API-compatible with the old one. Apart from that, I assumed that the starting point in terms of headers to reimplement would be the original Hyper headers module, and so that it would matter for everyone whether this PR gets merged or is left depending on further initiative.

seanmonstar:
Unfortunately, it can't just be merged, as it's a breaking change.

PR #1381, opened 2017-11-22
Shut down the connection when done with it

### LINKED ISSUE #1380: Conn should call shutdown on the inner stream when dropping
This is pretty important for TLS connections so that they can perform the TLS connection shutdown messaging to allow for session reuse.

sfackler:
This is important for TLS connections in particular

Closes #1380

I don't think this handles the client side properly.


seanmonstar:
Thanks!

PR #1390, opened 2017-11-29
feat(server): Allow keep alive to be turned off for a connection

### LINKED ISSUE #1365: Make graceful shutdown work at request rather than connection granularity
The graceful shutdown logic currently waits until all `Service`s have dropped (i.e. all connections have closed). However, with keep-alive this is overconservative - what we'd really like is to wait until pending requests have finished.

This can't be correctly modeled with the current API, however. The last point in a request/response pair that user code is aware of is when it sends the last bit of the response body through the body channel. It will still take some amount of time for that bit to be written out over the network, so we can't immediately shut down after that's done.

sfackler:
Closes #1365

### DIFF
--- a/src/proto/conn.rs
+++ b/src/proto/conn.rs
@@ -453,6 +453,14 @@ where I: AsyncRead + AsyncWrite,
     pub fn close_write(&mut self) {
         self.state.close_write();
     }
+
+    pub fn disable_keep_alive(&mut self) {
+        if self.state.is_idle() {
+            self.state.close_read();
+        } else {
+            self.state.disable_keep_alive();
+        }
+    }
 }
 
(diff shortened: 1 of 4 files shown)


sfackler:
This currently only allows you to disable keepalive rather than providing a toggle since the only use case I'm aware of involves turning it off and it seems a bit less risky to go in that direction only.

seanmonstar:
Hm, I wonder if this will work while the connection is currently "parked". I guess if you have the `Connection` future, and call this function, then a later call to `poll` will notice that the state is disabled, and end, right?

sfackler:
Nice catch - fixed and added a test for that case as well.

PR #1398, opened 2017-12-12
fix(proto): Shutdown quietly

sfackler:


sfackler:
On second thought, I don't think this is a good idea.

PR #1400, opened 2017-12-13
docs(server): Add a forms server example

jolhoeft:
Add an example program illustrating parsing a request body through
generating a response body.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ unicase = "2.0"
 num_cpus = "1.0"
 pretty_env_logger = "0.1"
 spmc = "0.2"
+url = "1.0"
 
 [features]
 default = ["server-proto"]
--- a/examples/params.rs
+++ b/examples/params.rs
@@ -1,0 +1,106 @@
+#![deny(warnings)]
+extern crate futures;
+extern crate hyper;
+extern crate pretty_env_logger;
+extern crate url;
+
+use futures::{Future, Stream};
+
+use hyper::{Get, Post, StatusCode};
+use hyper::header::ContentLength;
+use hyper::server::{Http, Service, Request, Response};
+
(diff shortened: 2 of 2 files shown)


jolhoeft:
The AppVeyor failures don't appear to be related to this PR

seanmonstar:
Hey neat! Do you think this sort of thing would be well suited to be a [guide](https://hyper.rs/guides/)?

jolhoeft:
Possibly. I wrote it because I realized I was misunderstanding how to get the response body out of the request body. I was thinking of it as a series of transforms on the request body until I had the response body, but couldn't figure out how to bail early with an error status if I needed to. I eventually realized you transform the request body into a response future. You create a new future/stream for the response body.

Having written that, it does sound like a very useful thing for the guide.

seanmonstar:
Yea, that's a good point. You tell me, would you be interested in turning this into a guide at https://github.com/hyperium/hyperium.github.io/tree/master/_guides/server?

jolhoeft:
I will give it a try. I expect it will take me a couple days. Clear prose is harder to write than code.

Do you want to merge this in separate from that? I see the example and the guide as two different but related things.

seanmonstar:
My personal (somewhat weakly held) opinion so far has been that "examples" can be hard to understand without context, and I've tried to not add too many, but instead any time I've wanted to make an example, tried to make it a guide instead.

jolhoeft:
Oddly my experience is the reverse. I look for examples before a detailed guide. I find a good example can give me a better high level view of how everything goes together. People have a wide range of learning styles, though, and I think we need both.

My approach on the guide is to explain this example in more detail, and add comments on how to approach Json processing, and possibly spawning threads for generating response bodies.

seanmonstar:
Fair enough, convinced! Thanks!

PR #1404, opened 2017-12-26
docs(server): Add an example of serving files

jolhoeft:
Here is an example of serving files in a separate thread. It should cover most cases where a separate thread is required, such as accessing a database.

There are actually two approaches illustrated here. One reads the whole file into memory, the other streams the file in chunks, but requires a second thread for the Response body. I don't like this two threaded approach, but haven't been able to figure out a way around it. Suggestions for a better approach would be appreciated.

I am working on an update to the guide pages to go with this example (which I will submit along with the Request body processing guide pages I'm working on.)

### DIFF
--- a/examples/send_file.rs
+++ b/examples/send_file.rs
@@ -1,0 +1,142 @@
+#![deny(warnings)]
+extern crate futures;
+extern crate hyper;
+extern crate pretty_env_logger;
+
+use futures::{Future, Sink};
+use futures::sync::{mpsc, oneshot};
+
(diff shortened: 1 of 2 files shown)


jolhoeft:
Two threads are not needed. I had a mental block, thinking I could only have one channel per thread, which isn't true. I've updated this example. There are still two versions, because the non-streaming version is simpler for small amounts of data/processing.

seanmonstar:
Awesome work!

PR #1406, opened 2017-12-26
docs(server): Add server example querying a web api

jolhoeft:


jolhoeft:
I don't understand the Travis errors. They are failing claiming I am not using `use std::ascii::AsciiExt;`, but only in nightly and beta. I am using that import though, and when I build with nightly/beta locally, everything is fine.

seanmonstar:
It's because starting in Rust 1.23 (which is currently in beta), the methods that were on `AsciiExt` are now inherent methods, and so the imported trait is not "used" on >=1.23. You need to slap a `#[allow(unused)]` on the `use std::ascii::AsciiExt` line.

jolhoeft:
Odd that I can't reproduce it locally, but it should be fixed now.

seanmonstar:
Thanks, these are really useful!

PR #1416, opened 2018-01-15
Guarantee file unchanged!

passchaos:


sfackler:
What is this protecting against?

sfackler:
Oh, nevermind. I see.

It seems really bad that this is reading 16 bytes at a time. That buffer should be like 1000 times larger than it is.

jolhoeft:
The small buffer is deliberate since the test file is so small. There is a comment about that, but further up the file.

passchaos:
Yeah, the buffer is too small to get a fast download speed, but more importly, the bound is necessary to get the consistent file.

passchaos:
In most cases, the size of the file is not a multiple of that of the 16, so, the last time to read a file to the buffer will not fill the buffer, without boundary, send files will be compared to the original file more than a few bytes.

jolhoeft:
Ok, I tested, and you are right, we are getting trailing data, but the browser is ignoring them. I was confusing it with a Vec.

seanmonstar:
Thanks!

PR #1421, opened 2018-01-20
Add the BREW method

Proximyst:
The BREW method as described by [RFC 7168](https://tools.ietf.org/html/rfc7168#section-2.1) should be added to Hyper as it is active and published for international use.


seanmonstar:
Look at the publish date of RFC 7168: it's an April Fool's joke.

PR #1427, opened 2018-01-26
"1.0 milestone" => "milestones"

kud1ing:
The 1.0 milestone has currently no issues but the 0.12 milestone has. So instead of adjusting the README.md for each upcoming mileston, i propose to just link to all milestones


seanmonstar:
Good idea, thanks!

Issue #304, opened 2015-02-09
HTTP/2 support


cactorium:
I'm guessing this would make good feature request :)

pyfisch:
Is it possible with current blocking design to write a good HTTP/2 implementation?

seanmonstar:
I don't see a specific stopper to implementing it with the blocking API.
However, we are exploring using async io for hyper before hyper 1.0.

andydude:
I am interested in being assigned to this, but I might look around the hyper source first.

drbo:
Hyper now uses asynchronous I/O, so HTTP/2 is no longer blocked in async I/O.

seanmonstar:
@stepancheg wow, exciting! I admit to not having much time to review all of it.

I encourage you to keep working on it, including the list you provided. hyper is far too close to a 0.11 release to try to squeeze in such a huge change. But we can definitely see how the project goes!

nathanaeljones:
HTTP/2 support is huge for performance. [Globally 82% of clients support it - 93% in the USA](https://caniuse.com/#search=http2).

Is HTTP/2 support planned for hyper 1.0?

seanmonstar:
Yes, we're working on the [h2](http://github.com/carllerche/h2) crate at the moment, and hyper 0.12 will be released with the h2 built in.

avkonst:
Could you please advise when hyper 0.12 is coming?

seanmonstar:
We wish to group together the breaking changes, and with tokio going through a reform, that is the current blocker.

szagi3891:
What changes are planned in the "Tokyo"?

rpjohnst:
@szagi3891 tokio-rs/tokio-rfcs#3

avkonst:
That is unfortunate dependency :( Is it feasible to have 2 releases with the first one releasing this feature?

carllerche:
@avkonst you can use the h2 crate as is today.

avkonst:
Thanks. It seems it would require to go one level lower, than the level of abstraction provided by hyper + iron for me.  I would expect that the web framework like Iron will benefit automatically from Hyper being http2 capable. Plus, it seems h2 does not handle http 1.1 and alpn, it says it is only http2 "parser".

carllerche:
[...]

PR #1428, opened 2018-01-26
feat(client): redesign the `Connect` trait

seanmonstar:
The original `Connect` trait had some limitations:

- There was no way to provide more details to the connector about how to
  connect, other than the `Uri`.
- There was no way for the connector to return any extra information
  about the connected transport.
- The `Error` was forced to be an `std::io::Error`.
- The transport and future had `'static` requirements.

As hyper gains HTTP/2 support, some of these things needed to be
changed. We want to allow the user to configure whether they hope to
us ALPN to start an HTTP/2 connection, and the connector needs to be
able to return back to hyper if it did so.

The new `Connect2` trait is meant to solve this.

- The `connect` method now receives a `Destination` type, instead of a
  `Uri`. This allows us to include additional data about how to connect.
- The `Future` returned from `connect` now must be a `Connected`, which
  wraps the transport, and includes possibly extra data about what
  happened when connecting.

The `Connect` trait is deprecated, with the hopes of `Connect2` taking
it's place in the next breaking release. For backwards compatibility,
any type that implements `Connect` now will automaticall implement
`Connect2`, ignoring any of the extra data from `Destination` and
`Connected`.

A step towards #304. 

--------

I've tested this branch with hyper-tls, and ensured that without a change, it still works, and that in a new branch, the new trait can be adopted. Unfortunately, ALPN isn't exposed in native-tls, so that would be needed before hyper-tls could make use of this.

cc @sfackler @carllerche

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -1,9 +1,9 @@
+//! Contains the `Connect2` trait, and supporting types.
 use std::error::Error as StdError;
 use std::fmt;
 use std::io;
 use std::mem;
 use std::sync::Arc;
-//use std::net::SocketAddr;
 
 use futures::{Future, Poll, Async};
 use futures::future::{Executor, ExecuteError};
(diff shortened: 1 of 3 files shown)


src/client/connect.rs
```diff
@@ -16,31 +16,74 @@ use tokio_service::Service;
...
+}
+
+impl Destination {
+    /// Get a reference to the requested `Uri`.
+    pub fn uri(&self) -> &Uri {
+        &self.uri
+    }
+
+    /// Returns whether this connection must negotiate HTTP/2 via ALPN.
+    pub fn h2(&self) -> bool {
```
seanmonstar: I suspect users to be able to configure a hyper `Client`, stating whether to allow both HTTP/1 and HTTP/2, or to require **only** HTTP/2. To allow both, ALPN needs to send both protocols. A connector would query the `Destination` to determine what protocols it should try for.

I wasn't actually sure what to name the methods. They are meant to represent the desire of the user, but some desires may be wants or preferences, versus musts. Maybe some combination in this list is better?

- `must_h2`
- `can_http1`
- `can_http2`
- `prefer_h2`

src/client/connect.rs
```diff
@@ -16,31 +16,74 @@ use tokio_service::Service;
...
+    /// Create new `Connected` type with empty metadata.
+    pub fn new(transport: T) -> Connected<T> {
+        Connected {
+            alpn: Alpn::Http1,
+            transport: transport,
+        }
+    }
+
+    /// Convert into the underlyinng Transport.
+    pub fn into_transport(self) -> T {
```
seanmonstar: This was added when I was prototyping in hyper-tls, and noticed that to wrap another `Connect2`, I needed to take the transport out of the `Connected`. It's kind of verbose, however, and loses whatever settings the previous connector may have set. Would `map<F, U>(self, f: F) where F: FnOnce(T) -> U` perhaps be a better fit?

src/client/connect.rs
```diff
@@ -16,31 +16,74 @@ use tokio_service::Service;
...
+        self.transport
+    }
+
+    /// Set that the connected transport negotiated HTTP/2 as it's
+    /// next protocol.
+    pub fn h2(&mut self) -> &mut Connected<T> {
```
seanmonstar: I would expect the `Connected` type have some methods to be able to set what protocol was actually negotiated. Should this be setters instead, `set_h2()` etc?

carllerche:
I don't have much context here. I'm not sure what exactly `Connect` instances are supposed to do for hyper. I don't know why passing a Uri is required.

As I'm sure you are aware, I have my stab at a connect trait [here](https://github.com/carllerche/tokio-connect/blob/master/src/lib.rs). It seems to pan out well as it is.

src/client/connect.rs
seanmonstar: Actually, this also seems like a good way for connectors to report back if it's an HTTP/1.1 proxy, and so the absolute-form should be used for h1.

@tafia This may interest you, as you developed hyper-proxy. Maybe something like `connected.set_proxy()` or `set_http1_proxy()`...

seanmonstar:
@carllerche The problem with the trait in tokio-connect is that it doesn't give enough information. hyper tries to provide a higher level API for users, but some of the details of how HTTP works depends on the connection layer. The tokio-connect trait doesn't allow passing any argument, so it cannot be given a host name or even a socket address.

The `Connect` trait in hyper is meant to be a little higher level, allowing users to swap out the connect/transport layer, while still getting an easier-to-use HTTP client. This PR tries to encapsulate what information hyper needs to give to connect layer, and what that can give back to hyper, in order to determine how exactly it can speak HTTP over it.

seanmonstar:
After thinking that a connector might return some info, like being an HTTP proxy, and another might wrap that connector to provide TLS, the proposed design of returning `Connected<T>` means that details from the underlying return value will likely be lost when the wrapper takes the transport out.

So it may be better to change the return type to `(T, Connected)` instead of `Connected<T>`, and allow wrapping connectors to just append details to the returned `Connected` metadata type.

seanmonstar:
The ALPN stuff has been deferred for when HTTP2 upgrades are added, but the change still allows for that to be added in the future. A way for connectors to signal that the connection is to a proxy has stayed. This has been merged into 0.12.x [here](https://github.com/hyperium/hyper/commit/a408c0ff65bb9a878cb44f27994699e406faa18c), without the deprecations and instead just changes to `Connect` directly.

PR #1430, opened 2018-01-31
More descriptive error on binding

asonix:
I had a friend who couldn't figure out why their rocket project would fail to start because `Permission Denied`. This change makes the error more understandable.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -20,7 +20,8 @@ use self::Error::{
     Io,
     Ssl,
     TooLarge,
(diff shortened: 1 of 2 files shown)


seanmonstar:
I'd worry that changing the error from `Error::Io` would break anyone handling the errors in their own frameworks. Could this just be handled in rocket?

asonix:
Yeah, this could probably be handled in Rocket. I'll look into that when I have some time

PR #1444, opened 2018-02-20
docs(headers): document the fact that <Basic as FromStr>::from_str takes a base-64 encoded string

mikeyhew:
I couldn't figure out why my "username:password" strings kept failing to parse
into a Basic auth header, until I realized that the implementation expects
it to be base-64 encoded, which would be the case if it was coming from HTTP.
I'm not sure if this is the best place to document it, but hopefully it will
make it more clear for other people / me when I forget.

Perhaps a better approach would be to document somewhere that all `FromStr` impls for
headers are there for parsing request headers, and not really for creating them.

### DIFF
--- a/src/header/common/authorization.rs
+++ b/src/header/common/authorization.rs
@@ -167,6 +167,7 @@ impl Scheme for Basic {
     }
 }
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks!

PR #1446, opened 2018-02-20
Drop the connecting item from pool when dropping an idle connection

pimeys:
So when we hit the keep-alive timeout, we cannot open a new connection because we had a leftover item in the connecting map. This seems to fix the issue, but I'm not sure is this how kosher and did I break something else.


seanmonstar:
Seems good!

PR #1450, opened 2018-02-24
 fix(server): Sleep on socket IO errors 

### LINKED ISSUE #1358: How do you write a resilient HTTP Hyper server that does not crash with "too many open files"?

klausi:
Fixes #1358 

This pull request introduces a new sleep_on_error configuration option. It is used to swallow IO errors when the Hyper server is accepting requests. That way developers do not have to worry about IO errors in case the server is under heavy load and runs out of available file descriptors. Better DoS protection.

* Is sleep_on_error a good name? I copied that from https://github.com/tailhook/tk-listen
* What should be the default value? Currently sleep_on_error is false for backwards compatibility and developers need to enable it. I would argue that Hyper should be secure by default and it should be set to true.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -19,7 +19,10 @@ fn main() {
             .with_body(PHRASE))
     }));
 
-    let server = Http::new().bind(&addr, new_service).unwrap();
+    let server = Http::new()
+        .sleep_on_errors(true)
+        .bind(&addr, new_service)
+        .unwrap();
     println!("Listening on http://{} with 1 thread.", server.local_addr().unwrap());
     server.run().unwrap();
 }
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -664,6 +664,7 @@ impl Stream for AddrIncoming {
     type Error = ::std::io::Error;
 
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        // Check if a previous timeout is active that was set by IO errors.
         if let Some(ref mut to) = self.timeout {
             match to.poll().expect("timeout never fails") {
                 Async::Ready(_) => {}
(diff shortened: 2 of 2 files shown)


seanmonstar:
Great, I love it!

I agree it should probably be enabled by default, and maybe someone may want to be able to configure the duration too, but this is a great first step.

klausi:
Oh, cool, this was merged fast! Sorry about the additional merge commit. I though you would use the squash merge method on Github and did not clean up the commits here. Will do that next time!

And a big thank you to @tailhook, I copied quite a bit of your code and docs!

PR #1453, opened 2018-02-27
feat(method): add link and unlink http methods

ParisLiakos:
Both are commonly used for REST APIs.

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -8,14 +8,14 @@ use http;
 
 use error::Error;
 use self::Method::{Options, Get, Post, Put, Delete, Head, Trace, Connect, Patch,
-                   Extension};
+                   Link, Unlink, Extension};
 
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR!

I'm not sure `LINK` and `UNLINK` are so common that they should be variants. Looking around at other HTTP libs, they don't seem to be included. The good news is that it's possible to use these with the `Extension` variant.

ParisLiakos:
Yeah, i already did that, `hyper::Method::Extension("UNLINK".to_string())` works fine, thanks :+1: 

If other HTTP libs dont include them they i was probably wrong :) I came across them frequently though.

I guess we should draw the line somewhere, otherwise we would probably have to include all of https://www.iana.org/assignments/http-methods/http-methods.xhtml

PR #1454, opened 2018-03-01
feat(client): introduce lower-level Connection API

### LINKED ISSUE #1449: A lower level Connection API for the Client
The `Client` is a higher level API that, among other things, provides a connection pool. Similar to how the server has higher and lower level APIs (`Server` vs `Serve`/`Connection`), the client should gain a lower level connection based API. This would allow finer grained control over when connections are made, and would allow external pool implementations (cc #1253).

### Implementation

There would be a new `hyper::client::Connection` type, which implements `Service`. It would represent a single connection bound to HTTP. It would not reconnect itself when it is closed, someone managing the connection would handle that.

seanmonstar:
Adds the API described in #1449.

Closes #1449

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -1,0 +1,456 @@
+//! Lower-level client connection API.
+//!
+//! The types in thie module are to provide a lower-level API based around a
+//! single connection. Connecting to a host, pooling connections, and the like
+//! are not handled at this level. This module provides the building blocks to
+//! customize those things externally.
+//!
+//! If don't have need to manage connections yourself, consider using the
+//! higher-level [Client](super) API.
+use std::fmt;
+use std::marker::PhantomData;
+
+use bytes::Bytes;
+use futures::{Async, Future, Poll, Stream};
+use futures::future::{self, Either};
+use tokio_io::{AsyncRead, AsyncWrite};
+
+use proto;
+use super::{dispatch, Request, Response};
+
+/// Returns a `Handshake` future over some IO.
+///
+/// This is a shortcut for `Builder::new().handshake(io)`.
+pub fn handshake<T>(io: T) -> Handshake<T, ::Body>
+where
+    T: AsyncRead + AsyncWrite,
+{
+    Builder::new()
+        .handshake(io)
+}
+
+/// The sender side of an established connection.
+pub struct SendRequest<B> {
+    dispatch: dispatch::Sender<proto::dispatch::ClientMsg<B>, ::Response>,
+
+}
+
+/// A future that processes all HTTP state for the IO object.
+///
+/// In most cases, this should just be spawned into an executor, so that it
+/// can process incoming and outgoing messages, notice hangups, and the like.
+pub struct Connection<T, B>
+where
+    T: AsyncRead + AsyncWrite,
+    B: Stream<Error=::Error> + 'static,
+    B::Item: AsRef<[u8]>,
+{
+    inner: proto::dispatch::Dispatcher<
+        proto::dispatch::Client<B>,
+        B,
+        T,
+        B::Item,
+        proto::ClientUpgradeTransaction,
+    >,
+}
+
+
+/// A builder to configure an HTTP connection.
+///
+/// After setting options, the builder is used to create a `Handshake` future.
+#[derive(Clone, Debug)]
+pub struct Builder {
+    h1_writev: bool,
+    keep_alive: bool,
+}
+
+/// A future setting up HTTP over an IO object.
+///
+/// If successful, yields a `(SendRequest, Connection)` pair.
+pub struct Handshake<T, B> {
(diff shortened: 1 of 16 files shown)


src/client/conn.rs
```diff
@@ -0,0 +1,456 @@
...
+}
+
+// ===== impl SendRequest
+
+impl<B> SendRequest<B>
+{
+    /// Polls to determine whether this sender can be used yet for a request.
+    ///
+    /// If the associated connection is closed, this returns an Error.
+    pub fn poll_ready(&mut self) -> Poll<(), ::Error> {
```
sfackler: Is the intent for this that you'd call it when pulling a connection out of a pool?

src/client/conn.rs
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    /// Poll the connection for completion, but without calling `shutdown`
+    /// on the underlying IO.
+    ///
+    /// This is useful to allow running a connection while doing an HTTP
+    /// upgrade. Once the upgrade is completed, the connection would be "done",
+    /// but it is not desired to actally shutdown the IO object. Instead you
+    /// would take it back using `into_parts`.
+    pub fn poll_without_shutdown(&mut self) -> Poll<(), ::Error> {
```
sfackler: Couldn't poll know that the connection's upgraded and not shut it down?

src/client/conn.rs
```diff
@@ -0,0 +1,456 @@
...
+    pub(super) fn h1_writev(&mut self, enabled: bool) -> &mut Builder {
+        self.h1_writev = enabled;
+        self
+    }
+
+    /// Set whether to use keep-alive on this connection.
+    ///
+    /// Default is true.
+    #[inline]
+    pub fn keep_alive(&mut self, enabled: bool) -> &mut Builder {
```
sfackler: Does this inject a `Connection: close` or just cause poll to return ready after the first response comes in?

src/client/conn.rs
seanmonstar: That is a fine time to do so, or even if you're not using a pool, and just have a single connection, it lets you know if the connection could process it right now. I think I need to alter this PR slightly so that it returns `NotReady` for h1 connections that aren't pipelining. It would also return `NotReady` if this were an h2 connection and you'd reached `SETTINGS_MAX_CONCURRENT_STREAMS `.

src/client/conn.rs
seanmonstar: It could, but that's probably a trap, since you may not choose to do anything special with a `Connection`, and just call `exe.spawn(conn.map_err(|e| debug!("conn error: {}", e))`. If you did that, and `poll` did what `poll_with_shutdown` does in this PR, then the transport would never have `shutdown` called.

If `Connection::Item` were `Parts`, then that could probably be easier, since then you'd need to handle the `Parts` in some way (even if just dropping) when spawning the future. But, in order to do that, `Connection` needs to hold the internal state in an `Option`, so it can remove it, and thus every poll requires a `self.inner.as_mut().unwrap()`. I figured by providing `poll_without_shutdown`, `Connection::poll` isn't bothered, and anyone could wrap `poll_with_shutdown` and an `Option` if they wanted that kind of API.

Does that reasoning seem convincing?

src/client/conn.rs
seanmonstar: Currently it tells the internal `h1::Conn` that after the first response, it cannot be used for a new request, and sets its state as `Closed`. This is the same as disabling `keep_alive` on the `Client` as well.

I've sometimes thought that it should probably set `Connection: close`, to be nice, but then, even if the header isn't set, once a response has been read, it's up to the client to use it again, or close the connection.

Interestingly, I think this option would do nothing if the connection were HTTP2. Or maybe it could send a `GOAWAY` after the request...

src/client/conn.rs
sfackler: If you did try to send a request, would that just be queued up on the connection becoming ready, or would it be an error?

src/client/conn.rs
sfackler: FWIW storing an Option and .as_mut().unwrap()'ing it is super common in futures/stream combinators so it wouldn't be that crazy. Seems fine to have two methods though.

src/client/conn.rs
sfackler: Given that it wouldn't do anything for h2, would it make more sense to not have this method at all and have users set Connection: close to manage keep alive? Hyper does pay attention if that header is set on outbound requests I think, right?

src/client/conn.rs
seanmonstar: If the connection weren't ready, you'd get an error. We start using this `poll_ready` -> `send` pattern in the h2 crate, and it felt very natural, and it's what `Sink` is changing to in futures 0.2.

sfackler: Ah cool, makes sense. I was thinking it wouldn't be an error, so I was a bit confused :)

Issue #1455, opened 2018-03-03
Make sleep_on_error timeout configurable, enable it by default + document it


klausi:
Follow-up to #1358 .

Now that we have internal IO error handling when a Hyper server is under heavy load (or DoS attack) we should make this configurable and enabled by default for a better developer experience.

- [ ] Make sleep_on_error timeout configurable (currently hard-coded to 10ms)
- [ ] Enable sleep_on_error per default
- [ ] Document it. https://docs.rs/hyper/0.11.21/hyper/server/struct.Http.html#method.sleep_on_errors is a pretty good start, should we say more about it? Maybe a footnote somewhere on https://hyper.rs/ ?

BatmanAoD:
Disclaimer: I am not a web developer, either frontend or backend.

Swallowing errors *by default* seems like a terrible idea to me. This is the best way to make unreliable and difficult to debug software. In fact, JavaScript's "swallow errors by default" design philosophy is perhaps the primary reason *why* I'm not a web developer; I was hoping that a full-stack Rust framework (with WebAssembly on the frontend) would make the idea more appealing to me, because Rust seems to have precisely the opposite design philosophy.

However, it does seem that for a DoS attack or heavy load, the server needs to be able to ignore requests for some short period of time. But the user should still be able to do something with the *specific errors* that are generated.

So my suggestion would be to permit users to specify callbacks to invoke when specific errors occur, or possibly a generic callback invoked when any ignored error occurs.

A reasonable default behavior might be to automatically set up these callbacks to simply log the error. That would accomplish the goal of making robust server setup easier, while also ensuring that there's still an easy way to discover and investigate possible bugs--or, in a production environment, actual attempted DoS attacks.

avkonst:
Completely agree with the above comment.

What is the behavior, when timeout is 0 ms?

I consider it totally appropriate to reject new connections and free related resources immediately if they can not be accepted due to lack of handlers allowed. It is unlikely that there will be anyone allowed in 10ms (or whatever small interval), but context for timeouts and inbound connections will be held in RAM. 

Upper layer should be notified about such rejections.

seanmonstar:
#1488 was just merged, and included some changes that touch this. Here's how it works now:

- `sleep_on_errors` is set to be default on
- When enabled, it will log at the `error!` level about the failures, so users can still see bad things are happening.
- The timeout duration itself isn't configurable, it's just 1 second.
[...]

PR #1457, opened 2018-03-03
feat(server): Make sleep_on_error duration configurable and enable it per default

klausi:
For issue #1455 .

This is a minor API break because I'm changing the type of the sleep_on_error() method. This method exists only for a very short amount of time and Hyper is pre-1.0 so changing this might be ok?

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -59,7 +59,7 @@ pub struct Http<B = ::Chunk> {
     max_buf_size: Option<usize>,
     keep_alive: bool,
     pipeline: bool,
-    sleep_on_errors: bool,
+    sleep_on_errors: Option<Duration>,
     _marker: PhantomData<B>,
 }
 
@@ -106,7 +106,7 @@ pub struct AddrIncoming {
     keep_alive_timeout: Option<Duration>,
     listener: TcpListener,
     handle: Handle,
-    sleep_on_errors: bool,
+    sleep_on_errors: Option<Duration>,
     timeout: Option<Timeout>,
 }
 
@@ -121,7 +121,7 @@ impl<B: AsRef<[u8]> + 'static> Http<B> {
             keep_alive: true,
             max_buf_size: None,
             pipeline: false,
-            sleep_on_errors: false,
+            sleep_on_errors: Some(Duration::from_millis(10)),
             _marker: PhantomData,
         }
     }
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -683,14 +685,14 @@ impl Stream for AddrIncoming {
                     return Ok(Async::Ready(Some(AddrStream::new(socket, addr))));
                 },
                 Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => return Ok(Async::NotReady),
-                Err(ref e) if self.sleep_on_errors => {
+                Err(ref e) if self.sleep_on_errors.is_some() => {
```
messense: I wonder if this can be rewritten as:

```rust
Err(ref e) if let Some(delay) = self.sleep_on_errors => {
```

so you don't need to use `.unwrap()` later.

klausi: Tried that, but the compiler does not allow `let` in a match arm, as far as I can see.

klausi:
The test fail on AppVeyor does not seem related to this change. `client_keep_alive_connreset()` has failed randomly before on other commits on AppVeyor, so I think we can ignore that in this issue.

src/server/mod.rs
```diff
@@ -183,9 +183,11 @@ impl<B: AsRef<[u8]> + 'static> Http<B> {
     /// (like "too many files open") may consume 100% CPU and a timout of 10ms
     /// is used in that case.
     ///
-    /// Default is false.
-    pub fn sleep_on_errors(&mut self, enabled: bool) -> &mut Self {
-        self.sleep_on_errors = enabled;
+    /// This can be disabled by setting the duration to None.
+    ///
+    /// Default is 10ms.
+    pub fn sleep_on_errors(&mut self, duration: Option<Duration>) -> &mut Self {
```
seanmonstar: We couldn't make this change for 0.11.x, but could be in 0.12.

src/server/mod.rs
seanmonstar: Could be done like this, right?

```rust
match self.listener.accept() {
    // Ok ...
    Err(e) => {
        if let Some(delay) = self.sleep_on_errors {
            // continue or delay
        }
        return Err(e);
    }
}
```

klausi:
@seanmonstar good idea with the `if let` in the match body, implemented that. Ready for review again!

seanmonstar:
So, the changes look right! Thanks!

However, as mentioned in a line comment, this is a breaking change, so couldn't be done in 0.11. I'd say this could be a PR against the 0.12.x branch, but with that switching to the new tokio, I'm unsure where timeouts will come from at the moment...

klausi:
I think we can put this on hold until the new Tokio is integrated and then rewrite this wherever it fits in.

seanmonstar:
With #1488 merged, a variant of this is now in place. Thanks!

PR #1470, opened 2018-03-20
refactor(lib): convert to futures 0.2.0-beta

### LINKED ISSUE #1448: Update to futures 0.2
It's not released yet, but this is just tracking that it needs to happen and is part of 0.12.

srijs:
Fixes #1448.

Noteworthy changes:

- Removed the dependency on `tokio-service`, instead copied over the `Service` and `NewService` traits for now. I looked into `tower`, but they're still on futures 0.1, and also made some changes to the `Service` trait.
- Not specifying an executor when creating a client will now cause a single background thread to be spawned to handle connection-related tasks. This change was needed since we couldn't rely on the current executor for `Pool::spawn_expired_interval`, but needed a `Send`-able executor instead.
- The logic in `proto::h1::io` to detect vectored I/O needed to adapt to changes in the `AsyncWrite` trait, I removed `WriteBufAuto` and am now relying on the result of `poll_vectored_write` instead.

Aside from that, relatively straight-forward changes. Oh, and of course...

-----

Blockers to merging:

- [ ] [use-after-free fix in tokio/futures2 glue](https://github.com/tokio-rs/tokio/pull/254)
- [ ] [`with_executor` combinator in futures](https://github.com/rust-lang-nursery/futures-rs/pull/902)
- [ ] [futures 2 compatibility in want](https://github.com/seanmonstar/want/pull/1)


src/proto/h1/conn.rs
```diff
@@ -475,25 +476,23 @@ where I: AsyncRead + AsyncWrite,
         }
     }
 
-    pub fn write_body(&mut self, chunk: Option<B>) -> StartSend<Option<B>, io::Error> {
+    pub fn write_body(&mut self, _cx: &mut task::Context, chunk: Option<B>) -> Poll<(), io::Error> {
```
seanmonstar: I think I prefer `cx` being the last argument, as it's "less important"... I filed https://github.com/rust-lang-nursery/futures-rs/issues/883 as well.

seanmonstar: Oh also, does this even need a `Context` argument? Seems it's not used?

src/proto/h1/decode.rs
```diff
@@ -108,7 +109,7 @@ impl Decoder {
                 loop {
                     let mut buf = None;
                     // advances the chunked state
-                    *state = try_ready!(state.step(body, size, &mut buf));
+                    *state = try_ready!(state.step(body, cx, size, &mut buf));
```
seanmonstar: Likewise here, I'd probably suggest the `cx` be the last argument, since we don't really need to notice it.

src/proto/h1/io.rs
```diff
@@ -180,27 +189,82 @@ where
...
+        let mut bufs = [From::from(PLACEHOLDER); 64];
+        let i = Buf::bytes_vec(&buf, &mut bufs[..]);
+        let n = try_ready!(io.poll_vectored_write(cx, &bufs[..i]));
+
+        let mut strategy = Strategy::Auto;
+        if i > 1 {
+            // If there's more than one IoVec available, attempt to
+            // determine the best buffering strategy based on whether
+            // the underlying AsyncWrite object supports vectored I/O.
+            if n == bufs[0].len() {
```
seanmonstar: Yea, this is how I'd probably do this too. :+1:

src/proto/h1/io.rs
```diff
@@ -164,14 +164,23 @@ where
...
         if self.flush_pipeline && !self.read_buf.is_empty() {
             //Ok(())
         } else if self.write_buf.remaining() == 0 {
-            try_nb!(self.io.flush());
+            try_ready!(self.io.poll_flush(cx));
         } else {
             loop {
-                let n = try_ready!(self.io.write_buf(&mut self.write_buf.auto()));
+                let (n, strategy) = try_ready!(write_buf(&mut self.io, cx, &mut self.write_buf));
+                if self.write_buf.strategy == Strategy::Auto {
```
seanmonstar: I think this can still be done with `WriteBufAuto`, that allows only doing the strategy checking as long as it's still `Auto`.

src/server/mod.rs
```diff
@@ -412,17 +410,24 @@ impl<S, B> Server<S, B>
...
                 inner: service,
                 info: Arc::downgrade(&info_cloned),
             };
             info_cloned.lock().unwrap().active += 1;
-            let fut = protocol.serve_connection(socket, s)
-                .map(|_| ())
-                .map_err(move |err| error!("server connection error: ({}) {}", addr, err));
-            spawn(fut);
-            Ok(())
+            let fut = protocol.serve_connection(socket, s).then(move |result| {
```
seanmonstar: I think this can be `FutureExt::recover` also.

src/server/mod.rs
```diff
@@ -412,17 +410,24 @@ impl<S, B> Server<S, B>
             let addr = socket.remote_addr;
             debug!("accepted new connection ({})", addr);
 
-            let service = new_service.new_service()?;
+            let service = match new_service.new_service() {
+                Ok(service) => service,
+                Err(err) => return future::Either::Left(future::err(err))
```
seanmonstar: Why does this return `Either` now? It's not immediately clear to me.

src/lib.rs
```diff
@@ -18,7 +18,7 @@
 
 extern crate bytes;
 #[macro_use] extern crate futures;
```
seanmonstar: I think there's a recommendation that libraries shouldn't depend on `futures` directly, but rather on the individual crates (like `futures-core`, `futures-channel`, etc). The idea is that hyper could keep working if some extension needed a breaking change, and we only ever publicly claimed to depend on `futures-core`...

It'd be good to take inventory of what public dependencies we may be exposing here, either directly or indirectly:

- `futures-core` since things like `FutureResponse` implement `Future`.
- `futures-io` since we accept `AsyncRead` and `AsyncWrite`...
- I think all usage of `futures-channel` is no longer publicly exposed, so not a public dependency, right?
- While we use `FutureExt` internally, any types returned from it aren't viewable from outside the crate, I think.

src/client/conn.rs
```diff
@@ -186,7 +187,7 @@ where
     pub fn send_request(&mut self, req: Request<B>) -> ResponseFuture {
         let inner = match self.dispatch.send(req) {
             Ok(rx) => {
-                Either::A(rx.then(move |res| {
+                Either::Left(rx.then(move |res| {
```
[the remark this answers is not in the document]
srijs: Looks like it, yeah!

- [ ] Use `.left()` and `.right()` combinators

src/client/connect.rs
```diff
@@ -320,14 +317,13 @@ impl Future for HttpConnecting {
...
-                        let work = dns::Work::new(host, port);
-                        state = State::Resolving(oneshot::spawn(work, executor));
+                        let host = ::std::mem::replace(host, String::new());
+                        state = State::Resolving(dns::Resolving::spawn(host, port, executor));
                     }
                 },
                 State::Resolving(ref mut future) => {
-                    match try!(future.poll()) {
-                        Async::NotReady => return Ok(Async::NotReady),
+                    match try!(future.poll(cx)) {
```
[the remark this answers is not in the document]
srijs: No worries, I'll fix it up!

- [ ] use `try_ready!` instead of explicit match

src/client/mod.rs
```diff
@@ -202,13 +202,18 @@ where C: Connect<Error=io::Error> + Sync + 'static,
...
                 connector.connect(dst)
-                    .from_err()
+                    .err_into()
                     .and_then(move |(io, connected)| {
                         conn::Builder::new()
                             .h1_writev(h1_writev)
                             .handshake_no_upgrades(io)
                             .and_then(move |(tx, conn)| {
-                                executor.execute(conn.map_err(|e| debug!("client connection error: {}", e)))?;
+                                executor.execute(conn.then(|result| {
```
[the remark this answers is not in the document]
srijs: Oh, neat!

- [ ] use `recover`

src/client/mod.rs
```diff
@@ -218,18 +223,18 @@ where C: Connect<Error=io::Error> + Sync + 'static,
...
-        let executor = self.executor.clone();
+        let race = checkout.select(connect).map(|either| {
+            either.either(|(pooled, _)| pooled, |(pooled, _)| pooled)
+        }).map_err(|either| {
+            // the Pool Checkout cannot error, so the only error
+            // is from the Connector
+            // XXX: should wait on the Checkout? Problem is
+            // that if the connector is failing, it may be that we
+            // never had a pooled stream at all
+            ClientError::Normal(either.either(|(e, _)| e, |(e, _)| e))
```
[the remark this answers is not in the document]
srijs: Yeah, `Either` used to have this useful [`split`](https://docs.rs/futures/0.1.18/futures/future/enum.Either.html#method.split), that's gone away in futures 0.2 unfortunately.

Maybe just replace it with a `match`?

src/client/mod.rs
```diff
@@ -254,13 +259,12 @@ where C: Connect<Error=io::Error> + Sync + 'static,
...
-                        // If the executor doesn't have room, oh well. Things will likely
-                        // be blowing up soon, but this specific task isn't required.
-                        let _ = executor.execute(future::poll_fn(move || {
-                            pooled.tx.poll_ready().map_err(|_| ())
-                        }));
-                    }
+                    //
+                    // If the executor doesn't have room, oh well. Things will likely
+                    // be blowing up soon, but this specific task isn't required.
+                    let _ = executor.execute(future::poll_fn(move |cx| {
```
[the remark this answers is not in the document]
srijs: I'll look into that!

- [ ] look into adding `SendRequest::is_ready` method

src/client/mod.rs
```diff
@@ -273,7 +277,7 @@ where C: Connect<Error=io::Error> + Sync + 'static,
     }
 
     fn schedule_pool_timer(&self) {
-        self.pool.spawn_expired_interval(&self.executor);
+        self.pool.spawn_expired_interval(&mut self.executor.clone())
```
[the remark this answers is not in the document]
srijs: It's needed because the method itself takes `&self`, so `self.executor` is `&Exec`. Cloning makes it owned and therefore mutable.

The method needs to take `&self` because it is called from `Client::request`, which takes `&self`. Maybe we could fix this up if we moved `schedule_pool_timer` into the constructor for `Client`?

src/client/mod.rs
```diff
@@ -632,65 +636,33 @@ impl<C: Clone, B> Clone for Config<C, B> {
...
-
-    impl Future for Background {
-        type Item = ();
-        type Error = ();
-
-        fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-            self.inner.poll()
-        }
+impl Default for Exec {
+    fn default() -> Exec {
```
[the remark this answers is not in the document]
srijs: I'm not sure about that, actually. I know that this is how `CpuPool` used to work, but I've recently worked with `ThreadPool` and I'm relatively sure that it will multiplex multiple futures onto a single thread.

Effectively each thread is a while loop that reads tasks off a channel and polls them once, then repeats. The waker for each task will cause it to be re-submitted to the channel so it'll get polled again.

Can you take a look at https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-executor/src/thread_pool.rs#L124 and let me know if your understanding matches mine?

src/lib.rs
```diff
@@ -18,7 +18,7 @@
...
-extern crate futures_cpupool;
+//extern crate futures_cpupool;
```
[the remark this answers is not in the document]
srijs: Good catch!

- [ ] delete `futures_cpupool` import

src/proto/h1/conn.rs
srijs: Oh yeah, in theory we could remove that one! I guess my thinking here was that since `Conn` is a public API, we might want to reserve the possibility of needing `Context` for future changes? Maybe that's a YAGNI though, so you tell me!

src/proto/h1/decode.rs
srijs: No problem!

- [ ] move `cx` to be the last argument

src/proto/h1/io.rs
srijs: Can you go into more detail about that? Why would we need `WriteBufAuto`?

src/server/mod.rs
srijs: It returns either because further down we return the `Spawn` future from the same function. The `Spawn` future is lazy and needs to be polled to spawn the future (since it's using the executor from the context to spawn).

src/lib.rs
srijs: Good points! We have the same situation with `tokio`, where I was gonna fix that up in another PR.

Your call whether to do it as part of this PR, or whether I'll follow up with a combined `futures` and `tokio` PR for it.

src/server/mod.rs
srijs: Will do!

srijs:
I'll take a look at `futures-compat`, thanks! One thing is that we'll also need compat between `tokio::io::Async{Read, Write}` and `futures::io::Async{Read, Write}`. Would you consider that in-scope for `futures-compat`, or would that live somewhere else?

aturon:
@srijs @seanmonstar 

The `unstable` bit is just waiting until futures 0.2.0 is published. We're putting out a `beta` this afternoon and otherwise are pretty much ready to go. So I would advise not investing time working around that flag.

src/client/mod.rs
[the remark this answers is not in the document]
seanmonstar: Ah OK, right. That `schedule_pool_timer` was introduced as a bit of a hack, and it's still being all hacky. I guess I'd move it to being called from `Checkout::poll`, so that it's only created once the client starts to get used, and there's a `Context` to grab.

src/client/mod.rs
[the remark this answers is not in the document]
srijs: Luckily my PR to add a `Context` arg to `future::lazy` just landed in time for futures 0.2 beta, so I fixed it using that instead!

srijs:
Addressed a bunch of stuff, as well as fixed everything to work with the futures 0.2 beta release!

@seanmonstar I was fighting the `conn_drop_prevents_pool_checkout` test this afternoon because it was flakey. I've made some changes that I _think_ are okay to make the test run deterministically, but I'd appreciate you taking a look at telling me if the changes are not okay.

src/proto/h1/decode.rs
srijs: On second thought, I guess I'll keep this open until https://github.com/rust-lang-nursery/futures-rs/issues/883 is settled?

src/client/mod.rs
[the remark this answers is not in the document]
srijs: Yeah, it is pretty unfortunate. I guess the way to think about it is `cx.spawn` should be used for short-lived futures (or futures that'll either likely complete within the lifetime of the task, or that are not essential to complete).

For long lives futures (such as pool timers and connection management), I believe we need to have the `Client` accept an executor that the user guarantees will be long-lived (such as the tokio runtime), or maintain a thread ourselves for this purpose.

srijs:
It's not immediately apparent to me why the 2/4 AppVeyor builds failed. Any pointers?

src/client/mod.rs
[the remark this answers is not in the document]
aturon: Oops, sorry, @seanmonstar and I talked about this last night but I neglected to follow up with a comment.

TL;DR: I think this was a design mistake with `block_on`, and I'm [correcting it](https://github.com/rust-lang-nursery/futures-rs/pull/890) by having it spawn tasks onto a global thread pool. Eventually that pool will be configurable as well.

But bottom line: we want you to be able to assume that you can use `cx.spawn` freely, even for long-running tasks.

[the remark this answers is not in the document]
seanmonstar: Yea, so I think that we should indeed just use `cx.spawn`, and not hold any executor inside the `Client`.

(The tests might need to be changed to stop using `block_on`. At least, it definitely won't work with the published 0.2 beta.)

src/proto/h1/decode.rs
seanmonstar: I don't want to burden you with my stylistic whims. If you want to leave things the way they are, I may just shuffle argument orders as I'm touching various files in the future.

src/proto/h1/io.rs
seanmonstar: I mostly was thinking of these reasons:

1. Allows the strategy swapping code to be separate from this, which is just wanting to write until blocked.
2. The checks to determine what strategy to switch to could probably only have to happen when strategy it still auto.

src/server/mod.rs
seanmonstar: Oh, weird! (That `spawn` API feels a little odd, but not your fault :))

seanmonstar:
> It's not immediately apparent to me why the 2/4 AppVeyor builds failed. Any pointers?

Hm, when it just exists like that with an error code, my assumption is that the process was aborted (either a double panic, or segfault). I'll restart and see poke around.

src/proto/h1/io.rs
srijs: Ah, cool! I can definitely factor out the strategy swapping code, I guess `WriteBufAuto` would now implement some sort of `write_into` method that takes a `AsyncWrite`?

- [ ] factor out strategy swapping code

src/server/mod.rs
srijs: - [ ] change to use `.left()` and `.right()`

src/client/mod.rs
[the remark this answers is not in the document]
srijs: Ah, that looks like a great way to solve the problem!

I'm a bit hesitant to go about changing all the tests, especially since once Aaron's PR is merged, `block_on` will be the simplest way to write that test code.

I guess I could try to shim the new `block_on` using `ThreadPool`?

seanmonstar:
So, searching for the exit code, its a segfault basically. Derefencing a bad pointer. This might be fixed in https://github.com/tokio-rs/tokio/pull/243

src/client/mod.rs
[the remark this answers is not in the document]
seanmonstar: If you wanted to update the `Cargo.toml` to point at the git repo of futures, to get things working, that's fine!

srijs:
Still seeing issues after using Carl's `races` branch. Commented on the PR in tokio.

srijs:
@seanmonstar Nice work on finding that use-after-free bug in tokio! 👏

I've targeted your branch with the fix, and at least locally the tests are running reliably now. Pushed the changes up to also check with Travis and AppVeyor.

src/client/mod.rs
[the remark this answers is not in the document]
srijs: I tried this, but then the types will mismatch with `futures 0.2.0-beta`, which `tokio` depends on. I'll try shimming it instead.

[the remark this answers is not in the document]
seanmonstar: You need to use the `[replace]` part, so that all dependencies use the same library also.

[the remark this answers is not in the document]
srijs: Yay, that works!

src/client/mod.rs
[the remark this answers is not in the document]
srijs: Moved it into `Checkout::poll` now.

srijs:
@seanmonstar Addressed your feedback so far, I think this should be ready for a second round of review!

src/client/dns.rs
```diff
@@ -6,27 +6,42 @@ use std::net::{
...
+impl Resolving {
+    pub fn spawn(host: String, port: u16, executor: &mut Executor) -> Resolving {
+        let (sender, receiver) = oneshot::channel();
+        executor.spawn(Box::new(lazy(move |_| {
+            debug!("resolving host={:?}, port={:?}", host, port);
+            let result = (host.as_ref(), port).to_socket_addrs()
+                .map(|i| IpAddrs { iter: i });
+            sender.send(result).ok();
+            Ok(())
+        }))).unwrap();
```
seanmonstar: It looks like the receiver will return an error if the sender was dropped, so we should be able to just ignore this spawn error, right?

src/client/mod.rs
```diff
@@ -85,9 +86,9 @@ impl<C, B> Client<C, B> {
     fn configured(config: Config<C, B>, exec: Exec) -> Client<C, B> {
         Client {
             connector: Arc::new(config.connector),
-            executor: exec,
+            executor: exec.clone(),
```
seanmonstar: We should be able to completely remove the executor field of the client, and always use `cx`.

(`HttpConnector` should still own its own executor, but that's because it needs to spawn blocking tasks. It should be enough to allow configuring the `HttpConnector` manually.)

srijs: - [ ] remove custom executor

srijs:
All done! I'm using my branch of futures momentarily until https://github.com/rust-lang-nursery/futures-rs/pull/902 is merged.

src/client/mod.rs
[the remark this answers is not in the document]
srijs: We can change this once https://github.com/bluss/either/pull/19 is merged!

seanmonstar:
@srijs Awesome work! I started a checklist in the original description to track what is blocking merging this. I realized I didn't know all of what that is, if you'd like to add to this list...

srijs:
@seanmonstar Good stuff, I added 2 PRs that we should get merged, one of them is https://github.com/seanmonstar/want/pull/1.

As to other merge blockers, would you consider anything that needs a release (and is sourced from git right now) a blocker, or is okay to fix up later as the new versions get published?

seanmonstar:
OK, so I'm thinking: screw it, we just merge this, and track that other stuff in new issues. Having it stagnate here doesn't help, and prevents working on newer things that need these changes. So, unless there's objections, I'll just merge as-is in a couple hours :D

xrl:
@seanmonstar @srijs why was this work reverted?

seanmonstar:
@xrl see here for why: https://github.com/hyperium/hyper/issues/1481

PR #1478, opened 2018-04-03
fix(lib): move to futures master

srijs:
Just a small patch, but my `with_executor` changes got merged into futures, so we can point to the main futures repo now instead.


seanmonstar:
Yay! Thanks!

Issue #1481, opened 2018-04-07
Determine futures version support


seanmonstar:
So, with the release of `futures` 0.2, it was revealed that [this is temporary state](https://www.reddit.com/r/rust/comments/8ac85w/futures_02_is_here/dwxm9on/), and that breakage was coming again real soon. Unfortunately, [0.2 support isn't fully there in Tokio](https://github.com/tokio-rs/tokio/issues/251), and it sounds like the recommendation from the Futures team is to just wait till 0.3 (or libstd or whatever).

Sooooo, we kind of still need to provide futures 0.1 support. I don't know if there is value in maintaining a 0.2 support behind a feature flag. It feels horrible to revert all the work that was done, but 0.12 cannot be released as-is.

I've spent some time on [futures-compat](https://github.com/seanmonstar/futures-compat), and it does allow compiling hyper master (with 0.2 futures) against tokio with no feature flag (using 0.1 futures). However, I don't think I'd personally want to release with 0.2 wrapping tokio in compat, since there is a penalty cost of cloning notifiers.

Perhaps the best action is to revert to a public 0.1 futures API, and publish `futures-compat` to allow anyone that wants to explore futures 0.2 to be able to do so, but production use of hyper doesn't degrade.

jimmycuadra:
It is out of the question to just wait for futures 0.3 before releasing hyper 0.12?

seanmonstar:
That is a possibility, but otherwise 0.12 is pretty close to ready. There'd also be much less breaking changes... Well, somewhat.

The question would be how long do we wait? Continue to hold off 1) upgrade to http crate, 2) upgrade to tokio, 3) http2, 4) `Send` Client?

The existing APIs in 0.11 are a little confusing to use.

scottlamb:
fwiw, I'm looking forward to the hyper 0.12.x API changes and would be sad if it were delayed for months because of this. futures 0.3 is apparently coming "relatively soon", but it will be nightly-only at first, so unsure when it's actually coming for stable.

jimmycuadra:
Another way to look at it... Given:

* The ecosystem is in major flux right now
* Anyone who's paying attention knows everything is pretty unstable until Rust 2018 in September
* futures has just set precedent for releasing a new version at the same time as announcing that its replacement is imminent

it's pretty reasonable to release an 0.12 as is. And maybe announce, similarly to futures, that an 0.13 will be coming sooner than it's been between previous versions of hyper.

[...]

PR #1482, opened 2018-04-07
Revert futures 0.2 changes

seanmonstar:
See #1481 

:cry:


srijs:
💔

From the CI errors, it looks like we might need to back-port some of the fixes I included for the new tokio behaviour. I'll see if I can figure out which ones they are!

kamyuentse:
On my local machine, the test `conn_drop_prevents_pool_checkout` can pass some time.
Seems there is a race somewhere, I am trying locate the problem.

srijs:
Yeah, it‘s a race condition I fixed as part of the futures 0.2 changes
that are being reverted here, should be a matter of backporting it!

tests/client.rs
```diff
@@ -1043,27 +1042,28 @@ mod dispatch_impl {
             let mut buf = [0; 4096];
             sock.read(&mut buf).expect("read 1");
             sock.write_all(b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n").unwrap();
-            //sock.read(&mut buf).expect("read 2");
+            sock.read(&mut buf).expect("read 2");
```
seanmonstar: Oh, this is what made it pass always :D

kamyuentse:
I want to know why removing that test case.

seanmonstar:
@kamyuentse I kept trying variants, and couldn't seem to get it passing 100% of the time without disabling the exact thing it was testing. And so then I read through it, trying to remember what the heck it was even testing, and realized it was because before inside `client/mod.rs` there was a `tx.start_send(req).unwrap()`. That doesn't exist anymore, and the test is really contrived.

Actually, I'm *fairly* certain the problem is the race condition in futures mpsc queue: https://github.com/rust-lang-nursery/futures-rs/issues/909

That'd make sense, since the test was basically dropping the `Receiver` in a runtime thread, and trying to use the `Sender` in the test thread, and occasionally seeing deadlocks as the message was never dropped.

kamyuentse:
@seanmonstar Thanks for your explanation! I don't know whether futures issue cause tokio worker panic, but I think I should cc @carllerche .

PR #1487, opened 2018-04-16
feat(client): add support to set SO_NODELAY on client HTTP sockets

### LINKED ISSUE #1473: TCP_NODELAY on client
How can I set nodelay on an HTTP client connection? If it's not exposed by hyper, can I get access to the underlying `TcpStream` so I can call [`TcpStream::set_nodelay`](https://tokio-rs.github.io/tokio-core/tokio_core/net/struct.TcpStream.html#method.set_nodelay)?

tee-too:
Add configuration on `HttpConnector` to set `SO_NODELAY` on client
HTTP sockets.

Closes #1473

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -170,6 +170,7 @@ pub struct HttpConnector {
     enforce_http: bool,
     handle: Option<Handle>,
     keep_alive_timeout: Option<Duration>,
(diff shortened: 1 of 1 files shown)


seanmonstar:
Awesome, thanks!

PR #1488, opened 2018-04-16
feat(server): re-design `Server` as higher-level API

### LINKED ISSUE #1263: Http::bind does not scale to uses where a Handle is needed

### LINKED ISSUE #1322: Creating Server with a Handle

seanmonstar:
The `hyper::Server` is now a proper higher-level API for running HTTP
servers. There is a related `hyper::server::Builder` type, to construct
a `Server`. All other types (`Http`, `Serve`, etc) were moved into the
"lower-level" `hyper::server::conn` module.

The `Server` is a `Future` representing a listening HTTP server. Options
needed to build one are set on the `Builder`.

As `Server` is just a `Future`, it no longer owns a thread-blocking
executor, and can thus be run next to other servers, clients, or
what-have-you.

Closes #1322
Closes #1263

BREAKING CHANGE: The `Server` is no longer created from `Http::bind`,
  nor is it `run`. It is a `Future` that must be polled by an
  `Executor`.

  The `hyper::server::Http` type has move to
  `hyper::server::conn::Http`.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -5,15 +5,15 @@ extern crate pretty_env_logger;
 extern crate tokio;
 
 use futures::Future;
-use futures::future::lazy;
 
 use hyper::{Body, Response};
-use hyper::server::{Http, const_service, service_fn};
+use hyper::server::{Server, const_service, service_fn};
 
 static PHRASE: &'static [u8] = b"Hello World!";
 
 fn main() {
     pretty_env_logger::init();
+
     let addr = ([127, 0, 0, 1], 3000).into();
 
     let new_service = const_service(service_fn(|_| {
@@ -21,13 +21,11 @@ fn main() {
         Ok::<_, hyper::Error>(Response::new(Body::from(PHRASE)))
     }));
 
-    tokio::run(lazy(move || {
-        let server = Http::new()
-            .sleep_on_errors(true)
-            .bind(&addr, new_service)
-            .unwrap();
+    let server = Server::bind(&addr)
+        .serve(new_service)
+        .map_err(|e| eprintln!("server error: {}", e));
 
-        println!("Listening on http://{} with 1 thread.", server.local_addr().unwrap());
-        server.run().map_err(|err| eprintln!("Server error {}", err))
-    }));
+    println!("Listening on http://{}", addr);
+
+    tokio::run(server);
 }
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -2,797 +2,175 @@
 //!
 //! A `Server` is created to listen on a port, parse HTTP requests, and hand
 //! them off to a `Service`.
+//!
+//! There are two levels of APIs provide for constructing HTTP servers:
+//!
+//! - The higher-level [`Server`](Server).
+//! - The lower-level [conn](conn) module.
 
 pub mod conn;
 mod service;
+mod tcp;
 
 use std::fmt;
-use std::io;
-use std::net::{SocketAddr, TcpListener as StdTcpListener};
-use std::sync::{Arc, Mutex, Weak};
+use std::net::SocketAddr;
 use std::time::Duration;
 
-use futures::task::{self, Task};
-use futures::future::{self, Either, Executor};
-use futures::{Future, Stream, Poll, Async};
-use futures_timer::Delay;
+use futures::{Future, Stream, Poll};
 use http::{Request, Response};
 use tokio_io::{AsyncRead, AsyncWrite};
-use tokio::spawn;
-use tokio::reactor::Handle;
-use tokio::net::TcpListener;
 pub use tokio_service::{NewService, Service};
 
 use body::{Body, Payload};
-use common::Exec;
-use proto;
-use self::addr_stream::AddrStream;
+// Renamed `Http` as `Http_` for now so that people upgrading don't see an
+// error that `hyper::server::Http` is private...
+use self::conn::{Http as Http_, SpawnAll};
 use self::hyper_service::HyperService;
+use self::tcp::{AddrIncoming};
 
-pub use self::conn::Connection;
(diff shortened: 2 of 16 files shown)


examples/hello.rs
```diff
@@ -5,29 +5,28 @@ extern crate pretty_env_logger;
...
         Ok::<_, hyper::Error>(Response::new(Body::from(PHRASE)))
     }));
 
-    tokio::run(lazy(move || {
-        let server = Http::new()
-            .sleep_on_errors(true)
-            .bind(&addr, new_service)
-            .unwrap();
+    let server = Server::bind(&addr)
+        .expect("bind to address")
```
seanmonstar: I'm strongly considering making the default `Server::bind` either a) panic if binding errors, or b) lazily binding on the first `poll` of the `Server`. Forcing users to deal with an error that isn't likely to happen every single time feels like an annoying speed bump. Anyone wanting to handle a bind error can bind a `TcpListener` themselves, and use `Builder::new`.

src/server/mod.rs
```diff
@@ -2,797 +2,164 @@
...
-    exec: Exec,
-    http2: bool,
-    keep_alive: bool,
-    max_buf_size: Option<usize>,
-    pipeline: bool,
-    sleep_on_errors: bool,
-}
+#[deprecated(note="import from hyper::server::conn, or consider hyper::Server")]
+#[doc(hidden)]
+pub type Http = Http_;
```
seanmonstar: I'm undecided on whether to keep this, or remove it. I did this since it allowed providing a hint in the message that things have changed, instead of simply `hyper::server::Http is private`. But it may be annoying to keep around for all of 0.12.x...

Issue #1474, opened 2018-03-30
Server cannot serve requests with small `max_buf_size`


yjh0502:
If `max_buf_size` is smaller than [INIT_BUFFER_SIZE](https://github.com/hyperium/hyper/blob/v0.11.24/src/proto/h1/io.rs#L13), then server will close a client connection immediately [1] because `max_buf_size` is always smaller than `INIT_BUFFER_SIZE` before receiving any data from a client.

[1] https://github.com/hyperium/hyper/blob/v0.11.24/src/proto/h1/io.rs#L119-L122

seanmonstar:
Woops! Is there a need to have a max buffer size of less than 8kb? There is a performance cost of having to do another `cmp::min` to determine the init buffer size...

Or could we just enforce 8kb as the minimum, regardless of config?

estk:
Hi!
I'd like to try to take this issue on. After pulling the code and looking around I have a few questions. It seems that the max_buf_size is set in a chain through server/conn.rs -> h1/conn.rs -> h1/io.rs. The actual use case seems to be something like this:

```
let http = Http::new().max_buf_size(...);
let conn = http.serve_connection(some_io, some_service);
```

So, where should the error be exposed to the user?
It would seem that from an api perspective it would be awkward to have max_buf_size() have any return value besides Self. So should serve_connection return a result that could potentially create a max_buf too small error? Both options dont seem great for me. Hoping the maintainers will have some insight. :)

seanmonstar:
@estk thanks for looking into this!

I think this should probably do this:

1. in `Http::max_buf_size`, compare the provided value with the minimum allowed, and use which is bigger.
2. If the value were smaller than allowed minimum, perhaps log a warning or error. It could even panic, using an `assert!`, I'm not sure which is best.
3. Add a line to the documentation of `max_buf_size` that the minimum is enforced.

PR #1489, opened 2018-04-17
Enforce minimum max buf size.

estk:
Attempts to address #1474 

Todo
- [ ] Tests


estk:
@seanmonstar Is this what you were hoping for?

src/server/conn.rs
```diff
@@ -134,10 +134,16 @@ impl Http {
         self
     }
 
+
     /// Set the maximum buffer size for the connection.
     ///
     /// Default is ~400kb.
+    /// The minimum value that h1 can handle is enforced and will panic if smaller.
```
seanmonstar: I'd put this in a `# Panics` section, and include the exact minimum, so that people can see what they need to guard against. Something like:

```
/// # Panics
///
/// The minimum value allowed is 8192. This method panics if the passed `max` is less than the minimum.
```

estk: Nice, didn't know about this. Thanks for the guidance!

PR #1490, opened 2018-04-17
feat(service): introduce hyper-specific `Service`

### LINKED ISSUE #1461: Update Service trait from tokio-service to tower

seanmonstar:
This introduces the `hyper::service` module, which replaces
`tokio-service`.

Since the trait is specific to hyper, its associated
types have been adjusted. It didn't make sense to need to define
`Service<Request=http::Request>`, since we already know the context is
HTTP. Instead, the request and response bodies are associated types now,
and slightly stricter bounds have been placed on `Error`.

The helpers `service_fn` and `service_fn_ok` should be sufficient for
now to ease creating `Service`s.

The `NewService` trait now allows service creation to also be
asynchronous.

These traits are similar to `tower` in nature, and possibly will be
replaced completely by it in the future. For now, hyper defining its own
allows the traits to have better context, and prevents breaking changes
in `tower` from affecting hyper.

Closes #1461

BREAKING CHANGE: The `Service` trait has changed: it has some changed
  associated types, and `call` is now bound to `&mut self`.

  The `NewService` trait has changed: it has some changed associated
  types, and `new_service` now returns a `Future`.

  `Client` no longer implements `Service` for now.

  `hyper::server::conn::Serve` now returns `Connecting` instead of
  `Connection`s, since `new_service` can now return a `Future`. The
  `Connecting` is a future wrapping the new service future, returning
  a `Connection` afterwards. In many cases, `Future::flatten` can be
  used.


src/service/service.rs
```diff
@@ -0,0 +1,165 @@
...
+use futures::{future, Future, IntoFuture};
+
+use body::Payload;
+use common::Never;
+use ::{Request, Response};
+
+/// An asynchronous function from `Request` to `Response`.
+pub trait Service {
+    /// The `Payload` body of the `http::Request`.
+    type ReqBody: Payload;
```
sfackler: Doesn't Hyper control what type the request body is?

seanmonstar: True. I'd like to bring the implementation back for `Client`, which will need to define the `ReqBody`. I'm just unsure about how best to move forward to `tower::Service`, which has a `poll_ready` function (and [may be changing soon](https://github.com/tower-rs/tower/issues/66))...

Issue #1191, opened 2017-05-27
Force push the gh-pages branch


dtolnay:
The output of `git log --graph --all --pretty=format:"%h%d %s"` is annoying because the first 577 lines are all "rebuild pages" commits. Following every commit to master there is a commit to gh-pages, so this long gh-pages history will always show up first. In docs.sh it should force-push the gh-pages branch so there is only ever 1 commit on that branch.

seanmonstar:
I've never tried to view the log like that before... This may be beyond my git knowledge, is there a way to wipe the history of a branch? Hm.. perhaps by finding the first commit of master, checking it out on gh-pages, then copying the contents over, and force pushing?

dtolnay:
It is beyond my git knowledge too! A long time ago I found this awesome alias that I stuck in `.gitconfig` and have been using ever since.

```
[alias]
    lg = log --graph --all --color --pretty=format:\"%x1b[31m%h%x1b[32m%d%x1b[0m%x20%s\" -16
```

![selection_054](https://cloud.githubusercontent.com/assets/1940490/26559258/652dfe74-4463-11e7-842e-2f2e0e433ee1.png)

---

[This may be helpful.](https://github.com/brson/rust-cookbook/blob/440a39436ed9402996f12100c20b051bb69dde49/deploy.sh#L53-L60). It is doing `mdbook build` instead of `cargo doc` but the force push part should work the same.

seanmonstar:
There was an attempt in #1495, but some concern around if gh-pages would build with dates in the past. I don't personally ever `git log --all` (my personal alias, `git l`, is just `git log --oneline -20 --no-merges`). As such, I'm going to close the issue, not because I'm against it (if someone provides a fix, cool!), but to remove it from the list of things that I look at to determine what is broken in hyper. Thanks!

PR #1495, opened 2018-04-21
chore(ci): hide gh-pages branch from git log --all

alsuren:
This is done by making the committer date sufficiently old.

28 years is chosen because that's how often weekdays line up
with days of the month reliably. This should continue to be
the case until 2100, when we next skip a leap year.

This should solve https://github.com/hyperium/hyper/issues/1191 without the need to force-push.

to test manually, I did 
```
GIT_AUTHOR_DATE=$((`date +%s`-60*60*24*(365*28+7))) GIT_COMMITTER_DATE=$((`date +%s`-60*60*24*365*20)) git commit -am "try breaking the date"
```
 on the gh-pages branch, and then did `git log --oneline --graph --all` and the gh-pages branch wasn't there (it turns out that breaking the date of the head commit is enough to remove the entire branch from view).

### DIFF
--- a/.travis/docs.sh
+++ b/.travis/docs.sh
@@ -13,15 +13,19 @@ cd deploy_docs
 git config user.name "Sean McArthur"
 git config user.email "sean.monstar@gmail.com"
 
+# Fake up the date to be ages ago, so that it doesn't pollute the output of
+#   git log --graph --all --oneline
+DATE=$((`date +%s`-60*60*24*(365*28+7)))
+
 if [ "$TRAVIS_TAG" = ""  ]; then
     rm -rf master
     mv ../target/doc ./master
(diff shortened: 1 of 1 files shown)


alsuren:
oh... github actually handles time-travelling really poorly doesn't it? (I have trouble when I re-order my commits when rebasing, and my test commit (HEAD^ on https://github.com/alsuren/hyper/commits/gh-pages) doesn't render at all).

I can write a `git filter-branch` script to rewrite the history if you want, or you can do a single force-push to flatten out all of the history into a single commit with an appropriate timestamp.

seanmonstar:
Thanks for coming up with this interesting hack :)

For now, I'm going to close, since it's not clear if gh-pages will continue to work correctly.

PR #1496, opened 2018-04-21
docs(README): Use HTTPS, update website link

salty-horse:
Some links used HTTP. The badge image, specifically, caused an HTTPS warning on crates.io.

I also updated the website link from https://hyperium.github.io to https://hyper.rs


seanmonstar:
Good catch, thanks!

PR #1497, opened 2018-04-23
feat(client): add support for title case header names at the socket level

### LINKED ISSUE #1492: Headers are lower-cased when sent and no option to disable this feature.
I am trying to use Hyper with a stubborn web service that does not like requests with headers that do not have the first letter of each word capitalized (`Content-Length` for example). I tried out some other HTTP client libraries for Rust (like `tk-http` which hasn't been updated recently and `mio_httpc` which doesn't work all the time with said web service), but Hyper doesn't have the issues as with the other libraries except for the header name mangling.

After reading the RFC regarding header casing, I see that header names are supposed to be case-insensitive, so Hyper is not at fault here and from looking at `HeaderMap` I see that it relies on headers being lower-cased for performance.

If it is not possible to accommodate case-sensitive header names, then could it be possible to directly provide an array of headers to be used in an HTTP request?

mbilker:
This introduces support for the HTTP/1 Client to write header names as title case when encoding
the request.

Need to work on tests.

Closes #1492

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -50,6 +50,7 @@ where I: AsyncRead + AsyncWrite,
                 error: None,
                 keep_alive: KA::Busy,
                 method: None,
+                title_case_headers: false,
                 read_task: None,
                 reading: Reading::Init,
                 writing: Writing::Init,
@@ -71,6 +72,10 @@ where I: AsyncRead + AsyncWrite,
 
     pub fn set_write_strategy_flatten(&mut self) {
         self.io.set_write_strategy_flatten();
+    }
+
+    pub fn set_title_case_headers(&mut self) {
+        self.state.title_case_headers = true;
     }
 
     pub fn into_inner(self) -> (I, Bytes) {
@@ -430,7 +435,7 @@ where I: AsyncRead + AsyncWrite,
         self.enforce_version(&mut head);
 
         let buf = self.io.write_buf_mut();
-        self.state.writing = match T::encode(head, body, &mut self.state.method, buf) {
+        self.state.writing = match T::encode(head, body, &mut self.state.method, self.state.title_case_headers, buf) {
             Ok(encoder) => {
                 if !encoder.is_eof() {
                     Writing::Body(encoder)
@@ -620,6 +625,7 @@ struct State {
     error: Option<::Error>,
     keep_alive: KA,
     method: Option<Method>,
+    title_case_headers: bool,
     read_task: Option<Task>,
     reading: Reading,
     writing: Writing,
@@ -649,6 +655,7 @@ impl fmt::Debug for State {
             .field("keep_alive", &self.keep_alive)
             .field("error", &self.error)
             //.field("method", &self.method)
+            //.field("title_case_headers", &self.title_case_headers)
             .field("read_task", &self.read_task)
             .finish()
     }
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -126,6 +126,7 @@ where
         mut head: MessageHead<Self::Outgoing>,
         body: Option<BodyLength>,
         method: &mut Option<Method>,
+        _title_case_headers: bool,
         dst: &mut Vec<u8>,
     ) -> ::Result<Encoder> {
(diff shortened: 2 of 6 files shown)


src/proto/h1/conn.rs
```diff
@@ -649,6 +655,7 @@ impl fmt::Debug for State {
             .field("keep_alive", &self.keep_alive)
             .field("error", &self.error)
             //.field("method", &self.method)
+            .field("title_case_headers", &self.title_case_headers)
```
seanmonstar: I'm thinking that perhaps this field shouldn't be logged, since the `State` gets dumped somewhat frequently to help debug the current state, and seeing this longer field being the same constant forever will just be noisy.

src/proto/h1/role.rs
```diff
@@ -170,7 +171,11 @@ where
             extend(dst, head.subject.canonical_reason().unwrap_or("<none>").as_bytes());
             extend(dst, b"\r\n");
         }
-        write_headers(&head.headers, dst);
+        if title_case_headers {
+            write_headers_title_case(&head.headers, dst);
+        } else {
+            write_headers(&head.headers, dst);
+        }
```
seanmonstar: Since there's no way to set this for the server yet (and I'm tempted to leave it out until someone asks for it), the bool check is kind of a waste. Whatcha think of removing it from the `ServerTransaction` version here?

src/proto/h1/role.rs
```diff
@@ -635,6 +645,40 @@ impl<'a> Iterator for HeadersAsBytesIter<'a> {
     }
 }
 
+// Write header names as title case. The header name is assumed to be ASCII,
+// therefore it is trivial to convert an ASCII character from lowercase to
+// uppercase. It is as simple as XORing the lowercase character byte with
+// space.
```
seanmonstar: I believe this will get a couple characters changed completely. Specifically, `|` and `~` will become `\\` and `^`, respectively. Probably not super common, but it'd be a semantic change...

Also, it'd be nice to include some tests of `title_case`, perhaps at the bottom of the file.

src/proto/h1/conn.rs
mbilker: I agree with that reasoning. Should it be commented like the `method` field to indicate it isn't logged or just get rid of the line entirely?

src/proto/h1/role.rs
mbilker: Oh, I forgot to think of those. A check to see if the byte falls between 97 (`a`) and 172 (`z`) should suffice.

src/proto/h1/role.rs
mbilker: Since `ServerTransaction` implements `Http1Transaction`, I cannot just remove it from the method signature as that would deviate from the trait definition, but I can prefix it with an underscore to signify that the boolean is unused.

mbilker:
I noticed that the encode methods are not tested at the bottom of the file, but they are heavily tested in the integration tests of `Client`. I will add some tests for the `title_case` method and some integration tests to check for title cased headers through the client request.

src/proto/h1/role.rs
```diff
@@ -857,6 +901,21 @@ mod tests {
         Client::decoder(&head, method).unwrap_err();
     }
 
+    #[test]
+    fn test_client_request_encode_title_case() {
```
seanmonstar: Great, awesome!

src/proto/h1/role.rs
mbilker: I read the octal field by accident from the man page. I fixed it to a `>= b'a'` and `<= b'z'`.

mbilker:
I modified the client integration test macros to accept a `title_case_headers` parameter and I added a test using said change to ensure the client is encoding the headers as title case.

If you feel that this change is unnecessary because of the unit test, then I can revert that file back to its original state.

mbilker:
Hmm. The tests passed on my computer, but on Travis they flat out fail or crash.

seanmonstar:
The tests in CI failing are unrelated. I'm battling a race condition in the futures mpsc channel implementation, and still haven't gotten it working always, it seems.

seanmonstar:
It looks good to me, thanks!

mbilker:
Thank you very much!

PR #1500, opened 2018-04-26
feat(client): support local bind for HttpConnector

### LINKED ISSUE #1498: Client feature proposal: set_local_address
Hi,

I am using the Hyper client on a host which has multiple network interfaces. I would like to issue an HTTP request over a specific interface. The syntax should be something like this:
```rust
    let bind_addr: Option<IpAddr> = Some("127.0.0.1".parse().unwrap());
    let client = Client::configure().set_local_address(bind_addr).build(&handle);
```
Below I propose:
* A configuration syntax.
* A way of implementing this architecturally, which involves changes to mio and tokio-core.

Would you be happy to receive a PR along these lines, and do you think the approach of enhancing mio and tokio-core is best, or should we simply change hyper?

Details follow.

Uncontroversially we need the following plumbing:
* in `client/mod.rs` we add `local_address: Option<IpAddr>` to `Config`, default it to `None`, add `set_local_address`, and make `build` set it on `HttpConnector` if present.
* Similarly, we add `local_address: Option<IpAddr>` to the fields of `HttpConnector` and provide a setter.
* To convey the local address to the point where the connection is done, the `Lazy` and `Resolving` states of `HttpConnecting` gain an extra tuple field `Option<IpAddr>`, and `ConnectingTcp` gains a `local_address` field. The same will also be required in the hyper-tls connector.

The interesting part is what we replace `self.current = tokio::net::TcpStream::connect(&addr, handle);` with in `ConnectingTcp`.

Doing it all purely in Hyper looks like this (replacing a few lines of `impl ConnectingTcp` around https://github.com/hyperium/hyper/blob/master/src/client/connect.rs#L426):
```rust
              if let Some(addr) = self.addrs.next() {
                debug!("connecting to {}", addr);
                let do_bind = |local: &Option<IpAddr>| {
                    let sock = match addr {
                        SocketAddr::V4(..) => TcpBuilder::new_v4(),
                        SocketAddr::V6(..) => TcpBuilder::new_v6(),
                    }?;
                    match local {
                        &None => (), // oops, should add special Windows knowledge here
                        &Some(ref local_addr) => sock.bind(SocketAddr::new(local_addr.clone(), 0)).map(|_| ())?,
                    };
                    sock.to_tcp_stream()
                };

                match do_bind(&self.local_address) {
                    Ok(tcp) => {
                        self.current = Some(mio::net::TcpStream::connect_stream(tcp, &addr, handle));
                        continue;
                    },
                    Err(e) => {
                        err = Some(e)
                    }
                };
            }
```

This direct approach is unpleasant because it violates the architectural layering and duplicates code. In particular, this code is taken almost verbatim from `mio::net::TcpStream::connect()` – and as you can see there, it should really include a [special Windows workaround](https://github.com/carllerche/mio/blob/master/src/net/tcp.rs#L97) here.

It also loses the zero-allocation property: Tokio’s `TcpStream::connect_stream` can only return a `Box<Future>`, so we have to change ConnectingTcp to use this boxed future rather than the bare Tokio `TcpStreamNew` (which we cannot construct outside Tokio).

How could we preserve the architectural layering?

In `mio::net::TcpStream` we should add a `connect_with_local_address` method beside `connect`, to encapsulate the Windows and socket knowledge. We should then expose that in a new `tokio::net::TcpStream::connect_with_local_address` method. Then this method is available to be used Hyper’s `ConnectingTcp`:
```rust
              if let Some(addr) = self.addrs.next() {
                debug!("connecting to {}", addr);
                self.current = tokio::net::TcpStream::connect_with_local_address(&addr, &local, handle);
```

I think this a cleaner approach architecturally, and it is also able to preserve zero-allocation. But it does involve making changes to three projects. Which is the best approach?

kw217:
Add `set_local_address` to the `HttpConnector`.
This configures the client to bind the socket to a local address of
the host before it connects to the destination. This is useful on
hosts which have multiple network interfaces, to ensure the request is
issued over a specific interface.

Includes an example to show how it is used and regression-check the API (copied from the existing client example). Please let me know if I should be adding any tests, and if so where.

Closes #1498

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -125,10 +125,11 @@ impl Connected {
 mod http {
     use super::*;
 
+    use std::borrow::Cow;
     use std::fmt;
     use std::io;
     use std::mem;
-    use std::net::SocketAddr;
+    use std::net::{IpAddr, SocketAddr};
     use std::sync::Arc;
     use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


src/client/connect.rs
```diff
@@ -246,6 +254,16 @@ mod http {
...
             self.nodelay = nodelay;
         }
+
+        /// Set that all sockets are bound to the configured address before connection.
+        ///
+        /// If `None`, the sockets will not be bound.
+        ///
+        /// Default is `None`.
+        #[inline]
+        pub fn set_local_address(&mut self, addr: Option<IpAddr>) {
```
seanmonstar: Just some bikeshedding: what is something like this more commonly called? A "local" address? The "bind" address? And, would it be better to call this an "address", or just "ip"?

I don't have much preference myself, as long as it's been considered.

examples/bound_client.rs
```diff
@@ -0,0 +1,56 @@
+#![deny(warnings)]
```
seanmonstar: I think my personal preference here would be to not include an entire example file, as it means another thing to adjust when changes are required. We don't have an example for every combination of configuration options...

src/client/connect.rs
kw217: It's normally called the local address; see bind(2) at https://linux.die.net/man/2/bind and http://pubs.opengroup.org/onlinepubs/009695399/functions/bind.html . Normally `bind` takes a `SockAddr`, but for this specific case of binding-before-connecting, it's vanishingly rare that you want to specify the port as well as the IP, which is why I have specified `IpAddr`. But it's still an address.

examples/bound_client.rs
kw217: No problem - it was just there to show what it would look like. I've removed it now.

seanmonstar:
Woops, test failures are because the `Cargo.toml` still references the example that you removed XD

kw217:
Right, sorry about that. Fixed now - back to the same timeout error you said not to worry about.

seanmonstar:
Thanks so much!

kw217:
Thank you!

PR #1503, opened 2018-04-27
Try h2 on h1 failure

estk:
TODO:
- [ ] get compiling!


estk:
So, I'm a bit stuck at the moment, my current strategy is to wrap the `ServiceFn` passed to `Http::serve_connection` in an `Arc` so that the h1 `Connection` can hold a reference to the service in case we want to try an upgrade. The problem is by doing this, somehow NewService is no longer Sync. See the error below for details.
So, that seems fine, but the thing that really has me scratching my head is that the `ServiceFn` though not in `Connection` directly before was by proxy since it was owned by the h1::Dispatcher. Why is `NewService` all of a sudden not `Sync` now?

I'm fairly new to rust so I feel like I'm a bit out of my depth here, and would really appreciate any help.


```
$ RUST_LOG=hyper=trace cargo test --test h2
   Compiling hyper v0.12.0-pre.0 (file:///Users/evan.simmons/rust/hyper)
error[E0277]: `<S as service::new_service::NewService>::Service` cannot be shared between threads safely
   --> src/server/conn.rs:536:42
    |
536 |                 self.serve.protocol.exec.execute(fut);
    |                                          ^^^^^^^ `<S as service::new_service::NewService>::Service` cannot be shared between threads safely
    |
    = help: the trait `std::marker::Sync` is not implemented for `<S as service::new_service::NewService>::Service`
    = help: consider adding a `where <S as service::new_service::NewService>::Service: std::marker::Sync` bound
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<<S as service::new_service::NewService>::Service>`
    = note: required because it appears within the type `server::conn::Connection<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Service>`
    = note: required because it appears within the type `futures::future::chain::Chain<futures::MapErr<server::conn::Connecting<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Future>, fn(<S as service::new_service::NewService>::InitError) -> error::Error {error::Error::new_user_new_service::<<S as service::new_service::NewService>::InitError>}>, server::conn::Connection<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Service>, [closure@src/server/conn.rs:534:31: 534:42]>`
    = note: required because it appears within the type `futures::AndThen<futures::MapErr<server::conn::Connecting<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Future>, fn(<S as service::new_service::NewService>::InitError) -> error::Error {error::Error::new_user_new_service::<<S as service::new_service::NewService>::InitError>}>, server::conn::Connection<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Service>, [closure@src/server/conn.rs:534:31: 534:42]>`
    = note: required because it appears within the type `futures::MapErr<futures::AndThen<futures::MapErr<server::conn::Connecting<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Future>, fn(<S as service::new_service::NewService>::InitError) -> error::Error {error::Error::new_user_new_service::<<S as service::new_service::NewService>::InitError>}>, server::conn::Connection<<I as futures::Stream>::Item, <S as service::new_service::NewService>::Service>, [closure@src/server/conn.rs:534:31: 534:42]>, [closure@src/server/conn.rs:535:30: 535:65]>`

For more information about this error, try `rustc --explain E0277`.

```

seanmonstar:
So, that specific error is probably a red herring, but I can explain the problem:

- There is a bounds that future being executed be `Send`. This is because the executor could be a thread pool, and so the future might move to a new thread.
- By wrapping something in `Arc`, that makes it possible to send copies to *different* threads at the same time. Both copies would be able to access `&T` from the `Arc<T>`.
- In order for that to be safe, `T` must be `Sync`, which states that all accesses of `&T` are thread-safe.
- Some examples of non-thread-safe `&T` access: `Rc::clone`, `Cell`, `RefCell`...
- Since those types can actually mutate `T` through a `&T`, but don't do so with synchronization, they don't implement `Sync`.
- Backing out again, that means that if you have an `Arc<T>`, in order for `Arc<T>: Send`, the type it is wrapping must be `T: Send + Sync`.

-----

I don't *think* you'll need an `Arc` here at all, but instead should be able to pop the service back out of the h1 dispatcher, just like the IO and read buffer.

estk:
@seanmonstar So, I've now gone down the deconstruction path, and I saw two options:

1. Implement and `into_inner()` on the trait.
1. Wrap the dispatch with an enum such that we can match on if it is a server and return an Option<Service> in the `Dispatcher::into_inner()`

By the looks of it, by wrapping in the enum I'd have to implement the Dispatch trait on it which seems like a lot of boilerplate and hence probably not the right way to go.

After giving the first option a try I find myself again a bit stuck and would love a tip if you have a sec, I'm getting the following compiler message which is a bit ironic since I dont really care what the type is on client since it never returns `Some`. Is there a way to tell the compiler that this *only* returns `None`? If not, is there any way to reconcile with the compiler, or should I just try the `enum` option?
```
$ RUST_LOG=hyper=trace cargo test --test h2
   Compiling hyper v0.12.0-pre.0 (file:///Users/evan.simmons/rust/hyper)
error[E0207]: the type parameter `S` is not constrained by the impl trait, self type, or predicates
   --> src/proto/h1/dispatch.rs:396:9
    |
396 | impl<B, S> Dispatch for Client<B>
    |         ^ unconstrained type parameter

For more information about this error, try `rustc --explain E0207`.

```

seanmonstar:
You shouldn't need to to much with the `Dispatch` trait. You already know the exact implementation of dispatch, `Dispatch<Server<...>>`. So, it should be possible to add an impl block for that type directly.

```rust
impl<generics> Dispatch<io, etc, Server<S>> {
    pub fn hello() {}
}
```

Though, it probably does mean the the `server:: Connection` needs to hold not a `Dispatcher` directly, but some sort of enum that can allow taking the dispatcher apart even in a `&mut self` context. An `Option` works, or occasionally a new enum can be better, depending on if there's more than 2 states, or the names would be ambigious.

estk:
Hokay, stuck again. Current issue is below, my guess is that since we are parameterizing on a specific `I`, we'll need to actually wrap that with a sum type that contains either `I` or `RewindedAddrStream<I>`. (There's definitely a better name for that but I'll leave it for cleanup.)


```
$ RUST_LOG=hyper=trace cargo test --test h2
   Compiling hyper v0.12.0-pre.0 (file:///Users/evan.simmons/rust/hyper)
error[E0308]: mismatched types
   --> src/server/conn.rs:394:43
    |
394 |         self.conn.get_or_insert(Either::B(h2));
    |                                           ^^ expected type parameter, found struct `server::tcp::rewinded_add_stream::RewindedAddrStream`
    |
    = note: expected type `proto::h2::server::Server<I, _, _>`
               found type `proto::h2::server::Server<server::tcp::rewinded_add_stream::RewindedAddrStream<I, std::io::Cursor<bytes::Bytes>>, _, _>`

For more information about this error, try `rustc --explain E0308`.

```

seanmonstar:
Thanks for sticking with this! I maybe should have labelled this hard instead of medium, it seems it's more involved than I originally thought XD

Yes, we'd need some sort of wrapper for the generic `AsyncRead + AsyncWrite` to allow "rewinding" and using the read buffer as the start of the h2 handshake. It should itself be able to be generic, not actually requiring the `AddrStream`. And if it's built into the internals of the `Conn` type, it's existence probably doesn't even need to be known. As in, the user would still just see it as `Conn<MyIoType, MyService>`. Does that make sense? I sometimes don't notice if I glossed over something with not enough detail.

estk:
Working, see pr description.

src/server/mod.rs
```diff
@@ -82,6 +83,12 @@ pub struct Builder<I> {
     protocol: Http_,
 }
 
+/// AsyncRewind is a rewindable AsyncRead+AsyncWrite trait
+pub trait AsyncRewind: AsyncRead + AsyncWrite {
```
seanmonstar: I think we can probably build the rewinding in internal, and not require that ever transport type a user might want to use have to implement this function. Basically, a private `struct Rewound<I>`. Here's a way I'd do it:

1. In the h1 case, nothing should be needed there. If there is a parse error, we can just use `deconstruct`/`into_inner` to get the original transport and the `read_buf` back.
2. If we've done number 1, and want to start the h2 handshake, then a `Rewound` would be constructed with those 2 pieces, and then `Rewound<I>` would *be* the `AsyncRead + AsyncWrite` type passed to the h2 handshake/connection.

That'd make the `server::Connection` type look kind of like this internally:

```rust
pub struct Connection<I, S> {
    inner: Option<
        Either<
            h1::Dispatcher<..., I>,
            h2::Server<Rewound<I>, ...>,
        >
    >
}
```

src/proto/h1/conn.rs
```diff
@@ -545,6 +552,9 @@ where I: AsyncRead + AsyncWrite,
     // - Client: there is nothing we can do
     // - Server: if Response hasn't been written yet, we can send a 4xx response
     fn on_parse_error(&mut self, err: ::Error) -> ::Result<()> {
+        if self.shutdown_on_error == false {
```
seanmonstar: This could probably be inside `T::on_error` instead. Hm, it could perhaps be useful if the `on_error` method were adjusted to allow accessing the current `&read_buf`. Then, `ServerTransaction::on_error` could check to see if it at least starts with the HTTP/2 magic prefix before deciding to bubble the error back up (and let `server::Connection` try h2), or to just return the HTTP/1 bad response.

estk:
You were totally right, it looks like it might be a bit of a nightmare to add the checking to the `Connection`.

estk:
@seanmonstar I'm a bit stuck again, it looks to me like things should be working, any idea why the ci seems to only pass on nightly?

seanmonstar:
@estk looking at the logs *right now*, it seems to be complaining about illegal syntax. It looks like a merge conflict was saved, as there's a line like `<<<<<<<<<<< HEAD` ;)

estk:
Lol, that’s embarrassing, i guess ill blame that on not having coffee yet. The build that i was referring to was the one before. Im mobile now but ill fix it up this evening

src/server/conn.rs
```diff
@@ -513,3 +553,165 @@ where
         }
     }
 }
+
+// TODO: different file? idk where to put this.
+mod rewindable {
```
estk: @seanmonstar, it looks like hyper as a project prefers to keep new files to a minimum, is this an appropriate place to put this?

tests/server.rs
```diff
@@ -39,6 +40,24 @@ fn tcp_bind(addr: &SocketAddr, handle: &Handle) -> ::tokio::io::Result<TcpListen
     TcpListener::from_std(std_listener, handle)
 }
 
+#[test]
+fn try_h2() {
```
estk: @seanmonstar this is a significant deviation from how other tests are performed, but I'm not really sure how else I would go about it. Any guidance here?

src/server/conn.rs
seanmonstar: There isn't an enforced preference, really... keeping things in less files came from two reasons: a) it can get annoying to have to import a bunch of the same stuff over and over, and b) sometimes some types make use of private fields of others (though this is less important now that there exists `pub(scope)`s.

If you think it's clearer in a separate file, that's perfectly fine :)

tests/server.rs
seanmonstar: It's fine! I've really wanted to take some time to clean up the tests, they've become a mess. They started with just a server helper, but eventually more specific details needed to be tested, and I never took the time to group up boilerplate.

src/server/conn.rs
```diff
@@ -74,31 +76,34 @@ pub(super) struct SpawnAll<I, S> {
...
 }
 
 /// Deconstructed parts of a `Connection`.
 ///
 /// This allows taking apart a `Connection` at a later time, in order to
 /// reclaim the IO object, and additional related pieces.
 #[derive(Debug)]
-pub struct Parts<T, S> {
+pub struct Parts<T, S> 
+    where S: Service {
```
seanmonstar: Is this new bounds required?

src/server/conn.rs
```diff
@@ -362,14 +369,37 @@ where
...
             Either::A(ref mut h1) => {
                 try_ready!(h1.poll_without_shutdown());
                 Ok(().into())
             },
             Either::B(ref mut h2) => h2.poll(),
         }
     }
+
+    fn try_h2(&mut self) -> Poll<(), ::Error> {
+        debug!("Trying to upgrade to h2");
```
seanmonstar: May want to demote this to a `trace!`, and remove the other two `debug!` logs in this function.

src/server/conn.rs
```diff
@@ -513,3 +553,165 @@ where
...
+mod rewindable {
+    use bytes::{Buf, BufMut, Bytes, IntoBuf};
+    use futures::{Async, Poll};
+    use std::io::{self, Read, Write};
+    use tokio_io::{AsyncRead, AsyncWrite};
+
+    #[derive(Debug)]
+    pub struct Rewindable<T>
+    where
+        T: AsyncRead + AsyncWrite,
```
seanmonstar: Ah, I see. If these bounds are removed, then they won't be needed in several other places as well. Usually, prefer bounds only on `impl`s that need them (sometimes it's unavoidable in a struct definition, if you need to refer to an associated type).

src/server/conn.rs
```diff
@@ -513,3 +553,165 @@ where
...
+    {
+        pub(super) fn new(tcp: T) -> Rewindable<T> {
+            Rewindable {
+                pre: None,
+                inner: tcp,
+            }
+        }
+        pub fn prepend(&mut self, bs: Bytes) {
+            self.pre.get_or_insert(bs);
```
seanmonstar: Prefer `self.pre = Some(bs)` instead, since `get_or_insert` does some matching on the original field, but we aren't using it.

You could put a `debug_assert!(self.pre.is_none())` before setting it, for correctness while tests are run.

src/server/conn.rs
```diff
@@ -513,3 +553,165 @@ where
...
+        T: AsyncRead + AsyncWrite,
+    {
+        #[inline]
+        unsafe fn prepare_uninitialized_buffer(&self, buf: &mut [u8]) -> bool {
+            self.inner.prepare_uninitialized_buffer(buf)
+        }
+
+        #[inline]
+        fn read_buf<B: BufMut>(&mut self, buf: &mut B) -> Poll<usize, io::Error> {
+            if let Some(ref mut bs) = self.pre {
```
seanmonstar: Perhaps here and in `read`, the `pre` should be changed to `None` once all the bytes have been read. That will mean then skipping the rest of this `if` block on latter reads, and allow the `Bytes` to be freed since it's no longer needed.

src/proto/h1/dispatch.rs
```diff
@@ -37,6 +37,21 @@ pub struct Client<B> {
...
 
+impl<S, Bs, I, T> Dispatcher<Server<S>, Bs, I, T>
+where
+    S: Service<ReqBody=Body, ResBody=Bs>,
+    S::Error: Into<Box<::std::error::Error + Send + Sync>>,
+    I: AsyncRead + AsyncWrite,
+    T: Http1Transaction,
+    Bs: Payload,
+{
+    pub fn deconstruct(self) -> (I, Bytes, S) {
```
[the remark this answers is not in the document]
estk: Absolutely!

src/server/conn.rs
estk: It seems not, ive removed it. 
So I know for the future, Is it a rust thing to not include bounds that will be satisfied but are not required? I would think that since this is an intermediate data structure we would want to be as specific as possible about what it contains for the end user.

src/server/conn.rs
estk: This is a great rust nugget of knowledge, thanks.

src/server/conn.rs
estk: Absolutely, actually I'm not super confident on what is in here, but getting close to something that seems right.

src/server/rewindable.rs
```diff
@@ -0,0 +1,185 @@
...
+    unsafe fn prepare_uninitialized_buffer(&self, buf: &mut [u8]) -> bool {
+        self.inner.prepare_uninitialized_buffer(buf)
+    }
+
+    // TODO(estk) this can be simplified! Question: can we just return what is in the prefix 
+    // or do we need to try to read from inner to fill the buf? If we could return early
+    // then we could have two separate code paths where we dont need to stage changes like
+    // here which adds complexity.
+    #[inline]
+    fn read_buf<B: BufMut>(&mut self, buf: &mut B) -> Poll<usize, io::Error> {
```
estk: @seanmonstar do you know if there are any guarantees around how many bytes should be written onto `buf`? If there are none we could drastically simplify this code.

src/server/rewindable.rs
```diff
@@ -0,0 +1,204 @@
...
+            Async::NotReady => panic!("should be ready"),
+            Async::Ready(cnt) => assert_eq!(5, cnt),
+        }
+
+        let read_buf = Bytes::from(&o1_cursor.into_inner()[..]);
+        stream.prepend(read_buf);
+
+        let mut o2_cursor = Cursor::new(o2);
+        stream.read_buf(&mut o2_cursor).unwrap();
+        stream.read_buf(&mut o2_cursor).unwrap();
```
estk: @seanmonstar This the crux of what im concerned about, is it okay that we're not reading out all bytes from a perf/correctness point of view?

seanmonstar: I don't actually understand the question, or what this test is testing :D

Mind adding some comments about what each part of the test is triggering or why?

src/server/rewindable.rs
```diff
@@ -0,0 +1,204 @@
...
+    unsafe fn prepare_uninitialized_buffer(&self, buf: &mut [u8]) -> bool {
+        self.inner.prepare_uninitialized_buffer(buf)
+    }
+
+    #[inline]
+    fn read_buf<B: BufMut>(&mut self, buf: &mut B) -> Poll<usize, io::Error> {
+        if let Some(bs) = self.pre.take() {
+            // If there are no remaining bytes, let the bytes get dropped.
+            if bs.len() > 0 {
+                let mut reader = bs.into_buf().reader();
```
seanmonstar: I think there may be a more idiomatic way to copy between a `Buf` and a `BufMut`. I think this may work:

```rust
let cnt = buf.remaining_mut();
let mut bs = bs.into_buf().take(cnt);
buf.put(&mut bs);

// bs.into_inner() stuff
```

It's unfortunate the dance needed for `BytesMut`. It's already been decided that it was mistake that it doesn't implement `Buf` itself, but it can't without a breaking change. It will in bytes 0.5.

src/server/rewindable.rs
estk: Haha sorry, was a bit of a non-question. I think ive answered it myself, I was wondering if there is an undocumented dependency of the `read_buf` method to read all the bytes possible into the `buf` in a single call.
How I'm doing it here is *one* of the following per call:
- Is there a prefix? Read as many bytes in the prefix as possible out, If there are some left, put those back in the prefix field.
- If there is not a prefix, read direct from the `inner`

In practice this means the buf passed in will not always be filled even though there may be bytes ready to read. I can imagine a situation where there is code relying on a small buffer being completely filled which could break.

src/server/rewindable.rs
estk: Excellent info as always.

src/server/rewindable.rs
seanmonstar: Oh, I see what you mean. Yea, this behavior is correct. You don't want to copy the buffered bytes, and try to a syscall in the same `read`. Because: what's the behavior if there is an error from the syscall? Return the error? But then the buffered bytes... what about those? It's messy.

Plus, anyone *relying* on a `read` to completely fill a buffer is definitely using IO incorrectly. You never know exactly how much data might be in the kernels buffers, and you don't know if more arrived just as it partially filled your buffer, so you just have to keep reading until your buffer is full, or your get `WouldBlock`.

seanmonstar:
There seems to be some sort of race condition in h2 that sometimes prevents the client connection from completely closing up, resulting in the test server and client just staring at each other until Travis gives up. 😭

estk:
@seanmonstar whew, it looks like we're approaching the finish line on this one. Any further comments or tests you'd like me to add?

src/proto/h1/conn.rs
```diff
@@ -529,8 +536,12 @@ where I: AsyncRead + AsyncWrite,
     // - Client: there is nothing we can do
     // - Server: if Response hasn't been written yet, we can send a 4xx response
     fn on_parse_error(&mut self, err: ::Error) -> ::Result<()> {
+
         match self.state.writing {
             Writing::Init => {
+                if self.has_h2_prefix() {
```
seanmonstar: Oh great idea, so that `Http1Transaction` doesn't need to get more complicated, I like it!

src/proto/h2/client.rs
```diff
@@ -60,6 +60,7 @@ where
         loop {
             let next = match self.state {
                 State::Handshaking(ref mut h) => {
+                    trace!("Polled client future, state is handshaking");
```
seanmonstar: I'd probably remove these traces, I'm sure they were useful developing the feature, but afterwards, debugging other things will likely mean these just feel more like noise.

src/server/conn.rs
```diff
@@ -74,31 +76,33 @@ pub(super) struct SpawnAll<I, S> {
 ///
 /// Polling this future will drive HTTP forward.
 #[must_use = "futures do nothing unless polled"]
-pub struct Connection<I, S>
+pub struct Connection<T, S>
 where
     S: Service,
+    T: AsyncRead + AsyncWrite,
```
seanmonstar: I believe these bounds can be removed from here and the `fmt::Debug` impl :)

src/server/conn.rs
```diff
@@ -74,31 +76,33 @@ pub(super) struct SpawnAll<I, S> {
...
 {
-    pub(super) conn: Either<
+    pub(super) conn: Option<
+        Either<
         proto::h1::Dispatcher<
             proto::h1::dispatch::Server<S>,
             S::ResBody,
-            I,
+            Rewind<T>,
```
seanmonstar: This is just a thought, feel free to tell me it's crazy: it may be possible to actually just pass `T` here, and not need `Rewind<T>` *only for h1*. It seems like the only time `rewind` is called is when deconstructing h1 and building an h2 dispatcher, so in `try_h2` is where it could be `let io = Rewind::new(io); io.rewind(read_buf); ...`. I have a hunch this will mean that this new feature doesn't affect the performance of HTTP1 at all, since the `read` calls would never need to check the rewind buffer.

estk: Yeah, this is something I thought about, I cant remember why I didn't do this, let me try to look through the code and jog my memory.

estk:
Haha thanks @seanmonstar, this has been an excellent learning experience so I'm not quite ready to give up yet. I've addressed the comments you made today, aside from the `.http1only()` suggestion, which I'm more happy to implement, but I probably wont have time today.

Unfortunately it seems like I haven't quite fixed the flakey-ness of the CI tests. It really has me baffled since I cannot reproduce it locally and it seems to happen on different rustc's every time. I've even gone to the trouble of running it in a debian docker container to see if I could replicate it there (no luck.) My suspicion is that it is a real issue, in h2 as you said, but without being able to reproduce it locally its very difficult to debug. If you could point me in the right direction there it would be super helpful. Also, if youre ready to just wrap this up yourself feel free, I'm sure I'll learn a ton from the code you add.

PR #1516, opened 2018-05-15
Avoid double boxing when wrapping a Stream in a Body

### LINKED ISSUE #1515: should Body implement From<Box<Stream<...>>?
I'm wondering if `Body` could provide something like this:

```rust
impl From<Box<Stream<Item=Chunk, Error=Box<::std::error::Error + Send + Sync>> + Send>> for Body {
    #[inline]
    fn from(stream: Box<Stream<Item=Chunk, Error=Box<::std::error::Error + Send + Sync>> + Send>)
            -> Body {
        Body::new(Kind::Wrapped(stream))
    }
}
```

in addition to or instead of `Body::wrap_stream`.

A couple reasons this seems handy while working on a crate that handles byte range serving:

   * I want it to support either using `Body` or plugging in your own `Payload` type. It needs to do stream ops like `unfold` and `flatten` (see [implementation](https://github.com/scottlamb/http-serve/blob/master/src/serving.rs)), so it ends up with a boxed stream that it wants to turn into your payload type. The most straightforward way to do this that I can think of would be to require this `From` implementation, but then it won't work with `Body`. I could add some extra custom trait and pre-implement it for `Body`, but then there's another thing you need to import for it to work / then either the `Payload` implementation has to depend on my crate or vice versa.
   * when I already have a boxed stream, wrapping it with another box (as `Body::wrap_stream` does) adds an unnecessary allocation.

lnicola:
Fixes #1515

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -421,15 +421,17 @@ impl From<Chunk> for Body {
     }
 }
 
-impl<S> From<Box<S>> for Body
-where
-    S: Stream + Send + 'static,
-    S::Error: Into<Box<::std::error::Error + Send + Sync>>,
-    Chunk: From<S::Item>,
+impl
+    From<Box<Stream<Item = Chunk, Error = Box<::std::error::Error + Send + Sync>> + Send + 'static>>
+    for Body
 {
     #[inline]
-    fn from(stream: Box<S>) -> Body {
-        Body::wrap_stream(stream)
+    fn from(
+        stream: Box<
+            Stream<Item = Chunk, Error = Box<::std::error::Error + Send + Sync>> + Send + 'static,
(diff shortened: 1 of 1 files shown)


lnicola:
I also tried the following approach, but it seems to overlap with the other `From` implementations.

```rust
impl<S> From<S> for Body
where
    S: Into<
        Box<Stream<Item = Chunk, Error = Box<::std::error::Error + Send + Sync>> + Send + 'static>,
    >,
{
    #[inline]
    fn from(stream: S) -> Body {
        Body::new(Kind::Wrapped(stream.into()))
    }
}
```

seanmonstar:
Thanks!

Issue #1522, opened 2018-05-25
examples/client.rs not building


stevedonovan:
Error says it all:
```
10 | use hyper::service::service_fn;
   |            ^^^^^^^ Could not find `service` in `hyper`
```
Am I missing something really obvious? (Using stable 1.26 here)

davidbarsky:
The examples directory is using the master branch of Hyper, which will be 0.12. The current release of Hyper on crates.io is 0.11.27, which lacks the `hyper::service` module. Since you're seeing this build error, I'm _assuming_ that you're pulling Hyper from crates.io and not the mainline branch of `hyperium/hyper`.

If you replace `hyper = "0.11"` in your Cargo.toml with `hyper = { git = "https://github.com/hyperium/hyper" }`, your project should build correctly.

stevedonovan:
Thanks, man! This was obvious in hindsight. Shall we keep this open as a marker until 0.12 lands on crates.io?

davidbarsky:
Maybe! That said, I've opened an PR to alleviate this issue.

PR #1525, opened 2018-05-29
docs(README): clarify examples/ rely on 0.12

davidbarsky:
A warning in the README—at least until 0.12 is released—might alleviate some confusion (e.g., #1522).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,3 +25,5 @@ hyper makes use of "async IO" (non-blocking sockets) via the
 Be aware that hyper is still actively evolving towards 1.0, and is likely
 to experience breaking changes before stabilising. You can also see the
 [issues in the upcoming milestones](https://github.com/hyperium/hyper/milestones).
+Currently, the `examples/` directory relies on Hyper 0.12, which is the current
+development branch _not_ the version on crates.io.


seanmonstar:
Thanks for the PR! I wonder, does the added text clarify much? It might be that I'm biased, but I'm used to example code at a certain commit in a repo to always work with that version of the code, not necessarily what is published.

Maybe there is some value in reminding people of that... What do you think of that message being in `examples/README.md` directly?

davidbarsky:
> Thanks for the PR! I wonder, does the added text clarify much? It might be that I'm biased, but I'm used to example code at a certain commit in a repo to always work with that version of the code, not necessarily what is published.

That's a fair point.

> Maybe there is some value in reminding people of that... What do you think of that message being in examples/README.md directly?

Could be! But if Hyper 0.12 will be released soon—the [milestone for 0.12](https://github.com/hyperium/hyper/milestone/4) suggests that the case—then maybe adding the README to examples _might_ be unnecessary. Either way, I'll update the PR (or open a new one...) to add the README to the examples directory.

davidbarsky:
Ah, didn't see https://github.com/hyperium/hyper/commit/2415ce96e198bafc107c38e53a7bbacc2d361dbf. Closing this PR as it's handled.

PR #1530, opened 2018-06-03
fix(server): add upgrading process to poll_without_shutdown()

### LINKED ISSUE #1529: Fallback to h2 does not work at poll_without_shutdown()
It seems that the patch for upgrading to h2 after failing parsing h1 is not applied to poll_without_shutdown side. Therefore, upgrading to h2 fails when using the method in the low level API directly.

ubnt-intrepid:
Fixes #1529.


seanmonstar:
Thanks for catching this!

PR #1533, opened 2018-06-03
docs(server): Update send_file example.

aaronriekenberg:
Adds a send_file_blocking example that uses tokio_threadpool::blocking (see https://docs.rs/tokio-threadpool/0.1.3/tokio_threadpool/fn.blocking.html).

It was a bit difficult for me to string together multiple examples to figure this out.  Probably is a common use-case to do a blocking operation without spawning a new thread for each request or having to run on some type of thread pool.


seanmonstar:
Thanks, I'm sure this would be useful! I actually think it might be better to just replace the existing send file example with this one, as it is better in most ways. What do you think?

aaronriekenberg:
Sounds good.  Updated to replace existing send_file.rs with this blocking example.  Thanks.

examples/send_file.rs
```diff
@@ -106,38 +58,38 @@ fn simple_file_send(f: &str) -> ResponseFuture {
...
-                        .unwrap())
-                    .expect("Send error on error reading file");
-            },
-        };
-    });
-
-    Box::new(rx.map_err(|e| io::Error::new(io::ErrorKind::Other, e)))
+    Box::new(
+        future::poll_fn(move || {
+            blocking(|| read_file_blocking(&filename))
```
seanmonstar: Should this instead show just using `tokio-fs` instead? I slightly worry people will miss this line and think they should just do blocking operations anywhere... Or do you think there is value in seeing the more general `blocking` tool?

Cargo.toml
```diff
@@ -34,6 +34,7 @@ tokio-executor = { version = "0.1.0", optional = true }
 tokio-io = "0.1"
 tokio-reactor = { version = "0.1", optional = true }
 tokio-tcp = { version = "0.1", optional = true }
+tokio-threadpool = { version = "0.1", optional = true }
```
seanmonstar: Since this is only used in the examples, I think this can just go in the `dev-dependencies`, yea?

examples/send_file.rs
aaronriekenberg: Fair point.  I updated the example to use tokio_fs and tokio_io to have the same behavior, which is probably a better example.

Cargo.toml
aaronriekenberg: Fixed.  tokio-threadpool is no longer needed.  Added tokio-fs to dev-dependencies.

seanmonstar:
Thanks so much, this looks great! ❤️

PR #1535, opened 2018-06-04
feat(server): add http1_only configuration to conn::Connection

### LINKED ISSUE #1512: Add an http1_only configuration for the server
The default server now supports both HTTP/1 and 2 automatically. Just like there is an option to set `http2_only`, there should probably also be an `http1_only` option, if for some reason someone didn't want to allow HTTP2 support.

ubnt-intrepid:
A new configuration `http1_only` to `Builder` and `Connection` are added, which indicates that the upgrading to h2 does not perform when a parsing error occurs.

Fixes #1512.

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -37,10 +37,21 @@ use error::{Kind, Parse};
 #[derive(Clone, Debug)]
 pub struct Http {
     exec: Exec,
-    http2: bool,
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the excellent PR, I appreciate the fine details, like the design of the `enum` and the doc comments! <3

Issue #1528, opened 2018-06-02
The echo example should have the functionality from the guide


seanmonstar:
The [echo guide](https://hyper.rs/guides/server/echo/) links to the `examples/echo.rs` file, but it doesn't include all the functionality in that guide. It should be updated to do so!

PR #1538, opened 2018-06-05
docs(server): update echo example with functionality from the guide

joshleeb:
Update the echo example based on the functionality explained in the Hyper guide: https://hyper.rs/guides/server/echo

CLOSES: #1528


seanmonstar:
Thank you so much, this is great!

PR #1543, opened 2018-06-05
Add Sender::abort

sfackler:
This allows a client to indicate that the request body should be cut off
in an abnormal fashion so the server doesn't simply get a "valid" but
truncated body.

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -11,7 +11,6 @@ use common::Never;
 use super::{Chunk, Payload};
 use super::internal::{FullDataArg, FullDataRet};
 
-
 type BodySender = mpsc::Sender<Result<Chunk, ::Error>>;
 
 /// A stream of `Chunk`s, used when receiving bodies.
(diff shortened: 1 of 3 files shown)


src/body/body.rs
```diff
@@ -303,6 +308,11 @@ impl Sender {
             .map_err(|err| err.into_inner().expect("just sent Ok"))
     }
 
+    /// Aborts the request in an abnormal fashion.
```
seanmonstar: Since this could be a body for a client request or server response, the doc should probably just say it "aborts the body".

sfackler:
I wrote a test, but something seems broken - Hyper's sending what looks like an HTTP/1.0-style EOF delimited request but with a version of HTTP/1.1:

screen shot 2018-06-05 at 3 33 39 pm

src/body/body.rs
sfackler: Fixed

sfackler:
Oh derp, i'm sending a get with a body..

seanmonstar:
Yay, thanks!

Issue #1542, opened 2018-06-05
Provide introspection on hyper::Error


jonhoo:
I have an application in which I want to try to set up a server on a specific port, but if that port is taken (i.e., an `AddrInUse` error is raised), it should fall back to any port (i.e., port 0). Unfortunately, since hyper's error type is now opaque (as of https://github.com/hyperium/hyper/commit/5d3c472228d40b57e47ea26004b3710cfdd451f3), I can no longer easily see if that was the cause of an error raised by `Server::try_bind`.

While I *could* construct a `TcpListener` myself and then pass that listener's `Incoming` to `Server::builder`, that's an unfortunate amount of extra boilerplate. Perhaps it would make sense to expose an `as_io_error(&self) -> Option<&io::Error>` on `hyper::Error`?

seanmonstar:
I saw mention of this in IRC, it seems you can't downcast the cause because `Error::cause()` doesn't return a `Error + 'static`, right?

What if `hyper::Error` were to get a `cause_static` (sorry, naming is hard) method, kind of like `std::io::Error::get_ref`, to allow downcasting?

jonhoo:
That'd be nice, but is there a particular reason you don't want to expose the `io::Error` through an explicit `as_io_error` method? As someone pointed out on IRC when I asked about it, `cause()` isn't really intended to be used as a way to figure out if a particular error occurred. Part of this is because causes form a _chain_, and you don't technically know how deep you have to go. And more importantly, the error nesting could *change* between versions.

seanmonstar:
I'm not especially against some form of `as_io_error`, I was just thinking that people may wish to access the errors they created themselves, like from a `Payload` or `Service`, and suggested something that might solve all cases.

jonhoo:
Oh, yeah, I think having both is probably a good idea! I think my concern about `cause` nesting would make it pretty difficult to extract user errors from behind a `hyper::Error` nonetheless, but at least it would be possible!

PR #1544, opened 2018-06-05
Add Error::cause2 and Error::into_error

sfackler:
cause2 is a bad name, but I couldn't think of anything obviously better
off the top of my head.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -125,6 +125,19 @@ impl Error {
         self.inner.kind == Kind::Closed
     }
 
+    /// Returns the error's cause.
+    /// 
+    /// This is identical to `Error::cause` except that it provides extra
+    /// bounds required to be able to downcast the error.
+    pub fn cause2(&self) -> Option<&(StdError + 'static + Sync + Send)> {
+        self.inner.cause.as_ref().map(|e| &**e)
+    }
+
+    /// Consumes the error, returning its cause.
+    pub fn into_cause(self) -> Option<Box<StdError + Sync + Send>> {
+        self.inner.cause
+    }
+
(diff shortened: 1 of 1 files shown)


seanmonstar:
Related #1542

jonhoo:
@sfackler `cause_static`?

jonhoo:
Also, could we include `as_io_error -> Option<&std::io::Error>` in this patch?

sfackler:
That's just `cause2().and_then(|e| e.downcast_ref::<io::Error>())` - not sure that needs to be a whole separate method.

jonhoo:
@sfackler no, that's not true. While it may be true *currently*, unless `hyper` explicitly guarantees that any `io::Error` will only be nested exactly one deep in the cause hierarchy, that could suddenly start failing as more error nesting is introduced.

jonhoo:
I'm not sure I agree? It's not inconceivable that `hyper` at some point got, say, a `tokio::Error` internally, which then in turn contains an `io::Error`. Or that it uses something like `failure`'s `.context()` which adds a layer of indirection to the `io::Error`. To your second point, I think `io::Error`s are sufficiently special to warrant them being exposed. For example, an error like `AddrInUse` the caller can often recover from by falling back to a non-fixed port.

seanmonstar:
While I don't foresee any plans to add additional types to the chain, the issue you describe could also happen directly inside hyper. We'd need some way to make sure `as_io_error` handled any intermediate types that were added, and it'd be just as easy to miss.

jonhoo:
Absolutely true! But as a consumer of the library, I'm unlikely to know if the internals change. Forcing users to go through `cause` for `io::Error` means that changing the internals of `hyper::Error` may be cause for a breaking semver change if it adds or removes nesting.

PR #1554, opened 2018-06-08
Make Body know about incoming Content-Length

### LINKED ISSUE #1545: Incoming HTTP1 bodies should know content_length and is_end_stream
For instance, if a server receives the follow request:

```
POST /foo HTTP/1.1
content-length: 15
```

The `Body` in the `Request` should know these things, neither of which are currently true:

- `Body::content_length()` should be `Some(15)`.
- `Body::is_end_stream()` should return true after `poll_data` has returned chunks of length totaling 15 bytes.

## Steps to fixing

- Adding an `Option<u64>` to the `Kind::Chan` variant of `Body`.
- Changing `h1::Conn::read_head` to return an `Option<BodyLength>` instead of `bool`.
- Updating the `h1::Dispatcher` to use the `Option<BodyLength>` to optionally set the new `Option<u64>` field of the `Chan` variant.
- Updating `Body::content_length` to return the new field from `Kind::Chan`.
- Updating `Body::poll_data` in the `Kind::Chan` variant to subtract from the new field the length of each chunk that passes by.
- Updating `Body::is_end_stream` in the `Kind::Chan` variant to return true **if**` the field is exactly `Some(0)`.

### Testing it works

- Add a test to `tests/client.rs` with a server replying with the bytes `HTTP/1.1 200 OK\r\ncontent-length: 5\r\n\r\nhello`.
- In the test, make a `GET` request to the server, and wait for the `Response<Body>`.
- Add an `assert_eq!(res.body().content_length(), Some(5))`.
- Add an `assert!(!res.body().is_end_stream())` (the body should be done yet, there's still 5 bytes to stream).
- Call `res.body_mut().poll_data()` to get the chunk. Assert it is 5 bytes for sanity.
- Finally, `assert!(res.body().is_end_stream())`, as the 5 bytes have now been streamed.

lnicola:
Fixes #1545.

I'm not sure about:

- what to do if the incoming body is longer than the content-length; is it ok to abort? I think Firefox does that
- the test (code and location)

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -35,6 +35,7 @@ pub struct Body {
 enum Kind {
     Once(Option<Chunk>),
     Chan {
+        content_length: Option<u64>,
         abort_rx: oneshot::Receiver<()>,
         rx: mpsc::Receiver<Result<Chunk, ::Error>>,
     },
@@ -85,6 +86,11 @@ impl Body {
     /// Useful when wanting to stream chunks from another thread.
     #[inline]
     pub fn channel() -> (Sender, Body) {
+        Self::new_channel(None)
+    }
+
+    #[inline]
+    pub(crate) fn new_channel(content_length: Option<u64>) -> (Sender, Body) {
         let (tx, rx) = mpsc::channel(0);
         let (abort_tx, abort_rx) = oneshot::channel();
 
@@ -93,8 +99,9 @@ impl Body {
             tx: tx,
         };
         let rx = Body::new(Kind::Chan {
-            abort_rx: abort_rx,
-            rx: rx,
+            content_length,
+            abort_rx,
+            rx,
         });
 
         (tx, rx)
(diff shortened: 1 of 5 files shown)


src/body/body.rs
```diff
@@ -93,6 +94,28 @@ impl Body {
...
+        });
+
+        (tx, rx)
+    }
+
+    /// Create a `Body` stream on indeterminate length with an associated sender half.
+    ///
+    /// Useful when wanting to stream chunks from another thread.
+    #[inline]
+    pub fn channel_with_length(content_length: u64) -> (Sender, Body) {
```
seanmonstar: For now, I'd suggest making this just `pub(crate)`.

Oh and then, these 2 functions can probably be reduced to calling a 3rd, private constructor, `fn new_channel(content_length: Option<u64>)`, to remove duplicate code, yea?

src/body/body.rs
```diff
@@ -188,13 +211,23 @@ impl Body {
...
+            Kind::Chan { content_length: ref mut len, ref mut rx, ref mut abort_rx } => {
                 if let Ok(Async::Ready(())) = abort_rx.poll() {
                     return Err(::Error::new_body_write("body write aborted"));
                 }
 
                 match rx.poll().expect("mpsc cannot error") {
-                    Async::Ready(Some(Ok(chunk))) => Ok(Async::Ready(Some(chunk))),
+                    Async::Ready(Some(Ok(chunk))) => {
+                        if let Some(ref mut len) = *len {
+                            match len.checked_sub(chunk.len() as u64) {
```
seanmonstar: If the constructor to set a length on the channel variant is private, we can rely on hyper's decoder to always limit the body size that is streamed (it enforces the length from the connection). Just for sanity, we could keep a debug assertion. Something like:

```rust
if let Some(ref mut len) = *len {
    debug_assert!(*len >= chunk.len());
    *len = *len - chunk.len();
}
```

lnicola:
Done, I think.

seanmonstar:
Thanks!

Issue #1551, opened 2018-06-06
HTTP2 should strip connection headers before sending


seanmonstar:
Connection-specific headers are forbidden in HTTP2, as that information has been moved into frame types of the h2 protocol. hyper should strip those headers before sending a request or response over HTTP2.

## Steps to fix

- Update `proto::h2::strip_connection_headers` to inspect the value of the `CONNECTION` header, and remove all headers it references. See [this code](https://github.com/runconduit/conduit/blob/v0.4.0/proxy/src/transparency/h1.rs#L73-L94) for how.
- Also remove other headers that h2 has determined are "connection" headers, such as those [in this code](https://github.com/carllerche/h2/blob/v0.1.9/src/frame/headers.rs#L758-L762).

joshleeb:
I'll start taking a look at this one 😄

SajjadPourali:
In my opinion, It has implemented inefficient, for example, if an individual needs to send the Host header manually, it is not possible to handle.
Look at the following code, which is alerady run on http://headers.me
```
use core::task::{Context, Poll};
use futures::future;
use hyper::{server::conn::Http, service::Service, Body};
use tokio::net::TcpListener;

struct Headers {
    addr: std::net::SocketAddr,
}

impl Service<hyper::Request<Body>> for Headers {
    type Response = hyper::Response<Body>;
    type Error = hyper::Error;
    type Future = future::Ready<Result<hyper::Response<Body>, Self::Error>>;

    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        Ok(()).into()
    }

    fn call(&mut self, req: hyper::Request<Body>) -> Self::Future {
        let mut final_response = format!("headers.me -> {}\r\n\r\n", self.addr.to_string());
        final_response.push_str(&format!(
            "{} {} {}\r\n",
            req.method().as_str(),
            req.uri().path_and_query().unwrap().as_str(),
            match req.version() {
                hyper::Version::HTTP_09 => "HTTP/0.9",
                hyper::Version::HTTP_10 => "HTTP/1.0",
                hyper::Version::HTTP_11 => "HTTP/1.1",
                hyper::Version::HTTP_2 => "HTTP/2.0",
                hyper::Version::HTTP_3 => "HTTP/3.0",
                _ => "Invalid",
            }
        ));

        for (key, value) in req.headers().iter() {
            final_response.push_str(&format!(
                "{}: {}\r\n",
                key,
[...]

PR #1555, opened 2018-06-08
Strip connection headers before sending

joshleeb:
Update `strip_connection_headers` function to strip connection headers
before sending.

Closes: #1551

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -1,10 +1,13 @@
 use bytes::Buf;
 use futures::{Async, Future, Poll};
 use h2::{Reason, SendStream};
+use http::header::{
+    HeaderName, CONNECTION, PROXY_AUTHENTICATE, PROXY_AUTHORIZATION, TE, TRAILER,
+    TRANSFER_ENCODING, UPGRADE,
+};
 use http::HeaderMap;
-use http::header::{CONNECTION, TRANSFER_ENCODING};
 
-use ::body::Payload;
+use body::Payload;
 
 mod client;
 mod server;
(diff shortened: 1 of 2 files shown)


src/proto/h2/mod.rs
```diff
@@ -13,13 +13,30 @@ pub(crate) use self::client::Client;
...
 
 fn strip_connection_headers(headers: &mut HeaderMap) {
-    if headers.remove(TRANSFER_ENCODING).is_some() {
-        trace!("removed illegal Transfer-Encoding header");
-    }
-    if headers.contains_key(CONNECTION) {
-        warn!("Connection header illegal in HTTP/2");
-        //TODO: actually remove it, after checking the value
-        //and removing all related headers
+    let connection_headers = vec![
```
lnicola: Make this an array, maybe?

joshleeb: :+1: That thought completely skipped my mind

src/proto/h2/mod.rs
```diff
@@ -13,13 +13,33 @@ pub(crate) use self::client::Client;
...
+            );
+            let header_contents = header.to_str().unwrap();
+
+            // A `Connection` header may have a comma-separated list of names of other headers that
+            // are meant for only this specific connection.
+            //
+            // Iterate these names and remove them as headers. Connection-specific headers are
+            // forbidden in HTTP2, as that information has been moved into frame types of the h2
+            // protocol.
+            for name in header_contents.split(',') {
```
lnicola: Actually it might be worth treating `Connection` separately, to avoid this.

src/proto/h2/mod.rs
```diff
@@ -13,13 +13,33 @@ pub(crate) use self::client::Client;
...
-        warn!("Connection header illegal in HTTP/2");
-        //TODO: actually remove it, after checking the value
-        //and removing all related headers
+    let connection_headers = [
+        CONNECTION,
+        TRANSFER_ENCODING,
+        UPGRADE,
+        HeaderName::from_lowercase(b"keep-alive").unwrap(),
+        HeaderName::from_lowercase(b"proxy-connection").unwrap(),
+    ];
```
lnicola: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection has a more complete set of headers.

src/proto/h2/mod.rs
seanmonstar: I agree, this part should only be done for the `CONNECTION` header. If someone had some weird custom `chunked: roflcopters` header, then it could be removed because of removing `Transfer-Encoding: chunked`.

joshleeb: Updated to remove the comma separated headers listed only in the `CONNECTION` header.

src/proto/h2/mod.rs
joshleeb: Updated this list to remove `CONNECTION`, and include the other headers mentioned in the mozilla docs you linked.

seanmonstar:
Great, thanks!

PR #1556, opened 2018-06-08
feat(http2): set Content-Length on outgoing messages

### LINKED ISSUE #1547: HTTP2 outgoing messages should set content-length if payload knows it
For instance, a server replying with `Response::new(Body::from("Hello, World!"))` over HTTP2 should set the `content-length: 13` header.

## Steps to fix

- In `proto::h2::server`, before calling `SendResponse::send_response`, check `body.content_length()`. If it returns `Some(len)`, check if the headers contain `CONTENT_LENGTH`, and if not, add the header with the serialized length.
- Do the same in `proto::h2::client`, before calling `SendRequest::send_request`.
- Serializing into a header value can make use of `hyper::headers::content_length_value`.
- A new function, `set_content_length_if_not_present`, could probably be added to `hyper::headers`, making use of `HeaderMap::entry`.

lnicola:
Fixes #1547

### DIFF
--- a/tests/server.rs
+++ b/tests/server.rs
@@ -353,6 +353,65 @@ mod response_body_lengths {
             expects_con_len: false,
         });
     }
+
+    #[test]
+    fn http2_auto_response_with_known_length() {
+        use hyper::body::Payload;
+
+        let server = serve();
+        let addr_str = format!("http://{}", server.addr());
+        server.reply().body("Hello, World!");
+
+        hyper::rt::run(hyper::rt::lazy(move || {
+            let client: Client<_, hyper::Body> = Client::builder().http2_only(true).build_http();
+            let uri = addr_str
+                .parse::<hyper::Uri>()
+                .expect("server addr should parse");
+
+            client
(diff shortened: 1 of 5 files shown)


tests/server.rs
```diff
@@ -353,6 +353,66 @@ mod response_body_lengths {
...
+        hyper::rt::run(hyper::rt::lazy(move || {
+            let client: Client<_, hyper::Body> = Client::builder().http2_only(true).build_http();
+            let uri = addr_str
+                .parse::<hyper::Uri>()
+                .expect("server addr should parse");
+
+            client
+                .get(uri)
+                .and_then(|res| {
+                    assert_eq!(res.body().content_length(), Some(13));
```
seanmonstar: At least the header can be checked for, right? `assert_eq!(res.headers()["content-length"], "13")`.

With that, these `#[ignore]`s could be removed, and just the checking of `res.body().content_length()` could be commented out with a TODO.

lnicola: Done

Issue #1546, opened 2018-06-06
Incoming HTTP2 bodies should know content_length


seanmonstar:
When receiving a body over HTTP2, the `Body::content_length` call should be able to say how big it is if the `content-length` header was present.

## Steps to fix

- Update `Kind::H2` variant of `Body` to include an `Option<u64>`.
- Update the `Body::h2` constructor to take a secong argument, `Option<u64>`.
- Update both `proto::h2::client` and `proto::h2::server` to look for and parse a `content-length` header before calling `Body::h2`.

joshleeb:
Is it worth `Body::content_length()` returning the content length of the received response even if the header isn't present?

I assuming that a request without a `content-length` header is still valid, and that we can simple figure out the content length by taking the length of the body.

lnicola:
I think this is about returning the value of the `Content-Length` header. `hyper` doesn't buffer the message, so if the other side doesn't send the header, `Body::content_length()` has no way of knowing what to return.

joshleeb:
That makes sense. Thanks @lnicola

seanmonstar:
Exactly as @lnicola said. This is about when a user receives a `hyper::Body` from hyper, the `<Body as Payload>::content_length()` call can try to give more information. This is especially useful if using that body to stream to a different message, such as in a proxy, having made a client request, and gotten a response back, and then returning that body as your own server's body.

As the documentation of `content_length` says, it's OK to not know the answer. It's just a hint to allow hyper to add a `content-length` header if not present. This in turn means that several common body cases will include a `content-length` header without forcing the user to remember it (like returning `Body::from("hello world")`.

lnicola:
@joshleeb Do you want to take this?

PR #1557, opened 2018-06-09
Get content-length for Incoming HTTP2 Bodies

joshleeb:
Update `Body::content_length` to return the content length of when `Body::Kind` is `H2`, instead of returning `None`. `Kind::H2` also holds the value of the `content-length` header as well as the `h2::RecvStream`.

Closes: #1546

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -306,8 +306,8 @@ impl Payload for Body {
         match self.kind {
             Kind::Once(Some(ref val)) => Some(val.len() as u64),
             Kind::Once(None) => Some(0),
-            Kind::Chan { content_length: len, .. } => len,
-            Kind::H2(..) => None,
+            Kind::Chan { content_length, .. } => content_length,
+            Kind::H2(_, content_length) => content_length,
             Kind::Wrapped(..) => None,
         }
     }
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -2,14 +2,14 @@ use std::borrow::Cow;
 use std::fmt;
 
 use bytes::Bytes;
-use futures::{Async, Future, Poll, Stream};
 use futures::sync::{mpsc, oneshot};
+use futures::{Async, Future, Poll, Stream};
 use h2;
 use http::HeaderMap;
 
 use common::Never;
-use super::{Chunk, Payload};
 use super::internal::{FullDataArg, FullDataRet};
+use super::{Chunk, Payload};
 use upgrade::OnUpgrade;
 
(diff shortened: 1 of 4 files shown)


src/body/body.rs
```diff
@@ -39,7 +39,7 @@ enum Kind {
         abort_rx: oneshot::Receiver<()>,
         rx: mpsc::Receiver<Result<Chunk, ::Error>>,
     },
-    H2(h2::RecvStream),
+    H2(h2::RecvStream, Option<u64>),
```
lnicola: Nit-pick: maybe swap these and make it a struct-like variant (`H2 { ... }`)

src/body/body.rs
```diff
@@ -212,19 +222,21 @@ impl Body {
...
-                h2.poll()
-                    .map(|async| {
-                        async.map(|opt| {
-                            opt.map(|bytes| {
-                                let _ = h2.release_capacity().release_capacity(bytes.len());
-                                Chunk::from(bytes)
-                            })
+            }
+            Kind::H2 {
+                content_length: ref mut _len,
```
seanmonstar: Since it's not needed, this pattern could be replace with just `..`.

lnicola:
@joshleeb if you get a chance, can you include a commit to enable the corresponding checks in the tests from #1556?

src/proto/h2/client.rs
```diff
@@ -110,6 +111,7 @@ where
                             if let Some(len) = body.content_length() {
                                 headers::set_content_length_if_missing(req.headers_mut(), len);
                             }
+                            let content_length = content_length_parse_all(req.headers());
```
seanmonstar: This looks like it sets `content_length` of the **Response**'s body by parsing the length from the **Request**. (Curious the tests didn't catch it...)

joshleeb: It looks like it does pick it up but I suspect that because the panic is not in the main test thread it doesn't cause the test to fail. 

https://travis-ci.org/hyperium/hyper/jobs/390534326#L825
https://travis-ci.org/hyperium/hyper/jobs/390534326#L825

I'll update the tests to fix this and remove:

```rust
if let Some(len) = body.content_length() {
  headers::set_content_length_if_missing(req.headers_mut(), len);
}
```

from the implementation.

seanmonstar:
Sorry, it seems github doesn't send notifications about rebased/ammended commits. I'll look now.

joshleeb:
Ahh damn sorry about that. Should be all fixed now 👍

seanmonstar:
Thanks!

PR #1558, opened 2018-06-10
feat(h2): implement flow control for h2 bodies

### LINKED ISSUE #1548: HTTP2 should respect flow control
Currently sending bodies will just push unbounded into the h2 buffer. While this works, unbounded buffering is not so great. Instead, hyper should check for existing window capacity before polling the body for data.

## Steps to fix

- Basically, update `proto::h2::PipeToSendStream` similar to how it was done in [this commit](https://github.com/tower-rs/tower-h2/commit/d9b3140e543af7887893cbf2f4337afd28d15839#diff-6f97a47ac7dba43dc7148725e622375fR82).

sfackler:
This is implemented sligntly differently than what the TODO comment
said, but matches with tower-h2's implementation. h2 already handles the
buffering of the pending chunk and capacity management for it, so we
just want to make sure there's some capacity available (i.e. the
previous chunk has been fully written out).

Closes #1548

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -103,20 +103,21 @@ where
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
             if !self.data_done {
-                // TODO: make use of flow control on SendStream
-                // If you're looking at this and thinking of trying to fix this TODO,
-                // you may want to look at:
-                // https://docs.rs/h2/0.1.*/h2/struct.SendStream.html
-                //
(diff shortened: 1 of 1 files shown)


sfackler:
This seems to flake, but only on x86_64 windows which is pretty weird....

sfackler:
I can't seem to repro locally on a Windows install :(

sfackler:
Ah this seems to be busted on master as well.

seanmonstar:
Thanks!

PR #1569, opened 2018-06-14
feat(client): Allow modifying the connect destination

### LINKED ISSUE #1564: Can't create or modify hyper::client::connect::Destination
I'm trying to write a connector that takes a destination, modifies it and passes it to the next connector. It seems this is currently not possible because there's no way to get a mutable reference to the internal state and only hyper can create new `Destination` objects.

My usecase is having a custom dns resolver as a Connector that resolves the domain before passing it to the final HttpConnector that creates the connection.

kpcyrd:
This resolves #1564.

As far as I can tell, this is a non breaking change that could be released in a 0.0.X upgrade. Please let me know if you prefer a different implementation. :)

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -56,6 +56,12 @@ pub(super) enum Alpn {
 */
 
 impl Destination {
+    /// Set the destination uri
+    #[inline]
+    pub fn set_uri(&mut self, uri: Uri) {
+        self.uri = uri;
+    }
+
     /// Get the protocol scheme.
     #[inline]
     pub fn scheme(&self) -> &str {


seanmonstar:
Thanks for starting this up!

This may be the simple way forward, I just want to discuss the pros and cons. In this case, a user can construct a `Uri` themselves, and then easily set it. What happens if they did so without a scheme or authority? `dest.set_uri(Uri::default()` would set it to `/`. Is that a problem? Should other connectors be able to handle empty string for `scheme` and `host`?

A possible alternative is to take a string instead, `fn set_uri(&mut self, s: &str) -> Result<(), hyper::Error>`. This could require absolute URIs, and return an error otherwise. But maybe it's fine to make connectors have to deal with empty strings.

Thoughts?

cc @sfackler

seanmonstar:
I like that. Is there a good use for having `set_scheme` as well?

sfackler:
I can't think of one off the top of my head, but it seems like we might as well be uniform with the setters.

kpcyrd:
There are no setters on `Uri`, if the api allows modifying parts of the Destination, should I try to build a new Uri with the value provided? Is there a way to ensure a Uri meets all requirements of a a valid destination?

seanmonstar:
You can convert the current URI into it's parts, and then you can parse the new one and insert it, and try to convert back into a URI. If it's illegal, that should return an error already.

seanmonstar:
I wrote the suggested changes up over in https://github.com/hyperium/hyper/pull/1572.

PR #1572, opened 2018-06-18
add `set_scheme`, `set_host`, and `set_port` for `Destination`

### LINKED ISSUE #1564: Can't create or modify hyper::client::connect::Destination

seanmonstar:
Closes #1564

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -6,9 +6,11 @@
 //!   establishes connections over TCP.
 //! - The [`Connect`](Connect) trait and related types to build custom connectors.
 use std::error::Error as StdError;
+use std::mem;
 
+use bytes::{BufMut, BytesMut};
 use futures::Future;
-use http::Uri;
+use http::{uri, Uri};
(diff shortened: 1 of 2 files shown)


src/client/connect.rs
```diff
@@ -79,6 +81,144 @@ impl Destination {
...
+    /// dst.set_host("some.proxy");
+    /// assert_eq!(dst.host(), "some.proxy");
+    /// # }
+    /// ```
+    ///
+    /// # Error
+    ///
+    /// Returns an error if the string is not a valid hostname.
+    pub fn set_host(&mut self, host: &str) -> ::Result<()> {
+        if host.contains(':') {
```
sfackler: This should check for `@` as well, right?

Issue #1549, opened 2018-06-06
HTTP2 should cancel quickly if stream is reset


seanmonstar:
When sending a body over HTTP2, if it takes a while to generate another chunk (so, `Payload::poll_data` returned `NotReady`), and the remote sends a `RST_STREAM`, the body should be canceled at that moment. Currently, hyper won't notice the `RST_STREAM` until once `Payload::poll_data` notifies the task that a new chunk is ready, and then hyper tries to write it via `send_data`.

In a similar vein, in the server, if a `Future` for a `Response` takes a while, we should also be sure to cancel that `Future` immediately if the client sends a `RST_STREAM` for it.

## Steps to fix

- Essentially, the fix is applying what was done in tower-h2, in [this commit](https://github.com/tower-rs/tower-h2/commit/a9eb8fcc4ef0941a8f04807a5a63dc3452ade7b9).

joshleeb:
I'd like to take this one on 😄

joshleeb:
@seanmonstar apologies for taking so long to get to this.

I've used the `tower-h2` code as a guide but I'm not sure if I'm on the right track.

PR #1573, opened 2018-06-20
feat(https): Quick cancel on RST_STREAM

joshleeb:
Update Http2 proto to cancel quick when the stream is reset, on an
`RST_STREAM` frame.

Closes: #1549

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -116,6 +116,13 @@ where
                             None => return Err(::Error::new_canceled(None::<::Error>)),
                         }
                     }
+                } else {
+                    if let Async::Ready(reason) =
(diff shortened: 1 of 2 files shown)


seanmonstar:
@joshleeb yep, that's the right direction! The new spawned task `H2Stream` polls the user's service future until a `Response` is ready:

https://github.com/hyperium/hyper/blob/ac1af8d15b768f5ba83345ff0af2f3a9a89d4ddf/src/proto/h2/server.rs#L177

Instead of just `try_ready`, that should be update such that if it isn't ready, we **also** check `self.reply.poll_reset`. If the remote sent a `RST_STREAM`, the `H2Stream` future can just finish up immediately (which makes it drop, dropping the service future, canceling any work it was doing).

seanmonstar:
Thanks!

PR #1584, opened 2018-06-27
Avoid socket thrash

DarrenTsung:
Hi, we're experiencing an issue with the latest version of `hyper` where lots of sockets are being placed in time-wait during periods of high load / high re-use of connections.

The problem is due to the race between the `checkout` and `connect` futures. If an idle connection is returned to the pool, the `checkout` future will resolve and the `connect` future will be dropped. If the `connect` future has started establishing a connection on a socket, that socket will be closed and enter time-wait state.

So if the time to establish a new connection is `X` and time to finish a request is `Y`, if `X > Y` a new connection will be discarded. Also the ratio of `X / Y` is the factor that must be exceeded by concurrent requests in order to change the number of connections, so if a request finishes in ~0.05s and a new connection takes ~0.5s to establish then any increase of load greater than your current number of connections (up to 10x) will cause up to ~9x connections to be created and closed.

screen shot 2018-06-27 at 11 27 15 am

```
(Above) A spike in socket churn during high load. 
Yellow: TCP time-wait sockets. Green: TCP in-use sockets.
```

---

My original approach was to race the `checkout` against a shortened `connect` future which is just the connector's `connect()` future. This didn't work because the `connect()` future returned by the `Connector` begins the process of establishing a connection on the socket in the middle of the future. 

So ideally we would be able to race the checkout until a socket is chosen for connection, but this would change the `Connect` trait API so I wasn't sure if it was the right change to make. Instead I set the checkout future to just check if an idle connection was ready or return `err_canceled`. 

This way if the `avoid_socket_thrash` variable is set, the `checkout` future will always win the race and will return a `Pooled<T>` if it exists or else run the `connect` future.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -113,6 +113,7 @@ pub struct Client<C, B = Body> {
     pool: Pool<PoolClient<B>>,
     retry_canceled_requests: bool,
     set_host: bool,
+    avoid_socket_thrash: bool,
     ver: Ver,
 }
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the great description! I've thought about this before, and wondered about inserting the "lost" connect tasks into the pool anyways, to allow other checkouts to use it. Think that would also help this issue out?

DarrenTsung:
Yeah, I think that would work as well although I wasn't sure how to continue running the `connect` future without blocking the request from completing. 

The other thing is you wouldn't want to "continue" the `connect` future if it hasn't started the binding a socket, which is some middle point in the future returned by the `Connector` type.

I guess an alternative would be to check if an idle connection exists, if not race the `connect` and `checkout`, finishing the `connect` future always?

seanmonstar:
That's true... A custom lazy future could be used for the `connect` call, that the `select` can then inspect to see if it had been started yet, to determine whether it should spawn into the executor to finish up the connect, I think.

seanmonstar:
FWIW, I have something locally now that does what I described, but coming up with a test case is tricky...

seanmonstar:
OK, finally got a test case working (bleck!), and filed PR https://github.com/hyperium/hyper/pull/1585.

I *believe* making better use of "lost" connect futures will result in better performance and system health overall, over disabling the race feature directly. What do you think?

DarrenTsung:
Awesome! Thanks for making a test case, I can't imagine that was easy. 

I think #1585 is the better solution and should be the best of both worlds (requests fast as possible without socket thrash). Thanks a ton @seanmonstar! I'll test it out and leave a comment here.

DarrenTsung:
Seems good to me - not seeing any socket thrash. :shipit:

PR #1585, opened 2018-06-28
fix(client): try to reuse connections when pool checkout wins

seanmonstar:
If a checkout wins, meaning an idle connection became available before
a connect future completed, instead of just dropping the connect future,
it spawns it into the background executor to allow being placed into
the pool on completion.

Alternative solution for #1584

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -91,6 +91,7 @@ use http::uri::Scheme;
 
 use body::{Body, Payload};
 use common::Exec;
+use common::lazy as hyper_lazy;
 use self::connect::{Connect, Destination};
 use self::pool::{Pool, Poolable, Reservation};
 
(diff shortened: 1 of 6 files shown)


src/client/mod.rs
```diff
@@ -318,9 +319,35 @@ where C: Connect + Sync + 'static,
             })
         };
 
-        let race = checkout.select(connect)
-            .map(|(pooled, _work)| pooled)
-            .or_else(|(e, other)| {
+        let executor = self.executor.clone();
+        let race = checkout.select2(connect)
```
DarrenTsung: I would leave a comment here that `.started()` being false depends on the order of the futures in the Select2 struct, given that future `A` is polled before future `B`. So we shouldn't change it to be
`connect.select2(checkout)` as it would break things.

Also if Select2's logic ever changes it would break things, but ¯\\_(ツ)_/¯.

seanmonstar: Thanks, added a comment for Future Someone.

Yes, this does depend on select's current logic... though the new test would be failed, and there is already another test to make sure the connect future is lazy, since we were otherwise seeing a lot of TCP streams started up and tossed immediately with a custom connector...

DarrenTsung:
Is there a release planned soon / can we get a release for this feature?

seanmonstar:
Yep! Just fixing up some other minor things I've noticed, but should be soon!

PR #1588, opened 2018-06-30
Add a client example getting and parsing json.

### LINKED ISSUE #1581: Add an example for client making HTTP GET call, then parsing body into JSON
The current example out there fails to show how to concatenate a body, then do anything with it.

I'm new to Rust and I've been struggling for 2 hours on something that I'd imagine most beginners would benefit from having a clear documented example on.

Here is a very verbose, no magic node.js example:

```javascript
var rp = require('request-promise');

async function getResponseBody(uri) {
  var options = {
    uri: uri,
    simple: false,
    resolveWithFullResponse: true
  };

  var response = await rp.get(options);

  return JSON.parse(response.body);
}

(async function() {
  console.log(await getResponseBody("http://api.icndb.com/jokes/random"));
})();
```

ozgurakkurt:
closes #1581

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,9 @@ spmc = "0.2"
 url = "1.0"
 tokio-fs = "0.1"
 tokio-mockstream = "1.1.0"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
 
 [features]
 default = [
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thank you! Would you mind also updating the `examples/README.md` with a short description about what it is?

seanmonstar:
Awesome thank you!

PR #1592, opened 2018-07-04
feat(ext): introduce extensions, starting with ConnectionInfo

### LINKED ISSUE #1402: Get socket parameters in Client or Server
When a `Client` makes a request, it may connect to one of many hosts associated with the URL.

I would like to have a way to access the IP address associated with the remote host. This is accessible from tokio via `NetSocket::peer_addr()`.

Logically, one should also be able to access the local address with `local_addr()`, and so it makes sense to support access to many of the apis made available on the socket-level from tokio.

seanmonstar:
- Adds `client::Builder::set_conn_info` to opt-in to having connection
  info added to `Response`s from clients.
- Adds `ext::ConnectionInfo` that allows querying types (like a
  `Response`) for connection info.

Closes #1402 

------

Still WIP...

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -10,7 +10,6 @@ static PHRASE: &'static [u8] = b"Hello World!";
 
 fn main() {
     pretty_env_logger::init();
-
     let addr = ([127, 0, 0, 1], 3000).into();
 
     // new_service is run for each connection, creating a 'service'
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -7,6 +7,7 @@
 //! - The [`Connect`](Connect) trait and related types to build custom connectors.
 use std::error::Error as StdError;
 use std::mem;
+use std::net::SocketAddr;
 
 use bytes::{BufMut, BytesMut};
 use futures::Future;
@@ -42,10 +43,11 @@ pub struct Destination {
 ///
 /// This can be used to inform recipients about things like if ALPN
 /// was used, or if connected to an HTTP proxy.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Connected {
     //alpn: Alpn,
     pub(super) is_proxied: bool,
+    pub(super) remote_addr: Option<SocketAddr>,
 }
 
 /*TODO: when HTTP1 Upgrades to H2 are added, this will be needed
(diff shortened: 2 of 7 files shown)


seanmonstar:
Some pieces of information would likely be very common. The two that come to mind are `remote_addr` and `local_addr` (is knowing the locally bound port sufficiently useful to be it's own thing)? However, a custom transport could have arbitrary information it could wish to advertise (like some details specific to a certain TLS implementation). I can think of a few ways that all this could be exposed:


### 1. `ConnectionInfo` + optional "extra" info in extensions

Have the "common" information be specific fields, and allow a user to get them from a `Response` by the API exposed here, `ConnectionInfo::get(&res)`. The "extra" custom transport data would just be available through the normal `res.extensions()` API.

- **Pros**:
  - It's very likely most transports can set the common data, like `remote_addr`, and so if that is all a user is interested in, then changing transports won't affect the existence of the values.
- **Cons**:
  - If a transport can provide both the common info, and wants to provide extra info, it results in multiple allocations, storing both `Box<ConnectionInfo as Any>` and `Box<SomeTransportExtra as Any>` separately in the extensions.

Example:

```rust
// a transport:
let info = Connected::new()
    .remote_addr(remote)
    .local_addr(local)
    .extra(RustlsExtraSomething::from(&conn));
```

```rust
// getting from a Response
let common = hyper::ext::ConnectionInfo::get(&res);
println!("remote={:?}, local={:?}", common.remote_addr(), common.local_addr());
let extra = res.extensions().get::<RustlsExtraSomething>();
```

### 2. A single custom "extra" info contains all the values a transport can provided.

Do away with the `ConnectionInfo` type proposed here, and just have **all** info from the transport exist as a single value in `res.extensions()`.

- **Pros**:
  - Reduces the amount of allocations needed to provide transport info in `extensions` to just 1.
- **Cons**:
  - Changing transports means needing to remember to change what type is fetched from `extensions`.

Example:

```rust
// for a transport
let info = Connected::new()
    .extra(RustlsExtraSomething::from(&conn));
```

```rust
// getting from a Response
if let Some(extra) = res.extensions().get::<RustlsExtraSomething>() {
    println!(
        "remote = {:?}, local = {:?}, cert = {:?}",
        extra.remote,
        extra.local,
        extra.cert,
    );
}
```

seanmonstar:
Alternative 2 is explored in #1594.

PR #1594, opened 2018-07-05
feat(client): adds `HttpInfo` to responses when `HttpConnector` is used

### LINKED ISSUE #1402: Get socket parameters in Client or Server

seanmonstar:
- Adds `client::connect::Connected::extra()`, which allows connectors to
  specify arbitrary custom information about a connected transport.

If a connector provides this extra value, it will be set in the
`Response` extensions.

Closes #1402

-----

This is an alternative to #1592

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -10,7 +10,6 @@ static PHRASE: &'static [u8] = b"Hello World!";
 
 fn main() {
     pretty_env_logger::init();
-
     let addr = ([127, 0, 0, 1], 3000).into();
 
     // new_service is run for each connection, creating a 'service'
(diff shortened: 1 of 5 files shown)


sfackler:
Anything blocking this?

seanmonstar:
@sfackler I'd been waiting on the work in Linkerd to test it out before merging, in case there were any changes needed. It also allows for people to tell me it's a good or stupid idea :D

kamalmarhubi:
@seanmonstar how long do you think it'll be for this to steep? I want to take a crack at https://github.com/seanmonstar/warp/issues/9 which I believe is ultimately blocked on this.

seanmonstar:
This is a PR for the client, so I don't think it will help with that feature in warp...

kamalmarhubi:
Serves me right for the very quick search and glance...

Issue #1316, opened 2017-09-10
Implement RFC6555 Happy Eyeballs support for dual stack IPv4 and IPv6 hosts


pkazmier:
[disclaimer: Rust experience < 1 week]

While writing, literally, my first Rust program, I needed an HTTP client to fetch XKCD comic info. After a little searching, I discovered Reqwest (and thus Hyper). Much to my dismay, I discovered that 25% of the time, my HTTP requests were timing out on my MacBook. To assist in my troubleshooting, I made a small test program to rule out user error given my limited Rust experience:

```rust
extern crate reqwest;
extern crate serde_json;

fn main() {
    let mut resp = reqwest::get("http://xkcd.com/1849/info.0.json").unwrap();
    let json: serde_json::Value = resp.json().unwrap();
    println!("Response: {:?}", json["num"]);
}
```

The problem still persisted (although not consistently), so I also wrote a small test program in Go to compare results. In addition, I fired up Wireshark on my laptop to get packet captures the next time the problem occurred. Here are my findings:

- `xkcd.com` has four A and four AAAA DNS resource records for IPv4 and IPv6 addresses.
- The Rust test program attempts to connect to one of the IPv6 addresses, but the remote host does not respond, which is why it times out. The packet capture confirms as it shows only a series of unacknowledged SYN packets being sent from my laptop.

```
     96 0.000000       MacBook               2a04:4e42:200::67     TCP      98     53620 → 443 [SYN] Seq=0 Win=65535 Len=0 MSS=1412 WS=32 TSval=456198563 TSecr=0 SACK_PERM=1
    104 1.001939       MacBook               2a04:4e42:200::67     TCP      98     [TCP Retransmission] 53620 → 443 [SYN] Seq=0 Win=65535 Len=0 MSS=1412 WS=32 TSval=456199563 TSecr=0 SACK_PERM=1
    113 1.002891       MacBook               2a04:4e42:200::67     TCP      98     [TCP Retransmission] 53620 → 443 [SYN] Seq=0 Win=65535 Len=0 MSS=1412 WS=32 TSval=456200563 TSecr=0 SACK_PERM=1
    115 1.001134       MacBook               2a04:4e42:200::67     TCP      98     [TCP Retransmission] 53620 → 443 [SYN] Seq=0 Win=65535 Len=0 MSS=1412 WS=32 TSval=456201564 TSecr=0 SACK_PERM=1
[...]

PR #1598, opened 2018-07-08
feat(client): implement rfc 6555 (happy eyeballs)

hban:
First PR, criticism welcome!

Update client connector to attempt a parallel connection using alternative address family, if connection using preferred address family takes too long.

Some questions:

* What should be a default value for timeout before we attempt other connections? Currently is 300ms.
* How to write a test for this? For manual testing I used edited /etc/hosts and iptables to simulate non-functioning IPv6. Maybe same could be done for Travis? Unsure about Appveyor.

Closes: #1316

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: rust
-sudo: false
+sudo: true # Required for functional IPv6 (forces VM instead of Docker).
 dist: trusty
 matrix:
     fast_finish: true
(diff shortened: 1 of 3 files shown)


scrogson:
It appears that RFC 6555 has been obsoleted by https://tools.ietf.org/html/rfc8305

hban:
Apparently Travis doesn't enable IPv6 by default (travis-ci/travis-ci#8361), so I copied workaround from one of the comments to make it work. I hope that's OK.

seanmonstar:
> Instead, I can write a test directly for ConnectingTcp. It would remove need for this test-only custom resolver, but would mix this integration test with other unit tests.

I think this way sounds much better! I have no problem with inner tests when testing very specific behavior that is difficult to simulate in the integration tests.

hban:
Done (and commits squashed).

PR #1600, opened 2018-07-10
feat(client): add ability to include `SO_REUSEADDR` option on sockets

### LINKED ISSUE #1599: Add `reuse_address(true)` to client connections
I'm experimenting with using hyper to run load tests and ephemeral port exhaustion is a real problem. Can `reuse_address(true)` be added to the client connection code ([here](https://github.com/hyperium/hyper/blob/5b5e3090955c1b6c1e7a8cb97b43de8d099f5303/src/client/connect.rs#L413), and [here](https://github.com/hyperium/hyper/blob/5b5e3090955c1b6c1e7a8cb97b43de8d099f5303/src/client/connect.rs#L425), I believe)?

Would making the change here also apply to tls connections made using hyper_tls?

EliSnow:
Closes #1599

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -403,11 +403,15 @@ mod http {
     use self::http_connector::HttpConnectorBlockingTask;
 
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Looks great, thanks!

PR #1604, opened 2018-07-17
feat(http): reexport `http`

sanmai-NL:
For parity with other reexports.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@ extern crate bytes;
 #[macro_use] extern crate futures;
 #[cfg(feature = "runtime")] extern crate futures_cpupool;
 extern crate h2;
-extern crate http;
+pub extern crate http;
 extern crate httparse;
 extern crate iovec;
 extern crate itoa;
@@ -36,6 +36,8 @@ extern crate want;
 #[cfg(all(test, feature = "nightly"))]
 extern crate test;
 
+#[deprecated(since = "0.12.7",
+note="Use items from the reexported `http` crate instead. ")]
 pub use http::{
     header,
     HeaderMap,


sanmai-NL:
One test failed locally:
```
failures:

---- client::connect::http::tests::client_happy_eyeballs stdout ----
thread 'client::connect::http::tests::client_happy_eyeballs' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `6`', src/client/connect.rs:1022:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    client::connect::http::tests::client_happy_eyeballs
```
Should be unrelated.

seanmonstar:
So far, I've figured people could depend on the `http` crate directly if they needed the less common types. What do you propose the exports would look like if re-exporting the whole crate?

sanmai-NL:
I’m not sure I understand your last sentence.

seanmonstar:
Ah, sorry. You asked about re-exporting all of `http`. What do propose the export would be? Just `hyper::http`? Or something different?

sanmai-NL:
Yes, just `hyper::http`; as that would be the effect of `pub extern crate ...;`, which I still see as the main way to reexport. It could indeed be something else with `pub use`.

seanmonstar:
I suppose a `pub extern crate http;` would be fine... Would you like to change this PR to do that?

sanmai-NL:
@seanmonstar, done. See also rust-lang-nursery/api-guidelines#176

seanmonstar:
Sorry for forgetting about this, I've made a tiny tweak and merged this manually to master. Thanks!

MingweiSamuel:
d55b5efb890ef04e37825221deae9c57e9e602fa

PR #1619, opened 2018-08-02
refactor(lib): replace Vec::set_len(0) with clear

lnicola:
These were trivially safe, but it's still a good idea to avoid `unsafe` here, even if just for PR purposes.


seanmonstar:
Ok! I noticed that issue in std was fixed that should optimize `clear` better. I guess it won't be available until... Rust 1.30?

lnicola:
I don't really expect that additional check to matter too much in the grand scheme of things.

Indeed, it missed 1.29 by a day.

PR #1622, opened 2018-08-03
replace usage of Box<Trait> with impl Trait

nivkner:
addresses some TODOs in the code

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -238,12 +238,11 @@ where
         }
     }
 
-    //TODO: replace with `impl Future` when stable
-    pub(crate) fn send_request_retryable(&mut self, req: Request<B>) -> Box<Future<Item=Response<Body>, Error=(::Error, Option<Request<B>>)> + Send>
+    pub(crate) fn send_request_retryable(&mut self, req: Request<B>) -> impl Future<Item = Response<Body>, Error = (::Error, Option<Request<B>>)> + Send
     where
         B: Send,
     {
-        let inner = match self.dispatch.try_send(req) {
+        match self.dispatch.try_send(req) {
             Ok(rx) => {
                 Either::A(rx.then(move |res| {
                     match res {
@@ -259,8 +258,7 @@ where
                 let err = ::Error::new_canceled(Some("connection was not ready"));
                 Either::B(future::err((err, Some(req))))
             }
-        };
-        Box::new(inner)
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


seanmonstar:
Thanks for starting this!

I've been very conservative with requiring new compiler versions, so I hadn't done this yet. I'd been meaning to look into if it was possible maybe with a macro and and config set at in `build.rs` to use `Box<Trait>` until 1.26, when it can become `impl Trait`.

nivkner:
added a `cfg` to allow building on rust 1.21 as requested

src/client/mod.rs
```diff
@@ -513,10 +606,27 @@ impl<C, B> fmt::Debug for Client<C, B> {
 
 /// A `Future` that will resolve to an HTTP Response.
 #[must_use = "futures do nothing unless polled"]
+#[cfg(impl_trait_available)]
+pub struct ResponseFuture<F> {
```
seanmonstar: Unfortunately this would be a breaking change, since any place someone has before something like `struct Foo(hyper::client::ResponseFuture)` would no longer compile, since they'd need to add the generic in. And with `impl Trait`, they'd have no way of ever define the exact type there, so the generic would need to be propagated to every other type holding one. So in the mean time, *this* case will need to stay with a inner boxed trait object.

(Supposedly it will be possible later to have `struct ResponseFuture { inner: existential type Future<..> }` or something like that, so there's hope!)

src/client/conn.rs
```diff
@@ -238,8 +238,32 @@ where
         }
     }
 
-    //TODO: replace with `impl Future` when stable
-    pub(crate) fn send_request_retryable(&mut self, req: Request<B>) -> Box<Future<Item=Response<Body>, Error=(::Error, Option<Request<B>>)> + Send>
+    #[cfg(impl_trait_available)]
+    pub(crate) fn send_request_retryable(&mut self, req: Request<B>) -> impl Future<Item = Response<Body>, Error = (::Error, Option<Request<B>>)> + Send
```
seanmonstar: Oh neat! I thought that even with `cfg`s, the older compiler would fail at the parsing step (which happens before `cfg`s are applied).

seanmonstar: It's not a requirement, I'm just thinking out loud here: do you know if there's a way that the code inside these two functions could only be written once, so that edits to one doesn't mean the other accidentally drifts?

nivkner:
Removed change to `ResponseFuture`

src/client/conn.rs
sinkuu: Type-position macro?

```rust
#[cfg(impl_trait_available)]
macro_rules! impl_trait_or_box {
    (ty: $($t:tt)+) => {
        impl $($t)+
    };
    (expr: $e:expr) => {
        $e
    };
}

#[cfg(not(impl_trait_available))]
macro_rules! impl_trait_or_box {
    (ty: $($t:tt)+) => {
        Box<$($t)+>
    };
    (expr: $e:expr) => {
        Box::new($e)
    };
}

fn method(...) -> impl_trait_or_box!(ty: Future + Send) {
    ...

    impl_trait_or_box!(expr: res)
}
```

Issue #1602, opened 2018-07-13
Create server instance by passing in TcpListener


yoshuawuyts:
Hi there!

I was looking for a way to create a Hyper server instance from an existing `TcpListener` instance, and [I couldn't find one](https://github.com/hyperium/hyper/blob/bfb2ab86441563b823f3052d1e158e6bcb0791a8/src/server/tcp.rs#L25-L45). I was wondering it might be possible to add a method to allow for this.

_Note: I might totally have missed it if there's already a method that allows for this. Apologies if any of this is redundant!_

## Motivation
Systemd has the capability of creating sockets and handing them off to applications using the `$LISTEN_FD` env var. Crates such as [listenfd](https://github.com/mitsuhiko/rust-listenfd) and [systemfd](https://github.com/mitsuhiko/systemfd) make clever use of this to provide an uninterrupted reload experience during development.

As part of the CLI WG we've written [clap-port-flag](https://github.com/rust-clique/clap-port-flag), which can create a socket from `--port`, `$PORT`, or `$LISTEN_FD`. It does this by exposing a `.bind()` method that contains all logic necessary to do this.

```rust
fn main() {
  let args = Cli::from_args();
  let _tcp_listener = args.port.bind().unwrap();
}
```
In order for Systemd's `$LISTEN_FD` to work with Hyper, a method would be needed to pass in a `TcpListener` instance.

## Prior Art
`actix-web` has a [`server.listen()`](https://docs.rs/actix-web/0.6.11/actix_web/server/struct.HttpServer.html#method.listen) method that takes a `TcpListener`. It also exposes a `server.bind()` method which acts similar to Hyper's `.bind()` method.

## Implementation
I propose adding a `hyper::server::Server::listen()` method with the following signature:
```rust
pub fn listen(listener: &TcpListener) -> Builder<AddrIncoming>
```
_Note: To be honest I'm not completely sure if `Builder<AddrIncoming>` would be the right return type here. I hope it is!_

## Other Considerations
It looks like `actix-web` also exposes a [`.listen_tls()`](https://docs.rs/actix-web/0.6.11/actix_web/server/struct.HttpServer.html#method.listen_tls) method which also accepts a TLS struct. I'm not sure what the interactions between TLS and external `TcpListener` initialization would be in Hyper.

## References
- https://github.com/mitsuhiko/rust-listenfd
- https://github.com/rust-clique/clap-port-flag (maintained by yours truly)
- https://github.com/mitsuhiko/systemfd
[...]

PR #1624, opened 2018-08-06
add Server::from_tcp()

yoshuawuyts:
Heya :wave:, this implements https://github.com/hyperium/hyper/issues/1602. I'm pretty sure there's a few rough edges to this PR, but I wanted to kick things off anyway.

### Error handling
The result type for `from_tcp()` is now:
```rust
Result<Builder<tokio_tcp::Incoming>, Box<dyn error::Error>> {
```
I wasn't sure what the best way to approach the errors was; any pointers here would be fantastic!

### Runtime Features
I wasn't sure what the runtime features are for (`no_std` related perhaps?), but figured I'd follow the prior art and add them where it seemed appropriate.

### Rustfmt
Rustfmt tends to run on every save; sorry if it changed things it shouldn't have. Happy to go back in and revert changes - but also happy to leave it if it doesn't matter :grin:

### Method name
I ended up going with `.from_tcp()` because I felt it was the most descriptive, but happy to change it to any other method if people feel that's more appropriate!

### Tests
I only tested this on an example application to make sure it works. What would be the best way to test this?

---

Hope this is helpful! :tada:

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -54,18 +54,20 @@ pub mod conn;
 #[cfg(feature = "runtime")] mod tcp;
 
 use std::fmt;
-#[cfg(feature = "runtime")] use std::net::SocketAddr;
+#[cfg(feature = "runtime")] use std::net::{SocketAddr, TcpListener as StdTcpListener};
+
 #[cfg(feature = "runtime")] use std::time::Duration;
 
 use futures::{Future, Stream, Poll};
 use tokio_io::{AsyncRead, AsyncWrite};
+#[cfg(feature = "runtime")] use tokio_reactor;
 
 use body::{Body, Payload};
 use service::{NewService, Service};
 // Renamed `Http` as `Http_` for now so that people upgrading don't see an
 // error that `hyper::server::Http` is private...
 use self::conn::{Http as Http_, SpawnAll};
-#[cfg(feature = "runtime")] use self::tcp::{AddrIncoming};
+#[cfg(feature = "runtime")] use self::tcp::AddrIncoming;
 
 /// A listening HTTP server that accepts connections in both HTTP1 and HTTP2 by default.
 ///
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for starting this!

> `Box<dyn error::Error>`

This is likely the reason CI is failing, it's not compatible with the minimum supported Rust version. I'd change to returning `hyper::Error`, using `hyper::Error::new_listen()`.

> Runtime features

These enable `tokio` support, it can be disabled if people wish to use hyper with a different runtime.

> Rustfmt

Last time I tried out rustfmt (admittedly a while ago), it broke some things. I've just pushed a `.rustfmt.toml` file to master that should disable rustfmt for everyone in hyper for now. When it's released stable, I'll re-enable it. That should also make it easier for you to work in the repo without making unrelated style changes.

> `Builder<tokio_tcp::Incoming>`

This is probably fine! Or... I mean, *maybe* it could be converted to the internal `AddrIncoming` automatically, which could make it easier to eventually configure the extra features `AddrIncoming` provides? I'm uncertain.

src/server/mod.rs
```diff
@@ -96,6 +106,19 @@ impl<I> Server<I, ()> {
     }
 }
 
+impl Server<tokio_tcp::Incoming, ()> {
+    /// Create a new instance from a `std::net::TcpListener` instance.
+    #[cfg(feature = "runtime")]
```
seanmonstar: It seems CI noticed that this `cfg` should go on the `impl` instead.

src/server/mod.rs
```diff
@@ -96,6 +106,19 @@ impl<I> Server<I, ()> {
...
+    pub fn from_tcp(
+        listener: net::TcpListener,
+    ) -> Result<Builder<tokio_tcp::Incoming>, ::Error> {
```
seanmonstar: I thought about it some more, and I do think it makes sense to convert into the internal `AddrIncoming` type instead. The changes will be easy, and then anyone using it will be able to take advantage of the default improvements like sleeping when max FDs is reached.

The changes are to add a `pub(super) fn from_tcp()` to `AddrIncomong`. The `new` function can be split, such that after getting the `std_listener`, it just calls `AddrIncoming::from_tcp` internally. And then this function also can just call `AddrIncoming::from_tcp(listener)`.

src/server/mod.rs
seanmonstar: Actually, since this should change to convert into an `AddrIncoming`, this whole `impl` can just be merge with the one below it...

src/server/mod.rs
miquels: hey, I was about to send in a similar patch. I wrote [hyper-tls-hack](https://github.com/miquels/hyper-tls-hack) , an implementation of AddrIncoming that supports TLS using tokio-tls. It has a `from_std_listener(std_listener: StdTcpListener,....` method, which I was about to split off and add to hyper's AddrIncoming. See [the code here](https://github.com/miquels/hyper-tls-hack/blob/master/src/lib.rs#L153) .

seanmonstar:
Thank you! \o/

yoshuawuyts:
Yayay, so happy this landed! Thanks for all the help! :tada:

PR #1625, opened 2018-08-07
Hyper failing with Tokio's CurrentThread executor

pimeys:
This came up recently.

Hyper: master
tokio-current-thread: 0.1.1

When executing this modified example with: `cargo run --example client http://httpbin.org`

The connection does not happen and returns an error: `Error executor failed to spawn task`

The default runtime works, so does tokio-core. I don't know does this issue belong here or to tokio.


seanmonstar:
The issue here is using `tokio-current-thread` by itself, when a runtime is needed instead.

PR #1629, opened 2018-08-10
chore(*): enable `#[deny(warnings)]` at only testing

ubnt-intrepid:
This PR intends to fix the error at `cargo doc` caused by warnings.


seanmonstar:
Thanks! Silly docs.rs....

Issue #1614, opened 2018-07-30
HTTP/1.0 and keep-alive bug


miquels:
I was testing my hyper-based server with `ab` (Apache Bench) and it turns out that if you enable the `-k` (_keepalive_) option, `ab` just hangs.

Turns out that `ab` uses `HTTP/1.0` requests. It sends a `Connection: keep-alive` header, and hyper does keep the connection alive, but it does not add a `Connection: keep-alive` header to the reply. `ab` assumes that the server does **not** support keep-alive, and waits for the server to close the connection, and as a result `ab` hangs.

Hyper should probably add a `Connection: keep-alive` header to the response if it is found in the request, the request protocol is `HTTP/1.0` and there's no `Connection` header in the Response object present yet.

seanmonstar:
Yikes, you're right! Seems there's two parts to this bug:

- When the version is `HTTP/1.0`, and the response didn't send `Connection: keep-alive`, it should have closed the connection.
- When coercing an `HTTP/1.1` response to `HTTP/1.0` because that's what the client speaks, a `Connection: keep-alive` header should be added.

Probably the best place to do this is in [`Conn::enforce_version`](https://github.com/hyperium/hyper/blob/069a32b1acc5ece35643c26cf5d9972dace5481c/src/proto/h1/conn.rs#L449). If enforcing 1.0, then check for the response version. If also 1.0, then check for `Connection: keep-alive`, and it not present, call `self.state.disable_keep_alive()`. If the response is 1.1, then translating to 1.1 should add `Connection: keep-alive` if `self.state.wants_keep_alive()` is true.

lambdasqd:
PR #1634 should fix this.

**before change**
`$ ab -k http://127.0.0.1:3000/
This is ApacheBench, Version 2.3 <$Revision: 1807734 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)...apr_pollset_poll: The timeout specified has expired (70007)
`

**after change**
`$ ab -k http://127.0.0.1:3000/
This is ApacheBench, Version 2.3 <$Revision: 1807734 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient).....done


Server Software:        
[...]

PR #1634, opened 2018-08-15
fix(server): add fix for keep-alive connection on http version 1.0

lambdasqd:
Change behaviour of connection or server response when the request is
version 1.0 and the Connection: keep-alive header is not present.

1. If the response is also version 1.0, then connection is closed if the
server keep-alive header is not present.
2. If the response is version 1.1, then the keep-alive header is added
when downgrading to version 1.0.

CLOSES: #1614


seanmonstar:
Awesome PR, thanks so much!

PR #1643, opened 2018-08-25
fix(http2): allow TE "trailers" request headers

### LINKED ISSUE #1642: TE "trailers" header incorrectly stripped from HTTP/2 requests
According to the HTTP/2 spec (https://http2.github.io/http2-spec/#rfc.section.8.1.2.2), connection-specific headers are not supposed to be in HTTP/2 messages, but a TE header is allowed for request messages as long as it only contains the value "trailers". `hyper` currently strips all connection-specific headers, including TE, regardless of whether the header is a TE "trailers" header belonging to an HTTP/2 request.

okready:
The HTTP/2 spec allows TE headers in requests if the value is
"trailers". Other TE headers are still stripped.

Closes #1642

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -108,7 +108,7 @@ where
                             }
                             let (head, body) = req.into_parts();
                             let mut req = ::http::Request::from_parts(head, ());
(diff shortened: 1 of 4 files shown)


seanmonstar:
Thanks!

PR #1646, opened 2018-08-29
Add `Body::set_beacon` method

dmexe:
`Body::set_beacon` attaches the beacon value to the body. The value will be drop when the whole body is successfully written to the socket (but before flush the socket) or when the body ends
with an error.

It useful for collecting metrics or logs for the request/response cycle. Currently, if some instrumentation wants to know the exact time when a response body writing is finished it needs to wrap existing `Payload` into another `Payload`. But it loses the `__hyper_full_data` optimization and hard to implement for high-level libraries such as https://github.com/seanmonstar/warp.

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::fmt;
+use std::any::Any;
 
 use bytes::Bytes;
 use futures::sync::{mpsc, oneshot};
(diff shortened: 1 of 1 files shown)


seanmonstar:
Oh woops, I missed the initial opening, sorry!

I think this can be largely accomplished by either a) implementing `Payload` for a custom type, or b) use a custom `Stream` and `Body::wrap_stream`.

dmexe:
@seanmonstar 
I've created RFC with the more generic solution
https://github.com/hyperium/http/issues/241

Issue #1517, opened 2018-05-15
Substitute DNS Resolver


DarrenTsung:
This relates to #1174. 

It would be great if we could substitute a resolver instead of using `to_socket_addrs` and `getaddrinfo`. It seems there was already a good amount of discussion around this in #1174, but I'm not sure what the conclusion was. 

It seems like there are three options:
1. Implement custom connector with `Connect` instead of HttpConnector
2. Add resolver type to HttpConnector (#1174)
3. Modifying tokio's `ClientProto`(?). I admit I'm not sure what this actually means for end-users.

Honestly I think 2) makes a lot of sense because we get to avoid any duplication and maintenance efforts to keep a custom connector up-to-date / bug-free with HttpConnector. But again, I'm not sure if 3) is better?

Context for this is: we're trying to update from an old fork of hyper and had previously swapped out the resolver in the fork :). 

@seanmonstar thoughts?

seanmonstar:
The `ClientProto` stuff is long gone, nothing to worry about there :)

I've seen others ask to replace the resolve in `HttpConnector` also, and it would be nice to be do so. The *only* thing that has held me back is that it would be another trait to need to stabilize. I don't have a good design for one, and haven't spent time trying to solve this, since so far I've figured that just copying the TCP connecting bits from the `HttpConnector` is pretty simple.

I'd be open to working this out, though!

pimeys:
I'd follow this issue and maybe even help:

https://github.com/tokio-rs/tokio/issues/363

Would be pretty nice to just use Tokio's DNS resolver in Hyper when it's ready. There are some unofficial crates to use in your custom connector that should not be that hard to add to your stack, e.g:

https://github.com/sbstp/tokio-dns

DarrenTsung:
Ah okay great, forget about the `ClientProto` stuff then.

I think thinking about a stable trait design is a fair point and might be more involved than I expected given the wide range of use-cases `hyper` needs to support.

However, if a trait-bound on `HttpConnector` is the goal, then the only logic we can replace is here, right?

```rust
                        let work = dns::Work::new(host, port);
                        state = State::Resolving(oneshot::spawn(work, executor));
```

The input on some hypothetical resolve function would be: `fn resolve(host: String, port: u16)` as per `dns::Work::new()` and it would return a `Future<Item=Iterator<Item=SocketAddr>, Error=io::Error>` (This is the same trait structure as #1174). 

[...]

PR #1648, opened 2018-09-06
feat(client): add HttpConnector::set_global_destination_only

niklasf:
Allow restricting destinations to addresses that appear to be
publicly routable. This can be a useful security measure when
connecting to user defined URLs.

---

Is this something you would consider including? Or if not, is there a more elegant way to build this outside of hyper than copying the entire HttpConnector implementation?

### DIFF
--- a/src/client/connect.rs
+++ b/src/client/connect.rs
@@ -450,6 +450,7 @@ mod http {
         nodelay: bool,
         local_address: Option<IpAddr>,
         happy_eyeballs_timeout: Option<Duration>,
+        global_destination_only: bool,
         reuse_address: bool,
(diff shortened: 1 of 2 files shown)


seanmonstar:
Sorry for taking so long to get to this, it slipped out of my notifications.

I'd probably be more in favor of #1517, which allows for plugging in some custom DNS resolving logic, instead of adding more and more knobs to the `HttpConnector` (especially since I tried searching for other HTTP libraries to see if they handled this, and most seem to just allow plugging in a resolver).

niklasf:
Makes sense. Thanks.

PR #1659, opened 2018-09-26
docs(examples): add proxy example

ahmedcharles:


examples/proxy.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::net::SocketAddr;
+use std::sync::Arc;
+
+fn main() {
+    pretty_env_logger::init();
+
+    let in_addr = ([127, 0, 0, 1], 3001).into();
+    let out_addr: SocketAddr = ([127, 0, 0, 1], 3000).into();
+
+    let client_arc = Arc::new(Client::new());
```
seanmonstar: The `Client` is basically a `Arc` internally, so no need to double arc it.

examples/proxy.rs
```diff
@@ -0,0 +1,47 @@
...
+    let new_service = move || {
+        let client = client_arc.clone();
+        // This is the `Service` that will handle the connection.
+        // `service_fn_ok` is a helper to convert a function that
+        // returns a Response into a `Service`.
+        service_fn(move |mut req| {
+            let uri_string = format!("http://{}/{}",
+                out_addr_clone,
+                req.uri().path_and_query().map(|x| x.as_str()).unwrap_or(""));
+            let uri = HttpTryFrom::try_from(uri_string).unwrap();
```
seanmonstar: Instead of the odd trait (`HttpTryFrom` will go away someday, whenever `TryFrom` actually stabilizes (so like 2038?)), this change just be `uri_string.parse()`.

ahmedcharles:
I did add it to the readme. I think?

examples/proxy.rs
ahmedcharles: Ok, I was wondering about that.

examples/proxy.rs
ahmedcharles: I'll give that a try.

ahmedcharles:
Anything else required?

seanmonstar:
Oh sorry, I didn't get a notification of the rebased commit. Thanks!

Issue #1539, opened 2018-06-05
Response headers are not capitalized in 0.12


resilar:
I'm running Hyper server and I noticed that capitalization of response headers was lost somewhere between the versions 0.11.27 and 0.12.1. Is this intentional? New behavior:

![capitalization.png](https://user-images.githubusercontent.com/18658078/40964579-2d1933de-68b4-11e8-84bd-41e6c9f059ea.png)

pimeys:
At least the RFC notes that the headers are case-insensitive:

https://tools.ietf.org/html/rfc2616#page-31

Ja huumeet.info, tori jne :D

seanmonstar:
Yes, this is expected. The `HeaderName` from the `http` crate keeps things lowercase for performance and compatibility with HTTP2.

HTTP/1 clients and servers **should** work just fine with any case, since RFC7230 declares them as case-insensitive. There are the occasional clients or servers that erroneously require title case headers. There was a config option added for the client to use title case, and if there's need, we can also add a server option.

logaritmisk:
I'm dealing with a "hard to please" client, and title case headers is one of the things I have to use. So this is a feature I would love to see.

seanmonstar:
Adding the option to the server is... kind of "easy", in that it can make use of the title case function used by the client encoding. The only tricky part I see is that the `Server::encode` function is highly optimized, so checking a boolean in a loop may be unfortunate. I wonder if there's something clever that can be done to avoid that.

ps1dr3x:
I'm having this problem too. I'm writing a proxy and this is causing the proxied request be different from the original request. Maybe i can make some tests and try to work on a pull request?

rolftimmermans:
I also would like this feature. I looked at the code but it seems header encoding for the server is not quite the same as for the client.

Can someone point me in the right direction of how this could be tackled? Then I can try to submit a PR.

PR #1665, opened 2018-09-28
feat(server): add title case header as a server config option

avborhanian:
This is an attempt at a fix for #1539.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -231,7 +231,12 @@ impl Http1Transaction for Server {
             msg.body,
             msg.req_method
         );
-        debug_assert!(!msg.title_case_headers, "no server config for title case headers");
+
+        let write_header_name = if msg.title_case_headers { 
+            title_case 
+        } else { 
+            extend
+        };
 
         // hyper currently doesn't support returning 1xx status codes as a Response
         // This is because Service only allows returning a single Response, and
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for starting on this! It definitely reads clearly.

I downloaded the branch to compare benchmarks, since I've so far not implemented this because I didn't want to noticeably hurt performance. It does look like this made a couple slower, with the biggest impact being writing of `content-length: x` when not present around ~33% slower. I haven't yet built the TechEmpower server to see if turns out matter in a request life cycle (I expect it will on the pipeline benchmark).

A possibly crazy idea I thought of was to make the function actually generic over some `EncodeHeaders` trait. That'd allow inlining and branching to be pulled up to the top, and could maybe allow the the lower-case version to not need to break up names from `b": "`... 🤷‍♂️

avborhanian:
Damn, much too slow! I'll keep benching/testing locally then. Is there a particular bench test I should look at making sure I get to a lower iter time? Or do I need to make one in this case for the content-length: 0 situation.

seanmonstar:
The benchmarks I ran are already in the code base, you can run them yourself (compare master to your branch) via `cargo bench --features nightly --lib proto::h1::role::tests::bench_server_encode`.

And to clarify, I don't mean to say the server itself is 33% slower, just those specific benchmarks. It's possible the difference is not really noticeable in the context of a full request/response and IO cycle.

rolftimmermans:
We are very much looking for this feature and do not mind to bear the performance costs, but I get that this might not be the same for everyone. 

Is it possible to merge this, but put the code behind a feature flag, so that it doesn't influence baseline performance in the default case?

Just wondering if there is an alternative to get this included without hurting the majority use case.

Issue #1650, opened 2018-09-07
Provide access to remote IP address in incoming server requests


klausi:
When writing a server in Hyper I should be able to retrieve the remote IP address from the request.

@seanmonstar mentioned in #1410 that an API similar to the client (#1402) might be relevant?

API usage could then look like this:

```rust
fn respond_with_sender_ip(request: Request<Body>) -> BoxFut {
    let remote_address = request
        .extensions()
        .get::<HttpInfo>()
        .expect("something something sets HttpInfo")
        .remote_addr();

    Box::new(future::ok(Response::new(Body::from(remote_address))))
}
```
(Please edit this to make the proposed API usage correct :-)

rainchen:
+1 for this request, ip address info is very useful for a network service

miquels:
See also https://github.com/carllerche/tower-web/issues/71

seanmonstar:
The idea of passing an argument to the `NewService` is being explored here: https://github.com/tower-rs/tower/issues/108

If that change is done, we could pass `&Incoming::Item` or something, and then the `NewService` could grab the remote address and anything else it wants (like TLS context info) to then be used in the constructed `Service`s.

kahing:
I need to support client side certificates and it'd be great if the incoming context contains enough information for me to do that as well

seanmonstar:
I did some exploration of converting `NewService` to `MakeService<Ctx>`, such that you can access the transport when creating a `Service`. The WIP is here: https://github.com/hyperium/hyper/commit/94cc8065490497dcaf6bbb9862c0b1f28bf10126

Here's how the API worked out (mostly so as to not be a breaking change for `NewService`):

```rust
let server = Server::bind(&addr)
    .serve(make_service_fn(|conn: &AddrStream| {
        let remote_addr = conn.remote_addr();
        service_fn_ok(move |_: Request<Body>| {
            Response::new(Body::from(format!("Hello, {}", remote_addr)))
        })
    }))
```

felipenoris:
I'm working on a reverse proxy using hyper (new version for https://github.com/brendanzab/hyper-reverse-proxy), and this is quite useful to implement `x-forwarded-for` header. The former version was hooked directly to `TcpListener`, which is quite low level.

PR #1666, opened 2018-09-29
Allowing the new service to see the socket

P-E-Meunier:
This is a more general solution to fix issue #1650.

Security features sometimes rely on logging the client's IP address.


seanmonstar:
This same question is tracked here in tower: https://github.com/tower-rs/tower/issues/108

seanmonstar:
This was a breaking change, but support has been added in a non-breaking fashion in #1692. Thanks!

PR #1670, opened 2018-10-10
feat(server): allow `!Send` Servers

### LINKED ISSUE #1667: Single threaded Server
The current implementation of `hyper::server::Server` is based on the Tokio's default runtime based on work stealing threadpool. It is appropriate for most use cases, but the value of `NewService` must satisfy the bounds on thread safety (i.e. `Send + 'static`).
Today, Tokio provides the single threaded runtime. If there is another implementation of `Server` based on it, it is possible to allow user to write `NewService` without satisfying thread safety (using `Rc` or borrowing to external instance).
Do you have any plan to implement such `Server` at this time?

seanmonstar:
Until this commit, servers have required that `Service` and their
`Future` to be `Send`, since the server needs to spawn some internal
tasks to an executor, and by default, that is `tokio::spawn`, which
could be spawning to a threadpool. This was true even if the user were
certain there was no threadpool involved, and was instead using a
different single-threaded runtime, like
`tokio::runtime::current_thread`.

This changes makes all the server pieces generic over an `E`, which is
essentially `Executor<PrivateTypes<Server::Future>>`. There's a new set
of internal traits, `H2Exec` and `NewSvcExec`, which allow for the type
signature to only show the generics that the user is providing. The
traits cannot be implemented explicitly, but there are blanket
implementations for `E: Executor<SpecificType>`. If the user provides
their own executor, it simply needs to have a generic `impl<F>
Executor<F> for MyExec`. That impl can have bounds deciding whether to
require `F: Send`. If the executor does require `Send`, and the
`Service` futures are `!Send`, there will be compiler errors.

To prevent a breaking change, all the types that gained the `E` generic
have a default type set, which is the original `tokio::spawn` executor.

Closes #1667

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,6 +115,11 @@ path = "examples/send_file.rs"
 required-features = ["runtime"]
 
 [[example]]
+name = "single_threaded"
+path = "examples/single_threaded.rs"
+required-features = ["runtime"]
+
+[[example]]
 name = "state"
 path = "examples/state.rs"
 required-features = ["runtime"]
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,6 +21,8 @@ parses it with serde and outputs the result.
 
 * [`send_file`](send_file.rs) - A server that sends back content of files using tokio_fs to read the files asynchronously.
 
+* [`single_threaded`](single_threaded.rs) - A server only running on 1 thread, so it can make use of `!Send` app state (like an `Rc` counter).
+
 * [`state`](state.rs) - A webserver showing basic state sharing among requests. A counter is shared, incremented for every request, and every response is sent the last count.
 
 * [`upgrades`](upgrades.rs) - A server and client demonstrating how to do HTTP upgrades (such as WebSockets or `CONNECT` tunneling).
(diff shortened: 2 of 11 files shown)


seanmonstar:
cc @carllerche

carllerche:
A very clever approach. I wonder if it can be simplified a bit, while maintaining the same strategy.

The goal is to keep the tasks that are spawned private while being able to pass in either a Send executor or a !Send executor.

Perhaps one strategy might be to define:

```rust
pub trait HyperExecutor<S, B, I, N, E, W>: Executor<H2Stream<S::Future, B>> + Executor<NewSvcTask<I, N, S, E, W>>
where S: Service,
           B: Payload
           W: Watcher<I, S, E>,
{
}
```

If `H2Stream` and `NewSvcTask` are private, it should be impossible to implement HyperExecutor, but making HyperExecutor public would allow the trait to be clicked in the docs and the trait could be documented to explain how it works and what the user should pass in.

seanmonstar:
Unfortunately, two traits are needed, since some cases don't require a `NewService` to even be in scope (such as `Http::serve_connection(io, svc)`).

carllerche:
If there is a strong reason to have two traits, so be it.

PR #1671, opened 2018-10-16
Split record_header_indices loop to work around rustc/LLVM bug, on relevant targets

SimonSapin:
rustc issue: https://github.com/rust-lang/rust/issues/55105

Steps to reproduce:

```
rustup target add armv7-linux-androideabi
RUSTFLAGS="-Ctarget-feature=+neon" cargo build --target armv7-linux-androideabi --release
```

Output without this change:

```
   Compiling hyper v0.12.11 (/home/simon/projects/servo-deps/hyper)
LLVM ERROR: ran out of registers during register allocation
```

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -920,13 +920,41 @@ struct HeaderIndices {
 
 fn record_header_indices(bytes: &[u8], headers: &[httparse::Header], indices: &mut [HeaderIndices]) {
     let bytes_ptr = bytes.as_ptr() as usize;
-    for (header, indices) in headers.iter().zip(indices.iter_mut()) {
-        let name_start = header.name.as_ptr() as usize - bytes_ptr;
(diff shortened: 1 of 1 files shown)


seanmonstar:
Yikes!

Since this is only an issue for Android ARM, what if looping twice included some `cfg`? That way any other platform doesn't need to see slowdown.

SimonSapin:
I’ve also reproduced the error with `armv7-unknown-linux-gnueabihf` but not: `aarch64-linux-android`, `aarch64-unknown-linux-gnu`, `arm-unknown-linux-gnueabihf`, or `arm-linux-androideabi`. So this looks specific to ARMv7 but not Android.

I’ve added corresponding `cfg`s, with a macro to avoid repeating stuff.

This PR works for me when building hyper by itself, let’s try it in Servo before merging.

seanmonstar:
Nice, the macro is pretty clever, thanks!

PR #1680, opened 2018-10-24
force close stream if keep-alive disabled. don't allocate when taking a buffer.

SergioBenitez:
Happy to expand on details if necessary.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -60,14 +60,15 @@ impl<R: Read> BufReader<R> {
     /// This operation does not copy the buffer. Instead, it directly returns
     /// the internal buffer. As a result, this reader will no longer have any
     /// buffered contents and any subsequent read from this reader will not
-    /// include the returned buffered contents.
+    /// include the returned buffered contents. Note that subsequent reads may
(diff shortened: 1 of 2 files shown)


seanmonstar:
The buffer change is super clear to me. The shutdown is less so. I'd expect that just dropping the streams would close the sockets. Is that sometimes not happening?

SergioBenitez:
I don't believe dropping the stream closes the socket unless the peer closes the socket as well. This results in socket leaks when you using something like NGINX on top of Hyper because the sockets are never closed. What's more, if the client simply ignores the `Connection: close` header (which is what NGINX does), then you also get a socket leak. Among other things, this makes it pretty easy to DoS the server by sending it data that it never reads.

seanmonstar:
Dropping a `TcpStream` *should* close it automatically, thanks to [this destructor](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/fd.rs#L291). I wonder if since the sockets are duplicated, it works less well. Anyway, I don't see an issue with actually calling it, so I'm fine merging.

PR #1690, opened 2018-10-31
Remove appveyor, add windows to Travis-CI

seanmonstar:
AppVeyor is just too slow.


seanmonstar:
Apparently the Travis support of Windows isn't stable enough. Restarted the build multiple times, and it hung every time. :sadtrombone:

PR #1692, opened 2018-11-01
Change `NewService` to `MakeService` with connection context

### LINKED ISSUE #1650: Provide access to remote IP address in incoming server requests

seanmonstar:
This adjusts the way `Service`s are created for a `hyper::Server`. The
`MakeService` trait allows receiving an argument when creating a
`Service`. The implementation for `hyper::Server` expects to pass a
reference to the accepted transport (so, `&Incoming::Item`). The user
can inspect the transport before making a `Service`.

In practice, this allows for things like getting the remote socket
address, or the TLS certification, or similar.

To prevent a breaking change, there is a blanket implementation of
`MakeService` for any `NewService`. Besides implementing `MakeService`
directly, there is also added `hyper::service::make_service_fn`.

Closes #1650

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -2,29 +2,23 @@
 extern crate hyper;
 extern crate pretty_env_logger;
 
-use hyper::{Body, Response, Server};
+use hyper::{Body, Request, Response, Server};
 use hyper::service::service_fn_ok;
 use hyper::rt::{self, Future};
-
-static PHRASE: &'static [u8] = b"Hello World!";
 
 fn main() {
     pretty_env_logger::init();
     let addr = ([127, 0, 0, 1], 3000).into();
 
-    // new_service is run for each connection, creating a 'service'
-    // to handle requests for that specific connection.
-    let new_service = || {
-        // This is the `Service` that will handle the connection.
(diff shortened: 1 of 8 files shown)


sfackler:
I mildly prefer `NewService2` to something like `MakeService` for this kind of thing - it more clearly identifies the new thing as just the replacement for the old and so reduces potential confusion as to which one to work with.

seanmonstar:
@sfackler it's to match what's discussed [here](https://github.com/tower-rs/tower/issues/108), that the name `NewService` may be less ideal than `MakeService`.

sfackler:
Ah cool, that makes sense 👍

PR #1699, opened 2018-11-05
Fix panic when parsing header names larger than 64kb

seanmonstar:


vorner:
Thank you. I don't know the internals, but from a naïve look at it, it looks fine.

One question, though. Was the problem present in http2 too? Because it seems there are only changes relating to http1.

Issue #1633, opened 2018-08-14
direct Unix-domain-socket support like the "hyperlocal" crate


njaard:
The [Hyperlocal](https://docs.rs/hyperlocal) crate allows a hyper server that listens on unix domain sockets. I find it essential because I use hyper proxied behind an nginx proxy. I prefer proxying from a service on a unix domain socket because then I can use the "native" namespace as opposed to maintaining my own list of ip addresses.

If I want my server to support `unix:` address prefixes, then I need to code to choose between a hyper's TCP and hyperlocal's socket spawner. And as a result, I need to understand all the `Server`-related traits to create a function that can do that.

seanmonstar:
hyper's `Server` can be constructed with any sort of "incoming" stream of connections, such as what hyperlocal provides. Considering this flexibility, and that using unix domain sockets isn't actually the common case, it has seemed sufficient to leave this functionality outside of the hyper crate. I'm going to close this, as essentially a duplicate of #126.

njaard:
I would normally agree, but fashioning a function that accepts a service and returns an appropriate (Boxed?) server based on what the address looks like is really difficult for someone who doesn't have an exceptionally good understanding of tokio and hyper's types.

At least there should be an example (or pre-baked) function for this in hyperlocal (for which this isn't the right place).

Also, people might be more inclined to use this feature if they knew how :)

njaard:
> hyper's Server can be constructed with any sort of "incoming" stream of connections, such as what hyperlocal provides

*Maybe* this statement is technically true, but it is still **impossible** to properly implement something like "hyperlocal", as "hyperlocal" demonstates, by not actually functioning correctly.

One cause seems to be that to replicate the functionality Hyper already has with Tcp sockets but instead with Unix Domain Sockets, one would need to have a copy of pretty much all of `hyper/src/server` and `hyper/src/common` (at least), since most of those are used by hyper's TCP support, and most of those are private symbols.

Issue #126, opened 2014-11-19
UNIX domain socket support


jgillich:
I would to to access an UNIX socket with hyper. `TcpStream` and `UnixStream` are very similar, but I assume it would still require some heavy changes (or, looking at `net::HttpStream`, maybe not?). So before I waste any time on this, do you guys think it would make sense to have this in hyper?

jimmycuadra:
I am also interested in this for use with the [https://github.com/coreos/fleet](fleet) API. At least some Ruby HTTP clients support requests to UNIX sockets, e.g. [Excon](https://github.com/excon/excon).

softprops:
This may be of interest to you folks. I just published a new crate called [hyperlocal](https://github.com/softprops/hyperlocal) which provides hyper interfaces for unix domain sockets. I realized that while working on a [docker client](github.com/softprops/shiplift) which also supports unix domain sockets that I could support both tcp and domain socket interfaces using hyper's client interface. I'll be rolling hyperlocal into that soon to replace the hand crafted http unix domain socket interface I'm currently using.

jimmycuadra:
That's awesome, softprops! Thank you.

jgillich:
Very cool :) I guess this can be closed then.

softprops:
I really want to emphasize how well designed hyper's internals were made. It made it very straight forward to extend it in this way. +100 @seanmonstar

myfreeweb:
Would be nice if hyper itself could listen on unix sockets, without an additional crate.

It also should support listening on user-provided (already opened) sockets (for both tcp and unix), e.g. acquired via socket activation or file descriptor passing.

Actually I think socket activation should be handled in hyper itself as well… even though there's a brilliant pun opportunity for naming a crate that does it ;-)

prepor:
It's an essential feature for http client library, most popular http client libraries support it. We also have already have support of unix sockets in tokio (https://github.com/tokio-rs/tokio-uds). Is there a chance that such feature will be accepted in case of PR?

seanmonstar:
I wouldn't expect hyper to explicitly include a UNIX socket implementation. It's already possible to just use any external implementation by using a custom `Connect` in the `Client`. hyper's goal is provide HTTP, and be generic over whatever the transport is underneath.

Being built in to a client would make sense for a more batteries-included kind of library, and for this exact feature, see https://github.com/seanmonstar/reqwest/issues/39

softprops:
[...]

PR #1702, opened 2018-11-07
add unix domain socket support

njaard:
This patch adds native unix domain socket support to the existing api.

# Why:

1. Most of the time, if you're hosting a service, you're using a reverse proxy to add SSL support and friends.
2. If you're using a reverse proxy, you should be using a unix domain socket to connect to the service: 
  * It's faster (Unix domain sockets have lower overhead)
  * You can dynamically swap services by starting the new one and then "mv"ing the socket file over the original.
  * You don't need to keep track of and forget the port.

Because it should be the default, it should thus be supported by default.

# API

`Server::bind` binds to a IP address. I added `Server::bind_unix` to bind to a `Path`.

# Feature gate

The feature gate is "uds" to be consistent with the package name `tokio-uds` The feature gate is disabled by default so that the maintainers can become comfortable with the naming. There's no runtime overhead unless the feature is used, so there's little reason to not enable it by default eventually (on Unix systems).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ tokio-reactor = { version = "0.1", optional = true }
 tokio-tcp = { version = "0.1", optional = true }
 tokio-threadpool = { version = "0.1.3", optional = true }
 tokio-timer = { version = "0.2", optional = true }
+tokio-uds = { version = "0.2", optional = true }
 want = "0.0.6"
 
 [dev-dependencies]
@@ -64,6 +65,9 @@ runtime = [
     "tokio-tcp",
     "tokio-threadpool",
     "tokio-timer",
+]
+uds = [
+    "tokio-uds",
 ]
 nightly = []
 __internal_flaky_tests = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ tokio-reactor = { version = "0.1", optional = true }
 tokio-tcp = { version = "0.1", optional = true }
 tokio-threadpool = { version = "0.1.3", optional = true }
 tokio-timer = { version = "0.2", optional = true }
-tokio-uds = { version = "0.2", optional = true }
 want = "0.0.6"
 
 [dev-dependencies]
@@ -66,11 +65,11 @@ runtime = [
     "tokio-threadpool",
     "tokio-timer",
 ]
-uds = [
-    "tokio-uds",
-]
 nightly = []
 __internal_flaky_tests = []
+
+[target.'cfg(unix)'.dependencies]
+tokio-uds = "0.2"
 
 [profile.release]
 codegen-units = 1
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -53,12 +53,12 @@
 pub mod conn;
 mod shutdown;
 #[cfg(feature = "runtime")] mod tcp;
-#[cfg(feature = "uds")] mod uds;
+#[cfg(all(unix,feature = "runtime"))] mod uds;
 
 use std::fmt;
 #[cfg(feature = "runtime")] use std::net::{SocketAddr, TcpListener as StdTcpListener};
 
-#[cfg(feature = "uds")] use std::os::unix::net::{SocketAddr as UnixSocketAddr,
+#[cfg(all(unix,feature = "runtime"))] use std::os::unix::net::{SocketAddr as UnixSocketAddr,
     UnixListener as StdUnixListener};
 
 #[cfg(feature = "runtime")] use std::time::Duration;
@@ -75,8 +75,8 @@ use service::{NewService, Service};
 use self::conn::{Http as Http_, NoopWatcher, SpawnAll};
 use self::shutdown::{Graceful, GracefulWatcher};
 #[cfg(feature = "runtime")] use self::tcp::AddrIncoming;
-#[cfg(feature = "uds")] use self::uds::UnixIncoming;
-#[cfg(feature = "uds")] use std::path::Path;
+#[cfg(all(unix,feature = "runtime"))] use self::uds::UnixIncoming;
+#[cfg(all(unix,feature = "runtime"))] use std::path::Path;
 
 /// A listening HTTP server that accepts connections in both HTTP1 and HTTP2 by default.
 ///
@@ -145,7 +145,7 @@ impl<S> Server<AddrIncoming, S> {
     }
 }
 
-#[cfg(feature = "uds")]
+#[cfg(all(unix,feature = "runtime"))]
 impl Server<UnixIncoming, ()> {
     /// Binds to the provided address, and returns a [`Builder`](Builder).
     ///
@@ -175,7 +175,7 @@ impl Server<UnixIncoming, ()> {
     }
 }
 
-#[cfg(feature = "uds")]
+#[cfg(all(unix,feature = "runtime"))]
 impl<S> Server<UnixIncoming, S> {
     /// Returns the local address that this server is bound to.
     pub fn local_addr(&self) -> &UnixSocketAddr {
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -53,9 +53,13 @@
 pub mod conn;
 mod shutdown;
 #[cfg(feature = "runtime")] mod tcp;
+#[cfg(feature = "uds")] mod uds;
 
 use std::fmt;
 #[cfg(feature = "runtime")] use std::net::{SocketAddr, TcpListener as StdTcpListener};
+
+#[cfg(feature = "uds")] use std::os::unix::net::{SocketAddr as UnixSocketAddr,
+    UnixListener as StdUnixListener};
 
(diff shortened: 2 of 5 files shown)


seanmonstar:
Thanks for the PR!

I still feel similarly as I did when discussing this in the other issues (#126, #1633), which is that I'm not sure this belongs in hyper directly, since while UDS does have its uses, it's not common enough to be part of hyper directly.

njaard:
First, I assert that it's not common enough because it's not easily available.

Also please consider how small this patch is compared to how complex hyperlocal is (while not even being correct or fully-featured!)

src/server/uds.rs
```diff
@@ -0,0 +1,233 @@
...
+use tokio_reactor::Handle;
+use tokio_uds::UnixListener;
+use tokio_timer::Delay;
+use std::path::Path;
+
+use self::addr_stream::AddrStream;
+
+/// A stream of connections from binding to an address.
+#[must_use = "streams do nothing unless polled"]
+pub struct UnixAddrIncoming {
```
seanmonstar: I'd change this to just `UnixIncoming`.

src/server/uds.rs
```diff
@@ -0,0 +1,233 @@
...
+    use std::io::{self, Read, Write};
+    use std::os::unix::net::SocketAddr;
+    use bytes::{Buf, BufMut};
+    use futures::Poll;
+    use tokio_uds::UnixStream;
+    use tokio_io::{AsyncRead, AsyncWrite};
+
+
+    #[derive(Debug)]
+    pub struct AddrStream {
```
seanmonstar: The name `AddrStream` probably makes less sense for UDS, it could be `UnixStream` or `UdsStream` or something.

src/server/uds.rs
```diff
@@ -0,0 +1,233 @@
...
+        pub(super) fn new(tcp: UnixStream, addr: SocketAddr) -> AddrStream {
+            AddrStream {
+                inner: tcp,
+                remote_addr: addr,
+            }
+        }
+
+        /// Returns the remote (peer) address of this connection.
+        #[inline]
+        pub fn remote_addr(&self) -> &SocketAddr {
```
seanmonstar: What even is the `SocketAddr` of a UDS?

src/server/mod.rs
```diff
@@ -366,3 +410,6 @@ impl<E> Builder<AddrIncoming, E> {
     }
 }
 
+#[cfg(feature = "uds")]
+impl<E> Builder<AddrIncoming, E> {
```
seanmonstar: I think this can be removed?

src/server/mod.rs
```diff
@@ -139,6 +145,44 @@ impl<S> Server<AddrIncoming, S> {
...
+        let handle = tokio_reactor::Handle::current();
+        UnixAddrIncoming::from_std(listener, &handle)
+            .map(Server::builder)
+    }
+}
+
+#[cfg(feature = "uds")]
+impl<S> Server<UnixAddrIncoming, S> {
+    /// Returns the local address that this server is bound to.
+    pub fn local_addr(&self) -> &UnixSocketAddr {
```
seanmonstar: What is a socket address of a UDS listener?

Cargo.toml
```diff
@@ -65,6 +66,9 @@ runtime = [
     "tokio-threadpool",
     "tokio-timer",
 ]
+uds = [
+    "tokio-uds",
```
seanmonstar: I wouldn't make a separate `uds` feature, just make this part of `runtime`.

src/server/uds.rs
njaard: [Not a lot](https://doc.rust-lang.org/std/os/unix/net/struct.SocketAddr.html), but I think it should be there for completeness/correctness.

src/server/uds.rs
njaard: `std` calls it a `UnixStream` (for `TcpStream`), so I wouldn't feel right about using the same name for something else. Maybe `PathIncoming`? That is, for consistency, where hyper's tcp implementation calls it "AddrSomething", we call this "PathSomething".

Cargo.toml
njaard: Well, in that case, it should be actually platform-dependent, since Windows doesn't have Unix Domain Sockets. How does one make a feature depend on another feature (runtime) and also a platform (unix)?

src/server/uds.rs
seanmonstar: But std doesn't define a `UnixIncoming` does it?

src/server/uds.rs
seanmonstar: Fair enough!

sfackler:
> Also please consider how small this patch is compared to how complex hyperlocal is (while not even being correct or fully-featured!)

The existence of an overly complex and not correct library doesn't seem to be a particularly convincing argument that something needs to be inlined into a separate library IMO.

src/server/uds.rs
njaard: std doesn't define `TcpIncoming` either :) - std has `std::os::unix::net::SocketAddr` and `std::net::SocketAddr` which is what the hyper's tcp `AddrIncoming` is supposed to wrap.

njaard:
> The existence of an overly complex and not correct library doesn't seem to be a particularly convincing argument that something needs to be inlined into a separate library IMO.

Well, you're not wrong, but I was intending the statement as a shorthand for the following points:

* I do not believe it's possible right now to implement this feature outside in a separate library. I've tried.
* If it were possible, I believe it would require a lot of duplication of code inside of hyper. As one example, the code that does protocol promotion from Http 1 to Http 2.
* It's easy to implement this feature correctly inside the current library without any overhead.

Also:
* I believe it's correct to do so, for the same reason that `std` supports the feature directly

sfackler:
What is insufficient about passing a stream of `UnixStream` into https://docs.rs/hyper/0.12.14/hyper/server/conn/struct.Http.html#method.serve_incoming?

njaard:
> What is insufficient about passing a stream of `UnixStream` into https://docs.rs/hyper/0.12.14/hyper/server/conn/struct.Http.html#method.serve_incoming?

To make one of those, I need to have a copy of [`SpawnAll`](https://github.com/hyperium/hyper/blob/master/src/server/conn.rs#L90) which requires a duplication of most of `hyper/server/conn/` and `hyper/common/`.

sfackler:
Why specifically do you need SpawnAll?

```rust
extern crate tokio; // 0.1.11
extern crate hyper; // 0.12.13
extern crate futures; // 0.1.25

use tokio::net::UnixListener;
use hyper::server::conn::Http;
use futures::{Future, Stream};
use hyper::{Response, Body};
use hyper::service;

fn main() {
    let incoming = UnixListener::bind("/tmp/my_socket")
        .unwrap()
        .incoming();
    
    let f = Http::new()
        .serve_incoming(incoming, || service::service_fn_ok(|_| Response::new(Body::empty())))
        .map_err(|e| println!("error: {}", e))
        .for_each(|conn| {
            let f = conn.map(|_| ())
                .map_err(|e| println!("error: {}", e));
            tokio::spawn(f);
            Ok(())
        });
        
    tokio::run(f);
}
```

njaard:
> Why specifically do you need SpawnAll?

I truly don't know! Why does hyper need SpawnAll?

seanmonstar:
> I do not believe it's possible right now to implement this feature outside in a separate library. I've tried.

I've used the `Server::builder` to integrate different listeners and socket types myself, so I know it's *possible*, and not too difficult. What prevents you from doing that in this case?

I don't think this should be merged because it's hard to do externally (it isn't!). However, I took a look at various other HTTP libraries, and found some built-in support for HTTP over UDS (I looked at nodejs and Go), and that was enough to convince me to make it *easier*.

njaard:
I have renamed the classes. I ended up not taking my own advice and I didn't use "Path" for any function, instead using `UnixIncoming` and friends externally and `uds` internally (so as to not contradict how tcp works, but also mapping sensibly from an unavoidable imbalance in the naming)

sfackler:
An even simpler version that uses `Server::builder` directly:

```rust
extern crate tokio; // 0.1.11
extern crate hyper; // 0.12.13
extern crate futures; // 0.1.25

use tokio::net::UnixListener;
use hyper::server::Server;
use futures::{Future, Stream};
use hyper::{Response, Body};
use hyper::service;

fn main() {
    let incoming = UnixListener::bind("/tmp/my_socket")
        .unwrap()
        .incoming();
    
    let f = Server::builder(incoming)
        .serve(|| service::service_fn_ok(|_| Response::new(Body::empty())))
        .map_err(|e| println!("error: {}", e));

    tokio::run(f);
}
```

njaard:
I'm really confused as to why @sfackler 's code above is so simple, yet my implementation that simply mirrors the tcp implementation is so complex.

seanmonstar:
@sfackler so do you think with how easy it is to put together externally, this should still remain outside of hyper?

sfackler:
Yeah that's where I lean.

PR #1703, opened 2018-11-09
chore(client): remove unused match

king6cong:


seanmonstar:
This match isn't actually unused, it's matching an an uninhabitable type, making the entire branch unreachable. It helps with optimization.

king6cong:
Great, got it. Thanks

king6cong:
Great, got it. Thanks

PR #1711, opened 2018-11-17
introduce Accceptor for representing the conversion of incoming I/Os

ubnt-intrepid:
Currently, the server API takes the stream of incoming connections as a `Stream` that returns an asynchronous I/O *directly*. This means that spawning of the task per connection will be done *after* the preparation of connection was completed (here, the "preparation" of connection is some *asynchronous* computation for acquiring a value that implements `AsyncRead+AsyncWrite` from a "raw" connection, such as TLS handshake). If the preparation of establishing a connection takes many time, the process for successor connections cannot be started.
This PR solves this problem by explicitly handling the preparation process within the implementation of `SpawnAll` and so on. In order to abstract the preparation process, a trait `Acceptor` is newly introduced.

### Alternatives
* does not change. the "parallelization"of preparation tasks are emulated by using [`FuturesUnordered`](https://docs.rs/futures/0.1/futures/stream/futures_unordered/struct.FuturesUnordered.html) or else.
    - in this case, preparations are handled within the same task and hence they cannot be scheduled by the background.
* replaces the trait bound of incoming stream with `I::Item: Future<Item = impl AsyncRead+AsyncWrite>`.
   - it affects existing code which assumes that `I::Item: AsyncRead + AsyncWrite`.

### Compatibility Notes
It includes some *breaking* changes in the low-level server API in `hyper::server::conn`.
* add type parameters
* the associated type `<Connection<I, F, E> as Future>::Error` will be changed from `F::Error` to `hyper::Error`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -131,6 +131,10 @@ name = "upgrades"
 path = "examples/upgrades.rs"
 required-features = ["runtime"]
 
+[[example]]
+name = "custom_acceptor"
+path = "examples/custom_acceptor.rs"
+required-features = ["runtime"]
 
 [[example]]
 name = "web_api"
--- a/examples/custom_acceptor.rs
+++ b/examples/custom_acceptor.rs
@@ -1,0 +1,71 @@
+#![deny(warnings)]
+extern crate hyper;
+#[macro_use]
+extern crate log;
+extern crate pretty_env_logger;
+extern crate tokio;
+
+use hyper::{Body, Response, Server};
+use hyper::service::service_fn_ok;
+use hyper::rt::{self, Future};
+
+fn main() {
+    pretty_env_logger::init();
+
+    let addr = ([127, 0, 0, 1], 3000).into();
+
+    let server = Server::bind(&addr)
+        .acceptor(acceptor::accept)
+        .serve(|| {
+            service_fn_ok(|_| {
+                Response::new(Body::from("Hello World!"))
+            })
+        })
+        .map_err(|e| eprintln!("server error: {}", e));
+
+    println!("Listening on http://{}", addr);
+
+    rt::run(server);
+}
+
+mod acceptor {
+    use std::io::{self, Read, Write};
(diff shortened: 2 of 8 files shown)


seanmonstar:
I *think* I get what this is trying to accomplish. Do you have a small example, like how this would be used for TLS?

ubnt-intrepid:
Here is an example of using TLS with [`tokio_tls::TlsAcceptor`](https://docs.rs/tokio-tls/0.2/tokio_tls/struct.TlsAcceptor.html):

```rust
extern crate hyper;
extern crate native_tls;
extern crate tokio_tls;

use hyper::rt::Future;

use native_tls::{Identity, TlsAcceptor as NativeTlsAcceptor};
use tokio_tls::TlsAcceptor;

fn main() {
    let der = std::fs::read("./identity.p12").unwrap();
    let cert = Identity::from_pkcs12(&der, "mypass").unwrap();
    let acceptor = NativeTlsAcceptor::builder(cert)
        .build()
        .map(TlsAcceptor::from)
        .unwrap();

    hyper::rt::run(
        hyper::Server::bind(&"127.0.0.1:3000".parse().unwrap())
            .acceptor(move |io| {
                acceptor.accept(io)
            })
            .serve(|| {
                hyper::service::service_fn_ok(|_| {
                    hyper::Response::new(hyper::Body::from("Hello World!"))
                })
            })
            .map_err(|e| eprintln!("server error: {}", e));
    );
}
```

[`tokio-rustls`](https://docs.rs/tokio-rustls/0.8/tokio_rustls/struct.TlsAcceptor.html#method.accept) and [`tokio-openssl`](https://docs.rs/tokio-openssl/0.3/tokio_openssl/trait.SslAcceptorExt.html#tymethod.accept_async) also provide the similar signature of method, so they could be used the same procedure.

ubnt-intrepid:
I'm closing this PR because the proposed method cannot have a way to access the TLS transports after the handshake process is completed.

PR #1721, opened 2018-11-25
Bugfix/1720

JeffLabonte:
### Scope 

`client` module

### Subject

* Change port to port_part
* Keep tests as is and make test work 

### Body

The port function is deprecated so the function is port_part now. 
Keep same behavior using new function.

image


JeffLabonte:
I might need some help to fix the commit-message-check error.

PR #1722, opened 2018-11-26
fix(client): fix deprecated errors on port from http

JeffLabonte:
### Scope 

`client` module

### Subject

* Change port to port_part
* Keep tests as is and make test work 

### Body

The port function is deprecated so the function is port_part now. 
Keep same behavior using new function.

image


seanmonstar:
Awesome, thank you!

PR #1723, opened 2018-11-27
feat(server): add `http1_half_close(bool)` option

### LINKED ISSUE #1716: Response future not dropped on disconnect
In certain cases I see Response future  is not dropped when the connection closes. I have a minimal reproduction here: https://github.com/luben/repro-hyper

The reproduction is simplified work queue - workers wait on `GET /next`, clients submit work to `POST /`, worker submit response to `POST /response`.  What happens is that if a worker restarts, it is left  registered but disconnected and it leads to next client request to hang. I added a small debugging class that tells me when the futures are dropped. It confirms that the response future from /next is not dropped on worker disconnect.

Changing the workers to use HTTP/2 instead of HTTP/1.1 (un-comment https://github.com/luben/repro-hyper/blob/master/src/bin/worker.rs#L16) solves the issue.

seanmonstar:
This option determines whether a read EOF should close the connection
automatically. The behavior was to always allow read EOF while waiting
to respond, so this option has a default of `true`.

Setting this option to `false` will allow Service futures to be canceled
as soon as disconnect is noticed.

Closes #1716

cc @luben

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -38,6 +38,7 @@ where I: AsyncRead + AsyncWrite,
         Conn {
             io: Buffered::new(io),
             state: State {
+                allow_half_close: true,
                 cached_headers: None,
                 error: None,
                 keep_alive: KA::Busy,
@@ -73,6 +74,10 @@ where I: AsyncRead + AsyncWrite,
 
     pub fn set_title_case_headers(&mut self) {
         self.state.title_case_headers = true;
+    }
+
+    pub(crate) fn set_disable_half_close(&mut self) {
+        self.state.allow_half_close = false;
     }
 
     pub fn into_inner(self) -> (I, Bytes) {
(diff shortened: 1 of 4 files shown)


luben:
LGTM, let me try to test it on the broker example.

luben:
Works as expected. One suggestion: can we surface it through the `server::Builder` as the rest of the connection config options - I did the change locally in order to test it.


```patch
index a0e0b01..4bcd618 100644
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -280,6 +280,14 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Set whether half-closed connections are considered open.
+    ///
+    /// Default is `true`.
+    pub fn http1_half_close(mut self, val: bool) -> Self {
+        self.protocol.http1_half_close(val);
+        self
+    }
+
```

seanmonstar:
Yep, we can do that!

src/proto/h1/conn.rs
```diff
@@ -117,7 +122,7 @@ where I: AsyncRead + AsyncWrite,
 
     fn should_error_on_eof(&self) -> bool {
         // If we're idle, it's probably just the connection closing gracefully.
-        T::should_error_on_parse_eof() && !self.state.is_idle()
+        !self.state.is_idle()
```
luben: This fixes the test but I am not sure it's the right approach:
```rust
(T::should_error_on_parse_eof() || !self.state.allow_half_close) && !self.state.is_idle()
```

PR #1724, opened 2018-11-27
feat(server): add `sleep_on_errors` builder method

starsheriff:
Added a method to allow setting the `sleep_on_errors` option for accepted connections using the `Builder<AddrIncoming, E>`.

I could not find any existing tests for the builders. Please point me in the right direction if I missed it, currently I assume that testing the builders is a bit out of scope.


src/server/mod.rs
```diff
@@ -378,5 +378,25 @@ impl<E> Builder<AddrIncoming, E> {
...
+    /// the application will likely close some files (or connections), and try
+    /// to accept the connection again. If this option is true, the error will
+    /// be logged at the error level, since it is still a big deal, and then
+    /// the listener will sleep for 1 second.
+    ///
+    /// In other cases, hitting the max open files should be treat similarly
+    /// to being out-of-memory, and simply error (and shutdown). Setting this
+    /// option to false will allow that.
+    ///
+    /// For more details see [`AddrIncoming::set_sleep_on_errors`]
```
seanmonstar: I'd probably include as a last sentence "Default is true.".

src/server/mod.rs
```diff
@@ -378,5 +378,25 @@ impl<E> Builder<AddrIncoming, E> {
         self.incoming.set_nodelay(enabled);
         self
     }
+
+    /// Set wether to sleep on accept errors.
```
seanmonstar: Typo: 'whether'

src/server/mod.rs
starsheriff: I didn't include this sentence. I thought that if this default changes it's easy to forget to update the documentation in other places. Hence, the link.

starsheriff:
changed the method name as suggested to `tcp_sleep_on_accept_errors`, longer but more self descriptive.

PR #1728, opened 2018-11-30
feat(service): export `hyper::service::MakeServiceRef`

### LINKED ISSUE #1727: Make the MakeServiceRef trait public
Hello

Would it be possible to make the `MakeServiceRef` trait public? It currently is the trait bound of [Builder::serve](https://docs.rs/hyper/0.12.16/hyper/server/struct.Builder.html#method.serve). If I want to decorate that/forward something into that method, I have quite hard time writing my trait bounds in a way it satisfies it. I ended up with this eventually, but it is a huge unreadable beast:

https://github.com/vorner/spirit/blob/8c00a34bfea3fa49f0ef27d734e2ca2439d0f68b/spirit-hyper/src/lib.rs#L133

Being able to reference the trait instead of copying the bounds from its blanket implementation would reduce it somewhat I think.

seanmonstar:
It's sealed, and has a blanket implementation, and so should only be
used as bounds. Even still, its hidden from the docs.

Closes #1727 

cc @vorner

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -25,10 +25,9 @@ use common::exec::{Exec, H2Exec, NewSvcExec};
 use common::io::Rewind;
 use error::{Kind, Parse};
 use proto;
-use service::Service;
+use service::{MakeServiceRef, Service};
 use upgrade::Upgraded;
 
-pub(super) use self::make_service::MakeServiceRef;
 pub(super) use self::spawn_all::NoopWatcher;
 use self::spawn_all::NewSvcTask;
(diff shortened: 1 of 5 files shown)


vorner:
Thanks, this will be of help :-) Though the hiding from the docs might discourage users from actually using it for their trait bounds (or discovering ‒ the only way I found it was that the docs of `serve` mentioned it and it didn't link anywhere) ‒ would it be better to show it and explain the purpose clearly and when it is appropriate to use?

PR #1734, opened 2018-12-16
Mention / link to reqwest in README

nicoburns:
Many people will find `hyper` when searching for a Rust HTTP crate. Reqwest is not nearly so easy to find unless you know what you are looking for, but is what a lot of people looking for a rust http library need. Let's point them in the right direction.

I recently came across a new Rust user on Reddit who had found the hyper crate, but was trying to use libcurl bindings to make http requests because they hadn't managed to find reqwest.


seanmonstar:
Thanks!

PR #1741, opened 2019-01-10
Add FromStr impl for Name

### LINKED ISSUE #1740: Add FromStr impl for hyper::client::connect::dns::Name
To allow people to construct a `Name`, such as when unit testing a `Resolve` implementation. Providing a `FromStr` impl instead of just `new(String) -> Name` allows us to add in validation in the future if need be.

faern:
Fixes #1740


seanmonstar:
Looks like nightly Rust booped itself. Thanks!

nlevitt:
🌞🌝 The `Name` is the mother of the ten thousand things.

PR #1743, opened 2019-01-11
Add useful trait impls to Name

faern:
When building a caching resolver it would be useful to be able to have `HashMap<Name, Vec<IpAddr>>` and similar. This is currently not possible because `Name` does not implement `Eq` or `Hash`. Here I add this and a few other standard traits that felt appropriate.


seanmonstar:
Sure, thanks!

PR #1745, opened 2019-01-12
Add `Destination::try_from_uri` to allow external clients to use `Connect`

LucioFranco:
This change adds a constructor field for `Destination` this allows the `Connect` trait implementors like `HttpConnector` to be used with the lower level `hyper::client::conn::*`. This was mostly needed for the first pass at the `tower-hyper` implementation and to allow it to use `HttpConnector` and other `Connect` impls as a way to build `tower_hyper::client::Connection`'s.

library for ref usage: https://github.com/LucioFranco/tower-hyper

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -61,6 +61,18 @@ pub(super) enum Alpn {
 }
 
 impl Destination {
+    /// Try to convert a `Uri` into a `Destination`
+    ///
+    /// # Error
+    ///
+    /// Returns an error if the uri contains no authority or
+    /// no scheme.
+    pub fn try_from_uri(uri: Uri) -> ::Result<Self> {
+        uri.authority_part().ok_or(::error::Parse::Uri)?;
+        uri.scheme_part().ok_or(::error::Parse::Uri)?;
+        Ok(Destination { uri })
+    }
+
     /// Get the protocol scheme.
     #[inline]
     pub fn scheme(&self) -> &str {
@@ -590,4 +602,3 @@ mod tests {
         assert_eq!(res2.extensions().get::<Ex2>(), Some(&Ex2("hiccup")));
     }
 }
-


seanmonstar:
I think I originally made assumptions that the `Uri` would *always* be in absolute-form, and so some of the setter methods may panic if that weren't the case. Before exposing this, perhaps it'd be a good idea to make sure the setters wouldn't panic if originally created with a non-absolute URI?

LucioFranco:
@seanmonstar good point! I will take a look. Do you think it may be good to just use the default `Uri::default` and then use the setters to actually set the values of the destination so that we can just reuse the check within the setters?

```rust
pub fn new(uri: Uri) -> Self {
	let dst = Destination { uri: Uri::default() };
	dst.set_scheme(uri.scheme())?;
	dst.set_host(uri.host())?;
	dst.set_port(uri.port())?;
	dst
}
```

seanmonstar:
I actually expect that code would panic, for sure. That's because setting the `scheme` and nothing else would be an error in `http::Uri`. Perhaps instead of `Destination::new(uri)`, there could be `try_from_uri(uri) -> Result<Self>`, and to be conservative, if scheme or host isn't set, return an error?

LucioFranco:
That would work, though I wonder if this may be something that could be added to http so `Uri::is_abosolute`? And I think that is the right way to go about it, I will add a check to make sure that `scheme` and `authority` exist. I assume authority will work since it will check host and port. 

Then once `TryFrom` is stable `Destination` can just implement `TryFrom<Uri>`.

LucioFranco:
@seanmonstar I've moved it to `try_from_uri` and now has error checking. I went with the `Parse::Uri` error as it seems to fit this the best without actually adding more errors to the struct. Though if you have other ideas I am open to that.

seanmonstar:
Thanks!

PR #1747, opened 2019-01-16
feat(client): Implement Tower's Service for Client

davidbarsky:
This is _relatively_ small. Open questions: 

- Should this behind a feature flag?
    - Until `DirectService` is the default in Tower?
    - ...past that point?

I _think_ additional middleware implementations should be [tower-rs/tower-hyper](https://github.com/tower-rs/tower-hyper), such as specific timeouts and retry policies. Building on that, I wonder if it makes sense to have a “batteries-included” middleware crate—I know my employer has specific best-practices around retries that are typically followed-company wide, but that probably doesn't belong in that crate.

Still to do before merging:

- [ ] Documentation and examples for Tower in Hyper

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ tokio-reactor = { version = "0.1", optional = true }
 tokio-tcp = { version = "0.1", optional = true }
 tokio-threadpool = { version = "0.1.3", optional = true }
 tokio-timer = { version = "0.2", optional = true }
+tower-service = "0.2"
 want = "0.0.6"
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


seanmonstar:
@sfackler wait what? Got an example, or a link to something I can read more about?

sfackler:
Here's an example in rust-postgres: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/Cargo.toml#L58

While we don't do it yet, we could have an equivalent bit-vec-06 or whatever dependency that works side-by-side with the existing one.

davidbarsky:
I'll close this ticket until https://github.com/tower-rs/tower/issues/136 is resolved. I think it's best that any experimentation along these lines occur in [tower-rs/tower-hyper](https://github.com/tower-rs/tower-hyper).

PR #1751, opened 2019-01-20
feat(client): add `conn::Builder::max_buf_size()`

### LINKED ISSUE #1748: set maximum header size client will accept
all headers are effectively stored in memory (as oppose to body which is streamed) and that's no way to cap how much memory is used. In my case I need to make many concurrent requests to servers I don't control and want to cap the amount of memory used by each request.

almielczarek:
This allows users to configure a limit to client connections' read and
write buffers.

Closes #1748

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -75,6 +75,7 @@ pub struct Builder {
     h1_writev: bool,
     h1_title_case_headers: bool,
     h1_read_buf_exact_size: Option<usize>,
+    h1_max_buf_size: Option<usize>,
     http2: bool,
 }
 
@@ -435,6 +436,7 @@ impl Builder {
             h1_writev: true,
             h1_read_buf_exact_size: None,
             h1_title_case_headers: false,
+            h1_max_buf_size: None,
             http2: false,
         }
     }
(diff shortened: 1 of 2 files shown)


src/client/conn.rs
```diff
@@ -75,6 +75,7 @@ pub struct Builder {
     h1_writev: bool,
     h1_title_case_headers: bool,
     h1_read_buf_exact_size: Option<usize>,
+    max_buf_size: Option<usize>,
```
seanmonstar: Since it's specific to HTTP1, what if the name were prefixed with `h1_`, like the others?

src/client/conn.rs
```diff
@@ -462,6 +464,24 @@ impl Builder {
...
     }
+
+    /// Set the maximum buffer size for the connection.
+    ///
+    /// Default is ~400kb.
+    ///
+    /// # Panics
+    ///
+    /// The minimum value allowed is 8192. This method panics if the passed `max` is less than the minimum.
+    pub fn max_buf_size(&mut self, max: usize) -> &mut Self {
```
seanmonstar: Since it's HTTP1 only, perhaps the method should be prefixed with `http1_`, like the other builder methods in `client/mod.rs`.

src/client/conn.rs
```diff
@@ -462,6 +464,24 @@ impl Builder {
         self.h1_read_buf_exact_size = sz;
         self
     }
+
+    /// Set the maximum buffer size for the connection.
+    ///
+    /// Default is ~400kb.
```
seanmonstar: We should probably document that setting this option conflicts with setting the `http1_read_buf_exact_size` option.

almielczarek:
Didn't think of that!

That seems to be the best option.

PR #1756, opened 2019-01-24
fix(http2): revert http2 refactor causing a client hang

seanmonstar:
This reverts commit 7b7dcc8f68227b1c2dac5dcab4a68348ebe70ca1.


sfackler:
This does fix the deadlock I'm seeing!

I'll try to split out a minimal repo.

seanmonstar:
@sfackler \o/ It'd be useful to add such a test to hyper to know when this workaround can correctly be removed.

PR #1761, opened 2019-02-11
Add max_buf_size setting in the server::Builder

luben:
Similarly to client::Builder::http1_max_buf_size

I am not sure for the name, should it be `http1_max_buf_size` or it applies to both h1 and h2?

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -302,6 +302,14 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Set the maximum buffer size.
+    ///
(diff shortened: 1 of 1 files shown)


luben:
Changed the commit message to fit the limits and force pushed it in my branch to update the PR.

seanmonstar:
Hm, good point, it only applies to the HTTP1 buffer, so it should probably be prefixed with `http1_`.

luben:
I will push the name change in a minute

seanmonstar:
Thanks!

PR #1762, opened 2019-02-11
Add `into_inner` to `AddrStream`

luben:
It consumes the `AddrStream` and returns the underlying TcpStream.

### DIFF
--- a/src/server/tcp.rs
+++ b/src/server/tcp.rs
@@ -219,6 +219,12 @@ mod addr_stream {
         pub fn remote_addr(&self) -> SocketAddr {
             self.remote_addr
         }
+
+        /// Consumes the AddrStream and returns the underlying IO object
+        #[inline]
+        pub fn into_inner(self) -> TcpStream {
+            self.inner
+        }
     }
 
     impl Read for AddrStream {


src/server/tcp.rs
```diff
@@ -219,6 +219,12 @@ mod addr_stream {
         pub fn remote_addr(&self) -> SocketAddr {
             self.remote_addr
         }
+
+        /// Consumes the AddrStream and return the underlying IO object
```
davidbarsky: Copy editing nit: ”Consumes the AddrStream and *returns* the underlying IO object“

davidbarsky:
I think these changes are good! I think it'll require @seanmonstar's sign-off, but I'm not opposed to this.

luben:
Thanks David, corrected it.

seanmonstar:
I'm not opposed, but while we wait for CI, I'm curious to hear what use case you have that needed this :D

luben:
I wanted to to an "upgrade" and get hands on the raw fd. Then redirect my stderr there. Obviously I needed some other logging around, including the remote address. I tried to do it with the low-level server::conn API but I couldn't get my head around how to do the upgrades.

seanmonstar:
Makes sense, thanks!

luben:
Thanks!

PR #1763, opened 2019-02-14
feat(service): allow `FnMut` with `service_fn`

seanmonstar:
`FnMut` is a super trait of `Fn`, so these bounds changes should not be breaking.

cc @sfackler I believe you'd asked about this before.


sfackler:
❤️

PR #1767, opened 2019-02-23
fix(server): add `poll_ready` to `Service` and `MakeService`

ubnt-intrepid:
This PR adds the method `poll_ready` to `Service` and `MakeService`, which provides the ability to these services for notifying the readiness whether to accept connections or requests.

These methods corresponds to `tower_service::Service::poll_ready` and prevents to ignore calling this method when using tower's services in `hyper::Server` or `hyper::server::conn::Connection`.

The added methods have the default implementation and has no effect on existing code that implements `Service`/`MakeService` directly.

### DIFF
--- a/src/service/service.rs
+++ b/src/service/service.rs
@@ -2,7 +2,7 @@ use std::error::Error as StdError;
 use std::fmt;
 use std::marker::PhantomData;
 
-use futures::{future, Future, IntoFuture};
+use futures::{future, Async, Future, IntoFuture, Poll};
 
 use body::Payload;
 use common::Never;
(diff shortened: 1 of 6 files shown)


src/service/service.rs
```diff
@@ -26,6 +26,11 @@ pub trait Service {
     /// The `Future` returned by this `Service`.
     type Future: Future<Item=Response<Self::ResBody>, Error=Self::Error>;
 
+    #[doc(hidden)]
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
```
seanmonstar: This should probably not be `doc(hidden)`, but include similar wording to [what it says in tower-service](https://docs.rs/tower-service/0.2.0/tower_service/trait.Service.html#tymethod.poll_ready).

ubnt-intrepid: I've added doc comment to them.

LucioFranco:
@seanmonstar do you think this would be a good time for hyper to move to the actual tower traits?

seanmonstar:
@LucioFranco I think we are close! Switching completely is breaking change, so maybe should be bundled with other breaking changes (like the `http` crate needs one...).

PR #1769, opened 2019-02-27
Implement Error::source when available

### LINKED ISSUE #1768: Implement Error::source when supported
The `source` method is the better replacement for `cause`. Hyper's `Error` should implement the method.

However, it's only available on newer Rust versions. To prevent *requiring* a new version, a build script can check `rustc --version` and set a `cfg` flag such that it's only implemented when available.

sfackler:
Closes #1768


sfackler:
Should we deprecate cause2 as well?

seanmonstar:
> Should we deprecate cause2 as well?

I think so, yes.

PR #1774, opened 2019-02-28
feat(server): add `http2_max_concurrent_streams` builder option

### LINKED ISSUE #1772: Add builder option for HTTP2 max concurrent streams
The `SETTINGS_MAX_CONCURRENT_STREAMS` option controls how many concurrent streams a peer is willing to receive. Since we don't yet has push promises, this only makes sense in the server case. The corresponding method exists on the `h2` type.

- [ ] `server::Builder::http2_max_concurrent_streams`
- [ ] `server::conn::Http::http2_max_concurrent_streams`

seanmonstar:
Closes #1772

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -302,6 +302,17 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Sets the [`SETTINGS_MAX_CONCURRENT_STREAMS`][spec] option for HTTP2
+    /// connections.
+    ///
+    /// Default is no limit (`None`).
+    ///
+    /// [spec]: https://http2.github.io/http2-spec/#SETTINGS_MAX_CONCURRENT_STREAMS
+    pub fn http2_max_concurrent_streams(mut self, max: impl Into<Option<u32>>) -> Self {
+        self.protocol.http2_max_concurrent_streams(max.into());
+        self
+    }
+
     /// Set the maximum buffer size.
     ///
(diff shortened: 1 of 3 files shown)


src/server/mod.rs
```diff
@@ -302,6 +302,17 @@ impl<I, E> Builder<I, E> {
...
     }
 
+    /// Sets the [`SETTINGS_MAX_CONCURRENT_STREAMS`][spec] option for HTTP2
+    /// connections.
+    ///
+    /// Default is no limit (`None`).
+    ///
+    /// [spec]: https://http2.github.io/http2-spec/#SETTINGS_MAX_CONCURRENT_STREAMS
+    pub fn http2_max_concurrent_streams(mut self, max: impl Into<Option<u32>>) -> Self {
+        self.protocol.http2_max_concurrent_streams(max.into());
```
kleimkuhler: Why do you have `max.into()` here, but

```
if let Some(max) = max.into {
    ...
}
```

in `src/server/conn`?

seanmonstar: Hm, didn't think much about it, really!

kleimkuhler: Oh okay!

kleimkuhler: Oh no this makes sense. `h2`'s `fn max_concurrent_streams` needs a value, while here you are passing in an `Option` to `conn::Builder`'s wrapper.

PR #1775, opened 2019-03-01
feat(client, server): set stream/conn. window size

### LINKED ISSUE #1771: Add builder options to adjust HTTP2 stream and connection window size
The corresponding methods already exist on the underlying `h2` builders.

- [ ] `client::Builder::http2_initial_stream_window_size`
- [ ] `client::Builder::http2_initial_connection_window_size`
- [ ] `client::conn::Builder::http2_initial_stream_window_size`
- [ ] `client::conn::Builder::http2_initial_connection_window_size`
- [ ] `server::Builder::http2_initial_stream_window_size`
- [ ] `server::Builder::http2_initial_connection_window_size`
- [ ] `server::conn::Http::http2_initial_stream_window_size`
- [ ] `server::conn::Http::http2_initial_connection_window_size`

kleimkuhler:
Add `fn http2_initial_stream_window_size` and `fn
http2_initial_connection_window_size` for client and server.

Closes #1771 

Signed-off-by: Kevin Leimkuhler <kevinl@buoyant.io>

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -504,12 +504,9 @@ impl Builder {
         self
     }
 
-    /// Sets the [`SETTINGS_INITIAL_WINDOW_SIZE`][spec] option for HTTP2
-    /// connection-level flow control.
+    /// Sets the max connection-level flow control for HTTP2
     ///
     /// Default is 65,535
-    ///
-    /// [spec]: https://http2.github.io/http2-spec/#SETTINGS_INITIAL_WINDOW_SIZE
     pub fn http2_initial_connection_window_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
         if let Some(sz) = sz.into() {
             self.h2_builder.initial_connection_window_size(sz);
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -15,6 +15,7 @@ use std::sync::Arc;
 use bytes::Bytes;
 use futures::{Async, Future, Poll};
 use futures::future::{self, Either, Executor};
+use h2;
 use tokio_io::{AsyncRead, AsyncWrite};
 
 use body::Payload;
@@ -77,6 +78,7 @@ pub struct Builder {
     h1_read_buf_exact_size: Option<usize>,
     h1_max_buf_size: Option<usize>,
     http2: bool,
+    h2_builder: h2::client::Builder,
 }
 
 /// A future setting up HTTP over an IO object.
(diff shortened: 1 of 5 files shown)


src/server/mod.rs
```diff
@@ -302,6 +302,28 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Sets the [`SETTINGS_INITIAL_WINDOW_SIZE`][spec] option for HTTP2
+    /// stream-level flow control.
+    ///
+    /// Default is 65,535
+    ///
+    /// [spec]: https://http2.github.io/http2-spec/#SETTINGS_INITIAL_WINDOW_SIZE
+    pub fn initial_stream_window_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
```
hawkw: This should probably be 
```suggestion
    pub fn http2_initial_stream_window_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
```
for consistency with the other functions added in this branch...

src/server/mod.rs
```diff
@@ -302,6 +302,28 @@ impl<I, E> Builder<I, E> {
...
+        self
+    }
+
+    /// Sets the [`SETTINGS_INITIAL_WINDOW_SIZE`][spec] option for HTTP2
+    /// connection-level flow control.
+    ///
+    /// Default is 65,535
+    ///
+    /// [spec]: https://http2.github.io/http2-spec/#SETTINGS_INITIAL_WINDOW_SIZE
+    pub fn initial_connection_window_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
```
hawkw:
```suggestion
    pub fn http2_initial_connection_window_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
```

src/client/conn.rs
```diff
@@ -485,6 +491,32 @@ impl Builder {
...
+    ///
+    /// [spec]: https://http2.github.io/http2-spec/#SETTINGS_INITIAL_WINDOW_SIZE
+    pub fn http2_initial_stream_window_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
+        if let Some(sz) = sz.into() {
+            self.h2_builder.initial_window_size(sz);
+        }
+        self
+    }
+
+    /// Sets the [`SETTINGS_INITIAL_WINDOW_SIZE`][spec] option for HTTP2
```
seanmonstar: This is more of a bit, but there isn't actually an HTTP2 setting for the initial *connection* window size. It *always* starts at 65,535, and this is set, we just immediately send  a `WINDOW_UPDATE` for the connection to make up the difference. So I'd just remove the setting reference, and just mention this changes the max connection flow control.

kleimkuhler: Okay that makes sense. I was back and forth on documentation here because of that situation, but glad to mention the difference here instead.

seanmonstar:
Thanks!

PR #1776, opened 2019-03-01
Make the make_service_fn example a full snippet

### LINKED ISSUE #1773: The make_service_fn example doesn't seem to work
I've tried to combine the server example and the make_service_fn example, in an obvious way. From everything I read in the documentation, I think this is how the thing should work, but it does not compile. Do I do something wrong, or is the example wrong? Would it be possible to extend the example with the actual usage of the `make_svc` variable?

```rust
use hyper::rt::Future;
use hyper::{Body, Response, Request, Server};
use hyper::service::{make_service_fn, service_fn_ok};
use tokio::net::TcpStream;

fn main() {
    // Construct our SocketAddr to listen on...
    let addr = ([127, 0, 0, 1], 3000).into();

    let make_svc = make_service_fn(|socket: &TcpStream| {
        let remote_addr = socket.peer_addr().unwrap();
        service_fn_ok(move |_: Request<Body>| {
            Response::new(Body::from(format!("Hello, {}", remote_addr)))
        })
    });

    // Then bind and serve...
    let server = Server::bind(&addr)
        .serve(make_svc);

    // Finally, spawn `server` onto an Executor...
    hyper::rt::run(server.map_err(|e| {
        eprintln!("server error: {}", e);
    }));
}
```

It fails to compile with:

```
    Checking example v0.1.0 (/home/vorner/example)
error[E0277]: the trait bound `hyper::service::make_service::MakeServiceFn<[closure@src/main.rs:10:36: 15:6]>: hyper::service::make_service::MakeService<&'a hyper::server::tcp::addr_stream::AddrStream>` is not satisfied
  --> src/main.rs:19:10
   |
19 |         .serve(make_svc);
   |          ^^^^^ the trait `hyper::service::make_service::MakeService<&'a hyper::server::tcp::addr_stream::AddrStream>` is not implemented for `hyper::service::make_service::MakeServiceFn<[closure@src/main.rs:10:36: 15:6]>`
   |
   = help: the following implementations were found:
             <hyper::service::make_service::MakeServiceFn<F> as hyper::service::make_service::MakeService<&'c Ctx>>
   = note: required because of the requirements on the impl of `hyper::service::make_service::MakeServiceRef<hyper::server::tcp::addr_stream::AddrStream>` for `hyper::service::make_service::MakeServiceFn<[closure@src/main.rs:10:36: 15:6]>`
```

vorner:
Include the creation of server too. Previously, the parameter type of
the closure had a different type than the default server provided, which
wasn't obvious to the user.

I think it's better to have the full example, both because the user has a working snippet to start from, and to make sure the types don't diverge during some refactorings.

Fixes #1773

### DIFF
--- a/src/service/make_service.rs
+++ b/src/service/make_service.rs
@@ -118,17 +118,33 @@ where
 ///
 /// # Example
 ///
-/// ```rust
+/// ```rust,no_run
+/// # #[cfg(feature = "runtime")] fn main() {
 /// use std::net::TcpStream;
-/// use hyper::{Body, Request, Response};
+/// use hyper::{Body, Request, Response, Server};
+/// use hyper::rt::{self, Future};
+/// use hyper::server::conn::AddrStream;
 /// use hyper::service::{make_service_fn, service_fn_ok};
 ///
-/// let make_svc = make_service_fn(|socket: &TcpStream| {
-///     let remote_addr = socket.peer_addr().unwrap();
+/// let addr = ([127, 0, 0, 1], 3000).into();
+///
+/// let make_svc = make_service_fn(|socket: &AddrStream| {
+///     let remote_addr = socket.remote_addr();
 ///     service_fn_ok(move |_: Request<Body>| {
 ///         Response::new(Body::from(format!("Hello, {}", remote_addr)))
 ///     })
 /// });
+///
+/// // Then bind and serve...
+/// let server = Server::bind(&addr)
+///     .serve(make_svc);
+///
+/// // Finally, spawn `server` onto an Executor...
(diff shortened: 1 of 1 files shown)


vorner:
Hmm. It says some of the commit messages are invalid, but I don't see the explanation ‒ the Details button leads to the https://gitcop.com/ home page, without any details :-(. So I don't see what to fix.

seanmonstar:
Ah bummer, the doctest is failing when default features are disabled (so no runtime).

vorner:
OK, I've took inspiration in some other example around to disable it without the features.

seanmonstar:
Thanks!

PR #1780, opened 2019-03-09
Added new module: body::take_bytes

anhbakhia:
This was an invalid pull request, I'm sorry.


quininer:
After switch to `tower`, this should work with [`BufStreamExt`](https://docs.rs/tokio-buf/0.1.0/tokio_buf/trait.BufStreamExt.html#method.limit).

PR #1784, opened 2019-03-17
fix(server): prohibit the length headers on successful CONNECT

### LINKED ISSUE #1783: Server should not send Content-Length or Transfer-Encoding on successful CONNECT
From the section 4.3.6 in RFC 7231 [^1]

> A server MUST NOT send any Transfer-Encoding or Content-Length header
> fields in a 2xx (Successful) response to CONNECT.  A client MUST
> ignore any Content-Length or Transfer-Encoding header fields received
> in a successful response to CONNECT.

The current server implementation does not treat the above condition and the field of Content-Length or Transfer-Encoding are sent to client.

[^1] : https://tools.ietf.org/html/rfc7231#section-4.3.6

ubnt-intrepid:
fixes #1783.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -233,13 +233,18 @@ impl Http1Transaction for Server {
         );
         debug_assert!(!msg.title_case_headers, "no server config for title case headers");
 
+        let mut wrote_len = false;
+
         // hyper currently doesn't support returning 1xx status codes as a Response
         // This is because Service only allows returning a single Response, and
         // so if you try to reply with a e.g. 100 Continue, you have no way of
         // replying with the latter status code response.
-        let is_upgrade = msg.head.subject == StatusCode::SWITCHING_PROTOCOLS
-            || (msg.req_method == &Some(Method::CONNECT) && msg.head.subject.is_success());
-        let (ret, mut is_last) = if is_upgrade {
+        let (ret, mut is_last) = if msg.head.subject == StatusCode::SWITCHING_PROTOCOLS {
+            (Ok(()), true)
+        } else if msg.req_method == &Some(Method::CONNECT) && msg.head.subject.is_success() {
+            // Sending content-length or transfer-encoding header on 2xx response
+            // to CONNECT is forbidden in RFC 7231.
+            wrote_len = true;
             (Ok(()), true)
         } else if msg.head.subject.is_informational() {
             warn!("response with 1xx status code not supported");
@@ -282,13 +287,12 @@ impl Http1Transaction for Server {
         }
 
         let mut encoder = Encoder::length(0);
-        let mut wrote_len = false;
         let mut wrote_date = false;
         'headers: for (name, mut values) in msg.head.headers.drain() {
             match name {
                 header::CONTENT_LENGTH => {
                     if wrote_len {
-                        warn!("transfer-encoding and content-length both found, canceling");
+                        warn!("unexpected content-length found, canceling");
                         rewind(dst);
                         return Err(::Error::new_header());
                     }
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks! Out of curiosity, do you know if the additional branches affect performance much? Likely the ones with least noise are in `proto::h1::role` at the bottom of the file.

ubnt-intrepid:
The benchmark result on my environment:

master:

```
test proto::h1::role::tests::bench_parse_incoming               ... bench:       3,199 ns/iter (+/- 1,142) = 274 MB/s
test proto::h1::role::tests::bench_parse_short                  ... bench:         282 ns/iter (+/- 10) = 141 MB/s
test proto::h1::role::tests::bench_server_encode_headers_preset ... bench:         246 ns/iter (+/- 13) = 439 MB/s
test proto::h1::role::tests::bench_server_encode_no_headers     ... bench:          34 ns/iter (+/- 0) = 2235 MB/s
```

this PR:

```
test proto::h1::role::tests::bench_parse_incoming               ... bench:       3,229 ns/iter (+/- 1,014) = 271 MB/s
test proto::h1::role::tests::bench_parse_short                  ... bench:         286 ns/iter (+/- 6) = 139 MB/s
test proto::h1::role::tests::bench_server_encode_headers_preset ... bench:         256 ns/iter (+/- 20) = 421 MB/s
test proto::h1::role::tests::bench_server_encode_no_headers     ... bench:          34 ns/iter (+/- 0) = 2235 MB/s
```

seanmonstar:
Oh yea, so it does affect slightly `bench_server_encode_headers_preset`. Since this condition should be very rare, I think we can structure the change so that the normal case doesn't see any extra branches.

src/proto/h1/role.rs
```diff
@@ -233,12 +233,15 @@ impl Http1Transaction for Server {
...
+        let is_connect_success = msg.req_method == &Some(Method::CONNECT)
+            && msg.head.subject.is_success();
+
         // hyper currently doesn't support returning 1xx status codes as a Response
         // This is because Service only allows returning a single Response, and
         // so if you try to reply with a e.g. 100 Continue, you have no way of
         // replying with the latter status code response.
         let is_upgrade = msg.head.subject == StatusCode::SWITCHING_PROTOCOLS
-            || (msg.req_method == &Some(Method::CONNECT) && msg.head.subject.is_success());
+            || is_connect_success;
```
seanmonstar: How about this, to reduce branches:

```rust
if msg.head.subject == StatusCode::SWITCHING_PROTOCOLS {
    (Ok(()), true)
} else if msg.req_method == &Some(Method::CONNECT) && msg.head.subject.is_success() {
    wrote_len = true; // will prevent content-length and transfer-encoding from being written
    (Ok(()), true)
}
```

ubnt-intrepid: Seems better.

ubnt-intrepid:
Fixed the patch to make the number of branch to be the same as the original.

<details>
<summary>benchmark result</summary>
master:

```
test proto::h1::role::tests::bench_parse_incoming               ... bench:       3,176 ns/iter (+/- 74) = 276 MB/s
test proto::h1::role::tests::bench_parse_short                  ... bench:         280 ns/iter (+/- 4) = 142 MB/s
test proto::h1::role::tests::bench_server_encode_headers_preset ... bench:         259 ns/iter (+/- 30) = 416 MB/s
test proto::h1::role::tests::bench_server_encode_no_headers     ... bench:          34 ns/iter (+/- 0) = 2235 MB/s
```

this PR:

```
test proto::h1::role::tests::bench_parse_incoming               ... bench:       3,193 ns/iter (+/- 107) = 274 MB/s
test proto::h1::role::tests::bench_parse_short                  ... bench:         280 ns/iter (+/- 8) = 142 MB/s
test proto::h1::role::tests::bench_server_encode_headers_preset ... bench:         245 ns/iter (+/- 20) = 440 MB/s
test proto::h1::role::tests::bench_server_encode_no_headers     ... bench:          34 ns/iter (+/- 0) = 2235 MB/s
```
</details>

seanmonstar:
Thanks!

Issue #1786, opened 2019-03-23
How do I use server::conn::Connection::poll_without_shutdown?


vi:
It seems to require type parameter `E` to implement some private trait, with a default bind to some private type.

The closest I can get:

```rust
use hyper; // 0.12.24
use hyper::server::conn::Connection;
use hyper::service::Service;
use tokio::prelude::{AsyncRead,AsyncWrite,Future};

struct MyWrapper<T:AsyncRead+AsyncWrite,S:Service,E>
(pub Connection<T,S,E>)
//where Connection<T,S,E>: Future<Item=(),Error=()>
;


impl<T:AsyncRead+AsyncWrite,S:Service,E> MyWrapper<T,S,E> where Connection<T,S,E> : Future<Item=(),Error=()> 
{
    fn poll(&mut self) {
        self.0.poll(); // works
        self.0.poll_without_shutdown(); // compile fail
    }
}
```

```
   = note: the method `poll_without_shutdown` exists but the following trait bounds were not satisfied:
           `hyper::common::exec::Exec : hyper::common::exec::H2Exec<<S as hyper::service::service::Service>::Future, _>`
```

Or is there another way to do HTTP upgrades without high-level API?

vi:
Looks like the answer is using `futures::future::poll_fn`, but it may be a problem to get IO back from Connection.

ubnt-intrepid:
Here is an example of retrieving I/O after `poll_without_shutdown`:

https://github.com/hyperium/hyper/blob/fc18b680a5656a0c31bc09c1c70571956a1fd013/tests/server.rs#L1157-L1162

The returned future from `poll_fn` has the signature `Future<Item = Parts<I, S>, Error = hyper::Error>` and you could retrieved the upgraded I/O as follows:

```rust
let mut conn = Some(http.serve_connection(stream, service));
poll_fn(move || -> hyper::Result<_> {
    try_ready!(conn.as_mut().unwrap().poll_without_shtudown());
    Ok(conn.take().unwrap().into_parts().into())
})
.and_then(|Parts { io, read_buf, service, .. }| {
    /* do some stuff... */
})
```

(It might be useful to be enable to write...)

```rust
poll_fn(move || conn.poll_with_upgrade()) // impl Future<Item = Parts<I, S>, Error=Error>
    .and_then(|Parts { io, read_buf, service, .. }| {
        ...
    })
```

vi:
Thanks, something like this should be in docs.

[...]

PR #1787, opened 2019-03-23
Add wrappers to use poll_without_shutdown without problems

vi:
Related issue: #1786.

Not sure about code style, etc.

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -570,11 +570,9 @@ where
 
     /// Prevent shutdown of the underlying IO object at the end of service the request,
     /// instead run `into_parts`. This is a convenience wrapper over `poll_without_shutdown`.
-    pub fn without_shutdown(self) -> impl Future<Item=Parts<I,S>, Error=::Error>
-            where S:'static, I:'static
-    {
+    pub fn without_shutdown(self) -> impl Future<Item=Parts<I,S>, Error=::Error> {
         let mut conn = Some(self);
-        futures::future::poll_fn(move || -> ::Result<_> {
+        ::futures::future::poll_fn(move || -> ::Result<_> {
             try_ready!(conn.as_mut().unwrap().poll_without_shutdown());
             Ok(conn.take().unwrap().into_parts().into())
         })
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -542,6 +542,10 @@ where
     /// upgrade. Once the upgrade is completed, the connection would be "done",
     /// but it is not desired to actally shutdown the IO object. Instead you
     /// would take it back using `into_parts`.
+    ///
+    /// Use [`poll_fn`](https://docs.rs/futures/0.1.25/futures/future/fn.poll_fn.html)
+    /// and [`try_ready!`](https://docs.rs/futures/0.1.25/futures/macro.try_ready.html)
+    /// to work with this function; or use the `without_shutdown` wrapper.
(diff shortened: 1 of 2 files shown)


vi:
Example server code: always return 101 status and continue as echo server:

```rust
use tokio::prelude::{Stream,Future,AsyncRead};
use tokio::net::tcp::{TcpListener,TcpStream};
use std::net::SocketAddr;
use hyper::server::conn;
use hyper::{Body,Response};
use std::io::{Error as IoError,ErrorKind as IoErrorKind};
use hyper::service::service_fn_ok;


fn main() -> Result<(),Box<dyn std::error::Error>> {
    let sa : SocketAddr = "127.0.0.1:1234".parse()?;
    let l = TcpListener::bind(&sa)?;
    let p = l.incoming().for_each(|c:TcpStream| {
        let mut h = conn::Http::new();
        h.http1_only(true);
        h.keep_alive(false);
        h.serve_connection(c, service_fn_ok(|_rq|{
            let mut r = Response::new(Body::empty());
            *r.status_mut() = hyper::StatusCode::SWITCHING_PROTOCOLS;
            r
        }))
        .without_shutdown()
            .map_err(|e|IoError::new(IoErrorKind::Other, e))
        .and_then(|hc| {
            let (r,w) = hc.io.split();
            tokio::io::copy(r,w)
            .map(drop)
        })
    }).map_err(|e|eprintln!("{}", e));
    tokio::runtime::run(p);
    Ok(())
}
```

seanmonstar:
What do you think of just returning `impl Future` instead of making new types? It can return `futures::future::poll_fn` internally...

vi:
> ... just returning `impl Future` instead of making new types...?

1. Maybe that type can also have `into_inner(self)` to allow undoing such conversion? This contradicts `impl Trait` usage.
2. As far as I remember, it is not recommended to use `impl Trait` in public APIs if possible, especially before `abstract type` lands. `impl Trait` is infections.

vi:
> `futures::future::poll_fn`

How do I return `Parts` in that case?

seanmonstar:
The upside is that if returning an `impl Future` ends up being a problem, it's not a breaking change to make it a fully nameable type. For now, I'd prefer to keep such small structs out of the public API.

You should be able to return `Parts` exactly the same way you implemented the future. Just put the connection in an Option, and then convert when the future is ready.

vi:
Refactored the pull request with `impl Future` based on [@ubnt-intrepid's comment](https://github.com/hyperium/hyper/issues/1786#issuecomment-476696825).

Now it also includes some doc addition to `poll_without_shutdown`.

src/server/conn.rs
```diff
@@ -564,6 +568,18 @@ where
         }
     }
 
+    /// Prevent shutdown of the underlying IO object at the end of service the request,
+    /// instead run `into_parts`. This is a convenience wrapper over `poll_without_shutdown`.
+    pub fn without_shutdown(self) -> impl Future<Item=Parts<I,S>, Error=::Error>
+            where S:'static, I:'static
```
seanmonstar: Are these `'static` bounds required? I wouldn't think so, since they aren't needed to call `poll_without_shutdown` nor `into_parts`...

vi: Looks like "no"... I was confused by `'static` bounds in `Connection`'s `Future` implementation.

They leaked into disappeared type `ConnectionNoShutdow` and before the refactoring it failed to compile without such bounds.

seanmonstar:
Thank you!

PR #1789, opened 2019-03-30
Allow to carry extensions from request to response

DoumanAsh:
As name implies, this PR adds swapping extensions out of `Request` to `Response`

I'm somewhat unsure if you'd like to see it or not, but it should help users to carry extra information specific to _particular_ request.
And it allow users to do so outside of `Connector`

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -323,15 +323,22 @@ where C: Connect + Sync + 'static,
                 return Either::A(future::err(ClientError::Normal(::Error::new_user_unsupported_request_method())));
             }
 
+            //Carry extensions to response
+            let mut req_ext = http::Extensions::new();
+            std::mem::swap(&mut req_ext, req.extensions_mut());
+
             let fut = pooled.send_request_retryable(req)
                 .map_err(ClientError::map_with_reused(pooled.is_reused()));
 
             // If the Connector included 'extra' info, add to Response...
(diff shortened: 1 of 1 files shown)


seanmonstar:
I *believe* I already have existing code that uses some extensions  as flags, and automatically setting them on the `Response` when the conditions of the response didn't match would mess up that logic...

However, it seems like setting the same extensions on the response would be possible with a `Service` middleware, ya?

DoumanAsh:
>I believe I already have existing code that uses some extensions as flags, and automatically setting them on the Response when the conditions of the response didn't match would mess up that logic

Eh is that so?
Can I still use extensions for my needs then? 😄 

I'm actually need it for my client library rather server side (I don't think `Service` applies there)
But if it would be better to remain out of hyper itself, we can close PR

seanmonstar:
> Can I still use extensions for my needs then?

Heh, of course! I didn't mean to emphasize *my* code, so much as if I've written it, I'd suspect others may have too.

And my reference to a `Service` is because I've been spending so much time in tower, where the client is treated like a `Service` as well (and will be eventually). Is it not possible to do what you want outside of hyper?

DoumanAsh:
It should be possible with some extra work as client consumes request, but not impossible I guess

DoumanAsh:
@seanmonstar Btw, I was using extensions to store `sec-websocket-key` which worked quite good😄

PR #1791, opened 2019-04-06
docs(server): show JSON deserialization in API example

olehermanse:
The previous version only showed a JSON GET API.
Deserializing the body of a POST request is not
trivial so the example should show it.

The new example takes the JSON body sent in a
POST request, deserializes it, adds a field, then
serializes it and sends it back.

### DIFF
--- a/examples/web_api.rs
+++ b/examples/web_api.rs
@@ -10,80 +10,88 @@ use hyper::{Body, Chunk, Client, Method, Request, Response, Server, StatusCode, 
 use hyper::client::HttpConnector;
 use hyper::service::service_fn;
 
-#[allow(unused, deprecated)]
-use std::ascii::AsciiExt;
-
 static NOTFOUND: &[u8] = b"Not Found";
-static URL: &str = "http://127.0.0.1:1337/web_api";
+static URL: &str = "http://127.0.0.1:1337/json_api";
 static INDEX: &[u8] = b"<a href=\"test.html\">test.html</a>";
-static LOWERCASE: &[u8] = b"i am a lower case string";
+static POST_DATA: &str = r#"{"original": "data"}"#;
 
-fn response_examples(req: Request<Body>, client: &Client<HttpConnector>)
-    -> Box<Future<Item=Response<Body>, Error=hyper::Error> + Send>
-{
+type GenericError = Box<dyn std::error::Error + Send + Sync>;
+type ResponseFuture = Box<Future<Item=Response<Body>, Error=GenericError> + Send>;
+
+fn client_request_response(client: &Client<HttpConnector>) -> ResponseFuture {
+     let req = Request::builder()
+         .method(Method::POST)
+         .uri(URL)
+         .header(header::CONTENT_TYPE, "application/json")
+         .body(POST_DATA.into())
+         .unwrap();
+
+     Box::new(client.request(req).from_err().map(|web_res| {
+         // Compare the JSON we sent (before) with what we received (after):
+         let body = Body::wrap_stream(web_res.into_body().map(|b| {
+             Chunk::from(format!("<b>POST request body</b>: {}<br><b>Response</b>: {}",
+                                 POST_DATA,
+                                 std::str::from_utf8(&b).unwrap()))
+         }));
+
+         Response::new(body)
+     }))
+}
+
+fn api_post_response(req: Request<Body>) -> ResponseFuture {
+    // A web api to run against
+    Box::new(req.into_body()
+        .concat2() // Concatenate all chunks in the body
+        .from_err()
+        .and_then(|entire_body| {
+            // TODO: Replace all unwraps with proper error handling
+            let str = String::from_utf8(entire_body.to_vec())?;
+            let mut data : serde_json::Value = serde_json::from_str(&str)?;
+            data["test"] = serde_json::Value::from("test_value");
+            let json = serde_json::to_string(&data)?;
+            let response = Response::builder()
+                .status(StatusCode::OK)
+                .header(header::CONTENT_TYPE, "application/json")
+                .body(Body::from(json))?;
+            Ok(response)
+        })
+    )
+}
+
+fn api_get_response() -> ResponseFuture {
+    let data = vec!["foo", "bar"];
+    let res = match serde_json::to_string(&data) {
+        Ok(json) => {
+            Response::builder()
+                .header(header::CONTENT_TYPE, "application/json")
+                .body(Body::from(json))
+                .unwrap()
+        }
+        Err(_) => {
+            Response::builder()
+                .status(StatusCode::INTERNAL_SERVER_ERROR)
+                .body(Body::from("Internal Server Error"))
+                .unwrap()
+        }
+    };
+
+    Box::new(future::ok(res))
+}
+
+fn response_examples(req: Request<Body>, client: &Client<HttpConnector>) -> ResponseFuture {
     match (req.method(), req.uri().path()) {
         (&Method::GET, "/") | (&Method::GET, "/index.html") => {
             let body = Body::from(INDEX);
             Box::new(future::ok(Response::new(body)))
         },
         (&Method::GET, "/test.html") => {
-            // Run a web query against the web api below
-
-            // build the request
-            let req = Request::builder()
-                .method(Method::POST)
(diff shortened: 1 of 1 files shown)


examples/web_api.rs
```diff
@@ -1,89 +1,99 @@
-#![deny(warnings)]
```
seanmonstar: Could we keep the deny so we write examples without warnings?

examples/web_api.rs
```diff
@@ -1,89 +1,99 @@
...
 extern crate futures;
 extern crate hyper;
 extern crate pretty_env_logger;
 extern crate serde_json;
 
 use futures::{future, Future, Stream};
+use serde_json::{from_str, Value, to_string};
```
seanmonstar: I feel these names are a bit too generic to be used without qualifying them (`to_string("foo")` sounds like `"foo".to_string()`), what if they were used qualified instead (so, like `serde_json::to_string("foo")`)?

examples/web_api.rs
olehermanse: Didn't know that's what it was for. Thanks!

examples/web_api.rs
olehermanse: Fair enough.

olehermanse:
@seanmonstar Thanks for the review. Your comments make sense, I have fixed both.

examples/web_api.rs
```diff
@@ -10,80 +10,91 @@ use hyper::{Body, Chunk, Client, Method, Request, Response, Server, StatusCode,
...
+     }))
+}
+
+fn api_post_response(req: Request<Body>) -> ResponseFuture {
+    // A web api to run against
+    Box::new(req.into_body()
+        // A body has multiple chunks, have to combine them before deserializing:
+        .fold(Vec::new(), |mut accumulator, chunk| {
+            accumulator.extend(&chunk[..]);
+            future::ok::<_, hyper::Error>(accumulator)
```
olehermanse: @seanmonstar Feel free to review this closer. I'm new to both hyper and rust. Is there a nicer way to convert the body chunks to a vector / string for use with `serde_json::from_str`?

seanmonstar: You can just use `into_body().concat2()` instead of the `fold`.

olehermanse: Thanks, fixed!

examples/web_api.rs
```diff
@@ -10,80 +10,87 @@ use hyper::{Body, Chunk, Client, Method, Request, Response, Server, StatusCode,
...
+         Response::new(body)
+     }))
+}
+
+fn api_post_response(req: Request<Body>) -> ResponseFuture {
+    // A web api to run against
+    Box::new(req.into_body()
+        .concat2() // Concatenate all chunks in the body
+        .and_then(|entire_body| {
+            // TODO: Replace all unwraps with proper error handling
```
seanmonstar: If the error type of `ResponseFuture` were changed to `Box<dyn std::error::Error + Send + Sync>`, then these `unwrap`s can be replaced with `?`. And the final `future::ok(response)` can just be `Ok(response)`.

olehermanse:
```
type GenericError = Box<dyn std::error::Error + Send + Sync>;
type ResponseFuture = Box<Future<Item=Response<Body>, Error=GenericError> + Send>;

fn client_request_response(client: &Client<HttpConnector>) -> ResponseFuture {
     let req = Request::builder()
         .method(Method::POST)
         .uri(URL)
         .header(header::CONTENT_TYPE, "application/json")
         .body(POST_DATA.into())
         .unwrap();

     Box::new(client.request(req).map(|web_res| {
         // Compare the JSON we sent (before) with what we received (after):
         let body = Body::wrap_stream(web_res.into_body().map(|b| {
             Chunk::from(format!("<b>POST request body</b>: {}<br><b>Response</b>: {}",
                                 POST_DATA,
                                 std::str::from_utf8(&b).unwrap()))
         }));

         Response::new(body)
     }))
}
```

```
olehermanse@OHMBP-2869 hyper $ cargo run --example web_api
   Compiling hyper v0.12.25 (/Users/olehermanse/code/hyper)
error[E0271]: type mismatch resolving `<futures::Map<hyper::client::ResponseFuture, [closure@examples/web_api.rs:29:39: 38:7]> as futures::Future>::Error == std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>`
  --> examples/web_api.rs:29:6
   |
29 | /      Box::new(client.request(req).map(|web_res| {
30 | |          // Compare the JSON we sent (before) with what we received (after):
31 | |          let body = Body::wrap_stream(web_res.into_body().map(|b| {
32 | |              Chunk::from(format!("<b>POST request body</b>: {}<br><b>Response</b>: {}",
...  |
37 | |          Response::new(body)
38 | |      }))
   | |________^ expected struct `hyper::Error`, found struct `std::boxed::Box`
   |
   = note: expected type `hyper::Error`
              found type `std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>`
   = note: required for the cast to the object type `dyn futures::Future<Item=hyper::Response<hyper::Body>, Error=std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>> + std::marker::Send`

error[E0271]: type mismatch resolving `<futures::AndThen<futures::stream::Concat2<hyper::Body>, std::result::Result<hyper::Response<hyper::Body>, hyper::Error>, [closure@examples/web_api.rs:45:19: 57:10]> as futures::Future>::Error == std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>`
  --> examples/web_api.rs:43:5
   |
43 | /     Box::new(req.into_body()
44 | |         .concat2() // Concatenate all chunks in the body
45 | |         .and_then(|entire_body| {
46 | |             // TODO: Replace all unwraps with proper error handling
...  |
57 | |         })
58 | |     )
   | |_____^ expected struct `hyper::Error`, found struct `std::boxed::Box`
   |
   = note: expected type `hyper::Error`
              found type `std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>`
   = note: required for the cast to the object type `dyn futures::Future<Item=hyper::Response<hyper::Body>, Error=std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>> + std::marker::Send`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0271`.

olehermanse@OHMBP-2869 hyper $ 
```

olehermanse: @seanmonstar That change causes other problems that I have no clue how to fix ^

seanmonstar: Oh sure. That's because `and_then` wants the error types to be the same, and the concat2 error is `hyper::Error`.

You can put a `.concat2().map_err(From::from)` so that the `and_then` can return a different error type. I know, it sounds like magic...

olehermanse: @seanmonstar Alright, thanks. I also found `.from_err()` which seems to do the same.

Then, `?` operator doesn't really work for futures:

```
olehermanse@OHMBP-2869 hyper $ cargo run --example web_api
   Compiling hyper v0.12.25 (/Users/olehermanse/code/hyper)
error[E0277]: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `std::ops::Try`)
  --> examples/web_api.rs:48:23
   |
48 |             let str = String::from_utf8(entire_body.to_vec())?;
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot use the `?` operator in a function that returns `futures::FutureResult<hyper::Response<hyper::Body>, _>`
   |
   = help: the trait `std::ops::Try` is not implemented for `futures::FutureResult<hyper::Response<hyper::Body>, _>`
   = note: required by `std::ops::Try::from_error`

For more information about this error, try `rustc --explain E0277`.

olehermanse@OHMBP-2869 hyper $ 
```

seanmonstar: You gotta change from `future::ok(response)` to just `Ok(response)`. The `?` will early-return `Err(err)` if it's an error, which is a `Result`. So anything else that is returned from the same closure must also be a `Result`, like `Ok(response)`.

olehermanse: Thanks, fixed.

PR #1793, opened 2019-04-10
impl Sink for Body::Sender, closes 1781

jxs:
I tried to find a general `Hyper::Error` to avoid calling unwrap, but didn't find any, to you have a suggestion or better solution?


seanmonstar:
If the channel is dropped in between calling poll_ready and send, it would error. You could return a closed error.

jxs:
ah! missed that, thanks Sean updated it!

Issue #1795, opened 2019-04-12
remove clippy warnings


bishtpawan:
**Clippy warnings**
Some  warnings encounter in role module like:
-use of len() methods which can be simplified more
-Some piece of code can be more simplified by using if let instead of match which makes code  verbose

PR #1796, opened 2019-04-12
refactor(role): code refactor

bishtpawan:
This is an attempt at a fix for #1795

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -1,15 +1,15 @@
 use std::fmt::{self, Write};
 use std::mem;
 
-use bytes::{BytesMut};
-use http::header::{self, Entry, HeaderName, HeaderValue};
+use bytes::BytesMut;
 use http::{HeaderMap, Method, StatusCode, Version};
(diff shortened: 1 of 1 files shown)


src/proto/h1/role.rs
```diff
@@ -156,7 +156,7 @@ impl Http1Transaction for Server {
                         is_te_chunked = true;
                         decoder = DecodedLength::CHUNKED;
                     }
-                },
+                }
```
seanmonstar: Was clippy complaining about these?

src/proto/h1/role.rs
```diff
@@ -308,40 +307,40 @@ impl Http1Transaction for Server {
                             encoder = Encoder::length(known_len);
 
                             #[cfg(debug_assertions)]
-                            {
```
seanmonstar: The diff makes this hard to tell, but did anything changes besides indentation? Could we revert the indent, so as to reduce rightward shift?

bishtpawan:
I think now it's easy to find changes.

PR #1799, opened 2019-04-23
Fix few case can't reuse connection

quininer:
I observed that in a few cases, hyper client cannot reuse h1 connections. After some log analysis, I think this is because [`PoolInner::put`](https://github.com/hyperium/hyper/blob/c0e08a41ccf18d82d2a90266a812261c3e52a202/src/client/pool.rs#L590) may happen between [`Pool::take`](https://github.com/hyperium/hyper/blob/c0e08a41ccf18d82d2a90266a812261c3e52a202/src/client/pool.rs#L203) and [`Pool::waiter`](https://github.com/hyperium/hyper/blob/c0e08a41ccf18d82d2a90266a812261c3e52a202/src/client/pool.rs#L309).

I believe merging these two operations will help.

### DIFF
--- a/src/client/pool.rs
+++ b/src/client/pool.rs
@@ -198,41 +198,6 @@ impl<T: Poolable> Pool<T> {
             .unwrap_or(0)
     }
 
-    fn take(&self, key: &Key) -> Option<Pooled<T>> {
-        let entry = {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Awesome find, thanks!

PR #1801, opened 2019-04-25
fix(server): skip automatic Content-Length header for HTTP 304 responses

### LINKED ISSUE #1797: Hyper should skip automatic Content-Length header for HTTP 304 responses
It's unnecessary, see [RFC 7232](https://tools.ietf.org/html/rfc7232#section-4.1).

Currently Hyper for empty bodies with Content-Length header unset automatically sets the value 0 which is incorrect for HTTP 304 responses, the correct values are either to not provide Content-Length (recommended by specification)  or to provide Content-Length of a full response.

jxs:
closes #1797

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -476,14 +476,20 @@ impl Http1Transaction for Server {
                 },
                 None |
                 Some(BodyLength::Known(0)) => {
-                    extend(dst, b"content-length: 0\r\n");
+                    if msg.head.subject != StatusCode::NOT_MODIFIED {
+                        extend(dst, b"content-length: 0\r\n");
+                    }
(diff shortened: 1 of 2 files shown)


jxs:
updated! dropped the body and not setting the content-length for `BodyLength::Known(len)`. But even if body isn't dropped on `match msg.body`it still is being dropped somewhere else.  
Added a new test case to assert body is dropped though

xfix:
HTTP 304 cannot have a body, so the case of HTTP 304 with body shouldn't happen, just saying. From RFC 7232.

> A 304 response cannot contain a message-body; it is always terminated by the first empty line after the header fields.

seanmonstar:
Thanks!

PR #1803, opened 2019-04-26
Fix error-handling state bug in SizedWriter

mattdw:
Not sure if you still care about 0.10.x at all, but because the Iron framework uses it, and we have an app still using Iron, I've fixed a bug that was causing data truncation on slow connections.

If socket writes ever failed (e.g. with ErrorKind::WouldBlock), HttpWriter::SizedWriter would still record those bytes as sent, which would result in a truncated write. Reordering those lines as per my commit allows higher layers to properly retry writing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### v0.10.16 (2019-04-26)
+
+#### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


seanmonstar:
I don't dedicate any time developing 0.10, but I'm happy to merge any bug fixes, so thanks! I'll take a look.

PR #1809, opened 2019-05-07
feat(error): add external `Error` kind

BroderickCarlin:
There is currently no means to instantiate a hyper Error from outside the scope of the hyper crate itself. The primary case where this can be beneficial is mapping Errors from other crates' futures such that they can work more naturally in the hyper ecosystem. This new Kind is not intended to provide detailed information on the error that occurred.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,9 +49,10 @@ pub(crate) enum Kind {
     BodyWrite,
     /// Error calling AsyncWrite::shutdown()
     Shutdown,
-
     /// A general error from h2.
     Http2,
+    /// An external error exposed for the purpose of mapping
+    External,
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! Do you happen to have an example where this would be helpful? Most places where a user is asked for an error have the bounds of `Into<Box<dyn std::error::Error>>`, allowing for *any* error so that it can be more descriptive.

BroderickCarlin:
Certainly. In my (albeit specific) use case I am attempting to do work on a repo that is fraught with code that has `impl`'d `Service` for many of it's data types. Within these `impl`'s it is very common to see `type Error = hyper::Error;`. One option would be for me to go through and convert everything to a new error type within the bounds `Into<Box<dyn std::error::Error>>`, however, this would involve a fair amount of extra work. For the specific case that inspired this PR I was attempting to quickly explore some new ideas and found the inability to create any sort of hyper Error object to be an unforeseen hurdle.

Issue #1808, opened 2019-05-06
Implement TryFrom for Destination


seanmonstar:
With `TryFrom` now stable, we can provide a `TryFrom<Uri>` for `Destination`. To prevent *requiring* the new Rust compiler, this can include a `#[cfg]` attribute that is set in the `build.rs`.

M3rs:
I'd like to give this a shot. I think I have the implementation working, but I don't know a lot about ```#[cfg]``` in ```built.rs```.

seanmonstar:
Similar to how it's currently checking for Rust 1.30, and if found, printing `cargo:rustc-cfg=error_source`, a new check can be added for Rust 1.34, and if found, print `cargo:rustc-cfg=try_from`. Then add the `#[cfg(try_from)]` attribute to the `impl TryFrom<Uri> for Destination`.

When doing so, it'd be best to only call `version().unwrap()` once, since that's invoking a child process to run `rustc -V`...

PR #1810, opened 2019-05-09
feat(client): Implement TryFrom for Destination

M3rs:
Add TryFrom<Uri> impl for Destination, for compiler version >= 1.34.
Add basic unit tests for TryFrom impl

Ref Issue: Implement TryFrom for Destination #1808

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -257,9 +257,7 @@ impl TryFrom<Uri> for Destination {
     type Error = ::error::Error;
 
     fn try_from(uri: Uri) -> Result<Self, Self::Error> {
-        uri.authority_part().ok_or(::error::Parse::Uri)?;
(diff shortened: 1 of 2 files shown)


src/client/connect/mod.rs
```diff
@@ -251,6 +252,17 @@ impl Destination {
...
 }
 
+#[cfg(try_from)]
+impl TryFrom<Uri> for Destination {
+    type Error = ::error::Error;
+
+    fn try_from(uri: Uri) -> Result<Self, Self::Error> {
+        uri.authority_part().ok_or(::error::Parse::Uri)?;
+        uri.scheme_part().ok_or(::error::Parse::Uri)?;
+        Ok(Destination { uri })
```
seanmonstar: I believe this can just call `Destination::try_from_uri` internally.

PR #1811, opened 2019-05-09
Add impl `http_body::Body` for `hyper::Body`

LucioFranco:
This adds a `http_body::Body` impl for hypers `Body`. This should
allow us to start moving to a more generic body trait based on
`BufStream` and `http-body`.

cc @seanmonstar


LucioFranco:
GitCop 😭 idk what is wrong.

seanmonstar:
Don't worry about GitCop, I always fix it as part of the squash step.

LucioFranco:
@seanmonstar so do you suggest a blanket impl of Body for T: Payload? Otherwise those changes seem fine, will get to them soon.

seanmonstar:
No, I mean instead of having `poll_trailers` have duplicate code, do like this:

```rust
impl http_body::Body for Body {
    fn poll_trailers(&mut self) -> Poll<Option<HeaderMap>, Self::Error> {
        Payload::poll_trailers(self)
    }
    // etc...
}
```

LucioFranco:
@seanmonstar I think I've addressed all your comments. Should be good for another pass.

src/body/body.rs
```diff
@@ -332,6 +322,35 @@ impl Payload for Body {
...
+        <Self as Payload>::is_end_stream(self)
+    }
+
+    fn size_hint(&self) -> SizeHint {
+        let mut hint = SizeHint::default();
+
+        let upper_limit = <Self as Payload>::content_length(self);
+
+        if let Some(upper) = upper_limit {
+            hint.set_upper(upper);
```
seanmonstar: I believe we can set the lower as well, since it's meant to be an exact size in `Payload`.

src/body/body.rs
```diff
@@ -183,54 +183,44 @@ impl Body {
...
                 delayed_eof: None,
                 on_upgrade: OnUpgrade::none(),
-            }))
+            })
+        })
     }
 
     fn poll_eof(&mut self) -> Poll<Option<Chunk>, ::Error> {
         match self.take_delayed_eof() {
-            Some(DelayEof::NotEof(mut delay)) => {
```
seanmonstar: Just checking, this is simply rustfmt?

LucioFranco: Ah damn totally could have been forgot to disable it for this PR oops. Would you like me to revert this?

src/body/body.rs
LucioFranco: so if I understand correctly, since we ahve the total content_length here so the exact size, lower would == upper?

seanmonstar: Yep, `Payload::content_length` is an exact size if `Some(_)`.

LucioFranco:
@seanmonstar I addressed your comment and reverted the rustfmt. Only issue is that h2 introduced breaking changes for 1.26. We can try to pin it to an older version?

seanmonstar:
Just finished publishing a new h2 version that should work on Rust 1.26, so I gave CI a nudge.

seanmonstar:
Woot, green! Thanks!

Issue #1814, opened 2019-05-12
server::Builder stream/connection window size methods work by reference


sfackler:
All of the other setters on the builder type operate by-value, but http2_initial_connection_window_size and http2_initial_stream_window size work by-reference.

seanmonstar:
Ah man, that's an unfortunate mistake. I'll assign this breaking change to be done in 0.13.

seanmonstar:
As an interim, we could add new methods that are by-value, with an underscore prefix... 🤷‍♂️

PR #1817, opened 2019-05-16
feat(server): add initial window builder methods that take self by-val

seanmonstar:
The current `Builder` methods `http2_initial_stream_window_size` and
`http2_initial_connection_window_size` take `&mut self`, while every
other builder method takes `self`. That breaks up the chaining of
options.

This patch adds two methods that configure the same option, but take
`self` instead, and have an underscore suffix (so,
`http2_initial_stream_window_size_`).

cc #1814


seanmonstar:
@sfackler any thoughts on whether the previous methods should get `#[deprecated]` attributes? Or is hiding them from docs enough?

sfackler:
I don't feel super strongly about deprecation. There's nothing *wrong* with using the old methods...

Issue #1818, opened 2019-05-24
HTTP/3 Support


uvd:
## Current status

HTTP/3 support is on the [roadmap](https://hyper.rs/contrib/roadmap/#http3). There is a working [h3 crate](https://github.com/hyperium/h3), and [reqwest](https://docs.rs/reqwest/latest/reqwest/#unstable-features) has unstable HTTP/3 support.

Much work is still needed, and volunteers are welcome.

Besides the issue tracker (which should store durable decisions and their context), there's this [h3 Discord channel](https://discord.gg/q5mVhMD) to coordinate work.

Here's some of the main steps:

- [ ] Harden the `h3` crate itself, such as fixing any straggling interop issues,
  and filling out the spec conformance tags we use for accountability.
- [ ] Proposal for (initially unstable) `hyper::rt::quic` integration, allowing
  people to bring their own QUIC.
- [ ] Write the `hyper::proto::http3` glue that translates hyper's connection
  patterns with the `h3` crate.

shenshouer:
Any plan for this?

quentusrex:
Since I followed the rabbit hole from Tokio and Tonic to here, the latest work appears to be over here currently:
https://github.com/djc/quinn/pull/734

Any best ways to support the latest efforts on http/3 support?

djc:
I have a branch going on here: https://github.com/djc/hyper/tree/quinn-h3. This adds (completely untested) support for HTTP/3 to the hyper server, but doesn't yet add any API to set it up. I have some ideas on how to do that. The work also more or less depends on #2182 and #2188 (which are currently included in that branch).

quentusrex:
@djc Looks like all of the tickets but https://github.com/hyperium/hyper/pull/2182 were merged. Does this mean that https://github.com/djc/hyper/tree/quinn-h3 is getting close?

djc:
I've been discussing it with @seanmonstar, as far as I can tell he's still figuring out how he wants to go about it.

seanmonstar:
We've put together a proposal here: https://github.com/hyperium/h3/pull/1

quentusrex:
With the proposal merged, what is the best way to see current status towards first MVP of the functionality? and with status, ways to support the effort?

djc:
To the best of my knowledge, work is ongoing in the h3 repo. There is a h3 channel in the Tokio Discord which you might join, as well.

[...]

PR #1824, opened 2019-06-10
[WIP] Start integration of H3 through quinn-h3

djc:
More as a discussion venue for now. If you agree with any prefix set of the branch let me know so that they might be merged (just cherry-pick them if you like, or we can go through a separate PR).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ include = [
 bytes = "0.4.4"
 futures = "0.1.21"
 futures-cpupool = { version = "0.1.6", optional = true }
-http = "0.1.15"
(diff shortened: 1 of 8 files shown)


seanmonstar:
cc #1818

hugecheese:
How does quinn-h3 compare to Cloudflare's [quiche](https://github.com/cloudflare/quiche)?

hugecheese:
^thoughts @seanmonstar @djc?

djc:
I have not looked into quiche much. As far as I know, it is more focused on integrating into C environments, and less on futures-based Rust scenarios. Their H3 implementation is probably more mature than the Quinn one for now.

djc:
I've started over on this (and the initial commit already landed separately).

PR #1827, opened 2019-06-14
feat(client): authority accessor on Destination

### LINKED ISSUE #1826: Destination should provide accessor to uri::Authority
It is unwieldy to derive an URI from `Destination`, e.g. for setting the Origin header.
One needs to get the string from `.host()` and destructure the `Option<u16>` retrieved by `.port()`, to construct an `Authority` for `uri::Builder` by parsing a temporary formatted `String` or a `BytesMut`.

Copying the authority part should be as simple as `dst.authority().unwrap().clone()`.

mzabaluev:
Add `authority()` getter method to `client::connect::Destination`.
This facilitates building URIs associated with the destination,
such as the value of the `Origin` header.

Closes #1826

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -96,6 +96,12 @@ impl Destination {
         self.uri.port_u16()
     }
 
+    /// Get the authority part of the URI.
+    #[inline]
+    pub fn authority(&self) -> Option<&uri::Authority> {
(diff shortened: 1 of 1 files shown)


seanmonstar:
As mentioned in #1826, this forces exposing the internal representation of `Destination`.

PR #1829, opened 2019-06-14
fix(http1): force always-ready connections to yield after a few spins

### LINKED ISSUE #1828: Bad tail latency: always-ready connection can starve others
We compare a [hyper hello-world server](https://github.com/denoland/deno/blob/3dff147d0ca1a2cd8d264d20a178d71cb38b1c4e/tools/hyper_hello.rs)'s performance against our own (dynamic language) web server. We've noticed that hyper has very bad tail latency - sometimes as much as 2 seconds.

You can see it here https://deno.land/benchmarks.html#max-latency
Screen Shot 2019-06-14 at 2 08 55 PM

We debugged this a bit, thinking at first it was our setup. But concluded that it was indeed a bug in hyper after making these changes: https://github.com/denoland/deno_third_party/compare/1902aba524c96f56c1dd654f04660a1fc3ebf80d...d8fb68ccfb72ad6a8b306dfcc3573efafa3c340f

cc @piscisaureus @carllerche

seanmonstar:
Adds a `YieldNow` helper, since I believe there may be a couple other places that could make use of this, like the accept loop in `SpawnAll`.

Closes #1828 

cc @ry

### DIFF
--- a/src/proto/h1/dispatch.rs
+++ b/src/proto/h1/dispatch.rs
@@ -7,7 +7,7 @@ use tokio_io::{AsyncRead, AsyncWrite};
 
 use body::{Body, Payload};
 use body::internal::FullDataArg;
-use common::Never;
+use common::{Never, YieldNow};
 use proto::{BodyLength, DecodedLength, Conn, Dispatched, MessageHead, RequestHead, RequestLine, ResponseHead};
 use super::Http1Transaction;
 use service::Service;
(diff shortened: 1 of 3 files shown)


src/proto/h1/dispatch.rs
```diff
@@ -112,21 +139,19 @@ where
...
-        } else {
-            Ok(Async::NotReady)
+        trace!("poll_loop yielding (self = {:p})", self);
+
+        match self.yield_now.poll_yield() {
+            Ok(Async::NotReady) => Ok(Async::NotReady),
+            // maybe with `!` this can be cleaner...
+            // but for now, just doing this to eliminate branches
+            Ok(Async::Ready(never)) |
+            Err(never ) => match never {}
```
olix0r: nit: whitespace `Err(never )`

ry:
Thanks @seanmonstar, that seemed to have fixed the problem!

Screen Shot 2019-06-17 at 1 42 54 PM

(From https://deno.land/benchmarks.html)

PR #1830, opened 2019-06-20
publicize connection pool api

nlevitt:
Can we make this stuff public? It's useful! I'm using it here: https://github.com/nlevitt/monie
Or is there a better way?

If you are amenable to this change, I can take a pass at documenting the "FIXME: document me" stuff.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -99,7 +99,7 @@ use self::pool::{Key as PoolKey, Pool, Poolable, Pooled, Reservation};
 pub mod conn;
 pub mod connect;
 pub(crate) mod dispatch;
-mod pool;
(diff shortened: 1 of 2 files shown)


seanmonstar:
Eek, several of those types and functions are sloppy *because* they aren't public. Also, some expose implementation details I'd rather not be frozen.

That said, I have wanted to make the pool stuff generic, such that it's basically a `MakeService`...

nlevitt:
> That said, I have wanted to make the pool stuff generic, such that it's basically a MakeService...

I like the sound of that, but not sure I'm up for attempting to implement it at the moment. I guess I'll keep doing what I've been doing, for the time being. :)

Issue #1759, opened 2019-02-03
CONNECT tunneling example


guskovd:
After reading https://medium.com/@mlowicki/http-s-proxy-in-golang-in-less-than-100-lines-of-code-6a51c2f2c38c, the idea was born to write a proxy server on rust.

In https://github.com/hyperium/hyper/blob/master/examples/README.md is written that the example https://github.com/hyperium/hyper/blob/master/examples/upgrades.rs can help.

But I do not really understand how this can be done. I think it would be great to write an example of how to redirect CONNECT requests.

schlamar:
Basically you should mix server + client example into one async request.

On a CONNECT server request, you create a client CONNECT request to the remote proxy. If that suceeds (OK in client response), you send OK in the server response. Then you upgrade server request and client response to get both raw connections. Then you pipe everything through in both directions (here is an example for that: https://github.com/tokio-rs/tokio/blob/599955f/tokio/examples/proxy.rs#L68).

This could get messy with futures, though.

seanmonstar:
There's now an `examples/http_proxy.rs` file in master!

PR #1835, opened 2019-06-27
docs(server): http_proxy example

euclidr:
I added a CONNECT tunneling example #1759

### DIFF
--- a/examples/http_proxy.rs
+++ b/examples/http_proxy.rs
@@ -1,27 +1,23 @@
 #![deny(warnings)]
-extern crate futures;
-extern crate hyper;
-extern crate tokio;
-extern crate tokio_tcp;
 
 use std::net::ToSocketAddrs;
 
-use futures::future;
+use futures_util::future;
+use futures_util::try_future::try_join;
 
-use hyper::rt::{self, Future};
-use hyper::service::service_fn;
+use hyper::service::{make_service_fn, service_fn};
 use hyper::upgrade::Upgraded;
-use hyper::{Body, Client, Method, Request, Response, Server};
+use hyper::{Body, Client, Error, Method, Request, Response, Server};
+use hyper::server::conn::AddrStream;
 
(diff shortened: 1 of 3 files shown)


examples/http_proxy.rs
```diff
@@ -0,0 +1,119 @@
...
+            if Method::CONNECT == req.method() {
+                // Recieved an HTTP request like:
+                // ```
+                // CONNECT www.domain.com:443 HTTP/1.1
+                // Host: www.domain.com:443
+                // Proxy-Connection: Keep-Alive
+                // ```
+                //
+                // When HTTP method is CONNECT we should return an empty body
+                // then we can eventually upgrade the connection and talk a new protocal.
```
ahmedcharles:
```suggestion
                // then we can eventually upgrade the connection and talk a new protocol.
```

euclidr:
@ahmedcharles there have been a lot of change since this example is written. `split` method of `AsyncRead` `AsyncWrite` in tokio-io is temporary removed, which make it hard to implement tunnel. 

I found tokio has plan to get `split` back, [see the issue](https://github.com/tokio-rs/tokio/issues/1256). I will fix this example once the `split` is back. Or are there better ways to implement tunnel?

#1884

ahmedcharles:
See: https://github.com/ahmedcharles/hyper/commit/cc37f87d6ebfd5a3db618852fc1788b2c88db515

It's hacky but it works. (Though, h2 doesn't build at the moment, so when I rebased, it failed to build.)

euclidr:
@ahmedcharles thanks, it works! I changed your code a little, `use tokio_tcp` to `use tokio_net::tcp`, the example runs correctly.

ahmedcharles:
No problem.

seanmonstar:
Thanks for doing this, and sorry for taking so long, I've rebased and merged this into master manually!

PR #1846, opened 2019-07-09
chore(lib): prepare master for 0.13.x changes

seanmonstar:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "hyper"
-version = "0.12.32" # don't forget to update html_root_url
+version = "0.13.0-a.0" # don't forget to update html_root_url
```
LucioFranco: do you prefer `a` over `alpha`?

seanmonstar:
Note that there is now a [0.12.x](https://github.com/hyperium/hyper/tree/0.12.x) branch that starts as last 0.12 release.

Cargo.toml
seanmonstar: I dunno, I just put it there so that version is considered different, I didn't expect to release it yet... :shrug:

PR #1852, opened 2019-07-10
Update the example hello

liufuyang:
Something like this?

There is a call on `make_service_fn` and `service_fn`. Is it possible to make it even simpler?

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -11,7 +11,7 @@ async fn hello(_: Request<Body>) -> Result<Response<Body>, hyper::Error> {
 }
 
 #[tokio::main]
-pub async fn main() {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thank you!

liufuyang:
@seanmonstar Thank you. BTW, there is no way to remove those `make_service_fn` and `service_fn`? 
I am no expert but it feels they seem like some boilerplates can be removed somehow? 🤔

seanmonstar:
Maybe eventually! It's that the server needs some `MakeService` type that will create a `Service` for each connection. It looks like boilerplate in such a tiny example, but so far in my experience, an application puts a lot more functionality into both pieces.

PR #1855, opened 2019-07-10
Update example echo - ready to merge

liufuyang:
I tried to update this as well but not sure how to deal with that `req.into_body().map(....`

Perhaps you have some example to how me how to update this example?

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -1,80 +1,80 @@
+#![feature(async_await)]
 #![deny(warnings)]
-extern crate futures;
+
 extern crate hyper;
 
-use futures::future;
-use hyper::rt::{Future, Stream};
-use hyper::service::service_fn;
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
-
-/// We need to return different futures depending on the route matched,
-/// and we can do that with an enum, such as `futures::Either`, or with
-/// trait objects.
-///
-/// A boxed Future (trait object) is used as it is easier to understand
-/// and extend with more types. Advanced users could switch to `Either`.
-type BoxFut = Box<dyn Future<Item = Response<Body>, Error = hyper::Error> + Send>;
+use hyper::service::{make_service_fn, service_fn};
+use futures_util::TryStreamExt;
 
 /// This is our service handler. It receives a Request, routes on its
 /// path, and returns a Future of a Response.
-fn echo(req: Request<Body>) -> BoxFut {
-    let mut response = Response::new(Body::empty());
+async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
 
     match (req.method(), req.uri().path()) {
         // Serve some instructions at /
         (&Method::GET, "/") => {
-            *response.body_mut() = Body::from("Try POSTing data to /echo");
+            Ok(Response::new(Body::from("Try POSTing data to /echo such as: `curl localhost:3000/echo -XPOST -d 'hello world'`")))
         }
 
         // Simply echo the body back to the client.
         (&Method::POST, "/echo") => {
-            *response.body_mut() = req.into_body();
+            Ok(Response::new(req.into_body()))
         }
 
-        // Convert to uppercase before sending back to client.
+        // Convert to uppercase before sending back to client using a stream.
         (&Method::POST, "/echo/uppercase") => {
-            let mapping = req.into_body().map(|chunk| {
+            let mapping = req.into_body().map_ok(|chunk| {
                 chunk
                     .iter()
                     .map(|byte| byte.to_ascii_uppercase())
                     .collect::<Vec<u8>>()
             });
-
-            *response.body_mut() = Body::wrap_stream(mapping);
+            Ok(Response::new(Body::wrap_stream(mapping)))
         }
 
         // Reverse the entire body before sending back to the client.
         //
         // Since we don't know the end yet, we can't simply stream
-        // the chunks as they arrive. So, this returns a different
-        // future, waiting on concatenating the full body, so that
-        // it can be reversed. Only then can we return a `Response`.
+        // the chunks as they arrive as we did with the above uppercase endpoint.
+        // So here we do `.await` on the future, waiting on concatenating the full body,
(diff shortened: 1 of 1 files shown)


examples/echo.rs
```diff
@@ -34,13 +28,15 @@ fn echo(req: Request<Body>) -> BoxFut {
         // Convert to uppercase before sending back to client.
         (&Method::POST, "/echo/uppercase") => {
             let mapping = req.into_body().map(|chunk| {
-                chunk
+                let chunk= chunk.unwrap();
```
seanmonstar: Oh yea, the old `map` is now `map_ok`.

examples/echo.rs
```diff
@@ -34,13 +28,15 @@ fn echo(req: Request<Body>) -> BoxFut {
...
+                let x = chunk
                     .iter()
                     .map(|byte| byte.to_ascii_uppercase())
-                    .collect::<Vec<u8>>()
-            });
+                    .collect::<Vec<_>>();
+                x
+            }).collect::<Vec<_>>().await;
```
seanmonstar: Hm, this `await` changes the behavior a little. Before, a response would be sent immediately, and the body would be streamed as it was received. By waiting on the body to complete here, a response is sent much later. We should probably keep to the `wrap_stream` approach.

liufuyang: Yeah, wondering how to do that, it seems that `wrap_stream` is not available anymore, it is commented out right now. Do you have any tips on what to do about it?

seanmonstar: Oh, woops! If you grab master again, I've just re-enabled `wrap_stream`!

examples/echo.rs
```diff
@@ -56,7 +48,7 @@ fn echo(req: Request<Body>) -> BoxFut {
                 response
             });
 
-            return Box::new(reversed);
+            return Ok(reversed);
```
liufuyang: how to deal with `concat2` above, it also seems to be a function that is missing

`no method named `concat2` found for type `hyper::Body` in the current scope`

examples/echo.rs
liufuyang: Using `wrap_stream` seems working now. Just this `concat2` issue below now...

examples/echo.rs
seanmonstar: It looks like that was renamed to `TryStreamExt::try_concat`.

liufuyang: hmmm.. I tried to go with it, but not really sure how :P 
Got some errors like this with my last commit c04d92f
```
error[E0277]: `(dyn hyper::rt::Stream<Item = std::result::Result<hyper::Chunk, std::boxed::Box<(dyn std::error::Error + std::marker::Send + std::marker::Sync + 'static)>>> + std::marker::Send + 'static)` cannot be shared between threads safely
  --> examples/echo.rs:78:10
   |
78 |         .serve(service);
   |          ^^^^^ `(dyn hyper::rt::Stream<Item = std::result::Result<hyper::Chunk, std::boxed::Box<(dyn std::error::Error + std::marker::Send + std::marker::Sync + 'static)>>> + std::marker::Send + 'static)` cannot be shared between threads safely
   |
```

seanmonstar: Okay, I checked out the branch, and finally understand the giant error messages. They're very confusing, but ultimately, since there was an `await` inside the `match req.uri()`, the `await` was forcing one of the state machine steps to own a borrow of `&Request`. Because of that, it was noticing the `&Request` is not `Sync`, thanks to the wrapped stream variant inside `Body`.

This will be very easy for users to bump into, and so probably I'll need to fix `Body` to require `Sync`...

seanmonstar: For reference, here's the issue in the rust-lang repo: https://github.com/rust-lang/rust/issues/57017

seanmonstar: Ok, I've merged to master a fix for `Body` that should remove that particular issue.

liufuyang: Well that sounds complicated. Not sure how much I could really understand but it's nice that you helping to solve it out :)

I removed onto master, now I seem to have this issue:
```
error[E0599]: no method named `into_future` found for type `tokio_sync::mpsc::bounded::Receiver<_>` in the current scope
   --> src/client/connect/dns.rs:134:14
    |
134 |             .into_future()
    |              ^^^^^^^^^^^
    |
    = note: the method `into_future` exists but the following trait bounds were not satisfied:
            `&mut tokio_sync::mpsc::bounded::Receiver<_> : futures_util::stream::StreamExt`
            `&tokio_sync::mpsc::bounded::Receiver<_> : futures_util::stream::StreamExt`
            `tokio_sync::mpsc::bounded::Receiver<_> : futures_util::stream::StreamExt`

```

seanmonstar: You may need to run `cargo update` to get the latest changes to tokio.

liufuyang: Hmmm... did that then I got this 😄 
```
error[E0277]: `dyn std::error::Error` cannot be sent between threads safely
  --> examples/echo.rs:67:1
   |
67 | #[tokio::main]
   | ^^^^^^^^^^^^^^ `dyn std::error::Error` cannot be sent between threads safely
   |
   = help: the trait `std::marker::Send` is not implemented for `dyn std::error::Error`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<dyn std::error::Error>`
   = note: required because it appears within the type `std::boxed::Box<dyn std::error::Error>`
   = note: required because it appears within the type `std::result::Result<(), std::boxed::Box<dyn std::error::Error>>`
```

liufuyang: Oh cool, looks like I just need to add
`async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> ` on the main. Nice.

liufuyang: looks like it is working now, let me tide up the PR a bit. Thanks for the help 😄

liufuyang:
@seanmonstar  Hey Sean, I have made this example work and cleanup the code a bit. Would you like take a look at it again? :)

PR #1868, opened 2019-07-13
Expose HttpConnecting future which is used by Connect impl

DoumanAsh:
I'd like to have access to future since it is anyway exposed through associated type of `Connect` implementation.
Otherwise it makes it hard to avoid boxing


DoumanAsh:
@seanmonstar ping? I wouldn't mind this future to be public unless you're planning to remove it :)

seanmonstar:
My hope is to eventually have this be an `impl Future`, once that stabilizes. I'd *love* to be able to rewrite a bunch of this `ConnecingTcp` stuff with `async`.

DoumanAsh:
You think it is going to be soon? 
Well I wouldn't mind to wait for it, so close it if you feel like that

PR #1860, opened 2019-07-12
docs(examples): Update web_api example to use async await

messense:
Ref #1849

Issue #1849, opened 2019-07-09
Update the examples for 0.13


seanmonstar:
Most of the `examples/*.rs` files still need to be updated to work with `std::future::Future`. Most can be more easily written using `#[tokio::main]`, like in [this example](https://github.com/tokio-rs/tokio/blob/master/tokio/examples/hello_world.rs).

liufuyang:
Can I try update a simple example? Looks like you already got the hello example updated, right?
https://github.com/hyperium/hyper/blob/master/examples/hello.rs

And I should try further update it to use `#[tokio::main]` ? :)

liufuyang:
I made a simple change like this: https://github.com/hyperium/hyper/pull/1852
let me know whether that is the right thing to do :)

PR #1872, opened 2019-07-16
Update all examples

weihanglo:
## What

Ref #1849 

- Update send_file example
- Fix lifetime issue of web_api example (PR #1860, thanks @messense )

### send_file

- [ ] Run `cargo run --example send_file`
- [ ] Open either `127.0.0.1:1337`. It must return a HTML identical to `examples/send_file_index.html`
- [ ] Click `index.html`.  then get a response of the the same index page.
- [ ] Click `big_file.html` then get a response of the the same index page.
- [ ] Click `no_file.html` then get a response of 404 not found.

### web_api

- [ ] Run `cargo run --example web_api`
- [ ] Run `curl 127.0.0.1:1337/` then get `<a href="test.html">test.html</a>`
- [ ] Run `curl 127.0.0.1:1337/test.html` then get `<b>POST request body</b>: {"original": "data"}<br><b>Response</b>: {"original":"data","test":"test_value"}`
- [ ] Run `curl 127.0.0.1:1337/json_api` then get `["foo","bar"]`
- [ ] Run `curl 127.0.0.1:1337/json_api -X POST -d '{"bar":"buz"}'` then get `{"bar":"buz","test":"test_value"}`


seanmonstar:
Do you know if the examples are being included again in the `--all-targets`, or do we need to uncomment the lines in the `Cargo.toml`?

weihanglo:
@seanmonstar  
I totally forgot to uncomment them.... They should own their build configurations. I will send a follow-up pull request.

PR #1879, opened 2019-07-20
Remove re-export of `tokio::main`

### LINKED ISSUE #1878: Remove re-export of `tokio::main`
You cannot use the `hyper::rt::main` attribute macro as it is without depending on `tokio` crate, so the re-export is not quite useful as a shorthand. I fear that it could be rather a source of confusion.

tesaguri:
Closes #1878.


seanmonstar:
Thank you!

PR #1882, opened 2019-07-25
Upgrade some lib tests to async/.await version

weihanglo:
## What

- Upgrade `client/connect/http.rs` to async/.await version
- Upgrade `client/pool.rs` to async/.await version
- Upgrade `client/dispatch.rs` to async/.await version

## How to test

- [ ] Run `cargo test --lib -- client::connect::http` and all tests should be passed.
- [ ] Run `cargo test --lib -- client::pool` and all tests should be passed.
- [ ] Run `cargo test --lib -- client::dispatch` and all tests should be passed.
- [ ] Run `cargo bench --lib client::dispatch --features nightly` and all benchmarks are executed properly.

## Benchmark Results

Running `cargo bench --lib client::dispatch --features nightly`

This pull request (e9c881c)

```bash
test client::dispatch::tests::giver_queue_cancel                ... bench:           8 ns/iter (+/- 0)
test client::dispatch::tests::giver_queue_not_ready             ... bench:         142 ns/iter (+/- 13)
test client::dispatch::tests::giver_queue_throughput            ... bench:         740 ns/iter (+/- 76)
```
---

v0.12.33

```bash
test client::dispatch::tests::giver_queue_cancel                ... bench:           9 ns/iter (+/- 2)
test client::dispatch::tests::giver_queue_not_ready             ... bench:         107 ns/iter (+/- 7)
test client::dispatch::tests::giver_queue_throughput            ... bench:         741 ns/iter (+/- 262)
```

Need to figure out why `giver_queue_not_ready` benchmark **got a performance reduction**.

### `PollOnce` utiltiy

Some tests call `Future::poll` directly to check the poll result once, and it is legacy `futures.rs` method and is not allowed to call on `std::future`. Hence, I made a [`PollOnce`](https://github.com/weihanglo/hyper/blob/e857361610a543d7a479aed9ee4c77b869a248bf/src/client/pool.rs#L847-L861) struct to simulate the behavior. I am not sure whether it is suitable or not. Any advice for improvement is welcome.

### [Solved] _test_pool_checkout_drop_cleans_up_waiters_ failed

<details>
   <summary>expand to reveal details </summary>

> [Solved] see a1097fb 

Continuing previous discussion, I encountered a problem when using `PollOnce` along with [_test_pool_checkout_drop_cleans_up_waiters_](https://github.com/weihanglo/hyper/blob/e857361610a543d7a479aed9ee4c77b869a248bf/src/client/pool.rs#L978-L1005) test case. The future `F` would move into `PollOnce<F>` then drop after awaiting `F` yield the result. Therefore, waiters count is always unchanged after awaiting futures.

```rust
rt.block_on(async move {
    poll_once1.await; // waiters count +1
    // checkout1 dropped. waiters count -1
    assert_eq!(pool.locked().waiters.get(&key).unwrap().len(), 1);
    poll_once2.await;  // waiters count +1
    // checkout2 dropped. waiters count -1
    assert_eq!(pool.locked().waiters.get(&key).unwrap().len(), 2);
});
```

Additionally, `drop` neither `poll_once` nor `checkout` can be dropped due to moving after awaiting futures. How should we fix this issue?

</details>

### `#[cfg(feature = "runtime")]`

I saw that some tests was labeled as #[cfg(feature = "runtime")]. Should all tests that need a runtime have to add this config?


seanmonstar:
Is it possible to just pass a reference to `PollOnce`, so it doesn't own and thus drop the original future? Like `PollOnce(&mut checkout1).await`?

For the runtime config, I don't yet know the answer. In 0.12, it was easy to just `wait` in several tests, and so they didn't need a runtime.

weihanglo:
> Is it possible to just pass a reference to `PollOnce`, so it doesn't own and thus drop the original future? Like `PollOnce(&mut checkout1).await`?

Done and test passed.

> For the runtime config, I don't yet know the answer. In 0.12, it was easy to just `wait` in several tests, and so they didn't need a runtime.

It seems that [`future::wait`](https://github.com/rust-lang-nursery/futures-rs/blob/0.1/src/future/mod.rs#L274-L300) also uses `futures::executor` as a "runtime" based on futures crate. I wonder what would happen if user run tests without runtime feature.

weihanglo:
Can we add tokio itself or [tokio-current-thread](https://crates.io/crates/tokio-current-thread) as **dev-dependencies**, so that feature flags would not affect any test needing a runtime?

src/client/connect/http.rs
```diff
@@ -571,60 +571,76 @@ impl ConnectingTcp {
...
+                connector.connect(dst).await.unwrap_err().kind(),
+                io::ErrorKind::InvalidInput,
+            );
+        })
     }
 
 
     #[test]
     fn test_errors_missing_scheme() {
+        let mut rt = Runtime::new().unwrap();
```
95th: `#[tokio::test]` will be a cleaner in async tests

seanmonstar:
Sorry for taking so long, I was on vacation for a few weeks. This is a great step forward! Looks like a rebase may be needed.

weihanglo:
@seanmonstar 
Updated. Thanks.

weihanglo:
I've update this PR several times. Would this be merged?

PR #1883, opened 2019-07-27
test(server): Update and re-enable the `tests/server.rs` test

jakubadamw:
These tests were temporarily disabled during the migration to the
`std::future::Future` type that's part of the stable Rust now.

These are updates to the tests that make them pass again.


seanmonstar:
I've rebased this manually into master, thanks so much!

PR #1889, opened 2019-08-05
fix(client): impl<T: Connect> Connect for Box<T>

khuey:
Self-explanatory

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -35,6 +35,15 @@ pub trait Connect: Send + Sync {
     fn connect(&self, dst: Destination) -> Self::Future;
 }
 
+impl<T: Connect + ?Sized> Connect for Box<T> {
+    type Transport = <T as Connect>::Transport;
+    type Error = <T as Connect>::Error;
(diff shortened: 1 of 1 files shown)


ahmedcharles:
Depends on #1890.

src/client/connect/mod.rs
```diff
@@ -35,6 +35,15 @@ pub trait Connect: Send + Sync {
     fn connect(&self, dst: Destination) -> Self::Future;
 }
 
+impl<T: Connect + ?Sized> Connect for Box<T> {
+    type Transport = <T as Connect>::Transport;
+    type Error = <T as Connect>::Error;
+    type Future = <T as Connect>::Future;
+    fn connect(&self, dst: Destination) -> Self::Future {
+        <T as Connect>::connect(self, dst)
```
seanmonstar: Does this recurse infinitely? I would have thought it would need to be `&**self`, but I could be wrong.

khuey: I'm explicitly invoking `<T as Connect>::connect` so it does not recurse (and `self` will deref coerce to the right value as an argument).

PR #1890, opened 2019-08-06
fix(server): worked-around deprecation of `poll_accept` method in tokio

yotamofek:
This fixes building with the tokio master branch head.
Refactoring will be required to actually use the new `accept` async fn.
Build was broken by: https://github.com/tokio-rs/tokio/commit/6d8cc4e4755abbd0baac9abf154837b9be011a07

### DIFF
--- a/examples/send_file.rs
+++ b/examples/send_file.rs
@@ -2,7 +2,7 @@
 #![deny(warnings)]
 
 use tokio::io::AsyncReadExt;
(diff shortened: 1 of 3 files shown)


yotamofek:
No worries,
glad I could help - It's exciting because it's my first contribution to the Rust ecosystem :)
Do you think there should be any performance implications to this change?

PR #1894, opened 2019-08-09
Fixes #1893 -- update to use the alpha release of tokio

### LINKED ISSUE #1893: Update to use tokio alpha from crates.io instead of from git

alex:


ahmedcharles:
This seems to rely on #1890

PR #1896, opened 2019-08-10
Fix failing build due to tokio changes

95th:


Byron:
@95th I am currently working on my own (async-await test program)[https://github.com/Byron/github-star-counter], and this PR worked perfectly for me. Is there a better way to get hyper to work now? Thank you!

Blub:
I'd expect a bigger change there, not sure. Tokio switched from a poll function to an async-fn, so this patch now creates a future and discards it if it's not ready, which is technically a weird thing to do.
So do several other active PRs currently as far as I can tell. While it technically may work for many of these cases, it feels... wrong IMO.

95th:
@Blub Agree. and it would have perf implications as well in the real world. I asked Carl Lerche about it. We need to make a bigger change so that we can directly use async fns. I made an unsuccessful attempt at it.

@Byron I closed this PR because there were others like it.

PR #1908, opened 2019-08-20
Update `Payload` to be a trait alias of `http_body::Body`

LucioFranco:
This changes Payload to be a trait alias of http_body::Body and seals Payload so that the only way to implement it is via http_body::Body. `content_length` was also removed in favor of the more flexible size hint api in http body.

If we're all happy with this I can cut a release of http body and push the cargo.toml change.

cc @seanmonstar @carllerche 

Signed-off-by: Lucio Franco <luciofranco14@gmail.com>

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -329,21 +329,18 @@ impl HttpBody for Body {
         match self.kind {
             Kind::Once(Some(ref val)) => {
                 let mut hint = SizeHint::default();
-                hint.set_lower(val.len() as u64);
-                hint.set_upper(val.len() as u64);
+                hint.set_exact(val.len() as u64);
                 hint
             },
             Kind::Once(None) => {
-                let mut hint = SizeHint::default();
-                hint.set_upper(0);
-                hint
+                SizeHint::default()
             },
             Kind::Wrapped(..) => SizeHint::default(),
             Kind::Chan { content_length, .. } | Kind::H2 { content_length, .. } => {
                 let mut hint = SizeHint::default();
 
                 if let Some(content_length) = content_length {
-                    hint.set_upper(content_length as u64);
+                    hint.set_exact(content_length as u64);
                 }
 
                 hint
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -6,7 +6,7 @@ use bytes::Bytes;
 use futures_core::{Stream, TryStream};
 use futures_channel::{mpsc, oneshot};
 use futures_util::TryStreamExt;
-//use tokio_buf::SizeHint;
+use http_body::{SizeHint, Body as HttpBody};
 use h2;
 use http::HeaderMap;
 
@@ -298,7 +298,7 @@ impl Default for Body {
     }
 }
 
-impl Payload for Body {
+impl HttpBody for Body {
     type Data = Chunk;
     type Error = crate::Error;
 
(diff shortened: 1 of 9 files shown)


src/proto/h1/dispatch.rs
```diff
@@ -268,19 +268,22 @@ where
...
-                    //
-                    // If so, we can skip a bit of bookkeeping that streaming
-                    // bodies need to do.
-                    if let Some(full) = body.__hyper_full_data(FullDataArg(())).0 {
-                        self.conn.write_full_msg(head, full);
-                        return Poll::Ready(Ok(()));
+
+                    let hint = body.size_hint();
+                    if hint.upper() == Some(hint.lower()) {
+                        if let (Some(mut body), _clear_body) = OptGuard::new(self.body_rx.as_mut()).guard_mut() {
```
LucioFranco: I believe this should replicate what `__hyper_full_data` did previously.

src/body/body.rs
```diff
@@ -298,22 +298,22 @@ impl Default for Body {
...
+impl HttpBody for Body {
     type Data = Chunk;
     type Error = crate::Error;
 
     fn poll_data(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Option<Result<Self::Data, Self::Error>>> {
         self.poll_eof(cx)
     }
 
-    fn poll_trailers(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Option<Result<HeaderMap, Self::Error>>> {
+    fn poll_trailers(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Result<Option<HeaderMap>, Self::Error>> {
```
seanmonstar: Was there some discussion that `poll_trailers` should return `Result<Option>` instead of `Option<Result>` because its not stream-like?

LucioFranco: Yeah @carllerche and I discussed this. And yes since it’s not a stream the api is different to indicate that.

src/proto/h2/client.rs
```diff
@@ -128,7 +128,7 @@ where
                     let (head, body) = req.into_parts();
                     let mut req = ::http::Request::from_parts(head, ());
                     super::strip_connection_headers(req.headers_mut(), true);
-                    if let Some(len) = body.content_length() {
+                    if let Some(len) = body.size_hint().upper() {
```
seanmonstar: If the lower hint isn't the same, then we can't know the actual content-length, right?

src/body/body.rs
```diff
@@ -326,21 +326,29 @@ impl Payload for Body {
...
+                let mut hint = SizeHint::default();
 
-    // We can improve the performance of `Body` when we know it is a Once kind.
-    #[doc(hidden)]
-    fn __hyper_full_data(&mut self, _: FullDataArg) -> FullDataRet<Self::Data> {
-        match self.kind {
-            Kind::Once(ref mut val) => FullDataRet(val.take()),
-            _ => FullDataRet(None),
+                if let Some(content_length) = content_length {
+                    hint.set_upper(content_length as u64);
```
seanmonstar: Does this also need to `set_lower`? Otherwise the range is `0..content_length`, right?

seanmonstar:
Looks good. I notice using the `SizeHint` is a little trickier... Would a `set_exact` and `exact() -> Option` which just handles setting/getting the lower and upper for us be helpful?

LucioFranco:
@seanmonstar yeah, I was thinking the same, will update http body then this PR.

src/body/body.rs
LucioFranco: I think this to me makes the assumption that we have an upper bound but since the lower bound is not the same, poll_data will not return `0..upper`, whereas if `lower == upper` that means poll_data already has up too lower and upper and will return the full chunk.

carllerche: The intent of size_hint is to represent the total amount of data in the stream. If the total size of the stream is known, then lower should == upper, even if all the data is not received.

I have considered adding a field to indicate the amount of data that is ready to be consumed, but that has not been done yet.

seanmonstar: Exactly, so `SizeHint` means *total* data, not just next chunk. The `content-length` header is *total*.

seanmonstar:
With https://github.com/hyperium/http-body/pull/14 merged, this can just use `set_exact` and `exact` instead.

src/proto/h1/dispatch.rs
seanmonstar: It's actually a little different, so I suggest just removing that optimization for now, as I have a simple idea to get it back.

src/proto/h2/client.rs
LucioFranco: done

LucioFranco:
@seanmonstar ok I applied the changes, and I think this is ready to go.

PR #1912, opened 2019-08-21
feat(client): change `Connect` trait into an alias for `Service`

### LINKED ISSUE #1902: Change `Connect` trait to alias `Service`
After #1782 is done, the `Connect` trait should be changed to basically an alias for a `Service<Destination>`. This would allow more composability, being able to use `ServiceExt` to customize connectors.

seanmonstar:
The `Connect` trait is now essentially an alias for
`Service<Destination>`, with a blanket implementation as such, and is
sealed.

Closes #1902

BREAKING CHANGE: Any manual implementations of `Connect` must instead
  implement `tower::Service<Destination>`.

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -3,6 +3,7 @@ use std::error::Error as StdError;
 use std::io;
 use std::mem;
 use std::net::{IpAddr, SocketAddr};
+use std::sync::Arc;
 use std::time::Duration;
 
 use http::uri::{Scheme, Uri};
@@ -13,7 +14,7 @@ use tokio_net::tcp::TcpStream;
 use tokio_timer::{Delay, Timeout};
 
 use crate::common::{Future, Pin, Poll, task};
-use super::{Connect, Connected, Destination};
+use super::{Connected, Destination};
 use super::dns::{self, GaiResolver, Resolve};
 #[cfg(feature = "runtime")] use super::dns::TokioThreadpoolGaiResolver;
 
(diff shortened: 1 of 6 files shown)


DoumanAsh:
Can you then re-export `tower::Service` in hyper then?

seanmonstar:
> Can you then re-export `tower::Service` in hyper?

Yes, I think it should be. On master right now, there is `hyper::service::Service`, which is basically an alias for `tower::Service<http::Request<B>>` (so sealed and a blanket impl). Perhaps that should be renamed (`hyper::service::alias::Service` or `hyper::service::HttpService` or something).

DoumanAsh:
Yeah, my library implements own connector now so ideally I'd prefer to depend on hyper traits, without need to keep tower-service up to date.
Also would be nice to have aliases for original traits like `Connect` that user need to implement rather than sealed so that we could just implement straight ahead it

src/client/connect/http.rs
```diff
@@ -216,51 +235,59 @@ impl<R: fmt::Debug> fmt::Debug for HttpConnector<R> {
     }
 }
 
-impl<R> Connect for HttpConnector<R>
+impl<R> tower_service::Service<Destination> for HttpConnector<R>
```
LucioFranco: we should probably support `TryFrom<Uri> for Destination` as well.

src/client/connect/mod.rs
```diff
@@ -24,26 +24,49 @@ use crate::common::{Future, Unpin};
...
     /// The connected IO Stream.
-    type Transport: AsyncRead + AsyncWrite + Unpin + Send + 'static;
+    type Transport: AsyncRead + AsyncWrite;
     /// An error occured when trying to connect.
     type Error: Into<Box<dyn StdError + Send + Sync>>;
     /// A Future that will resolve to the connected Transport.
-    type Future: Future<Output=Result<(Self::Transport, Connected), Self::Error>> + Unpin + Send;
-    /// Connect to a destination.
-    fn connect(&self, dst: Destination) -> Self::Future;
+    type Future: Future<Output=Result<(Self::Transport, Connected), Self::Error>>;
```
LucioFranco: is this a good time to move to http connection?

seanmonstar: Yea, I started working on that, but ran into some questions, so figured I'd just update this much and continue *that* in a new PR.

LucioFranco: sounds like a good plan 👍

src/client/connect/http.rs
LucioFranco: Yeah, I think that isn't a horrible idea...I think we'd want to error out early though if they don't provide a scheme and authority?

PR #1916, opened 2019-08-29
refactor(dns) migrate deprecated trim_{left,right} -> trim_{start,end}

ComputerDruid:
Since master is using the latest futures, I'm pretty sure we're safe to assume rustc > 1.30

### DIFF
--- a/src/client/connect/dns.rs
+++ b/src/client/connect/dns.rs
@@ -248,16 +248,7 @@ impl IpAddrs {
             let addr = SocketAddrV4::new(addr, port);
             return Some(IpAddrs { iter: vec![SocketAddr::V4(addr)].into_iter() })
         }
-        let host = {
-            // trim_left/trim_right deprecated...
-            // TODO: use trim_start/trim_end in Rust 1.30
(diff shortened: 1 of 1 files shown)


ComputerDruid:
The CI failure looks like it's from pulling tokio-net v0.2.0-alpha.3, not sure why it's doing that rather than following Cargo.lock though...

seanmonstar:
Thanks for the PR!

I've fixed master to pin to specific alphas, so hopefully breakage should be less.

ComputerDruid:
So, uh, gitcop seems like a good idea, but I can't figure out what it's complaining about here. When I click on details it just takes me to the homepage, which gives an example rule of subject < 72 chars. Is it supposed to be putting a comment here or something? I also can't seem to find a config file for where its rules are configured.

Sorry if I'm missing something obvious here.

seanmonstar:
Don't worry about it, I just use gitcop to remind me to adjust the message when squash-merging (if still curious, it was just missing a colon after `refactor(dns)`.

PR #1917, opened 2019-08-29
feat(client): change `GaiResolver` to use a global blocking threadpool

seanmonstar:
BREAKING CHANGE: Calls to `GaiResolver::new` and `HttpConnector::new` no
  longer should pass an integer argument for the number of threads.


sfackler:
Does the resolver interface need to exist at all anymore once `TcpStream::connect` does the resolution on its end?

seanmonstar:
Maybe `TcpStream::connect` gets smart enough, but for now, the connector stuff in hyper does more, like supporting happy eyeballs. Plus, with `Resolve` changed to a `Service`, one could easily use combinators to add in address blacklists/whitelists and things.

PR #1923, opened 2019-09-03
Remove `FutureExt::boxed` to unify project style

AtkinsChang:
introduced by #1890 while `Pin::new_unchecked` is used across this project which is more efficient

### DIFF
--- a/src/common/drain.rs
+++ b/src/common/drain.rs
@@ -1,9 +1,9 @@
 use std::mem;
 
+use futures_util::FutureExt as _;
 use tokio_sync::{mpsc, watch};
 
 use super::{Future, Never, Poll, Pin, task};
-use futures_util::FutureExt as _;
 
 // Sentinel value signaling that the watch is still open
 enum Action {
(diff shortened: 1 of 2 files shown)


src/server/tcp.rs
```diff
@@ -106,10 +105,9 @@ impl AddrIncoming {
         }
         self.timeout = None;
 
-        let mut accept_fut = self.listener.accept().boxed();
-
+        let mut accept = self.listener.accept();
         loop {
-            match accept_fut.poll_unpin(cx) {
+            match unsafe { Pin::new_unchecked(&mut accept) }.poll(cx) {
```
taiki-e:
```rust
let accept = self.listener.accept();
futures_util::pin_mut!(accept);

loop {
    match accept.poll_unpin(cx) {
```

src/common/drain.rs
```diff
@@ -100,9 +99,8 @@ where
         loop {
             match mem::replace(&mut me.state, State::Draining) {
                 State::Watch(on_drain) => {
-                    let mut recv_fut = me.watch.rx.recv_ref().boxed();
-
-                    match recv_fut.poll_unpin(cx) {
+                    let mut recv = me.watch.rx.recv_ref();
+                    match unsafe { Pin::new_unchecked(&mut recv) }.poll(cx) {
```
taiki-e:
```rust
let recv = me.watch.rx.recv_ref();
futures_util::pin_mut!(recv);

match recv.poll_unpin(cx) {
```

AtkinsChang:
replaced with `futures_util::pin_mut!`

Issue #1920, opened 2019-08-30
Consider `Stream`s instability


seanmonstar:
The `Stream` trait isn't stable like `Future` is, and hyper uses it in a couple places. Is it worth putting all usage behind some `unstable` feature flag?

- `impl Stream for Body`: this exists for convenience so users can use stream combinators. Easily optional.
- `Body::wrap_stream`: this exists to allow users to easily wrap stream combinators into an `HttpBody`. Easily optional.
- `hyper::Server::builder(impl Stream)`: this allows using a hyper server with a stream of incoming connections besides just TCP. This is pretty fundamental. An alternative could be to define an `Accept` trait that functions similarly. And then an optional `AcceptStream<impl Stream>` can exist unstably.

LucioFranco:
@seanmonstar I think the first two sound good. For the last one, I think the custom trait works well 👍.

seanmonstar:
This has been done.

PR #1928, opened 2019-09-05
feat(server): introduce `Accept` trait

seanmonstar:
The `Accept` trait is used by the server types to asynchronously accept
incoming connections. This replaces the previous usage of `Stream`.

Part of #1920

BREAKING CHANGE: Passing a `Stream` to `Server::builder` or
  `Http::serve_incoming` must be changed to pass an `Accept` instead. The
  `stream` optional feature can be enabled, and the a stream can be
  converted using `hyper::server::accept::from_stream`.

### DIFF
--- a/src/server/accept.rs
+++ b/src/server/accept.rs
@@ -1,0 +1,99 @@
+//! The `Accept` trait and supporting types.
+//!
+//! This module contains:
+//!
+//! - The [`Accept`](Accept) trait used to asynchronously accept incoming
+//!   connections.
+//! - Utilities like `poll_fn` to ease creating a custom `Accept`.
+
+#[cfg(feature = "stream")]
+use futures_core::Stream;
+
+use crate::common::{Pin, task::{self, Poll}};
+
+/// Asynchronously accept incoming connections.
+pub trait Accept {
+    /// The connection type that can be accepted.
+    type Conn;
+    /// The error type that can occur when accepting a connection.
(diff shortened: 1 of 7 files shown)


src/server/accept.rs
```diff
@@ -0,0 +1,99 @@
+//! The `Accept` trait, and supporting types.
```
LucioFranco:
```suggestion
//! The `Accept` trait and supporting types.
```

src/server/accept.rs
```diff
@@ -0,0 +1,99 @@
...
+    impl<F, IO, E> Accept for PollFn<F>
+    where
+        F: FnMut(&mut task::Context<'_>) -> Poll<Option<Result<IO, E>>>,
+    {
+        type Conn = IO;
+        type Error = E;
+        fn poll_accept(self: Pin<&mut Self>, cx: &mut task::Context<'_>)
+            -> Poll<Option<Result<Self::Conn, Self::Error>>>
+        {
+            unsafe {
```
LucioFranco: would it make sense to use pin-project here to remove this usage of unsafe?

seanmonstar: We could, this just seemed easier (and requires less compiler work). 🤷‍♂

LucioFranco: Yeah, this is safe anyways so I'm fine with keeping it.

PR #1932, opened 2019-09-07
Remove hyper::body::Chunk, replace with bytes::Bytes

### LINKED ISSUE #1931: Remove Chunk in favor of using Bytes directly
Compared to `Bytes`, `Chunk` is pointless and kinda frustrating:

* It's a trivial wrapper around `Bytes` but `Bytes` is exposed via the `From`/`Into` impl anyways so it's not like it's keeping it out of the public API
* `Chunk` adds 0 new functionality over `Bytes`, all of its methods forwards to `Bytes` anyway
* There's no way to split a `Chunk` without round-tripping through `Bytes`
* Any third-party crate that wants to build on `Stream<Item = Bytes>` needs special adapters to work with `hyper::Body`

That last one is a real problem for me because I'm finally updating `multipart-async` and trying to maximize interop by generifying the API as much as possible. Because `hyper::Body` doesn't implement `AsyncRead` I've instead standardized around a `Stream<Item = impl BodyChunk>` API where `BodyChunk` covers two operations: splitting and dereffing to `&[u8]`. My final constraint is that I want to be able to return the same `impl BodyChunk` type, mostly to make it easy to drop `multipart-async` into existing request handlers.

Without implementing `BodyChunk` specifically for `hyper::Chunk` while still having an applicable impl for it, I would need an impl that's generic over `T: From<Bytes>, Bytes: From<T>, T: AsRef<[u8]>`. My hangup on this is that I now cannot implement `BodyChunk` for `&[u8]` directly, which complicates testing as I have to convert all bytestring literals to `Bytes`. (I also want to avoid a generic impl based on `Into<Bytes>` as that operation may perform implicit copies and I'm trying to be near-zero-copy.)

I realize `Chunk` exists because it used to be a custom implementation of `Bytes` but at this point it can be changed to just be a type alias and basically nothing will break (except usages of `Chunk::into_bytes()`, ironically).

davidbarsky:
Resolves #1931 

This PR is a take at removing `hyper::body::Chunk` and replacing it with `bytes::Bytes`. I think that this PR should _probably_ rely on bytes 0.5, but that hasn't been published yet. Additionally, there are still mentions of "chunk" in non-public facing code; that should probably be scrubbed.

cc: @abonander


weihanglo:
Some lines in `h1::io` benchmark need to remove `crate::Chunk`.

https://github.com/hyperium/hyper/blob/db1ba078807b0f85dc75bb0c309a0f62bf2cfe7f/src/proto/h1/io.rs#L926-L929

lnicola:
Nb. `bytes 0.5` is now published.

davidbarsky:
@lnicola Thanks for the heads up!

seanmonstar:
We could move forward with this, thought there's some merge conflicts. I have a thing or two left that are distracting me, but I can fix this up after if you don't have time.

seanmonstar:
Fixed up here: https://github.com/hyperium/hyper/pull/2048

davidbarsky:
@seanmonstar Thanks for doing that! Sorry about that delay; I was pulled into an operation issue.

PR #1937, opened 2019-09-11
feat(client): allow connecting to IPv6 link-local addresses

ComputerDruid:


ComputerDruid:
Note: requires https://github.com/hyperium/http/pull/343 to work (patch in Cargo.toml)

src/client/connect/dns.rs
```diff
@@ -22,7 +22,7 @@ use crate::common::{Future, Never, Pin, Poll, Unpin, task};
 /// Resolve a hostname to a set of IP addresses.
 pub trait Resolve: Unpin {
     /// The set of IP addresses to try to connect to.
-    type Addrs: Iterator<Item=IpAddr>;
+    type Addrs: Iterator<Item=SocketAddr>;
```
erickt: I think this is modifying a public trait, so it'd be really nice if this could be considered for 0.13 inclusion, otherwise we might need to wait for the next minor version to not break users.

src/client/connect/dns.rs
```diff
@@ -245,6 +245,56 @@ impl Iterator for IpAddrs {
...
+}
+
+impl Ipv6AddrWithZone {
+    pub(super) fn try_parse(host: &str, port: u16) -> Option<Self> {
+        // IpV6 literals are always in []
+        if ! host.starts_with("[") {
+            return None;
+        }
+
+        let host = host.trim_start_matches('[').trim_end_matches(']');
```
erickt: This will ignore `[[fe80::1:2:3:4%25eth0]`, which isn't allowed as far as I can tell. Instead, I'd recommend:

```
if !host.starts_with('[') || !host.ends_with(']') {
    return None;
}
let host = &host[1..host.len()-1];
```

Also, there should be a test case for erring out if the brackets are incorrect.

src/client/connect/dns.rs
```diff
@@ -245,6 +245,56 @@ impl Iterator for IpAddrs {
...
+#[derive(Debug)]
+pub(super) struct Ipv6AddrWithZone {
+    pub(super) partial_addr: SocketAddrV6,
+    pub(super) as_str: String,
+}
+
+impl Ipv6AddrWithZone {
+    pub(super) fn try_parse(host: &str, port: u16) -> Option<Self> {
+        // IpV6 literals are always in []
+        if ! host.starts_with("[") {
```
erickt: nit: no space after the `!`.

src/client/connect/dns.rs
ComputerDruid: Note that this mirrors the IPv6 literal parsing, above.

erickt: true, but I think that'd be a bug there too :)

ComputerDruid:
Addressed the review comments, so this is ready for review again.

Uploaded as a separate commit, I can of course squash/rebase if desired.

seanmonstar:
I'm not up-to-speed on IPv6 zone IDs. It identifies a specific port? Or, why is this changing `Resolve` to return a `SocketAddr` instead of an `IpAddr`?

ComputerDruid:
It returns a `SocketAddr` because it needs `SocketAddrV6.scope_id` to be set correctly. https://doc.rust-lang.org/std/net/struct.SocketAddrV6.html#method.scope_id , which is used to determine which network interface to use to make the connection. Since generally all interfaces have `fe80::/64` link-local addresses, and link-local addresses are non-routable, there needs to be some disambiguation.

The only standard way I found that can turn a zone_id (on Linux, that's the interface name) into a scope_id (on Linux, the interface index) is getaddrinfo.

Some relevant info I found in man pages while I was digging: http://man7.org/linux/man-pages/man7/ipv6.7.html

> sin6_scope_id is an ID depending on the scope of the address.  It is new in Linux 2.4. Linux supports it only for link-local addresses, in that case sin6_scope_id contains the interface index (see netdevice(7))

The interface index can also be found in /proc/net/if_inet6 (where it's the second column). I don't use that in this CL since I'd rather use standard APIs, and getaddrinfo seems to do the trick.

src/client/connect/dns.rs
```diff
@@ -333,4 +384,34 @@ mod tests {
...
+        let uri = ::http::Uri::from_static("http://[::1%25eth0]:8080/");
+        let dst = super::super::Destination { uri };
+
+        let addr = Ipv6AddrWithZone::try_parse(
+            dst.host(),
+            dst.port().expect("port")
+        );
+
+        assert_eq!(addr.map(|a| format!("{:?}", a)), None);
+    }
```
erickt: We'll need a test to make sure clients don't pass the scope to servers.

seanmonstar: Do you mean a test in this PR that hyper remembers to strip it somewhere?

erickt: Yeah, at least in the HOST header (not sure if we need to do this elsewhere). Specifically in the HOST header (but anywhere else we might be sending the URL to a remote server). Otherwise, this could potentially open up some attack surfaces. I think [rfc7404](https://tools.ietf.org/html/rfc7404) speaks more about this.

erickt: Sorry, wrong rfc. The right one is [rfc6874](https://tools.ietf.org/html/rfc6874), which states in section 4:

> An HTTP client, proxy, or other intermediary MUST remove any ZoneID attached to an outgoing URI, as it has only local significance at the sending host.

seanmonstar:
Since this has a breaking change, it'd need to be merged before 0.13 is released (target is early next week). I still don't have my head around it entirely, but it sounds like this extra info cannot exist in an `IpAddr`, only a `SocketAddr`?

a0145:
Is this change still on the table? Without it, it doesn't seem possible to use a link-local address with hyper with more than one link-local nic. 

> I still don't have my head around it entirely, but it sounds like this extra info cannot exist in an `IpAddr`, only a `SocketAddr`?

Per the rust implementation of SocketAddrV6, the scope id is only contained in there.  It's not necessarily IP info but "connection setup" info which describes which interface to use for the connection, which is necessary when multiple link-local interfaces are present.

> I see, thanks for explaining! So, does this mean IPv6 addresses essentially allow encoding HttpConnector::set_local_addresses in a per-destination way?

If there were an equivalent `HttpConnector::set_local_interface(iface_id)`, technically that is what this notation does per destination.  By specifying the zone/interface, it's possible to select which interface to use for the connection which is necessary to disambiguate where you want to connect to.

Ref: https://tools.ietf.org/html/rfc4007#section-11
Ref: https://en.wikipedia.org/wiki/Link-local_address#IPv6

seanmonstar:
> Is this change still on the table?

I'm still open to the change, it was mentioned by contributors they wanted to add tests. Also, at this point, it'd be a breaking change to the "resolver" API, so either a way to provide it without breakage is needed, or this needs to wait till the next major version.

> If there were an equivalent `HttpConnector::set_local_interface(iface_id)`, technically that is what this notation does per destination.

Thanks, I understand now. Sounds useful!

seanmonstar:
So, as part of the breaking changes in 0.14, I'm pulling part of this change in #2346: the resolvers should return `SocketAddr`s instead of `IpAddr`s. I didn't want to pull in percent_decoding to decode a tiny part of the authority, but I believe a custom resolver can handle that if needed.

PR #1958, opened 2019-09-30
feat(client): add per IP address connection timeout

### LINKED ISSUE #1234: Add connect, read, and write timeouts to client

chewi:
It has previously been recommended to apply timeouts externally to hyper. The
hyper-timeout crate uses a special connector to apply read and write timeouts
but the connection timeout is still external.

This approach has one major drawback that only an implementation within hyper
can provide. If the requested hostname resolves to multiple IP addresses, hyper
internally tries to connect to each one in turn. If one of them doesn't respond
at all then it hangs until the OS-level timeout kicks in. This defaults to 127
seconds on Linux due to `tcp_syn_retries` being 6. hyper-timeout wraps around all
the connection attempts rather than each one individually so setting that
timeout to something shorter means that it will give up before trying the
second address. You could reduce `tcp_syn_retries` but this is kernel-wide so
this is far from ideal.

Therefore add a `set_connect_timeout` function to `HttpConnector`. It defaults to
`None`, preserving the earlier behaviour.

Closes #1234

----

I'm thinking a DNS timeout would also be a good idea so I'm working on that too. hyper-timeout's connect timeout covers that part of the process. It's arguably less crucial though. The default DNS timeout with glibc is 5 seconds each for 2 tries, which isn't as bad as 127 seconds, and reducing this system-wide has less of an impact than reducing `tcp_syn_retries`. Please let me know if you want that work in this PR or a new one.

I need this backported to 0.12 and I'm hoping you'll accept a PR for that when it's ready.

I'm still relatively new to Rust so please go easy on me! :baby:

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -10,7 +10,7 @@ use futures_util::{TryFutureExt, FutureExt};
 use net2::TcpBuilder;
 use tokio_net::driver::Handle;
 use tokio_net::tcp::TcpStream;
-use tokio_timer::Delay;
+use tokio_timer::{Delay, Timeout};
 
 use crate::common::{Future, Pin, Poll, task};
 use super::{Connect, Connected, Destination};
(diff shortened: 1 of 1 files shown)


chewi:
I've noticed that the error you get back if all the attempts time out is quite generic.

```
Error: Error(Connect, Kind(TimedOut))
```

When working on the DNS timeout, I get the same type. Should I make it more specific or do we not really care whether it was a DNS timeout or a connect timeout?

chewi:
Hmm, did Travis fail because it doesn't support IPv6? I can use IPv4 instead but `240.0.0.1` is merely reserved for future use, it's not an official black hole IP.

chewi:
Closing as #1972 was merged first.

PR #1963, opened 2019-10-01
feat(lib): add optional `tcp` feature, split from `runtime`

seanmonstar:
The `HttpConnector` and `AddrListener` types which make use of
`tokio::tcp` have been made their own optional feature. This allows
using them without requiring the *full* tokio runtime.


davidbarsky:
I like this change—it'd make the feature flag more clear as to what is actually is. Do you have any additional feature flags that you plan on introducing with Hyper 0.13? I'm curious as to what's on the docket.

seanmonstar:
>  Do you have any additional feature flags that you plan on introducing with Hyper 0.13?

Only as a good reason shows up. :D

PR #1974, opened 2019-10-13
cargo: bump minimum versions

mathstuf:
`bytes` is bumped due to the addition of `Bytes::advance` in 0.4.6. `h2`
0.1.13 introduced the `.is_io()` and `into_io()` methods.

Bumping `tokio-threadpool` to 0.1.16 removes a dependency on `rand` from
the crate which reduces the size and avoids minimum version dependency
issues in that dependency tree.

---
Trying to make `-Z minimum-versions` work on some of my crates. Will bump `bytes` on `master` as well since I suspect it is necessary there too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,13 +19,13 @@ include = [
 ]
 
 [dependencies]
-bytes = "0.4.4"
+bytes = "0.4.6"
 futures = "0.1.21"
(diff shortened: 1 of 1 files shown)


mathstuf:
Would it be possible to get a `0.12.36` release with this?

PR #1977, opened 2019-10-16
Expose close() method on body::Sender

ikatson:
Without this method it's impossible to finish the stream successfully, and the client
will be waiting forever for you to send new data.


seanmonstar:
You can finish the stream successfully by just dropping the `Sender`.

ikatson:
@seanmonstar thanks, that's awesome, and works! 

A question though arose when you told this: can't the underlying futures::mpsc::sender still contain some unflushed data? Answering myself: looks like everything will work fine as the receiver contains the reference to the underlying queue and will be woken up on drop.

Closing the PR.

PR #1978, opened 2019-10-17
docs(service): remove reference to service_fn_ok

chrisvittal:
This function has been otherwise removed and the link was broken.


seanmonstar:
Nice catch, thanks!

PR #1983, opened 2019-10-18
fix(server): ignore expect-continue with no body in debug mode

### LINKED ISSUE #1843: server debug panic on `Expect: 100-continue` without `Content-Length`
This is related to #1842 but seems distinct enough to warrant its own issue.

When doing Icecast streaming, the client typically sends an authorization header and an `Expect: 100-continue` to wait for authorization before trying to upload a stream. However, because of #1842 the hyper server decides the client is not actually going to send any data, and if the server is compiled in debug mode, this causes a panic `expect-continue needs a body`. When built in release mode the server ignores the `Expect: 100-continue` and handles the message as if it has no body.

You can trigger this using curl: `curl -X PUT http://127.0.0.1:5000/ -d test -H 'Expect: 100-continue' -H 'Content-Length:'`.

seanmonstar:
Closes #1843

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -151,7 +151,9 @@ where I: AsyncRead + AsyncWrite + Unpin,
         self.state.version = msg.head.version;
 
         if msg.decode == DecodedLength::ZERO {
-            debug_assert!(!msg.expect_continue, "expect-continue needs a body");
+            if log_enabled!(log::Level::Debug) && msg.expect_continue {
+                debug!("ignoring expect-continue since body is empty");
+            }
             self.state.reading = Reading::KeepAlive;
             if !T::should_read_first() {
                 self.try_keep_alive(cx);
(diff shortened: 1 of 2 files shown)


src/proto/h1/conn.rs
```diff
@@ -151,7 +151,9 @@ where I: AsyncRead + AsyncWrite + Unpin,
         self.state.version = msg.head.version;
 
         if msg.decode == DecodedLength::ZERO {
-            debug_assert!(!msg.expect_continue, "expect-continue needs a body");
+            if log_enabled!(log::Level::Debug) && msg.expect_continue {
```
sfackler: The `log_enabled!` here doesn't seem necessary, since there's no expensive computation going on in the block.

seanmonstar: True... I guess with `max-log-level` features, it could then just be eliminated... But I guess LLVM would also be able to eliminate `if some_bool { /* nothing */ }`. 🤷‍♂

sfackler: Yeah empty if blocks will go away.

Issue #1903, opened 2019-08-17
Change `Resolve` to alias `Service`


seanmonstar:
After #1782 is done, the `Resolve` trait should be changed to basically an alias for a `Service<Name>`. This would allow more composability, being able to use ServiceExt to customize resolvers.

bstrie:
Now that #1792 is closed, what do you have in mind by "basically an alias"? I assume this isn't referring to waiting for https://github.com/rust-lang/rust/issues/41517 , so what should `Resolve` become?

seanmonstar:
It'd be similar to `HttpService`, basically just bounds that mean `R: Service<Name, Response = impl Iterator<Item = IpAddr>>`. As the top comment mentions, that'd allow composing `Service`s, such as easily adding in a timeout layer.

PR #1987, opened 2019-10-21
feat(client): add resolve timeout to HttpConnector

chewi:
The recently-added connect timeout does not cover resolving hostnames, which
could also stall on an OS-level timeout if there are issues reaching the DNS
server.

As mentioned in #1972.

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -32,6 +32,7 @@ type ConnectFuture = Pin<Box<dyn Future<Output = io::Result<TcpStream>> + Send>>
 pub struct HttpConnector<R = GaiResolver> {
     enforce_http: bool,
     handle: Option<Handle>,
+    resolve_timeout: Option<Duration>,
     connect_timeout: Option<Duration>,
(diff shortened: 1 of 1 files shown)


chewi:
Ouch, serious rebase needed! I've done most of it but the pin stuff has me stumped. I'll try a bit longer.

seanmonstar:
Sorry for missing this! Though this reminds me, once `Resolve` is just a `Service` (#1903), a user could set a timeout on it individually:

```rust
use ServiceExt;

let resolver = GaiResolver::new().timeout(some_dur);
let connector = HttpConnector::with_resolver(resolver);
```

chewi:
Ah cool, I was vaguely aware that some kind of custom resolver work was going on but I didn't realise you could do that.

I'll close this then but would you accept a PR to add my timeout to 0.12? I take it with 0.13 near that you won't be backporting the Service feature and we're not quite ready for 0.13 yet.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #1942, opened 2019-09-15
Create an SSL example with tokio-tls


trezm:
I'm attempting to get a version of a hyper server running with TLS off of the master branch (i.e. with async/await.) This is proving harder than expected, and I'd love for there to be an updated example of a hyper server using the tls acceptor.

I'd be happy to update the code myself, but don't quite see how it fits in to an example like [this](https://github.com/hyperium/hyper/blob/master/examples/hello.rs). Where can I add in a a TcpStream -> TlsStream portion?

trezm:
So, was hacking on this a little more and now have this:

```rust
      let listener = TcpListener::bind(&addr).await.unwrap();
      let mut incoming = listener.incoming();
      let server = Builder
          ::new(incoming.filter_map(async |socket| {
            let stream: TcpStream = socket.unwrap();

            Some(
              Ok::<_, hyper::Error>(
                _arc_acceptor.clone().accept(stream).await.unwrap()
              )
            )
          }), Http::new())
          .serve(service);

      server.await?;
```

but am getting a few errors that I'm confused about:

```
    |
100 |                   _arc_acceptor.clone().accept(stream).await.unwrap()
    |                                         ^^^^^^ the trait `tokio_io::async_read::AsyncRead` is not implemented for `tokio_net::tcp::stream::TcpStream`
```
```
    |
105 |           .serve(service);
    |            ^^^^^ the trait `tokio_io::async_read::AsyncRead` is not implemented for `tokio_tls::TlsStream<tokio_net::tcp::stream::TcpStream>`
```
```
    |
107 |       server.await?;
    |       ^^^^^^^^^^^^ the trait `tokio_io::async_write::AsyncWrite` is not implemented for `tokio_tls::TlsStream<tokio_net::tcp::stream::TcpStream>`
    |
    = note: required because of the requirements on the impl of `std::future::Future` for `hyper::server::Server<futures_util::stream::filter_map::FilterMap<tokio_net::tcp::incoming::Incoming, impl std::future::Future, [closure@/Users/pete/dev/thruster/thruster-server/src/ssl_hyper_server.rs:94:37: 104:12 _arc_acceptor:_]>, hyper::service::make_service::MakeServiceFn<[closure@/Users/pete/dev/thruster/thruster-server/src/ssl_hyper_server.rs:78:37: 91:8 arc_app:_]>>`
    = note: required by `std::future::poll_with_tls_context`
```

[...]

PR #1990, opened 2019-10-22
feat(server): Add an ssl example

trezm:
Adds an example using hyper + ssl for a server.

Should resolve the closed issue #1942


shirshak55:
any reason to change 100644 → 100755 ?

trezm:
Totally, not sure why/how the permissions go changed on these files! Updated accordingly.

Darkspirit:
https://github.com/ctz/hyper-rustls/blob/master/examples/server.rs

trezm:
Note: I think this might currently be broken due to this change: https://github.com/hyperium/hyper/commit/71d088d3d0c062aa05459697f53bf15018cfd651, at least until tokio updates their dependencies to future 0.3.1 as well.

@Darkspirit that's awesome!! If you'd like to replace this with an example from that repo that'd be great. It just seems like for such a simple setup there should be an example to point people towards if they happen to want to have SSL on their server.

Darkspirit:
Thank you for trying to merge a tls server example at all! :) I would be very grateful if you could update this pull request to use (the well received) rustls instead of non-rusty native-tls. Rustls and its dependencies webpki and ring are for example used by time.cloudflare.com, Google Fuchsia and Baidu. (Ring is based on Boringssl, Google's improved fork of OpenSSL, but got lots of C and C++ replaced with Rust.) It's quite sad that native-tls is spreading a bit like cancer, especially because it doesn't use Rust at places where its safety guarantees are most useful.

trezm:
Oh totally! Wasn't even aware of the differences/benefits!! Always happier
to use the rusty alternative anyway!

trezm:
Checking back in on this -- I have a [PR open](https://github.com/tokio-rs/tokio/pull/1873) to the tokio-tls repo to add rustls as a tls option. That being said, it looks like that might take a while to get in, which is unfortunate.

While experimenting though, I also noticed that the way the code is currently written, a long-lived http call will block request execution. Does anyone have any ideas around this? Basically because `filter_map` does not continue accepting sockets, we would need a stream op that always takes the next item and forms a new stream whose order is based on the completion time of an async closure run on each element of the original stream.

twe4ked:
This looks awesome, it would be great to have an SSL example in the repo. I've had a search around an most of the examples I found of SSL/hyper are outdated because the async code has been moving so fast. Any updates on this?

trezm:
I'll try to get back on this now that https://github.com/tokio-rs/tls is a
real thing!

seanmonstar:
Thanks for the PR! I'd rather not include examples in the hyper repo selecting specific TLS libraries. I think something like this *could* be part of hyper-tls, once it gains a server API.

PR #2000, opened 2019-10-30
chore(lib): use dedicated enums for connection protocol versions

djc:
This should make it easier to add H3 functionality.


seanmonstar:
Looks clear to me, thanks!

PR #2007, opened 2019-11-09
chore(dependencies): update futures to 0.3.1

ComputerDruid:


ComputerDruid:
Two TODOs still in this PR:

- [ ]  We still end up keeping an extra dep on futures-util-preview alpha 19 so we can call `StreamExt::next()` on tokio's TCP listener, that should be fixable by depending on a newer (unreleased when I last checked) tokio.
- [ ] Also, should probably rename our `poll_cancel()` -> `poll_cancelled()` to match the upstream change.

I figured I'd upload as-is for now in case that's useful. I don't mind waiting to merge until those are fixed though, if that's preferable.

ComputerDruid:
Ah, I see I also missed non-default features. This isn't ready yet

ComputerDruid:
> Ah, I see I also missed non-default features. This isn't ready yet

That part should be fixed now.

tests/server.rs
```diff
@@ -15,10 +15,11 @@ use std::time::Duration;
...
 use futures_core::ready;
 use futures_core::future::BoxFuture;
-use futures_util::future::{self, Either, FutureExt};
-use futures_util::stream::StreamExt;
-use futures_util::try_future::{self, TryFutureExt};
-//use futures_util::try_stream::TryStreamExt;
+use futures_util::future::{self, Either, FutureExt, TryFutureExt};
+#[cfg(feature = "unstable-stream")]
+use futures_util::stream::StreamExt as _;
+// TODO: remove once tokio is updated to futures 0.3
```
ctaggart: It looks like we are just waiting for a new `0.2.0-alpha.7`. https://github.com/tokio-rs/tokio/issues/1708

ComputerDruid: Yes, that'd be ideal. I've been meaning to try against a git hash pin, but I haven't gotten around to it yet.

Importing both works but isn't very pretty.

PR #2015, opened 2019-11-15
feat(lib): change custom `executor`s to be `Fn`

### LINKED ISSUE #1944: Client builder's executor setter is confusing for new tokio 0.2 alpha
Currently in hyper, the `Builder::executor<E>()` setter requires a `for<'a> &'a E: tokio_executor::Executor` bound. Which, in latest tokio (`=0.2.0-alpha.4`), only `Sender` meets this bound.

However, `Sender` is now not exported anymore by tokio. There seems to be no easy way to set custom executor.

seanmonstar:
Instead of custom executors needing to be `tokio::Executor` or
`tokio::TypedExecutor`, they are now just `Fn`s.

BREAKING CHANGE: Configuring an `executor` for a client or server should
  now pass a closure or function to spawn the future.

Closes #1944

### DIFF
--- a/examples/single_threaded.rs
+++ b/examples/single_threaded.rs
@@ -41,7 +41,9 @@ async fn main() {
     let exec = current_thread::TaskExecutor::current();
 
     let server = Server::bind(&addr)
-        .executor(exec)
+        .executor(move |fut| {
+            exec.clone().spawn_local(Box::pin(fut)).unwrap();
+        })
         .serve(make_service);
 
     println!("Listening on http://{}", addr);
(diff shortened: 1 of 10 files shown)


seanmonstar:
A closure frequently couldn't be used since the executor needed to be able to be passed different future types. New solution in https://github.com/hyperium/hyper/pull/2015

LucioFranco:
@seanmonstar btw that points to the same PR?

seanmonstar:
Woops, fixed XD

PR #2016, opened 2019-11-18
feat(client): filter remote IP addresses by family of given local IP address

chewi:
It is not possible to connect to an IPv4 address from an IPv6 address or vice-versa so don't waste time trying. If no remote addresses match then a "missing connect error" will now occur.

### DIFF
--- a/src/client/connect/dns.rs
+++ b/src/client/connect/dns.rs
@@ -194,17 +194,25 @@ impl IpAddrs {
         None
     }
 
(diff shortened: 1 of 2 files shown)


kw217:
Thanks, I like this. If the server is bound to a specific address, there's no point in trying IP addresses from a different address family. If we're not bound, behaviour is unchanged (i.e., try those with the same AF as the first first, then those with the other AF). (I'm not a project collaborator, just a user.)

seanmonstar:
Thanks!

PR #2019, opened 2019-11-19
refactor: Use async/await in HttpConnector

### LINKED ISSUE #1984: Rewrite HttpConnector internals with async/await
The internal code handling the `HttpConnector` is messy from manual future implementations. I believe it could be much cleaner if written with `async fn` and `await`.

Marwes:
Closes #1984

### DIFF
--- a/src/client/connect/dns.rs
+++ b/src/client/connect/dns.rs
@@ -21,18 +21,16 @@
 //!     Ok::<_, Infallible>(iter::once(IpAddr::from([127, 0, 0, 1])))
 //! });
 //! ```
-use std::{fmt, io, vec};
 use std::error::Error;
-use std::net::{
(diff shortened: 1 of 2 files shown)


seanmonstar:
This is excellent work! I really need to get rustfmt running by default so it doesn't make diffs harder to read...

Marwes:
Ah :( seems like my format on save setup does not obey `.rustfmt.toml` anymore, pretty sure it used to...

gitcop doesn't explain which or why the commit messages are invalid so I am not sure what to do about it

seanmonstar:
Sorry I didn't get back to this sooner (I've been trying to get all the things released), and so merging the Tokio 0.2 upgrade has caused merge conflicts.

seanmonstar:
Thanks so much!

PR #2022, opened 2019-11-22
feat(client): add interface to clear connection pool

guoyunlong16:
This patch adds an interface to clear client connection pool for downstream users.  This is useful when the network situation changes, such as from Wi-Fi to cellular network. 

If we do not clear the connection pool when the network situation changes, the network may be temporarily unavailable.


guoyunlong16:
In our tests, only when switching Wi-Fi to cellular network on iOS, the network is temporarily unavailable. The network works well on iOS when switching cellular network to Wi-Fi and on Android in both cases. 

The problem may be caused by iOS.

PR #2030, opened 2019-12-04
Update Tokio, bytes, http, h2, and http-body

### LINKED ISSUE #2025: Update to tokio 0.2
- [ ] Blocked on `h2`: https://github.com/hyperium/h2/milestone/2

seanmonstar:
Closes #2025


Cargo.toml
```diff
@@ -68,13 +60,13 @@ default = [
 ]
 runtime = [
     "tcp",
-    "tokio",
+    #"tokio",
```
LucioFranco: no longer optional?

seanmonstar: It should be, adding back in the features.

Cargo.toml
```diff
@@ -68,13 +60,13 @@ default = [
...
     "net2",
-    "tokio-executor/blocking",
-    "tokio-net",
-    "tokio-timer",
+    "tokio/blocking",
+    "tokio/tcp",
+    "tokio/time",
 ]
 
 # unstable features
```
davidbarsky: Should `unstable-stream` be renamed to `stream` for the same reasons as in H2? https://github.com/hyperium/h2/pull/433

seanmonstar: Yes, though I'd rather do it in a separate PR.

jeromegn:
I've been playing with this branch and I noticed `Body` doesn't support `try_concat` anymore with futures 0.3 since `Chunk` would need to implement `Extend<u8>` according to the [new function signature](https://docs.rs/futures-util/0.3.1/futures_util/stream/trait.TryStreamExt.html#method.try_concat).

```
the trait bound `hyper::body::chunk::Chunk: std::iter::Extend<u8>` is not satisfied
required because of the requirements on the impl of `core::future::future::Future` for `futures_util::stream::try_stream::try_concat::TryConcat<&mut hyper::body::body::Body>`
```

seanmonstar:
> `Body` doesn't support `try_concat` anymore

True, this is because `Bytes` (and thus `Chunk`) no longer implement `Extend<u8>`, since they are meant to be a frozen item. I have plans to add some alternatives to the `http-body` trait.

o01eg:
@seanmonstar What about `BytesMut` ?

PR #2035, opened 2019-12-04
feat(client): remove `Destination` for `http::Uri` in connectors

seanmonstar:
This changes "connectors" to be called with a `Uri` instead of `Destination`. The `Destination` type wasn't providing anything extra over an `http::Uri`.

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -13,6 +13,12 @@ use ::http::{Response};
 #[cfg(feature = "tcp")] mod http;
 #[cfg(feature = "tcp")] pub use self::http::{HttpConnector, HttpInfo};
 
+/// Describes a type returned by a connector.
+pub trait Connection {
+    /// Return metadata describing the connection.
+    fn connected(&self) -> Connected;
+}
+
 /// Extra information about the connected transport.
 ///
 /// This can be used to inform recipients about things like if ALPN
@@ -167,7 +173,7 @@ pub(super) mod sealed {
     use tokio::io::{AsyncRead, AsyncWrite};
 
     use crate::common::{Future, Unpin};
-    use super::{Connected};
+    use super::{Connection};
 
     /// Connect to a destination, returning an IO transport.
     ///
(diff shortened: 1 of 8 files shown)


seanmonstar:
@LucioFranco The problem with this change is that it doesn't allow `HttpConnector` to be `MakeService`, since it still returns `(TcpStream, Connected)`.

I tried changing that part to be `T: HttpConnection`, but The `Connected::extra` API allows for constructing a type that can go in `http::Extensions` and still be `Clone`, which is required because we spawn the `Connection` task, and don't have access to it when filling in the `Response::extensions`.

Another downside to requiring `T: HttpConnection` is that any glue crates, like those providing HTTPS connectors, would need to newtype the connection because they can't implement a 3rd party trait for a 3rd party type.

I don't really know the best way forward with that...

LucioFranco:
@seanmonstar I think it makes sense for them to newtype? The issue is this means we can't have a generic set of connectors for things that are not hyper too :(

Can we provide the api for adding things into http extensions?

seanmonstar:
OK, here's an update: this introduces a `hyper::client::connect::Connection` trait and removes the tuple from the connector "service" response. This trait is only needed for `hyper::Client` integration, not for other lower-level usage.

src/client/connect/mod.rs
```diff
@@ -5,22 +5,18 @@
...
-/// A set of properties to describe where and how to try to connect.
-///
-/// This type is passed an argument to connectors.
-#[derive(Clone, Debug)]
-pub struct Destination {
-    pub(super) uri: Uri,
+/// Describes a type returned by a connector.
+pub trait Connection {
+    /// Return metadata describing the connection.
+    fn connected(&self) -> Connected;
```
LucioFranco: I feel like we could add something just like this to http-connection, then we could stay on the same trait?

LucioFranco:
So I wonder if we could just merge this ability to add something to the `Request::extensions` into the `http-connection` crate. Also for server side do we want to let it know that it has negotiated alpn? Not that we have to do this all here but just suggesting :)

src/client/connect/mod.rs
seanmonstar: Does something else currently use `http-connection`?

LucioFranco: Not exactly that I know of but its something we _could_ use for this and to standardize connectors. Just thought if we want to introduce it now is the time before `0.13`. I am happy to not do it though just thinking that if we want to have external tower connectors http-connection is the best since its easy to depend on without having to bring in hyper.

LucioFranco: It was being used in `tower-http` but not sure what the future of that is.

seanmonstar: I personally would feel rushed exposing the internals of this stuff to be usable cross-crate. It can be something to plan for next time?

LucioFranco: Sure! Happy to punt and disucss later. I don't think this is blocking for tonic either since I manually configure the http settings both client and server side.

PR #2037, opened 2019-12-04
fix failing test and example

alce:
This patch fixes a failing example and test.


alce:
This is not right. Got confused by feature flags I guess?

PR #2038, opened 2019-12-04
feat(body): create an empty body from a unit

apyrgio:
Add a From<()> implementation for Body structs, so that units can
be converted to empty bodies.

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -479,6 +479,13 @@ impl From<Cow<'static, str>> for Body {
     }
 }
 
+impl From<()> for Body {
+    #[inline]
+    fn from(_: ()) -> Body {
+        Body::empty()
+    }
+}
+
 impl Sender {
     /// Check to see if this `Sender` can send more data.
     pub fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<crate::Result<()>> {


seanmonstar:
This could be a cool addition! I wonder though, do you have any existing situations where this would be useful? Seeing `Body::empty()` seems very clear to me.

apyrgio:
The trigger for this addition was some code that converted backend results to HTTP responses. Consider the case where the backend returns `Result<T, Error>`, where T is `String` or `()`, depending on the operation. With the `From<()>` impl, the conversion is straight-forward:

```
fn to_response<T>(res: Result<T, Error>) -> Response<Body>
where
    Body: From<T>
{
    match res {
        Ok(t) => let body = Body::from(t); ...
        Err(e) => ...
    }
}
```

It's not life changing, but it seemed intuitive to me to automatically convert empty tuples to empty bodies.

seanmonstar:
Yea, that does look convenient... In other places, I've had small concerns about automatic conversion from `()`, since a function that returns nothing returns `()` by default, and it might be an accident to convert that (like I mentioned [here](https://github.com/seanmonstar/warp/issues/233#issuecomment-497452863)).

Still, @LucioFranco or @sfackler, do you have any opinion?

LucioFranco:
I generally prefer the explicit `Body::empty`, in the case of warp there it looks like having an explicit body type is better than implicitness. This increases more with async/await since things are even more hidden.

seanmonstar:
Yea, I'd suggest an app-specific `struct NoBody` and implement `From<NoBody> for hyper::Body` in those cases.

apyrgio:
Indeed, you have a point here. The unit type is ubiquitous and thus more prone to implicit conversions. Also, if it's any indication, I see that the standard library does not have `From<()>` impls.

I see that @LucioFranco also agrees that it's better to be more explicit in this case, so I'm closing this PR. Thanks for your comments.

PR #2043, opened 2019-12-05
feat(client): impl tower_service::Service for Client

davidbarsky:
As discussed in Discord.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -528,6 +528,26 @@ where C: Connect + Clone + Send + Sync + 'static,
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -62,6 +62,7 @@ use std::fmt;
 use std::mem;
 use std::sync::Arc;
 use std::time::Duration;
+use std::task::Context;
```
seanmonstar: `task` is imported below, so `poll_ready` can just use `&mut task::Context<'_>`.

davidbarsky: Fixed!

PR #2045, opened 2019-12-05
GitHub Actions: Name steps and use matrix.include for features etc.

nickelc:
I added `fail-fast: false` because i've seen random failures on windows and macOS for tests.
- `client::pool::tests::test_pool_timer_removes_expired`
- `client::connect::http::tests::client_happy_eyeballs`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,35 +10,64 @@ env:
 
 jobs:
   test:
-    name: Test
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


seanmonstar:
Do you know if it's possible to add the `__internal_happy_eyeballs_tests` to the features list only on Ubuntu?

nickelc:
I'll look into it.

nickelc:
Done but still random failures of `client::pool::tests::test_pool_timer_removes_expired`

seanmonstar:
Thanks!

PR #2046, opened 2019-12-05
refactor(body): make Chunk a type alias for bytes::Bytes

### LINKED ISSUE #1931: Remove Chunk in favor of using Bytes directly

bstrie:
Closes #1931

Now I don't know the first thing about Hyper's codebase, but the change itself appears to have been quite simple to achieve. `cargo test` passes, and `cargo +nightly test --all-features`, while it doesn't pass, fails in the same way it does on master, so that gives me a bit of confidence. :)

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -3,7 +3,6 @@ use std::borrow::Cow;
 use std::error::Error as StdError;
 use std::fmt;
 
-use bytes::Bytes;
 use futures_core::Stream; // for mpsc::Receiver
 use futures_channel::{mpsc, oneshot};
 #[cfg(feature = "stream")]
@@ -421,13 +420,6 @@ impl From<Chunk> for Body {
         } else {
             Body::new(Kind::Once(Some(chunk)))
         }
-    }
-}
-
-impl From<Bytes> for Body {
-    #[inline]
-    fn from(bytes: Bytes) -> Body {
-        Body::from(Chunk::from(bytes))
     }
 }
 
--- a/src/body/chunk.rs
+++ b/src/body/chunk.rs
@@ -1,6 +1,4 @@
-use std::fmt;
-
-use bytes::{Buf, Bytes};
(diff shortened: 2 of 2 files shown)


fd:
This PR seems to be superseded by https://github.com/hyperium/hyper/pull/2048

PR #2050, opened 2019-12-06
feat(body): add `body::aggregate` and `body::to_bytes` functions

seanmonstar:
Adds utility functions to `hyper::body` to help asynchronously
collecting all the buffers of some `HttpBody` into one.

- `aggregate` will collect all into an `impl Buf` without copying the
  contents. This is ideal if you don't need a contiguous buffer.
- `to_bytes` will copy all the data into a single contiguous `Bytes`
  buffer.

### DIFF
--- a/src/common/buf.rs
+++ b/src/common/buf.rs
@@ -1,0 +1,75 @@
+use std::collections::VecDeque;
+use std::io::IoSlice;
+
+use bytes::Buf;
+
+pub(crate) struct BufList<T> {
+    bufs: VecDeque<T>,
+}
+
+impl<T: Buf> BufList<T> {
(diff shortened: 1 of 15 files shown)


src/common/buf.rs
```diff
@@ -0,0 +1,75 @@
+use std::collections::VecDeque;
```
LucioFranco: can we move these to an external crate? I'd like to use this in tonic.

seanmonstar: I've been wanting something like this externally too. @carllerche suggested maybe in something like `bytes-util` or `tokio-codec` or somewhere. 🤷‍♂

PR #2066, opened 2019-12-10
chore(CI): add cargo doc step to CI

seanmonstar:
The `cargo doc` step was failing, and CI didn't notice. This should help catch doc failures earlier.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,7 +30,6 @@ jobs:
           command: fmt
           args: --all -- --check
 
-
   test:
     name: Test ${{ matrix.rust }} on ${{ matrix.os }}
     needs: [style]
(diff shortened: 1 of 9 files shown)


seanmonstar:
This may not pass because of https://github.com/rust-lang/rust/issues/64705

.github/workflows/CI.yml
```diff
@@ -96,3 +95,29 @@ jobs:
         with:
           command: test
           args: --benches ${{ matrix.features }}
+
+      - name: Build docs
+        if: matrix.build-only != true
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
```
nickelc:
```suggestion
          command: doc
```

seanmonstar: Oops, I decided to just move it to a separate job completely, no need to build docs on all targets.

PR #2069, opened 2019-12-11
Add Body::to_vec() convenience method

djc:
Reusing this error is a little ugly, let me know if this makes sense and how you'd want it to look.

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -158,6 +158,31 @@ impl Body {
         Body::new(Kind::Wrapped(Box::pin(mapped)))
     }
 
+    /// Read the entire body into a `Vec<u8>`.
+    ///
+    /// This is useful if you have a short response that you want to read into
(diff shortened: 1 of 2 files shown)


nickelc:
Have you seen [`hyper::body::to_bytes(Body)`](https://docs.rs/hyper/0.13.0/hyper/body/fn.to_bytes.html) or [`hyper::body::aggregate(Body)`](https://docs.rs/hyper/0.13.0/hyper/body/fn.aggregate.html)?

djc:
Ah, no! Managed to miss those. Maybe I'll add a pointer to them in the `Body` docs.

seanmonstar:
I'd rather we just have the `body::{aggregate, to_bytes}` helpers for now. We could add more references to them in the docs, though.

djc:
Makes sense!

PR #2073, opened 2019-12-11
feat(client): expose `hyper::client::connect::Connect` trait alias

seanmonstar:
This is *just* a "trait alias". It is automatically implemented for all
`Service<Uri>`s that have the required bounds. It's purpose being public
is to ease setting trait bounds outside of hyper. Therefore, it doesn't
have any exposed associated types, to prevent otherwise relying on any
super-traits that hyper requires.

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -637,10 +637,13 @@ mod tests {
 
     use ::http::Uri;
 
-    use super::super::sealed::Connect;
+    use super::super::sealed::{Connect, ConnectSvc};
     use super::HttpConnector;
 
-    async fn connect<C>(connector: C, dst: Uri) -> Result<C::Transport, C::Error>
+    async fn connect<C>(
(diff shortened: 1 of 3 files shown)


LucioFranco:
I wonder if it makes sense to work off of `MakeConnection`, I think this might make it a bit easier to support things accross the tower ecosystem? Though this should be fine for now since hyper is the main tower client (maybe only) right now. But I know I ran into this issue with `tonic`, ill have to use this trait or enumerate the entire trait bounds.

seanmonstar:
> I wonder if it makes sense to work off of `MakeConnection`, I think this might make it a bit easier to support things accross the tower ecosystem? 

I don't think that will do what this does: allow someone to add the bounds without *depending* on the extra bounds.

LucioFranco:
Also my worry about things accross the ecosystem isn't an issue if we lift the other trait alias.

PR #2080, opened 2019-12-17
Updated tokio version -> 0.2.4

birarvin:
Updated tokio version to be compatible with tokio dependency for https://crates.io/crates/tokio-compat


CryZe:
0.2 and 0.2.4 are semver compatible, meaning that hyper already uses 0.2.4

PR #2087, opened 2019-12-20
Chunk method to detect if it is the final part of real http-chunk or not

inv2004:
feat(client): Chunk method to detect if it is the final part of real http-chunk or not

During chunked transfer - hyper can return only part of http-chunk and it does not provide ability to detect if it is just part of the chunk or not. The solution could be to fill chunks in the buffer until you know that you read full size of the http-chunk-len. This PR adds Chunk::is_end_of_chunk(&self) -> bool method which return true - if you read http-chunk-len.

```rust
let mut buf = Vec::new();
body.for_each(move |chunk| {
    let is_final = chunk.is_end_of_chunk();
    buf.extend_from_slice(&chunk.into_bytes());

   if !is_final {
      return Ok(())
   }

   println!("buf contains valid json");
   buf.clear();
}
```

### DIFF
--- a/src/body/chunk.rs
+++ b/src/body/chunk.rs
@@ -9,6 +9,7 @@ use bytes::{Buf, Bytes};
 /// A `Chunk` can be easily created by many of Rust's standard types that
 /// represent a collection of bytes, using `Chunk::from`.
 pub struct Chunk {
+    is_final: bool,
     /// The buffer of bytes making up this body.
     bytes: Bytes,
 }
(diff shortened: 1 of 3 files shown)


sfackler:
Why is the end of the stream not a sufficient indicator? In this example you'd just move the println out after the for_each.

inv2004:
Sorry for bad example. Server sends several jsons: one per http-chunk.

seanmonstar:
Intermediaries (proxies and gateways) may re-interpret the chunked encoding, so this isn't reliable. Additionally, HTTP2 and 3 don't use chunked encoding. And in 0.13, we got rid of the `Chunk` type in favor of just `Bytes`.

PR #2089, opened 2019-12-22
feat(client): impl tower_service::Service for &Client

tesaguri:
There's a blanket implementation of `tower_service::Service` for `&mut S` but not for `&S`. Since `hyper::Client::request` takes `&self` only, `&Client` can implement `Service`, allowing users to take advantage of Tower's abstraction without additional ownership restrictions.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -562,6 +562,26 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


tesaguri:
Rebased and fixed the build.

seanmonstar:
Thank you! Sorry for forgetting about this one.

PR #2091, opened 2019-12-30
add 'bytes' method to body to concatenate bytes

danieleades:


danieleades:
build is failing due to use of deprecated `Error::description`.

also 'gitcop' is unhappy about something

seanmonstar:
There's already `hyper::body::to_bytes` to handle *any* `impl HttpBody`.

danieleades:
ah yep, missed that.

PR #2094, opened 2019-12-30
Mild refactoring

danieleades:


src/proto/h1/encode.rs
```diff
@@ -202,48 +202,51 @@ impl Encoder {
     }
 }
 
+impl<B> EncodedBuf<B>
+where
+    B: Buf,
+{
+    fn buf(&self) -> &dyn Buf {
```
seanmonstar: Do we know if this negatively affects performance at all?

src/proto/h1/io.rs
```diff
@@ -168,12 +168,9 @@ where
                     }
                 }
             }
-            match ready!(self.poll_read_from_io(cx)).map_err(crate::Error::new_io)? {
-                0 => {
-                    trace!("parse eof");
-                    return Poll::Ready(Err(crate::Error::new_incomplete()));
-                }
-                _ => {}
+            if let 0 = ready!(self.poll_read_from_io(cx)).map_err(crate::Error::new_io)? {
```
seanmonstar: `if let 0 =` reads odd to me. Should just be `if $expr == 0`?

src/proto/h1/io.rs
```diff
@@ -325,35 +321,33 @@ impl ReadStrategy {
...
-                        // A read within the current range should cancel
-                        // a potential decrease, since we just saw proof
-                        // that we still need this size.
+        if let ReadStrategy::Adaptive {
+            ref mut decrease_now,
+            ref mut next,
+            max,
+            ..
+        } = *self
+        {
```
seanmonstar: Does rustfmt suggest this? I personally prefer braces after the expression instead of a newline usually...

src/proto/h1/encode.rs
danieleades: I can't think of a reason why it should, except that i don't have `#inline` sprinkled on the methods. I could add them, but since i don't what effect it will have, it would feel a bit cargo culty...

danieleades: totally open to input here- also I can just scrap it if there's any performance concerns

src/proto/h1/io.rs
danieleades: copy that, will update

src/proto/h1/io.rs
danieleades: yeah this is rustfmt. I notice that this project doesn't have a `.rustfmt.toml`, this could be configured to avoid the newline in cases like this.

danieleades:
no worries, will update the PR.
would you prefer me to squash the commits?

seanmonstar:
For the commits, do what's easiest for you. I can manage when merging.

src/proto/h1/io.rs
danieleades: done.

danieleades:
some of these changes are mine, most of them are clippy lints. My goal here was really to make clippy quiet enough that it could be used in CI for this project.

All the 'deny's are now fixed, which means a clippy check could be added to CI today.
Clippy is still 'warning' about the following-

- structs with `new()` methods, that don't implement default.
    
    I guess adding this would create a hefty amount of duplication, but could make occasionally simplify things in calling code (ie `unwrap_or_default()` and similar)

- overly complex types

    such as `Poll<Option<crate::Result<Connecting<IO, S::Future, E>>>>`. personally i'm not sure that splitting this into typedefs outside the impl block makes this more readable, but hey.

- everywhere that matches on `Err(_)` (which is a lot of places). Clippy complains that the specific error type isn't being handled. fair enough i guess

- passing `Copy` types by value, and also cloning them. there's a lot of this around. I kind of like this if there's ever any chance of the `Copy` derive being removed in future, and sometimes it's more explicit. But there would be an immeasurable performance improvement by fixing this. (i've not touched these in this PR)

- `proto::h1::role::encode`. i mean jesus, fair enough. There's quite a few monolithic functions than should be split up, and a lot of these might benefit from async/await syntax. I think it's out of scope of this pull request, and probably above my paygrade until i'm more familiar with the codebase.

[cargo-scout](https://github.com/o0Ignition0o/cargo-scout) looks like a cool solution- this would apply clippy lints only to *changed* code on pull requests. It's not stable yet, but would be a nice incremental adoption strategy. The other options would be to either accept clippy warnings, or explicitly allow the above lints.

src/proto/h1/encode.rs
danieleades: i've put the `#[inline]`s back. the situation should now be no better or worse than before.

src/proto/h2/mod.rs
```diff
@@ -47,16 +47,15 @@ fn strip_connection_headers(headers: &mut HeaderMap, is_request: bool) {
...
+    } else if headers.remove(TE).is_some() {
+        warn!("TE headers illegal in HTTP/2 responses");
     }
 
     if let Some(header) = headers.remove(CONNECTION) {
+        static STATIC_CONNECTION: HeaderName = CONNECTION;
         warn!(
             "Connection header illegal in HTTP/2: {}",
-            CONNECTION.as_str()
+            STATIC_CONNECTION.as_str()
```
seanmonstar: I don't understand this change. Why is this better?

src/proto/h1/encode.rs
danieleades: no problemo

src/proto/h2/mod.rs
danieleades: this is a subtle one. It's the result of [this clippy lint](https://rust-lang.github.io/rust-clippy/master/#declare_interior_mutable_const), which is a hard failure.

danieleades: i've moved this change to a separate pull request.

PR #2097, opened 2020-01-03
create a static reference to interior mutable const

danieleades:


danieleades:
this is the result of the following clippy lint which i have blindly followed
https://rust-lang.github.io/rust-clippy/master/#declare_interior_mutable_const

I can't see any statistically significant difference in the benches on my machine.

seanmonstar:
Huh, that's an odd one.

1. This really looks like a false positive on clippy's part. This method call doesn't mutate the constant...
2. Looking at the surrounding code, I think this log may be wrong! It should be logging the `header` value that was just removed.

sfackler:
https://github.com/rust-lang/rust-clippy/issues/3825

PR #2101, opened 2020-01-04
style: use just std instead of ::std in paths

faern:
It was a long time ago one had to write `::std` to make rustc recognize the path. Now just `std` works fine.


faern:
What do you prefer the "Scope" part of the commit message be when it touches multiple parts? Or would you prefer I split this change up into one commit per component? Gitcop is angry.

seanmonstar:
Don't worry about it, I'll fix it up in the merge.

PR #2104, opened 2020-01-08
feat: Implement Clone/Copy on ServiceFn and MakeServiceFn

Marwes:
And expose them in docs. Since Clone and Copy can't be propagated out
via `impl Service` I believe it is better to expose them now.
`tower::service_fn` already implement `Clone` and `Copy.

### DIFF
--- a/src/service/mod.rs
+++ b/src/service/mod.rs
@@ -45,5 +45,5 @@ pub(crate) use self::http::HttpService;
 pub(crate) use self::make::{MakeConnection, MakeServiceRef};
 pub(crate) use self::oneshot::{oneshot, Oneshot};
 
-pub use self::make::{make_service_fn, MakeServiceFn};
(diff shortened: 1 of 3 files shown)


src/service/mod.rs
```diff
@@ -45,5 +45,5 @@ pub(crate) use self::http::HttpService;
 pub(crate) use self::make::{MakeConnection, MakeServiceRef};
 pub(crate) use self::oneshot::{oneshot, Oneshot};
 
-pub use self::make::make_service_fn;
-pub use self::util::service_fn;
+pub use self::make::{make_service_fn, MakeServiceFn};
```
seanmonstar: I'd prefer to keep them un-exported for the time being.

Marwes: Done (though I find un-exported types like this makes unnecessarily hard to understand what traits they implement and how they implement them).

Issue #2103, opened 2020-01-08
http_proxy.rs example not work when handle https proxy


kaiwk:
`https_proxy=127.0.0.1:8100 curl -i https://google.com` gives `400 Bad Request`

seanmonstar:
Yes, the example prints this line to your console: `CONNECT host is not socket addr: google.com:443`.

The example doesn't include DNS.

PR #2105, opened 2020-01-09
fix(example): add dns resolve for http_proxy

kaiwk:
see #2103

### DIFF
--- a/examples/http_proxy.rs
+++ b/examples/http_proxy.rs
@@ -5,6 +5,7 @@ use std::net::SocketAddr;
 
 use futures_util::future::try_join;
 
+use hyper::client::connect::dns::{resolve, GaiResolver};
 use hyper::service::{make_service_fn, service_fn};
 use hyper::upgrade::Upgraded;
 use hyper::{Body, Client, Method, Request, Response, Server};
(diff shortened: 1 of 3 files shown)


examples/http_proxy.rs
```diff
@@ -1,7 +1,7 @@
 #![deny(warnings)]
 
 use std::convert::Infallible;
-use std::net::SocketAddr;
+use std::net::{SocketAddr, ToSocketAddrs};
```
davidbarsky: Tokio now has an asynchronous DNS resolution function: https://docs.rs/tokio/0.2.8/tokio/net/fn.lookup_host.html. Maybe consider using that instead, as std's `to_socket_addrs()` is a blocking operation.

seanmonstar: Or another option, that doesn't require enabling another Tokio feature, is to just use `hyper::client::connect::dns::GaiResolver`.

kaiwk: > Tokio now has an asynchronous DNS resolution function: https://docs.rs/tokio/0.2.8/tokio/net/fn.lookup_host.html. Maybe consider using that instead, as std's `to_socket_addrs()` is a blocking operation.

it seems we need upgrade `tokio` to use `lookup_host`.

I  `pub` the  `resolve` method in `dns.rs`, is it appropriate?

shirshak55:
is there any reason not to use to_socket_addr which comes with std library?
https://github.com/shirshak55/http-to-socks/blob/master/src/main.rs#L74

shirshak55:
Is this pr going to be merged?

seanmonstar:
I'd rather not make some of these functions part of the public API. The example includes in its comments what it can and cannot do, and even with DNS resolution, it won't work over HTTPS. I think for that reason it's better to show it in a limited capacity, and imply that making a full proxy is more involved.

PR #2119, opened 2020-01-27
fix(http1): only send `100 Continue` if request body is polled

### LINKED ISSUE #838: Change 100-continue behavior to send when Body has been polled

seanmonstar:
Before, if a client request included an `Expect: 100-continue` header,
the `100 Continue` response was sent immediately. However, this is
problematic if the service is going to reply with some 4xx status code
and reject the body.

This change delays the automatic sending of the `100 Continue` status
until the service has call `poll_data` on the request body once.

Closes #838

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -11,7 +11,7 @@ use futures_util::TryStreamExt;
 use http::HeaderMap;
 use http_body::{Body as HttpBody, SizeHint};
 
-use crate::common::{task, Future, Never, Pin, Poll};
+use crate::common::{task, watch, Future, Never, Pin, Poll};
 use crate::proto::DecodedLength;
 use crate::upgrade::OnUpgrade;
 
@@ -33,7 +33,7 @@ enum Kind {
     Once(Option<Bytes>),
     Chan {
         content_length: DecodedLength,
-        abort_rx: oneshot::Receiver<()>,
+        want_tx: watch::Sender,
         rx: mpsc::Receiver<Result<Bytes, crate::Error>>,
     },
     H2 {
@@ -79,11 +79,13 @@ enum DelayEof {
 /// Useful when wanting to stream chunks from another thread. See
 /// [`Body::channel`](Body::channel) for more.
 #[must_use = "Sender does nothing unless sent on"]
-#[derive(Debug)]
 pub struct Sender {
-    abort_tx: oneshot::Sender<()>,
+    want_rx: watch::Receiver,
     tx: BodySender,
 }
+
+const WANT_PENDING: usize = 1;
+const WANT_READY: usize = 2;
 
 impl Body {
     /// Create an empty `Body` stream.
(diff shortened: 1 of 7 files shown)


src/body/body.rs
```diff
@@ -460,19 +469,28 @@ impl From<Cow<'static, str>> for Body {
 impl Sender {
     /// Check to see if this `Sender` can send more data.
     pub fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<crate::Result<()>> {
-        match self.abort_tx.poll_canceled(cx) {
```
seanmonstar: Seems this was important for detecting when the receiver is dropped while the connection is waiting on the socket.

sfackler:
Does this need to be handled for http2 as well?

seanmonstar:
Could be an enhancement, but the HTTP2 side doesn't currently have any handling of `Expect: 100-continue`. I believe sending a `RST_STREAM` is preferred there, which hyper *does* handle.

sfackler:
Do you mean the HTTP2 standard doesn't define 100-continue semantics, or that h2 doesn't deal with it right now? RST_STREAM doesn't quite cover the same use cases, since the client doesn't have a robust way to know if it should send the body or not other than I guess waiting for a little bit to see if it gets a RST.

It seems like some implementations do support it, FWIW. Don't feel super strongly about it though: https://github.com/dotnet/corefx/issues/31312

seanmonstar:
>  that h2 doesn't deal with it right now? 

This one. I haven't done much looking into what HTTP2 says about `expect: 100-continue`, I just figured `RST_STREAM` can serve a similar purpose, and h2 *does* handle that (a `RST_STREAM` is sent if the body is dropped).

i110:
Do you have a plan to make the same thing possible with HTTP/2? According to HTTP specifications, an h2 server, like h1, must return a 100 continue response to `Expect: 100-continue`. Also, in practical terms beyond the spec, 100 continue allows for more precise control of body transmission timing than RST_STREAM. I need it right now and I'm thinking of creating a patch if there's no ongoing implementation or concerns. If you have any suggestions, please let me know.

PR #2127, opened 2020-01-30
feat(server): add poll_peek to AddrStream

daxpedda:
This was specifically added to peek for the first byte to determine if the connection is HTTP or HTTPS.
I copied the documentation from tokio.

### DIFF
--- a/src/server/tcp.rs
+++ b/src/server/tcp.rs
@@ -223,6 +223,15 @@ mod addr_stream {
         pub fn into_inner(self) -> TcpStream {
             self.inner
         }
+
+        /// Attempts to receive data on the socket, without removing that data from the queue, registering the current task for wakeup if data is not yet available.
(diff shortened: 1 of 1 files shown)


daxpedda:
Rebased and fixed the build.

src/server/tcp.rs
```diff
@@ -223,6 +223,15 @@ mod addr_stream {
         pub fn into_inner(self) -> TcpStream {
             self.inner
         }
+
+        /// Attempts to receive data on the socket, without removing that data from the queue, registering the current task for wakeup if data is not yet available.
```
seanmonstar:
```suggestion
        /// Attempt to receive data on the socket, without removing that data
        /// from the queue, registering the current task for wakeup if data is
        /// not yet available.
```

seanmonstar:
Thanks!

PR #2138, opened 2020-02-22
feat(http2): add adaptive window size support using BDP

seanmonstar:
This adds support for calculating the Bandwidth-delay product when using
HTTP2. When a DATA frame is received, a PING is sent to the remote.
While the PING acknowledgement is outstanding, the amount of bytes of all
received DATA frames is accumulated. Once we receive the PING
acknowledgement, we calculate the BDP based on the number of received
bytes and the round-trip-time of the PING. If we are near the current
maximum window size, the size is doubled.

It's disabled by default until tested more extensively.

cc #1960

### DIFF
--- a/src/proto/h2/bdp.rs
+++ b/src/proto/h2/bdp.rs
@@ -1,0 +1,190 @@
+// What should it do?
+//
+// # BDP Algorithm
+//
+// 1. When receiving a DATA frame, if a BDP ping isn't outstanding:
+//   1a. Record current time.
+//   1b. Send a BDP ping.
+// 2. Increment the number of received bytes.
+// 3. When the BDP ping ack is received:
+//   3a. Record duration from sent time.
+//   3b. Merge RTT with a running average.
+//   3c. Calculate bdp as bytes/rtt.
+//   3d. If bdp is over 2/3 max, set new max to bdp and update windows.
+//
+//
+// # Implementation
+//
+// - `hyper::Body::h2` variant includes a "bdp channel"
+//   - When the body's `poll_data` yields bytes, call `bdp.sample(bytes.len())`
+//
+
+use std::sync::{Arc, Mutex, Weak};
+use std::task::{self, Poll};
+use std::time::{Duration, Instant};
+
+use h2::{Ping, PingPong};
+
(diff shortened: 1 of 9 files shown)


src/proto/h2/bdp.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub(crate) fn is_enabled(&self) -> bool {
+        self.shared.strong_count() > 0
+    }
+}
+
+impl Estimator {
+    pub(super) fn poll_estimate(&mut self, cx: &mut task::Context<'_>) -> Poll<WindowSize> {
+        let mut inner = self.shared.lock().unwrap();
+        if !inner.ping_sent {
+            // XXX: this doesn't register a waker...?
```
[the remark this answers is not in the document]
seanmonstar: Well it shouldn't need to, since sending the ping will wake up the h2 connection task, which is the same one... It's just that `poll_estimate` is assuming that even though the function doesn't otherwise know.

src/proto/h2/bdp.rs
```diff
@@ -0,0 +1,190 @@
...
+            return None;
+        }
+
+        // average the rtt
+        let rtt = seconds(rtt);
+        if self.samples < 10 {
+            // Average the first 10 samples
+            self.rtt += (rtt - self.rtt) / (self.samples as f64);
+        } else {
+            self.rtt += (rtt - self.rtt) / 0.9;
```
goffrie, line 160: Is this trying to be an exponential moving average? Should this be `* 0.9` then?

seanmonstar: Yea, I inversed it by accident. It's fixed in the latest release.

goffrie: Ah my bad, I didn't notice because I was using an old release.

src/proto/h2/bdp.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        // calculate the current bandwidth
+        let bw = (bytes as f64) / (self.rtt * 1.5);
```
loyd, line 164: Can somebody explain why 1.5?

PR #2139, opened 2020-02-26
feat(proto): update to time 0.2

koushiro:


seanmonstar:
Thanks for the PR!

I wonder, how "weighty" is time v0.2? We really only need the ability to format a `SystemTime` to a single specific format...

Cargo.toml
```diff
@@ -52,7 +52,7 @@ serde_json = "1.0"
 tokio = { version = "0.2.2", features = ["fs", "macros", "io-std", "rt-util", "sync", "time", "test-util"] }
 tokio-test = "0.2"
 tower-util = "0.3"
-url = "1.0"
+url = "2.0"
```
jhpratt: I'm by no means the maintainer of hyper, but I would imagine keeping this separate would be best.

src/proto/h1/date.rs
```diff
@@ -52,18 +52,18 @@ impl CachedDate {
...
-    fn update(&mut self, now: time::Timespec) {
+    fn update(&mut self, now: OffsetDateTime) {
         self.pos = 0;
-        let _ = write!(self, "{}", time::at_utc(now).rfc822());
+        let _ = write!(self, "{}", now.format("%a, %d %b %Y %H:%M:%S GMT"));
         debug_assert!(self.pos == DATE_VALUE_LENGTH);
-        self.next_update = now + Duration::seconds(1);
-        self.next_update.nsec = 0;
+        let nanosecond = self.next_update.time().nanosecond();
+        self.next_update = now + Duration::seconds(1) - Duration::nanoseconds(nanosecond as i64);
```
jhpratt: Easier to just add the number of nanoseconds directly. It may be worth checking the optimized output of this, though, as the compiler may be able to optimize away the `x - x`.

```suggestion
        self.next_update = now + Duration::nanoseconds(1_000_000_000 - now.nanosecond() as i64);
```

src/proto/h1/date.rs
```diff
@@ -52,18 +52,18 @@ impl CachedDate {
...
         if now > self.next_update {
             self.update(now);
         }
     }
 
-    fn update(&mut self, now: time::Timespec) {
+    fn update(&mut self, now: OffsetDateTime) {
         self.pos = 0;
-        let _ = write!(self, "{}", time::at_utc(now).rfc822());
+        let _ = write!(self, "{}", now.format("%a, %d %b %Y %H:%M:%S GMT"));
```
jhpratt: This needs to be explicitly set to UTC. Without this, the output value may not be semantically correct.

```suggestion
        let _ = write!(self, "{}", now.to_offset(UtcOffset::UTC).format("%a, %d %b %Y %H:%M:%S GMT"));
```

Cargo.toml
```diff
@@ -31,7 +31,7 @@ h2 = "0.2.1"
 itoa = "0.4.1"
 log = "0.4"
 pin-project = "0.4"
-time = "0.1"
+time = { version = "0.2.7", default-features = false, features = ["std"] }
```
jhpratt: There's no particular reason to require 0.2.7. 0.2.4 would make the most sense, given it lowered the MSRV to 1.34 (from 1.40 beforehand).

src/proto/h1/date.rs
koushiro: The `OffsetDateTime::now()` returns the current date and time in UTC.
I don't think it's necessary to set to UTC explicitly.

Cargo.toml
koushiro: make sense

src/proto/h1/date.rs
jhpratt: It's zero-cost to convert to UTC if it already is, so I think it would be smart to add it anyways. Otherwise you're in premature optimization territory, as you may as well only be storing the `PrimitiveDateTime` if you're already assuming the time zone.

koushiro: Well, setting the offset to UTC explicitly is not bad at least. I will add it :)

tomaka:
As someone who is experimenting with the `wasm32-wasi` target, `time` v0.1 is problematic because [it does the time gathering itself](https://github.com/time-rs/time/blob/v0.1/src/sys.rs). I have to use a local override at the moment to make it work.

`time` v0.2, on the other hand, delegates everything to `std::time::SystemTime`, which does work on wasi.

jhpratt:
While I don't believe it should block this PR landing, a 0.3 is in progress, which will have very different formatting. It's not done yet, so I can't speak as to how well the compiler will optimize it away, but it should be _very_ low cost, assuming the format is known at compile-time (which it sounds like it is).

seanmonstar:
I've filed #2253 to remove `time` 0.1. I was concerned about the "weight" of the new version of time, specifically it has an extra dependency, a procedural macro, and a bunch of time manipulation code that hyper didn't need. I've opted to use `httpdate`, which still uses `SystemTime` internally, but has much less code to compile.

jhpratt:
Just to address the concern about additional weight, time 0.3 will have far more aggressive feature flags. The macros will be fully rewritten to be lightweight (eliminating syn and quote) and being a flag. I also recently ran `cargo diet` to eliminate publishing unnecessary files, though that's not yet published (or even pushed up I believe).

The extra dependency is standback, which compiles down to _literally_ nothing on the most recent stable — its sole purpose is to backport APIs, so all the compiler would have to do it tokenize and determine what compiler is being used via a build script.

I've actually considered putting the arithmetic behind an on-by-default flag, such that `no-default-features` would essentially leave you with just the struct definitions and inherent methods, nothing else. Formatting will be placed behind a flag in 0.3 as well.

Of course, httpdate is still probably more efficient, given that it's a fixed format. It's pretty difficult to surpass that, I'd imagine. Just throwing this out there to indicate that most of the issues with regard to weight will be improved! Thanks for your work on hyper and everything else.

seanmonstar:
That's excellent to hear!

PR #2142, opened 2020-02-27
feat(client): rename `client::Builder` pool options

seanmonstar:
- Renamed `keep_alive_timeout` to `pool_idle_timeout`.
- Renamed `max_idle_per_host` to `pool_max_idle_per_host`.
- Deprecated `keep_alive(bool)` due to confusing name. To disable the
  connection pool, call `pool_max_idle_per_host(0)`.


seanmonstar:
This is partly motivated by wanting to add HTTP2 keep-alive options, and the names `keep_alive(bool)` and `keep_alive_timeout` would be confusing as they have different semantics.

PR #2143, opened 2020-02-27
Seal `Resolve` and make it public

LucioFranco:
This follows `Connect`'s path and should make it a bit easier to discover this trait.

Signed-off-by: Lucio Franco <luciofranco14@gmail.com>

### DIFF
--- a/src/client/connect/dns.rs
+++ b/src/client/connect/dns.rs
@@ -32,7 +32,7 @@ use std::{fmt, io, vec};
 use tokio::task::JoinHandle;
 use tower_service::Service;
 
-pub(super) use self::sealed::Resolve;
+pub use self::sealed::Resolve;
 
 /// A domain name to resolve into IP addresses.
 #[derive(Clone, Hash, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


LucioFranco:
@seanmonstar honestly, I think these should be some sort of direct trait alias or even just using Service bound directly.

Can we get rid of all the weird internal apis and just use trait aliases that are not sealed?

seanmonstar:
It's possible... What makes you prefer that?

LucioFranco:
@seanmonstar  I think there is a lot of confusion around implementing these lower level traits. I think one way to solve this is to still use trait aliases but make it a bit more clear what is going on. For example, we should make `H2Exec` public and write docs on how to implement it. So when a user clicks on it they are redirected there and can easily find the info they need to implement it. 

I also believe that the blanket impl should block any direct impls of the trait? 

Mostly, just curious how we could improve the UX around these traits since they have caused a lot of friction in the past.

davidbarsky:
Since I'm the person who bothered @LucioFranco about enough that he opened this PR: my view on:

> What do you think of the extra non-sense we do for Connect, like hiding the associated types behind an un-exported extra trait, and making the methods with an internal-only argument? Am I too paranoid?

...is that this approach is a bit overkill. While I understand the need for defensively designed interfaces and misuse-resistant code, I feel that these public-but-unusable traits aliases are far too high of a barrier and ergonomic hit. I also believe that the risk of someone implementing  `Connect` or `Resolve` and that introducing additional difficulties for non-breaking version bumps are _greatly_ outweighed by the accessibility and ergonomic wins of making these traits easy and possible to implement.

For instance, I attempted to update [hyper-rustls](https://github.com/ctz/hyper-rustls) to Hyper 0.13 prior to `hyper-rustls` being updated and the _biggest_ hurdle I faced are the sealed trait aliases. They were high enough of a hurdle that I eventually abandoned the update attempt.

(I apologize that this comment isn't as well-edited it should be. I'm pretty tired today.)

djc:
I very strongly agree with @davidbarsky on especially this:

> I also believe that the risk of someone implementing Connect or Resolve and that introducing additional difficulties for non-breaking version bumps are greatly outweighed by the accessibility and ergonomic wins of making these traits easy and possible to implement.

seanmonstar:
I just looked through the source, and it looks like `Resolve` is only referenced by `HttpConnector::new_with_resolver(r)`. The docs there don't help, as it just says to pass a `Resolve` argument. I *think* the best thing to do is to change that documentation to point at the `dns` module docs, which already explains how to implement `Resolve`. I'll file an issue.

And if it continues to confuse people, I'll concede that my paranoia is less important.

PR #2149, opened 2020-03-09
fix(http1): try to drain connection buffer if user drops Body

seanmonstar:


seanmonstar:
This should help https://github.com/seanmonstar/reqwest/issues/844

PR #2186, opened 2020-04-18
Force tokio to be at least 0.2.5

Eijebong:
Without this, you can end up with tokio 0.2.4 and hyper 0.13.5 in your
project, leading to a compile error like this:

```
error[E0599]: no method named `try_recv` found for struct `tokio::sync::mpsc::unbounded::UnboundedReceiver<client::dispatch::Envelope<T, U>>` in the current scope
   --> src/client/dispatch.rs:161:26
    |
161 |         match self.inner.try_recv() {
    |                          ^^^^^^^^ method not found in `tokio::sync::mpsc::unbounded::UnboundedReceiver<client::dispatch::Envelope<T, U>>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `hyper`.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ log = "0.4"
 pin-project = "0.4"
 time = "0.1"
 tower-service = "0.3"
-tokio = { version = "0.2", features = ["sync"] }
+tokio = { version = "0.2.5", features = ["sync"] }
 want = "0.3"
 
 # Optional


nightmared:
I'm curious, why not use `version = "~0.2.5"`. You fear some feature may disappear in the 0.2.* branch ?
That way you could get a potentially more up-to-date version of tokio.

Eijebong:
"0.2.5" and "^0.2.5" are equivalent (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) Which means that you can use 0.2.y with `y > 5` for tokio.

nightmared:
Oh I see, thanks for the explanation !

Issue #2159, opened 2020-03-24
Why does Body need to be Sync?


rkuhn:
In https://github.com/hyperium/hyper/pull/2048/files#diff-8fd688c7233290bc8ae01519ef707f1eR58 the Body was changed to require a `Sync` bound. This trickles down into `Body::wrap_stream`, which now requires the Stream to be Sync as well. A Stream struct basically is the description of a task out of which elements can be pulled. Requiring this task to be Sync amounts to saying that the `poll_next` function needs to be reentrant, but that runs counter to its signature which promises a `&mut` self reference (i.e. an exclusive reference). Somewhere in here something is broken, could you enlighten me exactly where the breakage lies?

sfackler:
That is not the commit that made the change, this is: https://github.com/hyperium/hyper/commit/4441372121e8b278ac773ddd4e408a642dadf2d8.

Reentrancy is not discussed in the definition of Sync that I'm aware of. The standard definition has to do with safe concurrent access to a shared reference to a type.

seanmonstar:
The reason is described in the comment, pointing at https://github.com/rust-lang/rust/issues/57017.

Without the requirement, this code fails to compile in a very confusing way:

```rust
match res.status() {
    StatusCode::OK => {
        let _chunk = res.body_mut().data().await;
    },
    _ => (),
}
```

rkuhn:
@sfackler Thanks for the more precise link! Being able to safely share a reference between multiple threads includes the possibility of those threads concurrently making use of the reference, which in this case implies reentrancy of the one method offered by said reference.

@seanmonstar Good to know, so my conclusion is that the design of async/await forces Sync in unexpected places. Considering that Sync is “Send for references”, it becomes obvious that the async/await syntax is using these marker types in a way that overapproximates the problem: while it is true that these objects may be sent to other threads, the usual implication of Sync is concurrent access, which is lexically impossible in an async function body. As long as the async scheduler takes care to insert the appropriate optimizer and CPU barriers, an async function body should be regarded as a single-threaded piece of code, so both Send and Sync are not exactly the right abstractions to model what is going on. Do you think it makes sense to discuss this in a different place or forum closer to Rust core? I am rather new in this ecosystem and still learning its ways.

seanmonstar:
You could bring it up in the issue I referenced, or if it only loosely matches, try internals.rust-lang.org.

rkuhn:
done at internals, thanks!

PR #2187, opened 2020-04-19
feat(body): remove Sync bound for Body::wrap_stream

rkuhn:
A stream wrapped into a Body previously needed to implement `Sync` so
that the Body type implements this autotrait as well (which is needed
due to limitations in async/await). Since a stream only offers one
method that is called with an exclusive reference, this type is
statically proven to be Sync already. In theory it should be fine to add
an `unsafe impl Sync`, but this commit instead adds a Mutex that is
never locked to prove that this approach is correct.

This makes it easier to construct response bodies for client code.

For more discussion on this see [rust-internals](https://internals.rust-lang.org/t/what-shall-sync-mean-across-an-await/12020/12).

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -11,6 +11,7 @@ use futures_util::TryStreamExt;
 use http::HeaderMap;
 use http_body::{Body as HttpBody, SizeHint};
 
+use crate::common::sync_wrapper::SyncWrapper;
 use crate::common::{task, watch, Future, Never, Pin, Poll};
 use crate::proto::h2::ping;
 use crate::proto::DecodedLength;
@@ -42,13 +43,11 @@ enum Kind {
         content_length: DecodedLength,
         recv: h2::RecvStream,
     },
-    // NOTE: This requires `Sync` because of how easy it is to use `await`
-    // while a borrow of a `Request<Body>` exists.
-    //
-    // See https://github.com/rust-lang/rust/issues/57017
     #[cfg(feature = "stream")]
     Wrapped(
-        Pin<Box<dyn Stream<Item = Result<Bytes, Box<dyn StdError + Send + Sync>>> + Send + Sync>>,
+        SyncWrapper<
+            Pin<Box<dyn Stream<Item = Result<Bytes, Box<dyn StdError + Send + Sync>>> + Send>>,
+        >,
     ),
 }
 
@@ -156,12 +155,12 @@ impl Body {
     #[cfg(feature = "stream")]
     pub fn wrap_stream<S, O, E>(stream: S) -> Body
     where
-        S: Stream<Item = Result<O, E>> + Send + Sync + 'static,
+        S: Stream<Item = Result<O, E>> + Send + 'static,
         O: Into<Bytes> + 'static,
         E: Into<Box<dyn StdError + Send + Sync>> + 'static,
     {
         let mapped = stream.map_ok(Into::into).map_err(Into::into);
-        Body::new(Kind::Wrapped(Box::pin(mapped)))
+        Body::new(Kind::Wrapped(SyncWrapper::new(Box::pin(mapped))))
     }
 
     /// Converts this `Body` into a `Future` of a pending HTTP upgrade.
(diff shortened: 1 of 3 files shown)


rkuhn:
@seanmonstar This is the follow-up to #2159, WDYT?

rkuhn:
Well, now that it is proven, the alternative is to just `unsafe Impl Sync` (which could break later) or use the SyncWrapper from that rust-internals thread (that should compile to nothing). Which one would you prefer?

rkuhn:
I decided to take a shot at the SyncWrapper: https://github.com/rust-lang/rust/pull/71529

seanmonstar:
We could have a `SyncWrapper` internal to hyper here. Along with comments reminding us what it's sane/safe.

rkuhn:
@seanmonstar following some discussion on zulip I published [sync_wrapper](https://crates.io/crates/sync_wrapper) and updated this PR — is this okay?

rkuhn:
Hmm, the test failure is not related to this change AFAICS, and I also cannot reproduce it when running the same command locally …

seanmonstar:
Could we inline the type into the module? It seems small enough that we don't need to add the cost of downloading/compiling another dependency...

rkuhn:
We sure could do that, but if we kept the docs then the increase in size of `hyper` wouldn’t be much less than the 6kB taken by sync_wrapper (compilation time basically not measurable — 150ms by itself for release). I’ll definitely use SyncWrapper for other purposes as well; let me know what you think.

seanmonstar:
The type seems small enough to me to inline. There's costs to having a separate crate, some real (new network calls to download metadata and source, having to link libraries), and perceived (users get grumpy as the dependency count gets bigger).

rkuhn:
Okay, will probably do it tomorrow.

rkuhn:
I pushed the version which has the SyncWrapper in `common`.

lnicola:
Note that this was a breaking change (see the PR linked above).

seanmonstar:
How is this a breaking change? We've made the bounds less strict, so everything that matched before should match still.

lnicola:
I think it can break code that exposes those bounds:

```rust
fn foo() -> impl Sync + Send { } // remove Send

// client code
fn bar() -> impl Sync + Send { foo() }
```

seanmonstar:
Sure, that would be a breaking change, but that's not what we did. We changed `fn wrap_stream<T: Blah + Send + Sync>() - >Body`  to `fn wrap_stream<T: Blah + Send> -> Body`. The `Body` type is still `Sync`. Nothing *should have* broken.

lnicola:
Right, I had to take another look. The issue is that that this impl changed: https://github.com/hyperium/hyper/pull/2187/files#diff-b032ea26c55c1f4f1c2c32cd030e0bc0L405.

rkuhn:
Hmm, the `From` instance now applies to more inputs, how can that be a problem?

rkuhn:
Or asked differently: can you show some compiler error demonstrating what you mean by “breaking change”? That would make it easier for me to grasp.

lnicola:
```
error[E0277]: the trait bound `hyper::Body: std::convert::From<std::boxed::Box<(dyn futures::Stream<Item = std::result::Result<_, _>> + std::marker::Send + std::marker::Sync + 'static)>>` is not satisfied
   --> src/main.rs:528:21
    |
528 |                     http_serve::serve(crf, &request)
    |                     ^^^^^^^^^^^^^^^^^ the trait `std::convert::From<std::boxed::Box<(dyn futures::Stream<Item = std::result::Result<_, _>> + std::marker::Send + std::marker::Sync + 'static)>>` is not implemented for `hyper::Body`
    | 
   ::: /home/grayshade/.cargo/registry/src/github.com-1ecc6299db9ec823/http-serve-0.2.2/src/serving.rs:80:15
    |
80  |     B: Body + From<Box<dyn Stream<Item = Result<Ent::Data, Ent::Error>> + Send + Sync>>,
    |               ------------------------------------------------------------------------- required by this bound in `http_serve::serve`
    |
    = help: the following implementations were found:
              <hyper::Body as std::convert::From<&'static [u8]>>
              <hyper::Body as std::convert::From<&'static str>>
              <hyper::Body as std::convert::From<bytes::Bytes>>
              <hyper::Body as std::convert::From<std::borrow::Cow<'static, [u8]>>>
            and 4 others
```

rkuhn:
Oh, this is a limitation of the Rust type system that I wasn’t aware of: the actually implemented From instance covers the needs of the one you ask for, so it should be fine, but rustc doesn’t seem to understand that.

rkuhn:
Out of interest: why is this second part of the bound present? Shouldn’t `B: Body` allow you to do everything needed?

lnicola:
I'm not sure, but it doesn't compile without them :sweat_smile:. That crate is rather generics-heavy, and it's not impossible that some bounds are unnecessary.

PR #2188, opened 2020-04-19
refactor: use HttpBody with extra bounds instead of Payload trait

djc:
I've done most of the legwork, but I'm confused about the interaction of the error types that the compiler is complaining about at this point. If someone else wants to take a look, that'd be great.


djc:
Fixed up the remaining issues.

seanmonstar:
Looks great, thanks! I noticed the error types grew some extra `+ Send` bounds, so I've removed those in #2202.

djc:
Cool, thanks!

PR #2204, opened 2020-05-16
feat(lib): Move from `log` to `tracing` in a backwards-compatible way

davidbarsky:
I've moved Hyper from `log` to `tracing`. Existing `log`-based users shouldn't notice a difference, but `tracing` users will see higher performance when filtering data. This isn't the _end_  of the `tracing` integration that can happen in `Hyper` (e.g., Hyper can start using spans, typed fields, etc.), but _something_ is better than nothing. I'd rather address those points, including examples, in followups.

I've attached a screenshot of the `hello` example working, but the logged information is pulled from `tracing`, not `log`.

Screen Shot 2020-05-16 at 1 23 19 PM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ http-body = "0.3.1"
 httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
-tracing = { version = "0.1", default-features = false, features = ["std"] }
+tracing = { version = "0.1", default-features = false, features = ["std", "log"] }
 pin-project = "0.4"
 time = "0.1"
 tower-service = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
-pin-project = "0.4"
+pin-project = "0.4.20"
 time = "0.1"
 tower-service = "0.3"
 tokio = { version = "0.2.5", features = ["sync"] }
@@ -82,7 +82,6 @@ __internal_happy_eyeballs_tests = []
 features = [
     "runtime",
     "stream",
-    "log"
 ]
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,8 +58,7 @@ url = "1.0"
 [features]
 default = [
     "runtime",
-    "stream",
-    "log"
+    "stream"
 ]
 runtime = [
     "tcp",
@@ -70,9 +69,6 @@ tcp = [
     "tokio/blocking",
     "tokio/tcp",
     "tokio/time",
-]
-log = [
-    "tracing/log",
 ]
 
 # `impl Stream` for things
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ http-body = "0.3.1"
 httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
-tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
+tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project = "0.4"
 time = "0.1"
 tower-service = "0.3"
@@ -59,6 +59,7 @@ url = "1.0"
 default = [
     "runtime",
     "stream",
+    "log"
 ]
 runtime = [
     "tcp",
@@ -69,6 +70,10 @@ tcp = [
     "tokio/blocking",
     "tokio/tcp",
     "tokio/time",
+]
+log = [
+    "tracing/log",
+    "tracing/std"
 ]
 
 # `impl Stream` for things
@@ -82,6 +87,7 @@ __internal_happy_eyeballs_tests = []
 features = [
     "runtime",
     "stream",
+    "log"
 ]
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ http-body = "0.3.1"
 httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
-tracing = { version = "0.1", default-features = false, features = ["std", "log"] }
+tracing = { version = "0.1", features = ["log"] }
 pin-project = "0.4"
 time = "0.1"
 tower-service = "0.3"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -29,7 +29,7 @@ http-body = "0.3.1"
 httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
-log = "0.4"
+tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
```
hawkw, line 32: It would be good to expose the `log` feature optionally (though it should default to on so we don't break existing code that expects `log` records emitted by Hyper), so that `tracing` users who don't need `log` can disable it. Otherwise, depending on Hyper will _always_ enable the `log` feature, even when users don't need it.

davidbarsky: hmm, solid point. will fix.

seanmonstar: Something to consider, though, is that if the logging enabled becomes an optional feature, then those who have `default-features = false` will have the logs disabled when they upgrade. 

(I find cargo's default-features to be annoying in this regard, but what we gonna do.)

davidbarsky: Darn, that's an annoying thing. Are you okay with breaking those people in a minor release? Or more generally, how _should_ this be handled?

Cargo.toml
```diff
@@ -70,6 +71,10 @@ tcp = [
     "tokio/tcp",
     "tokio/time",
 ]
+log = [
+    "tracing/log",
+    "tracing/std"
```
hawkw, line 76: i suspect that `tracing/std` can always be on, but i don't actually think _hyper_ needs any of the functionality behind that feature, so we really don't care if it's enabled or not

seanmonstar:
As an update, I'm personally fine with proceeding here, as long as we don't break people who already have `default-features = false`. We can't make it so a a simple `cargo update` means their logs suddenly disappear.

Is having the `tracing/log` feature *required on* a deal breaker (at least until the next breaking release)?

hawkw:
> Is having the `tracing/log` feature _required on_ a deal breaker (at least until the next breaking release)?

In my book, it's totally fine to always the `tracing/log` feature, now that it no longer breaks the build if the `tracing` import is renamed (we fixed the macro issue that caused this in tokio-rs/tracing#740). 

As long as Hyper doesn't also enable the `tracing/log-always` feature, the logs will still be disabled by default for users who have set a `tracing` subscriber, so they won't have to pay the performance cost of filtering out duplicate `log` records. When no subscriber has been set, the log records will still be emitted normally. And, users who are using both `tracing` _and_ `log` and want to emit both can enable `tracing/log-always`.

The only added cost of the `tracing/log` feature now is that the `log` crate will always be compiled in, but it's widespread enough that I think it's probably pretty hard to avoid. In the future, it would be nice if users could opt out of it entirely if they are using `tracing` exclusively, but I don't think that's a dealbreaker.

davidbarsky:
Personally, it’s not a dealbreaker. Most users of `tracing` already have `log` somewhere in their dependency graph.

Cargo.toml
davidbarsky: I think it might be off is `default-features` is set to false; I suspect that pulling in syn + quote via instrument might be a bigger issue for hyper. In either case, I don't care _too_ much.

davidbarsky:
I've rebased and explicitly opted-in into `log` in `tracing`. I've also opted into the `std` feature to enable recording errors as values, but still set `default-values = false` to avoid pulling in `syn` if its unneeded.

Cargo.toml
```diff
@@ -59,6 +59,7 @@ url = "1.0"
 default = [
     "runtime",
     "stream",
+    "log"
```
seanmonstar, line 62: This should be removed (and the `log = ["tracing/log"]` as well).

Cargo.toml
```diff
@@ -29,7 +29,7 @@ http-body = "0.3.1"
 httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
-tracing = { version = "0.1", default-features = false, features = ["std", "log"] }
+tracing = { version = "0.1", features = ["log"] }
```
hawkw, line 32: if we're not using attributes, i think it's better to leave it off --- this means that if users opt out of it, it will still be pulled.

Cargo.toml
davidbarsky: should be resolved

Cargo.toml
seanmonstar: What do you mean? Doesn't it need to be kept so that users who were already using some `log`ger will still get logs?

davidbarsky: fixed!

davidbarsky: they'll keep getting logs, but they won't get https://docs.rs/tracing/0.1.15/tracing/attr.instrument.html.

davidbarsky: Or rather, _Hyper_ won't be pulling in that macro.

Cargo.toml
```diff
@@ -29,8 +29,8 @@ http-body = "0.3.1"
 httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
-log = "0.4"
-pin-project = "0.4.20"
+tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
+pin-project = "0.4"
```
seanmonstar, line 33: This needs to stay 0.4.20

Cargo.toml
```diff
@@ -82,6 +82,7 @@ __internal_happy_eyeballs_tests = []
 features = [
     "runtime",
     "stream",
+    "log"
```
seanmonstar, line 85: Can be removed.

seanmonstar:
Let me just confirm here, this is the dependency line that should be merged?

```toml
tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
```

The default-features should be disabled, and `std` should be enabled?

davidbarsky:
> The default-features should be disabled, and std should be enabled?

Yup! Conversation I had with Eliza:

```
endsofthreads: just to be clear on hyper: `default-features = false, features = ["std", "log"]`, right
mycoliza: should work
```

Sorry about the back and forth!

Cargo.toml
davidbarsky: Removed.

seanmonstar:
Thanks David!

hawkw:
Yay! 🙌

Issue #2205, opened 2020-05-18
net2 has been deprecated, use socket2 instead


Shadow53:
A GitHub action using RUSTSEC recently created an issue with the above description on a project I work on, and I thought I would inform the dependencies that use `net2`. Here's a link to the [related advisory](https://rustsec.org/advisories/RUSTSEC-2020-0016.html).

seanmonstar:
We can certainly upgrade (I don't quite understand why it's that relevant to rust-sec :shrug:).

Shadow53:
I'm not very familiar with rust-sec, but maybe the authors requested it because it's fairly commonly used? :man_shrugging: Glad you're aware of the deprecation, though.

jyn514:
I can help work on this.

PR #2206, opened 2020-05-18
Switch from net2 to socket2

jyn514:
Closes https://github.com/hyperium/hyper/issues/2205


jyn514:
The doc failure doesn't seem related to this change, maybe it was broken in a recent nightly?

PR #2211, opened 2020-05-28
feat(http2): allow configuring the HTTP/2 frame size

goffrie:
The default of 16K is taken from h2.

PR #2214, opened 2020-06-02
feat(http2): configure HTTP/2 frame size in the high-level builders too

goffrie:
Oops, I missed this in #2211.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -342,6 +342,16 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Sets the maximum frame size to use for HTTP2.
+    ///
+    /// Passing `None` will do nothing.
+    ///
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -342,6 +342,16 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Sets the maximum frame size to use for HTTP2.
+    ///
+    /// Passing `None` will do nothing.
+    ///
+    /// If not set, hyper will use a default.
+    pub fn http2_max_frame_size(&mut self, sz: impl Into<Option<u32>>) -> &mut Self {
```
seanmonstar, line 350: This builder uses by-value style.

goffrie: Oops, fixed.

PR #2216, opened 2020-06-03
fix(server): skip automatic Content-Length headers when not allowed

### LINKED ISSUE #2215: Hyper should skip automatic Content-Length header for HTTP 1xx responses
Similar to issue #1797 hyper should not set a content-length header for a number of different scenario's. Quoting [RFC 7230, section 3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2) 6th paragraph: 

> A server MUST NOT send a Content-Length header field in any response
> with a status code of 1xx (Informational) or 204 (No Content).  A
> server MUST NOT send a Content-Length header field in any 2xx
> (Successful) response to a CONNECT request (Section 4.3.6 of
> [RFC7231]).

At this moment the content-length header is set for 101 Switch Protocols responses. The resulting issue is that e.g. with a warp WebSocket server, a .Net client will drop the connection after 100 seconds: https://stackoverflow.com/a/44553521/85514 . 

I believe the solution should be similar to [b342c38f], and I'm willing to make a PR to do that. However, I'm not familiar enough with the code base to see if that would fix the problem / has any other implications.

psmit:
Closes #2215


psmit:
Note that some old tests fail at this moment. I'll fix those after getting feedback whether this is the right way to fix this issue.

seanmonstar:
It's probably somewhere in that code, for sure! I'd suggest starting by modifying this [test case](https://github.com/hyperium/hyper/blob/2354a7eec352b1f72cd8989d29d73dff211403a1/tests/server.rs#L1331) to check there is no `content-length` header, which should fail without your changes.

psmit:
Thanks for the hint. I adjusted 2 test cases and now the PR is ready for review.

tests/server.rs
```diff
@@ -498,7 +498,7 @@ fn head_response_doesnt_send_body() {
     let mut response = String::new();
     req.read_to_string(&mut response).unwrap();
 
-    assert!(response.contains("content-length: 11\r\n"));
+    assert!(!has_header(&response, "content-length"));
```
seanmonstar, line 501: I believe we want to keep this behavior. It's correct and expected that a `HEAD` request will be able to check what the `content-length` *would be* if sending a `GET`.

psmit: You are very correct. I have now reverted this test and made sure the right behavior is followed for HEAD requests.

psmit:
@seanmonstar  I have modified the code accordingly. Could you please review it again?

src/proto/h1/role.rs
```diff
@@ -511,7 +511,7 @@ impl Http1Transaction for Server {
             encoder = match msg.body {
                 Some(BodyLength::Unknown) => {
                     if msg.head.version == Version::HTTP_10
-                        || !Server::can_chunked(msg.req_method, msg.head.subject)
+                        || !Server::can_have_content_length(msg.req_method, msg.head.subject)
```
seanmonstar, line 514: I *think* this specific one should still be `can_chunked`, since if this is false, the `else` branch will set chunked. I'm a little worried that this change didn't trigger one of the response body tests...

krallin:
@psmit just curious: are you still interested in driving this? We're affected by this issue as well and would love to see the fix go in. Thanks! :)

psmit:
Hi @krallin, my apologies, I had not realized that this PR wasn't merged yet and your reminder got lost in my mail box. I'll be working now to get this PR finished.

src/proto/h1/role.rs
psmit: Looking to this again, I agree. I have changed it now and I think the PR is complete.

krallin:
Is anything blocking this PR?

psmit:
> Is anything blocking this PR?

Not to my knowledge, we just have to wait for the review to be done.

seanmonstar (approved):

seanmonstar:
Sorry, I got distracted. This looks good to me, let's go! Thanks for sticking with it <3

krallin:
@seanmonstar  any plans to cut a new release? We'd love to stop having to use a fork of Hyper just to get this :)

seanmonstar:
I was hoping to get #2473 merged today or tomorrow, and then do a release. If it gets delayed, I'll release anyways. So either way, there should be a 0.14.5 today or tomorrow.

krallin:
awesome, thank you! :)

Issue #1875, opened 2019-07-17
Unreachable code reached in server with invalid Content-Length header


jethrogb:
This test doesn't pass on 0.10.x:
```patch
diff --git a/src/server/request.rs b/src/server/request.rs
index c632d27..2c05e72 100644
--- a/src/server/request.rs
+++ b/src/server/request.rs
@@ -304,4 +304,22 @@ mod tests {
         assert_eq!(read_to_string(req).unwrap(), "1".to_owned());
     }
 
+    /// Tests that when an invalid Content-Length header doesn't crash the
+    /// server.
+    #[test]
+    fn test_invalid_content_length() {
+        let mut mock = MockStream::with_input(b"\
+            POST / HTTP/1.1\r\n\
+            Host: example.domain\r\n\
+            Content-Length: x\r\n\
+            \r\n\
+            I'm a bad request.\r\n\
+        ");
+
+        // FIXME: Use Type ascription
+        let mock: &mut NetworkStream = &mut mock;
+        let mut stream = BufReader::new(mock);
+
+        let _ = Request::new(&mut stream, sock("127.0.0.1:80"));
+    }
 }
```
Apparently, older versions of cURL do this.

seanmonstar:
That's unfortunate, I don't remember why at all. 

What server is running this? hyper 0.10 ended 2 years ago, I'm sure there's other bugs floating in there too ;_;

jethrogb:
> hyper 0.10 ended 2 years ago

What do you mean by this? There was a release in April.

seanmonstar:
Ah, so there was! I meant 0.11 came out 2 years ago (and 0.12 a year ago). I don't have any time to maintain 0.10 myself. I can publish small bug fixes that are sent.

PR #2217, opened 2020-06-05
Avoid panic when Content-Length is invalid

mzohreva:
This would fix issue #1875 for 0.10 branch.

cc @jethrogb


seanmonstar:
Thanks for the PR! Thinking more about it, I'm probably not going to publish more releases for 0.10.x. It's several years since it stopped seeing support, and being sure that a new release works properly is not something I have time for.

jethrogb:
How does April 2019 qualify as "several years"?

PR #2220, opened 2020-06-06
refactor(lib): remove pin related unsafe code

taiki-e:


src/service/oneshot.rs
```diff
@@ -20,49 +18,43 @@ where
...
+                    return fut.poll(cx);
                 }
-                State::Tmp => unreachable!(),
+                StateProj::Tmp => unreachable!(),
             }
 
-            match mem::replace(&mut me.state, State::Tmp) {
-                State::NotReady(mut svc, req) => {
-                    me.state = State::Called(svc.call(req));
+            match me.state.as_mut().project_replace(State::Tmp) {
```
taiki-e, line 55: [`project_replace`](https://docs.rs/pin-project/0.4/pin_project/attr.pin_project.html#project_replace) is an optional method recently added.

> It is equivalent to `Pin::set`, except that the unpinned fields are moved and returned, instead of being dropped in-place.
> 
> ```rust
> fn project_replace(self: Pin<&mut Self>, other: Self) -> ProjectionOwned;
> ```
>
> The `ProjectionOwned` type is identical to the `Self` type, except that all pinned fields have been replaced by equivalent `PhantomData` types.

src/service/oneshot.rs
```diff
@@ -20,49 +18,43 @@ where
 // is ready, and then calling `Service::call` with the request, and
 // waiting for that `Future`.
 #[allow(missing_debug_implementations)]
+#[pin_project]
 pub struct Oneshot<S: Service<Req>, Req> {
+    #[pin]
     state: State<S, Req>,
 }
 
+#[pin_project(Replace, project = StateProj, project_replace = StateProjOwn)]
```
taiki-e, line 27: It is not ideal that we need to have both `Replace` and `project_replace = <name>` as `project_replace = <name>` implies that a `project_replace` method is generated. 
I'll fix it soon.

seanmonstar: Do you mean I should wait to merge, or just some time later?

taiki-e: Oh sorry, [a fix for this was already merged](https://github.com/taiki-e/pin-project/pull/243), but I forgot to release a new version.

taiki-e: Updated to use a new version pin-project which includes a fix for this.

PR #2225, opened 2020-06-08
docs(examples): add example for integration testing a client/server

robinst:
I wrote a small library for a client that does some requests, and I
wanted to write an integration test for it (in `tests/`) and run against
a test server that will automatically shut down at the end of the test.

Being new to async hyper, it took me a while to figure out how to write
it. An example like this would have helped.

The same style of test can also be used if the code being tested is a
server.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -35,6 +35,8 @@ pretty_env_logger = "0.4"
 
 * [`http_proxy`](http_proxy.rs) - A simple HTTP(S) proxy that handle and upgrade `CONNECT` requests and then proxy data between client and remote server.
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the PR!

I'm kind of worried that the number of examples keeps growing, and that can cause confusion to a newcomer about which ones to look at. So, I'd lean towards leaving this one out.

robinst:
Ok. Hopefully people can find at least this PR if they need it and not lose about half a day like me trying to figure out how to set up tests.

svenstaro:
If it's still possible, I'd still very much like to see this merged. I think this shows a useful pattern.

Issue #2224, opened 2020-06-08
The HTTP proxy example doesn't work.


Jason5Lee:
When I run the that example and try to `all_proxy=127.0.0.1:8080 curl https://www.domain.com/` , it outputs `curl: (56) Received HTTP code 400 from proxy after CONNECT` , and the proxy server logs `CONNECT host is not socket addr: www.domain.com:443`

anderson916:
Same here. 

`
curl -i https://www.google.com
HTTP/1.1 400 Bad Request
content-length: 35
date: Fri, 19 Jun 2020 01:39:49 GMT

curl: (56) Received HTTP code 400 from proxy after CONNECT
`
http proxy stdout:

`
req: Request { method: CONNECT, uri: www.google.com:443, version: HTTP/1.1, headers: {"host": "www.google.com:443", "user-agent": "curl/7.64.1", "proxy-connection": "Keep-Alive"}, body: Body(Empty) }
CONNECT host is not socket addr: www.google.com:443
`

seanmonstar:
The error message gives a hint: this simple proxy requires the `host` header to be a socket address.

shirshak55:
You can use trust dns or std library socket addr or Hyper GaiResolver too?

PR #2226, opened 2020-06-08
Fix http proxy example

Jason5Lee:
Related issue: [#2224](https://github.com/hyperium/hyper/issues/2224).


seanmonstar:
This fails since the DNS feature of tokio isn't enabled.

Jason5Lee:
> This fails since the DNS feature of tokio isn't enabled.

I have `features = ["full"]` set, and by adding dns feature to it, the problem still exists.

seanmonstar:
Closing since it seems to have stalled.

PR #2227, opened 2020-06-09
Allow passwords with colon in Basic Authorization headers

mzohreva:
From [RFC 7617](https://tools.ietf.org/html/rfc7617) section 2:
> the first colon in a user-pass string separates user-id and password
> from one another; text after the first colon is part of the password.

cc @jethrogb


seanmonstar:
Thanks for the PR! Same as in #2217, I don't expect to push out any more 0.10.x releases.

PR #2229, opened 2020-06-11
fix(client): don't panic in DNS resolution when task cancelled

dae:
If the runtime is dropped while a DNS request is being made, it can
lead to a panic. This patch checks if the task was cancelled, and
returns a generic IO error instead of panicking in that case.

The following code reproduces the problem on my macOS machine:

```
use hyper::Client;
use tokio::runtime;

type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>>;

fn main() {
    let rt = runtime::Builder::new()
        .threaded_scheduler()
        .core_threads(1)
        .enable_all()
        .build()
        .unwrap();

    // spawn a request and then drop the runtime immediately
    rt.spawn(fetch_url());
}

async fn fetch_url() -> Result<()> {
    let url: hyper::Uri = "http://example.com".parse()?;
    let client = Client::builder().build_http::<hyper::Body>();

    let res = client.get(url).await?;
    println!("Response: {}", res.status());

    Ok(())
}
```

If you think this should be handled differently, please let me know.

### DIFF
--- a/src/client/connect/dns.rs
+++ b/src/client/connect/dns.rs
@@ -141,7 +141,13 @@ impl Future for GaiFuture {
         Pin::new(&mut self.inner).poll(cx).map(|res| match res {
             Ok(Ok(addrs)) => Ok(GaiAddrs { inner: addrs }),
             Ok(Err(err)) => Err(err),
-            Err(join_err) => panic!("gai background task failed: {:?}", join_err),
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks, very clear explanation!

PR #2235, opened 2020-06-23
Some simplifications

Marwes:
Using `ready!` in a few places, and avoids getting a future only to poll it manually.


Marwes:
Gitcop offers no explanation for why the commit message is invalid :S

seanmonstar:
Thanks!

Issue #2178, opened 2020-04-09
Expose an `AddOrigin` service


LucioFranco:
Its common in more lower level service configurations that we need to add an origin to each request. We should provide this utility since we already have all the types imported.

seanmonstar:
Sure, this could be `hyper::service::AddOrigin` (`WithOrigin`?).

mayankshah1607:
Hi! I'd love to take up this issue. This is the first time I'm contributing to a Rust project, so I'd really appreciate any form of help. Thanks :)

mayankshah1607:
Hey @seanmonstar @LucioFranco 
I looked into the code, but I just have a quick (or rather noob) question - where and how exactly is this utility to be implemented? Do we have to implement this one as another trait, or a helper function to set the origin header on `Request`?
Thanks :)

gokulchandra:
First version of what a potential solution for this could look like: https://github.com/hyperium/hyper/pull/2242

gokulchandra:
@seanmonstar I have updated the solution for this.. #2256 .. Keen to get your feedback.

seanmonstar:
In the back of my mind, I was thinking that we want to start pruning things from hyper to prepare for 1.0, and this addition doesn't seem essential enough to add. That's different from what we were thinking last year when the issue was written. I believe something like this could easily exist, but probably part of the `tower` set of crates, instead of hyper.

PR #2242, opened 2020-07-10
[WIP] feat(client): add origin config to client builder

gokulchandra:
Took an attempt to provide a solution for #2178 .

I went with the assumption that we'd only want to set when Hyper is used as a Http client.
The `Client::Builder` now has an additional config to accept a value for setting the `Origin` header. Once configured it gets added for every request made.

```
let client = Client::builder()
            .add_origin("http://secret-mission".parse().unwrap())
            .build()
```
Let me know if you think this makes sense. Can iterate on it as required.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -56,7 +56,7 @@ use std::time::Duration;
 use futures_channel::oneshot;
 use futures_util::future::{self, Either, FutureExt as _, TryFutureExt as _};
 use http::header::{HeaderValue, HOST, ORIGIN};
-use http::uri::{Scheme};
+use http::uri::Scheme;
 use http::{Method, Request, Response, Uri, Version};
 
 use self::connect::{sealed::Connect, Alpn, Connected, Connection};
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the PR!

The issue #2178 may not have been super clear. It was meant that we should provide some `hyper::service::AddOrigin` middleware. Something that looks like this (I'm going to skip some imports and things):

```rust
pub struct AddOrigin<S> {
    scheme: http::uri::Scheme,
    auth: http::uri::Authority,
    service: S,
}

impl<Req, S: Service<Req> Service<Req> for AddOrigin {
    // ...
}
```

It's similar to what was done in this commit: https://github.com/tower-rs/tower-http/commit/51cef91f758c1aae672876c2ac1dde93e47e63aa

cc @LucioFranco

gokulchandra:
Ah okay. I misunderstood that. So the intention is to provide a `AddOrigin` middleware agnostic of whether the library is used as a client or server. 

Thanks for the reference, let me update this PR.

seanmonstar:
I'm going to close this PR for now, but feel free to open if you want to make the changes!

gokulchandra:
Fair enough. Will raise a PR once I am done with the changes.

PR #2256, opened 2020-07-29
feat(lib): provide add_origin middleware.

### LINKED ISSUE #2178: Expose an `AddOrigin` service

gokulchandra:
closes #2178

Provides a middleware `add_origin` that sets the configured `Authority` and `Scheme` value to be sent down to all inner services. 

@seanmonstar Let me know if you think this makes sense. Can update as required.

### DIFF
--- a/src/service/add_origin.rs
+++ b/src/service/add_origin.rs
@@ -1,0 +1,60 @@
+use super::Service;
+use crate::common::{task, Poll};
+use crate::{Request, Uri};
+use http::uri::{Authority, Parts, Scheme};
+
+/// Wraps an HTTP service, injecting authority and scheme on every request.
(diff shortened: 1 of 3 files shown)


gokulchandra:
@seanmonstar Could you have look at this PR? Happy to update is requested..

seanmonstar:
Sorry for the slow review. In the back of my mind, I was thinking that we want to start pruning things from hyper to prepare for 1.0, and this addition doesn't seem essential enough to add. That's different from what we were thinking last year when the issue was written. I believe something like this could easily exist, but probably part of the `tower` set of crates, instead of hyper.

PR #2261, opened 2020-08-04
feat(client): propagate tracing spans

hawkw:
Hyper's clients drive the connection to the server in a background task,
and dispatch requests to it from the `Client` handle types using
channels. This means that the actual work of sending a request and
receiving a response is _not_ performed on the same task as where the
user calls `Client::request`

When `tracing` is used in user code, it can be valuable to see the
`tracing` span context where the call to `Client::request` was made in
Hyper's debug output as well. This would allow associating proto errors
that Hyper logs with the individual request...and the semantic context
in the user's application where that request was made. However, because
the actual work of making the request takes place in a separate task,
the `tracing` span context is not propagated by default.

This branch fixes this by capturing the span context when a request is
made, and propagating it over the channel to be entered by the
background task.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ httparse = "1.0"
 h2 = "0.2.2"
 itoa = "0.4.1"
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
+tracing-futures = { version = "0.2", default-features = false, features = ["std-future"]}
 pin-project = "0.4.20"
 tower-service = "0.3"
 tokio = { version = "0.2.11", features = ["sync"] }
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -1,5 +1,6 @@
 use futures_util::future;
 use tokio::sync::{mpsc, oneshot};
+use tracing::Span;
 
 use crate::common::{task, Future, Pin, Poll};
 
@@ -80,7 +81,7 @@ impl<T, U> Sender<T, U> {
         }
         let (tx, rx) = oneshot::channel();
         self.inner
-            .send(Envelope(Some((val, Callback::Retry(tx)))))
+            .send(Envelope(Some((val, Callback::Retry(tx), Span::current()))))
             .map(move |_| rx)
             .map_err(|mut e| (e.0).0.take().expect("envelope not dropped").0)
     }
@@ -91,7 +92,11 @@ impl<T, U> Sender<T, U> {
         }
         let (tx, rx) = oneshot::channel();
         self.inner
-            .send(Envelope(Some((val, Callback::NoRetry(tx)))))
+            .send(Envelope(Some((
+                val,
+                Callback::NoRetry(tx),
+                Span::current(),
+            ))))
             .map(move |_| rx)
             .map_err(|mut e| (e.0).0.take().expect("envelope not dropped").0)
     }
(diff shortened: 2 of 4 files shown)


seanmonstar:
To be sure I understand, at the moment, this just helps users who are making spans *outside* hyper, to see the child spans internally. This is because hyper isn't yet making any spans inside `Client::request` (though I expect it should soon).

hawkw:
> To be sure I understand, at the moment, this just helps users who are making spans _outside_ hyper, to see the child spans internally. This is because hyper isn't yet making any spans inside `Client::request` (though I expect it should soon).

Yup, that's correct. I was starting to add some spans inside of Hyper's internals, but realized that this was the first step to actually emitting our own.

seanmonstar:
I was trying this out by instrumenting the client example's `fetch_url` future with a span, but didn't notice any internal events being part of the span... Maybe I'm misunderstanding what this does...

hawkw:
> I was trying this out by instrumenting the client example's `fetch_url` future with a span, but didn't notice any internal events being part of the span... Maybe I'm misunderstanding what this does...

Hmm, I got it to work. Although it's not propagated to the connection task yet; perhaps that should have its own span.
Screenshot_20200805_152920
with the following change to the example:
```patch
diff --git a/Cargo.toml b/Cargo.toml
index d87359a..1f4468a 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,8 @@ tokio = { version = "0.2.2", features = ["fs", "macros", "io-std", "rt-util", "s
 tokio-test = "0.2"
 tokio-util = { version = "0.3", features = ["codec"] }
 tower-util = "0.3"
+tracing = { version = "0.1", default-features = false, features = ["log", "std", "attributes"] }
+tracing-subscriber = "0.2"
 url = "1.0"

 [features]
diff --git a/examples/client.rs b/examples/client.rs
index 2333746..654e29a 100644
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -10,7 +10,7 @@ type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>

 #[tokio::main]
 async fn main() -> Result<()> {
-    pretty_env_logger::init();
+    tracing_subscriber::fmt::init();

     // Some simple CLI args requirements...
     let url = match env::args().nth(1) {
@@ -32,13 +32,14 @@ async fn main() -> Result<()> {
     fetch_url(url).await
 }

+#[tracing::instrument]
 async fn fetch_url(url: hyper::Uri) -> Result<()> {
     let client = Client::new();

     let mut res = client.get(url).await?;

-    println!("Response: {}", res.status());
-    println!("Headers: {:#?}\n", res.headers());
+    tracing::info!("Response: {}", res.status());
+    tracing::info!("Headers: {:#?}\n", res.headers());

     // Stream the body, writing each chunk to stdout as we get it
     // (instead of buffering and printing at the end).
@@ -46,8 +47,8 @@ async fn fetch_url(url: hyper::Uri) -> Result<()> {
         let chunk = next?;
         io::stdout().write_all(&chunk).await?;
     }
-
-    println!("\n\nDone!");
+    println!("\n");
+    tracing::info!("Done!");

     Ok(())
 }
```

seanmonstar:
Yea, I see that too. I thought this was hooking up the request span with the connection, seems I misunderstood.

seanmonstar:
This is desirable, but doesn't seem to work quite as expected in this PR. Would welcome trying out in a new PR.

PR #2262, opened 2020-08-05
Add spans for parse_headers and encode_headers

seanmonstar:
Add `trace` level spans to record time spent parsing and encoding headers.

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -474,7 +474,7 @@ where
         self.enforce_version(&mut head);
 
         let buf = self.io.headers_buf();
-        match T::encode(
+        match super::role::encode_headers::<T>(
             Encode {
                 head: &mut head,
                 body,
--- a/src/proto/h1/io.rs
+++ b/src/proto/h1/io.rs
@@ -149,7 +149,7 @@ where
         S: Http1Transaction,
     {
         loop {
-            match S::parse(
+            match super::role::parse_headers::<S>(
                 &mut self.read_buf,
                 ParseContext {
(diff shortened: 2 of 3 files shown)


src/proto/h1/mod.rs
```diff
@@ -93,3 +93,26 @@ impl Wants {
...
+    let span = trace_span!("parse_headers");
+    let _s = span.enter();
+    T::parse(bytes, ctx)
+}
+
+fn encode_headers<T>(enc: Encode<'_, T::Outgoing>, dst: &mut Vec<u8>) -> crate::Result<Encoder>
+where
+    T: Http1Transaction,
+{
+    let span = trace_span!("encode_headers");
```
hawkw, line 115: TIOLI: we could mimic the existing `{role=server}`/`{role=client}` logging in a more structured way like this:
```suggestion
    let span = trace_span!("parse_headers", role = %std::any::type_name::<T>());
    let _s = span.enter();
    T::parse(bytes, ctx)
}

fn encode_headers<T>(enc: Encode<'_, T::Outgoing>, dst: &mut Vec<u8>) -> crate::Result<Encoder>
where
    T: Http1Transaction,
{
    let span = trace_span!("encode_headers", role = %std::any::type_name::<T>());
```

seanmonstar: The role stuff was from `log` days. I think having a top level with the role would be the right way, now, ya?

hawkw: Yeah, that would be ideal --- I thought this might be worth adding for now until a higher-level span is added?

PR #2264, opened 2020-08-10
fix(server): update proto::h1::end_body to return Result<()>

jxs:


src/proto/h1/conn.rs
```diff
@@ -601,13 +602,14 @@ where
                             Writing::KeepAlive
                         }
                     }
-                    Err(_not_eof) => Writing::Closed,
+                    Err(_not_eof) => return Err(crate::Error::new_closed()),
```
seanmonstar, line 605: I'd probably make this something like `Error::new_body_write_aborted()`.

seanmonstar:
Ok, so I understand what is happening with the failing test now. The problem is that in some states, the response needs to close the connection when the body is finished (close-delimited). This is when there is no `content-length` and no `transfer-encoding`. In that case, it's actually correct to just "close" the writing side, instead of dying in an error.

So, it seems we'd want to identify between "closing the body at this point is an actual error" vs "closing the body just means the socket should close cleanly".

seanmonstar:
To differentiate, we could return an `enum` that could be matched in the `_not_eof` case there.

jxs:
updated `Encoder::end` to not return `Err` on `CloseDelimited`, only return `Err` when there's `content-length` left.
Updated `Encoder` eof tests, to match the behavior you described above: when there is no content-length and no transfer-encoding it's correct to close the writing side, so `Encoder::end()` should not return an `Err`

src/proto/h1/conn.rs
```diff
@@ -601,13 +602,18 @@ where
                             Writing::KeepAlive
                         }
                     }
-                    Err(_not_eof) => Writing::Closed,
+                    Err(_not_eof) => {
+                        return Err(crate::Error::new_user_body(
```
seanmonstar, line 606: Why not just `new_body_write_aborted()`?

jxs: to get the same `Error` as when user calls `tx.abort()`. with `Err(crate::Error::new_body_write_aborted())`  `Display` is: `body write aborted`, when combining both `Display` is `error from user's HttpBody stream: body write aborted`.
Want me to leave only `new_body_write_aborted()`?

seanmonstar: Oh, now I see it, my bad. This is right!

PR #2269, opened 2020-08-13
docs(server): example update

dujar:
missing from https://hyper.rs/guides/server/echo/

otherwise compilation error:

use futures_util::TryStreamExt;
| ^^^^^^^^^^^^ use of undeclared type or module futures_util


seanmonstar:
Thanks for the PR! I don't believe this is necessary, as it is used lower down in the file (the `map_ok`).

Issue #2272, opened 2020-08-19
Add example to hyper::client::conn module docs


seanmonstar:
The documentation for `hyper::client::conn` is rather sparse. It would be helpful to give an overview of how to use the types in that module, by adding a simple example. The example could mention "after getting some transport", use `handshake`, spawn the `Connection` task, and then send a request.

PR #2276, opened 2020-08-27
docs(server): Docs for conn modules in server and client

sapessi:
[WIP] Adding doc examples for the lower-level `conn` modules in `server` and `client`.

Beginning to address #2272.


sapessi:
redo with the corrected commit messages

PR #2277, opened 2020-08-27
docs(server): Example for conn modules

sapessi:
Added a module-level, basic example for the `Http` struct in the `server::conn` module. This PR is a work in progress

Begins to address #2272

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -23,11 +23,15 @@
 //!     let mut tcp_listener = TcpListener::bind(addr).await?;
 //!     loop {
 //!         let (tcp_stream, _) = tcp_listener.accept().await?;
-//!         Http::new()
-//!             .http1_only(true)
-//!             .keep_alive(true)
-//!             .serve_connection(tcp_stream, service_fn(hello))
-//!             .await?;
+//!         tokio::task::spawn(async move {
+//!             if let Err(http_err) = Http::new()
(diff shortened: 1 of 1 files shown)


davidbarsky:
I think this is a good start, but I'd _maybe_ add an example that spawns the incoming connection on [a dedicated task](https://github.com/davidbarsky/stuff/blob/master/src/main.rs#L46-L73), but I'd leave that up to Sean.

src/server/conn.rs
```diff
@@ -7,6 +7,34 @@
...
+//! async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+//!     let addr: SocketAddr = ([127, 0, 0, 1], 8080).into();
+//!
+//!     let mut tcp_listener = TcpListener::bind(addr).await?;
+//!     loop {
+//!         let (tcp_stream, _) = tcp_listener.accept().await?;
+//!         Http::new()
+//!             .http1_only(true)
+//!             .keep_alive(true)
+//!             .serve_connection(tcp_stream, service_fn(hello))
```
seanmonstar, line 29: Yea, I'd suggest that this future be put in a `tokio::spawn`, since otherwise the loop could only handle 1 connection at a time.

sapessi: Updated example

seanmonstar:
Thanks!

PR #2278, opened 2020-09-09
Initial C API for hyper

seanmonstar:
This provides an initial C API for hyper, motivated by #2265. 

This starts with essentially the `hyper::client::conn` API exposed (server and client pool aren't yet). As part of it's merging, this is considered **unstable**.

### DIFF
--- a/capi/include/hyper.h
+++ b/capi/include/hyper.h
@@ -1,0 +1,554 @@
+#ifndef _HYPER_H
+#define _HYPER_H
+
+#include <stdint.h>
+#include <stddef.h>
+
+#define HYPER_ITER_CONTINUE 0
+
+#define HYPER_ITER_BREAK 1
+
+#define HYPER_HTTP_VERSION_NONE 0
+
+#define HYPER_HTTP_VERSION_1_0 10
+
+#define HYPER_HTTP_VERSION_1_1 11
+
+#define HYPER_HTTP_VERSION_2 20
+
+#define HYPER_IO_PENDING 4294967295
+
+#define HYPER_IO_ERROR 4294967294
+
+#define HYPER_POLL_READY 0
+
+#define HYPER_POLL_PENDING 1
+
+#define HYPER_POLL_ERROR 3
+
+typedef enum {
+  /*
+   All is well.
+   */
+  HYPERE_OK,
+  /*
+   General error, details in the `hyper_error *`.
+   */
+  HYPERE_ERROR,
+  /*
+   A function argument was invalid.
+   */
+  HYPERE_INVALID_ARG,
+  /*
+   The IO transport returned an EOF when one wasn't expected.
+
+   This typically means an HTTP request or response was expected, but the
+   connection closed cleanly without sending (all of) it.
+   */
+  HYPERE_UNEXPECTED_EOF,
+  /*
+   Aborted by a user supplied callback.
+   */
+  HYPERE_ABORTED_BY_CALLBACK,
+  /*
+   An optional hyper feature was not enabled.
+   */
+  HYPERE_FEATURE_NOT_ENABLED,
+} hyper_code;
+
+typedef enum {
+  /*
+   The value of this task is null (does not imply an error).
+   */
+  HYPER_TASK_EMPTY,
+  /*
+   The value of this task is `hyper_error *`.
+   */
+  HYPER_TASK_ERROR,
+  /*
+   The value of this task is `hyper_clientconn *`.
+   */
+  HYPER_TASK_CLIENTCONN,
+  /*
+   The value of this task is `hyper_response *`.
+   */
+  HYPER_TASK_RESPONSE,
+  /*
+   The value of this task is `hyper_buf *`.
+   */
+  HYPER_TASK_BUF,
+} hyper_task_return_type;
+
+typedef struct hyper_executor hyper_executor;
+
+typedef struct hyper_io hyper_io;
+
+typedef struct hyper_task hyper_task;
+
+typedef struct hyper_body hyper_body;
+
+typedef struct hyper_buf hyper_buf;
+
+typedef struct hyper_clientconn hyper_clientconn;
+
+typedef struct hyper_clientconn_options hyper_clientconn_options;
+
+typedef struct hyper_context hyper_context;
+
+typedef struct hyper_error hyper_error;
+
+typedef struct hyper_headers hyper_headers;
+
+typedef struct hyper_request hyper_request;
+
+typedef struct hyper_response hyper_response;
+
+typedef struct hyper_waker hyper_waker;
+
+typedef int (*hyper_body_foreach_callback)(void*, const hyper_buf*);
+
+typedef int (*hyper_body_data_callback)(void*, hyper_context*, hyper_buf**);
+
+typedef int (*hyper_headers_foreach_callback)(void*, const uint8_t*, size_t, const uint8_t*, size_t);
+
+typedef size_t (*hyper_io_read_callback)(void*, hyper_context*, uint8_t*, size_t);
+
+typedef size_t (*hyper_io_write_callback)(void*, hyper_context*, const uint8_t*, size_t);
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+/*
+ Returns a static ASCII (null terminated) string of the hyper version.
+ */
+const char *hyper_version(void);
+
+/*
+ Create a new "empty" body.
+
+ If not configured, this body acts as an empty payload.
+ */
+hyper_body *hyper_body_new(void);
+
+/*
+ Free a `hyper_body *`.
+ */
+void hyper_body_free(hyper_body *body);
+
+/*
+ Return a task that will poll the body for the next buffer of data.
+
+ The task value may have different types depending on the outcome:
+
+ - `HYPER_TASK_BUF`: Success, and more data was received.
+ - `HYPER_TASK_ERROR`: An error retrieving the data.
+ - `HYPER_TASK_EMPTY`: The body has finished streaming data.
+
+ This does not consume the `hyper_body *`, so it may be used to again.
+ However, it MUST NOT be used or freed until the related task completes.
+ */
+hyper_task *hyper_body_data(hyper_body *body);
+
+/*
+ Return a task that will poll the body and execute the callback with each
+ body chunk that is received.
+
+ The `hyper_buf` pointer is only a borrowed reference, it cannot live outside
+ the execution of the callback. You must make a copy to retain it.
+
+ The callback should return `HYPER_ITER_CONTINUE` to continue iterating
+ chunks as they are received, or `HYPER_ITER_BREAK` to cancel.
+
+ This will consume the `hyper_body *`, you shouldn't use it anymore or free it.
+ */
+hyper_task *hyper_body_foreach(hyper_body *body, hyper_body_foreach_callback func, void *userdata);
+
+/*
+ Set userdata on this body, which will be passed to callback functions.
+ */
+void hyper_body_set_userdata(hyper_body *body, void *userdata);
+
+/*
+ Set the data callback for this body.
+
+ The callback is called each time hyper needs to send more data for the
+ body. It is passed the value from `hyper_body_set_userdata`.
+
+ If there is data available, the `hyper_buf **` argument should be set
+ to a `hyper_buf *` containing the data, and `HYPER_POLL_READY` should
+ be returned.
+
+ Returning `HYPER_POLL_READY` while the `hyper_buf **` argument points
+ to `NULL` will indicate the body has completed all data.
+
+ If there is more data to send, but it isn't yet available, a
+ `hyper_waker` should be saved from the `hyper_context *` argument, and
+ `HYPER_POLL_PENDING` should be returned. You must wake the saved waker
+ to signal the task when data is available.
+
+ If some error has occurred, you can return `HYPER_POLL_ERROR` to abort
+ the body.
+ */
+void hyper_body_set_data_func(hyper_body *body, hyper_body_data_callback func);
+
+/*
+ Create a new `hyper_buf *` by copying the provided bytes.
+
+ This makes an owned copy of the bytes, so the `buf` argument can be
+ freed or changed afterwards.
+ */
+hyper_buf *hyper_buf_copy(const uint8_t *buf, size_t len);
+
+/*
+ Get a pointer to the bytes in this buffer.
+
+ This should be used in conjunction with `hyper_buf_len` to get the length
+ of the bytes data.
+
+ This pointer is borrowed data, and not valid once the `hyper_buf` is
+ consumed/freed.
+ */
+const uint8_t *hyper_buf_bytes(const hyper_buf *buf);
+
+/*
+ Get the length of the bytes this buffer contains.
+ */
+size_t hyper_buf_len(const hyper_buf *buf);
+
+/*
+ Free this buffer.
+ */
+void hyper_buf_free(hyper_buf *buf);
+
+/*
+ Starts an HTTP client connection handshake using the provided IO transport
+ and options.
+
+ Both the `io` and the `options` are consumed in this function call.
+
+ The returned `hyper_task *` must be polled with an executor until the
+ handshake completes, at which point the value can be taken.
+ */
+hyper_task *hyper_clientconn_handshake(hyper_io *io, hyper_clientconn_options *options);
+
+/*
+ Send a request on the client connection.
+
+ Returns a task that needs to be polled until it is ready. When ready, the
+ task yields a `hyper_response *`.
+ */
+hyper_task *hyper_clientconn_send(hyper_clientconn *conn, hyper_request *req);
+
+/*
+ Free a `hyper_clientconn *`.
+ */
+void hyper_clientconn_free(hyper_clientconn *conn);
+
+/*
+ Creates a new set of HTTP clientconn options to be used in a handshake.
+ */
+hyper_clientconn_options *hyper_clientconn_options_new(void);
+
+/*
+ Free a `hyper_clientconn_options *`.
+ */
+void hyper_clientconn_options_free(hyper_clientconn_options *opts);
+
+/*
+ Set the client background task executor.
+
+ This does not consume the `options` or the `exec`.
+ */
+void hyper_clientconn_options_exec(hyper_clientconn_options *opts, const hyper_executor *exec);
+
+/*
+ Set the whether to use HTTP2.
+
+ Pass `0` to disable, `1` to enable.
+ */
+hyper_code hyper_clientconn_options_http2(hyper_clientconn_options *opts, int enabled);
+
+/*
+ Frees a `hyper_error`.
+ */
+void hyper_error_free(hyper_error *err);
+
+/*
+ Get an equivalent `hyper_code` from this error.
+ */
+hyper_code hyper_error_code(const hyper_error *err);
+
+/*
+ Print the details of this error to a buffer.
+
+ The `dst_len` value must be the maximum length that the buffer can
+ store.
+
+ The return value is number of bytes that were written to `dst`.
+ */
+size_t hyper_error_print(const hyper_error *err, uint8_t *dst, size_t dst_len);
+
+/*
+ Construct a new HTTP request.
+ */
+hyper_request *hyper_request_new(void);
+
+/*
+ Free an HTTP request if not going to send it on a client.
+ */
+void hyper_request_free(hyper_request *req);
+
+/*
+ Set the HTTP Method of the request.
+ */
+hyper_code hyper_request_set_method(hyper_request *req, const uint8_t *method, size_t method_len);
+
+/*
+ Set the URI of the request.
+ */
+hyper_code hyper_request_set_uri(hyper_request *req, const uint8_t *uri, size_t uri_len);
+
+/*
+ Set the preferred HTTP version of the request.
+
+ The version value should be one of the `HYPER_HTTP_VERSION_` constants.
+
+ Note that this won't change the major HTTP version of the connection,
+ since that is determined at the handshake step.
+ */
+hyper_code hyper_request_set_version(hyper_request *req, int version);
+
+/*
+ Gets a reference to the HTTP headers of this request
+
+ This is not an owned reference, so it should not be accessed after the
+ `hyper_request` has been consumed.
+ */
+hyper_headers *hyper_request_headers(hyper_request *req);
+
+/*
+ Set the body of the request.
+
+ The default is an empty body.
+
+ This takes ownership of the `hyper_body *`, you must not use it or
+ free it after setting it on the request.
+ */
+hyper_code hyper_request_set_body(hyper_request *req, hyper_body *body);
+
+/*
+ Free an HTTP response after using it.
+ */
+void hyper_response_free(hyper_response *resp);
+
+/*
+ Get the HTTP-Status code of this response.
+
+ It will always be within the range of 100-599.
+ */
+uint16_t hyper_response_status(const hyper_response *resp);
+
+/*
+ Get the HTTP version used by this response.
+
+ The returned value could be:
+
+ - `HYPER_HTTP_VERSION_1_0`
+ - `HYPER_HTTP_VERSION_1_1`
+ - `HYPER_HTTP_VERSION_2`
+ - `HYPER_HTTP_VERSION_NONE` if newer (or older).
+ */
+int hyper_response_version(const hyper_response *resp);
+
+/*
+ Gets a reference to the HTTP headers of this response.
+
+ This is not an owned reference, so it should not be accessed after the
+ `hyper_response` has been freed.
+ */
+hyper_headers *hyper_response_headers(hyper_response *resp);
+
+/*
+ Take ownership of the body of this response.
+
+ It is safe to free the response even after taking ownership of its body.
+ */
+hyper_body *hyper_response_body(hyper_response *resp);
+
+/*
+ Iterates the headers passing each name and value pair to the callback.
+
+ The `userdata` pointer is also passed to the callback.
+
+ The callback should return `HYPER_ITER_CONTINUE` to keep iterating, or
+ `HYPER_ITER_BREAK` to stop.
+ */
+void hyper_headers_foreach(const hyper_headers *headers,
+                           hyper_headers_foreach_callback func,
+                           void *userdata);
+
+/*
+ Sets the header with the provided name to the provided value.
+
+ This overwrites any previous value set for the header.
+ */
+hyper_code hyper_headers_set(hyper_headers *headers,
+                             const uint8_t *name,
+                             size_t name_len,
+                             const uint8_t *value,
+                             size_t value_len);
+
+/*
+ Adds the provided value to the list of the provided name.
+
+ If there were already existing values for the name, this will append the
+ new value to the internal list.
+ */
+hyper_code hyper_headers_add(hyper_headers *headers,
+                             const uint8_t *name,
+                             size_t name_len,
+                             const uint8_t *value,
+                             size_t value_len);
+
+/*
+ Create a new IO type used to represent a transport.
+
+ The read and write functions of this transport should be set with
+ `hyper_io_set_read` and `hyper_io_set_write`.
+ */
+hyper_io *hyper_io_new(void);
+
+/*
+ Free an unused `hyper_io *`.
+
+ This is typically only useful if you aren't going to pass ownership
+ of the IO handle to hyper, such as with `hyper_clientconn_handshake()`.
+ */
+void hyper_io_free(hyper_io *io);
+
+/*
+ Set the user data pointer for this IO to some value.
+
+ This value is passed as an argument to the read and write callbacks.
+ */
+void hyper_io_set_userdata(hyper_io *io, void *data);
+
+/*
+ Set the read function for this IO transport.
+
+ Data that is read from the transport should be put in the `buf` pointer,
+ up to `buf_len` bytes. The number of bytes read should be the return value.
+
+ It is undefined behavior to try to access the bytes in the `buf` pointer,
+ unless you have already written them yourself. It is also undefined behavior
+ to return that more bytes have been written than actually set on the `buf`.
+
+ If there is no data currently available, a waker should be claimed from
+ the `ctx` and registered with whatever polling mechanism is used to signal
+ when data is available later on. The return value should be
+ `HYPER_IO_PENDING`.
+
+ If there is an irrecoverable error reading data, then `HYPER_IO_ERROR`
+ should be the return value.
+ */
+void hyper_io_set_read(hyper_io *io, hyper_io_read_callback func);
+
+/*
+ Set the write function for this IO transport.
+
+ Data from the `buf` pointer should be written to the transport, up to
+ `buf_len` bytes. The number of bytes written should be the return value.
+
+ If no data can currently be written, the `waker` should be cloned and
+ registered with whatever polling mechanism is used to signal when data
+ is available later on. The return value should be `HYPER_IO_PENDING`.
+
+ Yeet.
+
+ If there is an irrecoverable error reading data, then `HYPER_IO_ERROR`
+ should be the return value.
+ */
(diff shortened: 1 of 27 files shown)


bagder:
Thanks for this!

(It's a little hard to get into the API for an outsider with zero knowledge of how hyper works.)

Let's dive in. I'm focusing on reading `client.c` to understand how the API is to be used.

`hyper_executor_poll()` is called when select() has returned for a particular socket, right? When doing multiple concurrent connections, the mapping between the socket with activity and the corresponding `exec` handle needs to be done by the application.  So for HTTP/2 streams over the same connection, do we simply create multiple 'exec' contexts for the same socket?

Shouldn't hyper also tell the client somehow what to wait for in terms of socket event? Like the socket needs to be found writable when it wants to send the initial request? The current client seems to only set `all_fds->write` when it gets called to write to the socket and it gets EAGAIN back. But select() won't return for writable *before* that... ?

seanmonstar:
> (It's a little hard to get into the API for an outsider with zero knowledge of how hyper works.)

Yea, apologies about that. Part of this effort will need to better document how to actually *use* this API, but not just what the types look like.

> do we simply create multiple 'exec' contexts for the same socket?

An `exec` can process any number of "tasks" at a time. So you can manage multiple connections, multiple transactions with a single `exec`. The `exec` is a type that holds onto all registered futures, makes `wakers` to poll them with, and when polled again, knows which `wakers` were notified to only poll the relevant tasks.

> Shouldn't hyper also tell the client somehow what to wait for in terms of socket event?

This is done at the `hyper_io` callback level. The callbacks determine any way they want to register their IO and be able to notify the associated `waker` when ready. This allows a user to pick select, epoll, kqueue, IOCP, some sort of memory channel, whatever.

It might be a bug in my `client.c` to not have set the `all_fds->write` initially. That's what I get for having a program that I can't actually run yet...

bagder:
No need to apologize, I'm mostly pointing it out to highlight where perhaps some extra documentation might be needed down the line!

sdroege:
Doesn't help to figure out how the C API could look like, but you might want to take a look at [cargo-c](https://github.com/lu-zero/cargo-c) and [cbindgen](https://github.com/eqrion/cbindgen). The former to handle all the boring build/installation parts of a C library, the latter for autogenerating the C header file from the Rust code. `cargo-c` uses `cbindgen` by default to create the header.

bagder:
I think maybe parts of this work is about making a *sensible* C API that would suit users rather than just a blunt conversion from Rust as easy as possible. I don't think a tool can do that...

sdroege:
> I don't think a tool can do that...

That's also what none of the tools does. I just mentioned them so you can focus exactly on that part instead of boring build questions and `extern "C"/#[repr(C)]` to C header translations and keeping those in sync.

seanmonstar:
As an update: there is now sufficient Rust FFI code in the `hyper-capi` crate to make the `client.c` file compile and run, sending a request to httpbin, and getting response headers back.

The next piece I'll be tackling is the body API, to be able to stream the response body (and send a request body).

hyper-capi/README.md
```diff
@@ -0,0 +1,13 @@
+# C API for hyper
+
+This provides a C API to use the hyper library.
+
+## Testing
+
+```
+cargo build
+
+gcc -o client examples/client.c -I./include -L./target/debug -lhyper_c
```
bagder, line 10: How exactly is the `hyper_c` lib getting made? I've run `cargo build` in the root dir but I get neither a .so nor a .a file built?

hyper-capi/examples/client.c
```diff
@@ -0,0 +1,289 @@
...
+	}
+	if (hyper_request_set_uri(req, (uint8_t *)"/", sizeof("/") - 1)) {
+		printf("error setting uri\n");
+		return 1;
+	}
+
+	hyper_headers *req_headers = hyper_request_headers(req);
+
+	hyper_str name = { .buf = "host", .len = 4};
+	hyper_str val = { .buf = "httpbin.org", .len = sizeof("httpbin.org") - 1 };
```
bagder, line 182: These lines causes compiler warnings with `-pedantic` since `buf` is defined as unsigned while a const string is not:
~~~
examples/client.c: In function ‘main’:
examples/client.c:181:28: warning: pointer targets in initialization of ‘const uint8_t *’ {aka ‘const unsigned char *’} from ‘char *’ differ in signedness [-Wpointer-sign]
  181 |  hyper_str name = { .buf = "host", .len = 4};
      |                            ^~~~~~
examples/client.c:181:28: note: (near initialization for ‘name.buf’)
examples/client.c:182:27: warning: pointer targets in initialization of ‘const uint8_t *’ {aka ‘const unsigned char *’} from ‘char *’ differ in signedness [-Wpointer-sign]
  182 |  hyper_str val = { .buf = "httpbin.org", .len = sizeof("httpbin.org") - 1 };
      |                           ^~~~~~~~~~~~~
examples/client.c:182:27: note: (near initialization for ‘val.buf’)

bagder (commented):
The use of `hyper_str` for strings  is not very "C style" but I think I can live with it. Just two points about it apart from the compiler warning:

1. A string in C typically means it cannot contain zero bytes but these ones probably can, right? It then needs proper documentation about that detail since in C programs we will likely convert to/from actual C strings with such limitations...
2. The use of `hyper_str` is still inconsistent in the current incarnation where some functions still take a `uint8_t *` + `size_t` length and some take a `hyper_str *`.

hyper-capi/include/hyper.h
```diff
@@ -0,0 +1,286 @@
...
+hyper_task *hyper_clientconn_handshake(hyper_io *io, hyper_clientconn_options *options);
+
+// Send a request on the client connection.
+//
+// Returns a task that needs to be polled until it is ready. When ready, the
+// task yields a `hyper_response *`.
+hyper_task *hyper_clientconn_send(hyper_clientconn *client, hyper_request *request);
+
+// Creates a new set of HTTP clientconn options to be used in a handshake.
+hyper_clientconn_options *hyper_clientconn_options_new();
```
bagder, line 76: In C, you should put a `void` between those parentheses, like `(void)` if there's no argument. `-Wstrict-prototypes` will tell you:
~~~
./include/hyper.h:76:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
   76 | hyper_clientconn_options *hyper_clientconn_options_new();
      | ^~~~~~~~~~~~~~~~~~~~~~~~
./include/hyper.h:92:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
   92 | hyper_io *hyper_io_new();
      | ^~~~~~~~
./include/hyper.h:132:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  132 | hyper_request *hyper_request_new();
      | ^~~~~~~~~~~~~
./include/hyper.h:232:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  232 | hyper_executor *hyper_executor_new();
      | ^~~~~~~~~~~~~~

seanmonstar:
> The use of hyper_str for strings is not very "C style"

Yea, I initially hoped it would help with functions that needed multiple strings (like passing a name and value), and as a return value. I had hoped this was possible (and nicer, giving pseudo named arguments), but it's not:

```c
hyper_headers_set(headers, hyper_str { .buf = "foo", .len = 3 }, hyper_str { .buf = "bar", .len = 3 });
```

I also assumed it'd be the best way to allow a `hyper_headers_get(headers, name)` return value. Perhaps the more "C style" is to pass in "out" arguments? `hyper_header_get(headers, name, name_len, &value, &value_len)`?

comex:
> Yea, I initially hoped it would help with functions that needed multiple strings (like passing a name and value), and as a return value. I had hoped this was possible (and nicer, giving pseudo named arguments), but it's not:
> 
> ```c
> hyper_headers_set(headers, hyper_str { .buf = "foo", .len = 3 }, hyper_str { .buf = "bar", .len = 3 });
> ```

It is possible; you just need to put the type in parentheses: `(hyper_str){.buf = "foo", .len = 3}`

This is a C99 compound literal.  The version without parentheses is actually valid in C++20, but not C.

bagder:
Don't assume everyone can and will use such statements though.

In my case, I have a code base written in C89 so even though this API will require C99, my code will not and we will not be doing these "modern" fancy compound literals.

hyper-capi/include/hyper.h
```diff
@@ -0,0 +1,318 @@
...
+hyper_task_return_type hyper_task_type(hyper_task *task);
+
+// Set a user data pointer to be associated with this task.
+//
+// This value will be passed to task callbacks, and can be checked later
+// with `hyper_task_userdata`.
+void hyper_task_set_data(hyper_task *task, void *userdata);
+
+// Retrieve the userdata that has been set via `hyper_task_set_data`.
+void *hyper_task_userdata(hyper_task *task);
```
bagder, line 288: Wouldn't it be more consistent if it was `hyper_task_set_userdata` as well? I mean then "userdata" would be common among the function names and the argument names.

bagder:
Can I also ask for a `hyper_version()` call (the name can of course be different) that can return the version of the currently used Hyper?

bagder:
Two questions on the sample client:

1. Since the `EXAMPLE_HANDSHAKE` state is the first state set just before the loop, can't all this logic done in this state within the loop then just as well be done before the loop starts?
2. Having the client use separate states for sending and body confuses me. A server can very well start sending the response while the request is in progress, they can't be mutually exclusive! So, shouldn't they then be a single state? And without "handshake" as a state, then we're down to a single state...

bagder:
For some reason, the example client fails to download my local test file that's 17MB big, or anything that is larger. Small files seems to be fine.

If I strace the hang, I can see it read/write for a while (512KB buffers?! wow) and then just stall on a select(). Highly reproducible.

hyper-capi/include/hyper.h
seanmonstar: That's true. In the rest of the C API, it's been just `*_set_data`, but then wondered if `hyper_task_data` would be to similar to `hyper_task_value`.

seanmonstar:
> `hyper_version()`

Sure. Is it for checking functionality, and thus return an integer of the version, or for display purposes, and so just wants a string? Or maybe both?

> `EXAMPLE_HANDSHAKE` state 

Yep, it originally was done before the loop, since an HTTP/1 handshake is ready immediately. I moved it into the loop once I got state tracking in the tasks, since HTTP/2 or 3 handshakes could require IO (if wanting to check an h2 prior knowledge before sending the request, for example). That difference is simply just setting a `clientconn_option` (not exposed just yet).

> A server can very well start sending the response while the request is in progress,

Yep, and hyper supports that. The `send` task resolves as soon as the server gives back response headers. The continued writing of a request body (which would be set on the `hyper_request *`) can still be happening while the response body is read back. The response body is only read off the transport as its polled (like with `hyper_body_foreach`), otherwise backpressure is applied by leaving it in the transports buffer.


>  fails to download my local test file that's 17MB big,

Sounds weird to me, I'll give that a try.

> (512KB buffers?! wow)

Heh, that's part of hyper's "adaptive read buffer", that will try to improve throughput by increasing the read buffer if after a few reads the full buffer was filled each time, up to a configurable maximum. When on localhost, that'll reach the max pretty quickly. We can expose that know on `hyper_clientconn_options`.

seanmonstar:
> fails to download my local test file that's 17MB big,

Found the issue. The executor implementation (`FuturesUnordered`) recently received "starvation protection", forcing a yield after 32 iterations inside its poll. This was meant to allow systems to check other tasks. However in hyper-capi, the `FuturesUnordered` is the entire executor, it's not sharing with other tasks, so I added a check for this condition, and now I can download larger files.

bagder:
You might be on top of this already and I might be totally premature here (so feel free to ignore and just give this the amount of attention you think it deserves), but something in the current setup/example is not really optimal for performance. I'm mentioning it here anyway so that you're at least aware of it:

After the fix mentioned above was merged, I modified the example to download 8GB on Linux from my localhost HTTP/1.1 server (to /dev/null). On my particular host this took **150 seconds**. My curl command line tool does the same transfer in slightly over **2 seconds**.

bagder:
##   hyper_version()

> Sure. Is it for checking functionality, and thus return an integer of the version, or for display purposes, and so just wants a string? Or maybe both?

I was thinking primarily for displaying purposes so that users can tell/show what they're using. But maybe version checking will be interesting too...

seanmonstar:
> something in the current setup/example is not really optimal for performance

I haven't done benchmarking of this example vs the top level `example/client.rs`, but that time difference sounds off. My first thought is that `cargo build` doesn't do any optimizations by default, only with `cargo build --release`, and then probably also needing to set optimizations in `gcc`. But if you're already doing that, then something else is **really** off...

bagder:
Yes, this is probably something else than just optimization enabled or not, which is why I thought I'd mention it. It feels like there's a mistake somewhere.

seanmonstar:
Just to double check, I grabbed a localhost server with a 500mb file (my Linux host is rather limited), I ran `time client 127.0.0.1 3000 >/dev/null` both with and without optimizations.

- Without, 24.4s.
- With, 0.6s.

I had to run `cargo build --release`, but then also noticed I needed to update the Makefile to link to `target/release` instead of `target/debug` to finally see the improvement. So probably worth adding a `client-debug` and `client-release` target.

bagder:
Gosh! Okay, then I probably just saw that. I could not even imagine that huge differences between debug and release. I'll rerun my tests soon to verify. Thanks!

bagder:
Yeps, confirmed. The debug version is indeed **75 times slower** for me...

hyper-capi/examples/client.c
```diff
@@ -0,0 +1,303 @@
...
+					printf("error setting method\n");
+					return 1;
+				}
+				if (hyper_request_set_uri(req, STR_ARG(path))) {
+					printf("error setting uri\n");
+					return 1;
+				}
+
+				hyper_headers *req_headers = hyper_request_headers(req);
+				hyper_headers_set(req_headers,  STR_ARG("host"), STR_ARG(host));
```
bagder, line 214: Question on this: I can use `hyper_headers_set()` if I only add a single header with the name, but if I want to add two headers with the same name I would need to make the second one with `hyper_headers_add` ?

hyper-capi/examples/client.c
```diff
@@ -0,0 +1,303 @@
...
+	printf("http handshake ...\n");
+
+	// We need an executor generally to poll futures
+	hyper_executor *exec = hyper_executor_new();
+
+	// Prepare client options
+	hyper_clientconn_options *opts = hyper_clientconn_options_new();
+	hyper_clientconn_options_exec(opts, exec);
+
+	hyper_task *handshake = hyper_clientconn_handshake(io, opts);
```
bagder, line 174: `handshake` is never freed later in the example code. Is that a mistake or does hyper handle that itself somehow?

bagder: Ah no, now I think I get it. That's the `task` returned from `hyper_executor_poll` in the first invoke, right?

hyper-capi/include/hyper.h
```diff
@@ -0,0 +1,318 @@
...
+// Frees options not passed to a handshake.
+void hyper_clientconn_options_free(hyper_clientconn_options *options);
+
+// HTTP IO
+
+// Create a new IO type used to represent a transport.
+//
+// The read and write functions of this transport should be set with
+// `hyper_io_set_read` and `hyper_io_set_write`.
+hyper_io *hyper_io_new(void);
```
bagder, line 87: No `hyper_io_free()` ? What if I want to discard it?

hyper-capi/include/hyper.h
```diff
@@ -0,0 +1,318 @@
...
+	HYPER_POLL_READY,
+	HYPER_POLL_PENDING
+} hyper_poll;
+
+// HTTP ClientConn
+
+// Starts an HTTP client connection handshake using the provided IO transport
+// and options.
+//
+// Both the `io` and the `options` are consumed in this function call.
```
bagder, line 55: Is that also guaranteed if the function returns a `NULL`/failure? Would be good to have that clarified.

jplatte:
> Yeps, confirmed. The debug version is indeed **75 times slower** for me...

This is not such a surprising thing with Rust projects, since without explicit overrides, every [transitive] dependency except the standard library (which is usually shipped only in optimized binary form) is compiled in debug mode by default.

hyper-capi/examples/client.c
seanmonstar: Correct, `hyper_executor_push` takes ownership of the task. If it gets popped later with `hyper_executor_poll`, that gives ownership back and you can free it. If you were to free the executor while it still owned some tasks, it'd free them at the same time.

hyper-capi/examples/client.c
seanmonstar: Yes, `_set` overwrites any previous value. `_add` appends, even if it's empty. You could think of `_set` as combining `_remove` and `_add` (perhaps `_remove` is missing from the C API...).

So it'd also be fair to have just used `hyper_headers_add` in this case.

bagder: Ah yes, so I can just use `*_add()`, thanks!

hyper-capi/src/task.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+// ===== impl Waker =====
+
+
+ffi_fn! {
+    fn hyper_context_waker(cx: *mut Context<'_>) -> *mut Waker {
+        let waker = unsafe { &mut *cx }.waker().clone();
+        Box::into_raw(Box::new(Waker { waker }))
```
bdonlan, line 330: I note that this requires allocation on every waker construction. Would it be better to create a two-pointer-wide struct on the C side to hold the waker directly and allow us to avoid this extra allocation?

hyper-capi/src/io.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+extern "C" fn read_noop(
+    _userdata: *mut c_void,
+    _: *mut Context<'_>,
+    _buf: *mut u8,
+    _buf_len: size_t,
+) -> size_t {
+    0
```
bdonlan, line 57: These stubs should perhaps return `IO_ERROR` instead.

hyper-capi/src/http_types.rs
```diff
@@ -0,0 +1,132 @@
...
+            Err(_) => {
+                hyper_code::Kaboom
+            }
+        }
+    }
+}
+
+ffi_fn! {
+    fn hyper_request_headers(req: *mut Request<Body>) -> *mut HeaderMap {
+        unsafe { &mut *req }.headers_mut()
```
bdonlan, line 59: Is this sound from a rust memory model perspective? We're returning a pointer to an inner field not accessed through an `UnsafeCell`. I'm not sure how the full aliasing model of rust works with pointers; is it an issue that the mutable reference we got that pointer from has gone out of scope here? Or is it fine as long as we don't dereference `req` again until we're done with this `*mut HeaderMap`? Do we run into theoretical soundness issues if C code asks for the headers first and then asks for e.g. `hyper_request_set_uri`?

This is all a bit unlikely unless LTO crosses between C and Rust, of course, but with LLVM that's a possibility.

You could of course sidestep these soundness issues by having `hyper_headers_set` take `*mut Request<Body>` instead.

hyper-capi/src/http_types.rs
```diff
@@ -0,0 +1,132 @@
...
+
+ffi_fn! {
+    fn hyper_headers_foreach(headers: *const HeaderMap, func: IterFn, userdata: *mut c_void) {
+        for (name, value) in unsafe { &*headers }.iter() {
+            let name_ptr = name.as_str().as_bytes().as_ptr();
+            let name_len = name.as_str().as_bytes().len();
+            let val_ptr = value.as_bytes().as_ptr();
+            let val_len = value.as_bytes().len();
+
+            if IterStep::Continue != func(userdata, name_ptr, name_len, val_ptr, val_len) {
```
bdonlan, line 108: Note that we have UB if C returns a value not defined in the enum.

hyper-capi/include/hyper.h
```diff
@@ -0,0 +1,318 @@
...
+// task yields a `hyper_response *`.
+hyper_task *hyper_clientconn_send(hyper_clientconn *client, hyper_request *request);
+
+// Free a `hyper_clientconn *`.
+void hyper_clientconn_free(hyper_clientconn *client);
+
+// Creates a new set of HTTP clientconn options to be used in a handshake.
+hyper_clientconn_options *hyper_clientconn_options_new(void);
+
+// Set the client background task executor.
```
bdonlan, line 73: Hmm. What is this used for? We don't have any hooks for setting timeouts (for example), so I'm not sure what need there would be for a background executor at this time.

hyper-capi/src/http_types.rs
seanmonstar: Eek, nice catch! I had assumed with a `repr(C)`, it would just pretend to be a `c_int`, but I just double checked the Rust Reference, and it'd still be UB. Fixed by changing to just a constant and asking for `c_int`s from the C side.

hyper-capi/src/task.rs
seanmonstar: It is unfortunate, but I hoped not too badly. I'm unsure if the exact size of `std::task::Waker` can be relied upon... It does say that it is only composed of the data pointer and the vtable pointer, but it doesn't explicitly guarantee that won't ever change...

hyper-capi/include/hyper.h
seanmonstar: hyper puts the "connection" itself into a separate task that is always polled, to handle hang up detection, and any other meta stuff such as when using HTTP/2. The `clientconn` that is returned is a handle to send requests into that task.

hyper-capi/src/lib.rs
```diff
@@ -21,7 +21,14 @@ pub enum IterStep {
...
 struct AssertSendSafe<T>(T);
 
 unsafe impl<T> Send for AssertSendSafe<T> {}
+
+static VERSION_CSTR: &str = concat!(env!("CARGO_PKG_VERSION"), "\0");
+
+ffi_fn! {
+    fn hyper_version() -> *const u8 {
+        VERSION_CSTR.as_ptr()
+    }
```
bagder, line 27: In my build this returns `0.0.0` - I presume that's not expected?

seanmonstar: It's expected at this very moment, since it's the value inside `hyper-capi/Cargo.toml`. This is one point I've been thinking about with regards to making this a separate crate, it's technically a different version than the `hyper` crate. I'll be writing up the pros and cons about that in a separate issue.

But I've also been wondering, since this is to help users be able to report exactly what version of stuff they used, I suppose it may be helpful to include a couple of the versions of the bigger dependencies in hyper as well (like `h2`, hyper's HTTP/2 implementation that is a sub crate).

hyper-capi/src/task.rs
seanmonstar: I've brought this up to [wg-async-foundation](https://rust-lang.zulipchat.com/#narrow/stream/187312-wg-async-foundations/topic/size_of.3A.3A.3CWaker.3E()).

hyper-capi/src/lib.rs
MCOfficer: It's pretty ugly, but if i'm reading this correctly, semver allows you to specify the associated `hyper` version as metadata of the `hyper-api` version: `0.0.1+hyper.0.3.18`.

bagder: In my case I'm interested in the conveying of version information to users so that users subsequently can tell us that info back in a future when we're debugging weird things and we're curious what versions the user is actually using. "semver" has little value in that context. But as @seanmonstar points out, the version numbers of the most important sub-dependencies can certainly have value.

MCOfficer: I was under the impression that this would always be `hyper-api`'s crate version, which has to adhere to semver. If it can be anything, all the better.

hyper-capi/include/hyper.h
```diff
@@ -0,0 +1,323 @@
...
+
+// Adds the provided value to the list of the provided name.
+//
+// If there were already existing values for the name, this will append the
+// new value to the internal list.
+void hyper_headers_add(hyper_headers *headers,
+                       const uint8_t* name,
+                       size_t name_len,
+                       const uint8_t *value,
+                       size_t value_len);
```
bagder, line 186: Both `hyper_headers_set` and `hyper_headers_add` return `void` here. What happens when they can't allocate memory for adding the headers to the internal lists? Surely they must be able to return error?

hyper-capi/examples/client.c
bagder: ... except that `_add()` doesn't seem to be provided yet:
~~~
client.c:(.text+0x666): undefined reference to `hyper_headers_add'
/usr/bin/ld: client.c:(.text+0x68e): undefined reference to `hyper_headers_add'
collect2: error: ld returned 1 exit status

bagder:
**hyper_headers_set**: when I set the header `Host: 127.0.0.1`, hyper sends it off as `host: 127.0.0.1` even when used over HTTP/1.1. Is this a hyper-c issue or a generic Hyper issue? (I'm not really sure how to deal with issues I find when working on this new C API). I am of the strong opinion that it isn't hyper's job to modify the headers I tell it to use. HTTP headers are by definition case insensitive but there are known issues with many server side implementations that will assume and insist on specific casing of headers. Those are server issues but I want my client to be able to work with them nonetheless.

lnicola:
> title-case headers

See https://github.com/hyperium/hyper/issues/1492 for context. This [is available](https://docs.rs/hyper/0.13.8/hyper/client/struct.Builder.html#method.http1_title_case_headers) in the `hyper` API, but doesn't seem to be exposed yet in the C bindings.

bagder:
I would really like to see a additional (separate?) client example that receives data in a streaming manner and one that sends (POST/PUT) in a streaming manner, to help me figure out how the API is to be used for that.

hyper-capi/include/hyper.h
seanmonstar: This was a mistake in the header file, they did indeed have a `hyper_code` return type in the Rust code, and I've updated the header file.

I've been searching this morning to see if there's a tool that validates that functions defined in header files exist in a compiled library, and so far haven't found one.

sdroege: > I've been searching this morning to see if there's a tool that validates that functions defined in header files exist in a compiled library, and so far haven't found one.

I'm not aware of one either, but `cbindgen` could create the header file automatically for `pub extern "C" fn`s if you like one of the C API styles it supports. That also prevents implementation and header to get out of sync.

seanmonstar: I am trying out `cbindgen` right now, if all works this could be a good way to prevent these mistakes.

bagder:
The prototype is now fixed for `hyper_headers_add`, but the function doesn't exist... =)

bagder:
Note that the example code doesn't build anymore after the recent API changes.

seanmonstar:
I just noticed that as well, and fixed that. The `cbindgen` step is manual at the moment, I'll be adding a CI task to verify it's been checked for a commit.

hyper-capi/src/lib.rs
seanmonstar: I've written up those thoughts into an [this issue](https://github.com/hyperium/hyper/issues/2299) (the version string is relevant, but a smaller point to what I wrote up).

capi/include/hyper.h
```diff
@@ -105,6 +107,9 @@ hyper_task *hyper_body_data(hyper_body *body);
  The `hyper_buf` pointer is only a borrowed reference, it cannot live outside
  the execution of the callback. You must make a copy to retain it.
 
+ The callback should return `HYPER_ITER_CONTINUE` to continue iterating
+ chunks as they are received, or `HYPER_ITER_BREAK` to cancel.
```
bagder, line 111: Ah yes, this reminds me of another detail. What if my callback fails in a way that calls for erroring out completely, like out of memory? `HYPER_ITER_BREAK` aborts the iteration but `hyper_headers_foreach` doesn't return an error code. Do I need to set my own state variable in the struct I point `userdata` to, or can we have a dedicated return code that makes `hyper_headers_foreach` return an error code?

seanmonstar: What's the more common pattern? I'm fine with either option.

bagder: As a user, it is slightly easier and more convenient to have a dedicated return code for **"error"** and have `hyper_headers_foreach()` be able to return success/error based on that. That'd be my preference. But if there's a reason not to do that, I can still deal with it without trouble.

capi/include/hyper.h
```diff
@@ -270,6 +270,26 @@ void hyper_response_free(hyper_response *resp);
  */
 uint16_t hyper_response_status(const hyper_response *resp);
 
+/*
+ Get a pointer to the reason-phrase of this response.
+
+ This buffer is not null-terminated.
```
bagder, line 276: So how is the buffer terminated? How can we tell where the reason-phrase ends? Is it (CR)LF?

lnicola: See two lines below.

bagder: Sorry, I must be blind! =)

capi/include/hyper.h
```diff
@@ -0,0 +1,545 @@
...
+ Adds the provided value to the list of the provided name.
+
+ If there were already existing values for the name, this will append the
+ new value to the internal list.
+ */
+hyper_code hyper_headers_add(hyper_headers *headers,
+                             const uint8_t *name,
+                             size_t name_len,
+                             const uint8_t *value,
+                             size_t value_len);
```
bagder, line 407: Question: when I add a request header *without* a value, with `value_len` set to zero, Hyper seems to still insert a space after the colon. This means I can't produce an outgoing header without a trailing space if there's no value for the header. Is this intended or something that could be considered to be changed? It's a little detail that makes it difficult for me to generate the HTTP request with Hyper as flexible as I can with curl (and it's a wasted byte over the wire).

seanmonstar: Huh. That must be because the encoder is essentially a loop of `buf += name + ": " + val + "\r\n"`. I suppose we could add a check if the value is empty. Doesn't seem like something that is very common, but I understand curl already supports it.

bagder: There are web services "out there" that wants a header without value, and yes curl has specific support to provide such a header. In our test suite we verify the entire protocol "dump" as issued by the client - which in my case now subsequently makes the curl+hyper client issue an extra space for this case and therefore the test case fails...

capi/include/hyper.h
```diff
@@ -0,0 +1,545 @@
...
+ This buffer is not null-terminated.
+
+ This buffer is owned by the response, and should not be used after
+ the response has been freed.
+
+ Use `hyper_response_reason_phrase_len()` to get the length of this
+ buffer.
+ */
+const uint8_t *hyper_response_reason_phrase(const hyper_response *resp);
+
```
bagder, line 339: Are we any closer to this function being able to return the actually used reason instead of a fixed canonical version? This is after all information provided from the server that clients can use/inspect...

seanmonstar: I got distracted with other things, hoping that the canonical string was enough for most test cases. Grabbing the actual reason should be a quick add, aiming to get that in today.

capi/include/hyper.h
seanmonstar: Yea, figured. Added a check.

bagder:
:champagne: - hooray, thanks @seanmonstar!

hyper-capi/src/http_types.rs
seanmonstar: I've checked with @pnkfelix, and he believes that what this is doing is currently safe. It *would* be invalid for a user in C to call this multiple times and access the `*mut HeaderMap` simultaneously, but simply returning this pointer should be sound (and shouldn't require an `UnsafeCell`).

seanmonstar:
This has been squashed and cleaned up as we prepare to merge to master. As part of the merge, I've marked **the API as unstable**, requiring a special `cfg` and documenting it as such. There are still some things to figure out, but I'll file follow up issues regarding those. I'm aiming to click "merge" tomorrow.

bagder:
I'm all buckled up and ready! :grin:

simpleearn (approved):

PR #2283, opened 2020-09-17
body::Sender: Add a note about abnormal body closing

dbrgn:
According to @seanmonstar, bodies being closed in a normal fashion when an error occurs in the spawned task holding the sender is a frequent gotcha. It caught me as well.

This note would have prevented a lot of debugging in my case.

(Let me know if you'd prefer different wording!)

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -75,10 +75,19 @@ enum DelayEof {
     Eof(DelayEofUntil),
 }
 
-/// A sender half used with `Body::channel()`.
+/// A sender half created through [`Body::channel()`].
 ///
-/// Useful when wanting to stream chunks from another thread. See
-/// [`Body::channel`](Body::channel) for more.
+/// Useful when wanting to stream chunks from another thread.
+///
+/// ## Body Closing
+///
+/// Note that the request body will always be closed normally when the sender is dropped (meaning
+/// that the empty terminating chunk will be sent to the client). If you desire to close the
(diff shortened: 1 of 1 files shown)


dbrgn:
Hm, gitcop complains about my commit message, but the "Details" link doesn't tell me what's wrong :confused: I guess it requires conventional style?

seanmonstar:
Don't worry about gitcop, it's not tagged "required", it's just a reminder for me to fix when squashing.

src/body/body.rs
```diff
@@ -75,10 +75,19 @@ enum DelayEof {
...
+/// A sender half created through [`Body::channel()`].
 ///
-/// Useful when wanting to stream chunks from another thread. See
-/// [`Body::channel`](Body::channel) for more.
+/// Useful when wanting to stream chunks from another thread.
+///
+/// ## Body Closing
+///
+/// Note that the request body will always be closed normally when the sender is dropped (meaning
+/// that the empty terminating chunk will be sent to the client). If you desire to close the
```
seanmonstar, line 85:
```suggestion
/// that the empty terminating chunk will be sent to the remote). If you desire to close the
```

It could be server -> client, or client -> server.

dbrgn: Good point!

seanmonstar (approved):

seanmonstar:
Thanks!

PR #2285, opened 2020-09-18
Add a way to force queued strategy

### LINKED ISSUE #2282: Provide a way to force the queue writev strategy
There is `http1_writev(bool)` currently to allow someone to force alwaysing using the "flatten" strategy of writing. However, when calling `http1_writev(true)`, that just keeps the default, which is "auto". There should be a way for users to specify "always use the queue strategy, don't use auto".

The most logical way would be to keep "auto" as the default, calling `http1_writev(false)` would force "flatten", and calling `http1_writev(true)` would force "queue". This can be done by changing the builders to hold `http1_writev: Option<bool>` internally.

Snarpix:
This fixes #2282

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -75,7 +75,7 @@ where
 #[derive(Clone, Debug)]
 pub struct Builder {
     pub(super) exec: Exec,
-    h1_writev: bool,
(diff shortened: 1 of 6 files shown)


seanmonstar (approved):
Perfect, thanks!

PR #2288, opened 2020-09-24
For 0.10.x: fix(buffer): fix maybe_reserve() not working after take_buf()

strohel:
This is a fix for the ancient pre-async 0.10.x branch (the bug is not present in 0.11.0+).
I was hesitant whether to submit a patch for such an old branch at all, but I'll let you decide whether you still support it.

The bug affects Rocket 0.4.*, [whose ecosystem seems to be the last major user of hyper 0.10.x](https://lib.rs/crates/hyper/rev). Rocket 0.5 with up-to-date hyper dependency is in the works, but there are no indications it will be finished any time soon.

The problem demonstrates as Rocket's inability to read second request on
a keep-alive connection. I.e. this fixes [a bug where keep-alive connections
don't work at all](https://github.com/SergioBenitez/Rocket/issues/580) in current Rocket 0.4.5. More details in the commit message.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -99,7 +99,10 @@ impl<R: Read> BufReader<R> {
     #[inline]
     fn maybe_reserve(&mut self) {
         let cap = self.buf.capacity();
-        if self.cap == cap && cap < MAX_BUFFER_SIZE {
+        if cap == 0 {
+            trace!("reserving initial {}", INIT_BUFFER_SIZE);
+            self.buf = vec![0; INIT_BUFFER_SIZE];
+        } else if self.cap == cap && cap < MAX_BUFFER_SIZE {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for filing! There's been a few recent PRs targeting 0.10.x. In those, I felt it was best to close, since 0.10.x stopped seeing active development 3 years ago. This change does look small, but I'm inclined to follow the same precedent. With no development, and very little testing, I'd hate to publish a new release that might require more support, or breaks something else unexpected...

I think all I can do is suggest upgrading to a newer version of hyper. Sorry if that's disappointing.

strohel:
Cool, that's totally understandable.

To communicate this, you may want to remove the 0.10.x branch from the repo - it would not possible to open a PR for it and the v0.10.16 tag still references all its commits.

PR #2290, opened 2020-09-27
Add accessors to `Connected` fields

sfackler:
These are useful when unit testing connector layers that do things like handling proxies, so you don't need to spin up an entire HTTP request to see if the returned connection state is correct or not. Ideally we'd expose an accessor for the extra info as well, but that's not currently in a form that's easy to work with.

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -81,7 +81,7 @@
 //! [`Connection`]: Connection
 use std::fmt;
 
-use ::http::Response;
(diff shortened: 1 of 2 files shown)


sfackler:
I think the way to expose the extras would just be

```rust
pub fn get_extras(&self, extensions: &mut Extensions) { ... }
```

A bit weird, but it's really the only way you'd be able to interact with it anyway.

seanmonstar (approved):
:shipit:

PR #2292, opened 2020-09-29
fix(http): fix panic when addrs in ConnectingTcpRemote is empty

### LINKED ISSUE #2291: 'missing connect error' panic when connecting to IPv6-only domain using a socket bound to IPv4 address
This example uses `reqwest` for simplicity:

```rust
#[tokio::main]
async fn main() -> () {
    let builder = reqwest::ClientBuilder::new();
    let builder = builder.local_address(std::net::IpAddr::from([0, 0, 0, 0]));
    let client = builder.build().unwrap();
    match client.get("http://ipv6.google.com").send().await {
        Ok(r) => {
            println!("Response: {:?}", r);
        }
        Err(e) => {
            println!("Error: {:?}", e);
        }
    }
}
```

Error message:

```
thread 'main' panicked at 'missing connect error', /home/alexwl/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.13.8/src/client/connect/http.rs:544:13
```

IPv6-only domains (only `AAAA` DNS records, not `A`) like `ipv6.google.com`, are rare, but they do exist in the wild.

The problem is that `split_by_preference` removes IPv6 addresses from `IpAddrs` when the `local_addr` is IPv4:

https://github.com/hyperium/hyper/blob/523d66a41ffe0429530241c381070fb490618a4a/src/client/connect/dns.rs#L203-L225

`connect` panics when the `self.addrs` is empty and the `err` is `None`:

https://github.com/hyperium/hyper/blob/523d66a41ffe0429530241c381070fb490618a4a/src/client/connect/http.rs#L524-L545

I think `connect` should return an error instead of panicking when the `err` is `None`.

For example:

```rust
match err {
    Some(e) => Err(e),
    None => Err(std::io::Error::new(
        std::io::ErrorKind::NotConnected,
        "Network unreachable"
    ))
}
```

alexwl:
Closes #2291

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -541,7 +541,13 @@ impl ConnectingTcpRemote {
             }
         }
 
-        Err(err.take().expect("missing connect error"))
+        match err {
+            Some(e) => Err(e),
+            None => Err(std::io::Error::new(
+                std::io::ErrorKind::NotConnected,
+                "Network unreachable",
+            )),
+        }
     }
 }
 


seanmonstar (approved):

seanmonstar:
Thanks so much!

PR #2307, opened 2020-10-23
Allow library consumer to drain a connection

teozkr:
A draining connection will no longer be used for new requests, but allow existing requests to finish. As discussed on Discord earlier (https://discordapp.com/channels/500028886025895936/670880858630258689/768499073644756992).

This feels like the wrong layer to add this on, since `Connected` is otherwise unaware of any connection pooling going on. It also feels like a smell to me that this will break if the pool starts refreshing `Connection::connected()` for any reason. The more appropriate place would probably be to store it on the `PoolClient`, but that's getting constructed and destroyed all over the place, so I'm not really sure about what guarantees we have there about the object identity over time.

This is also missing a bunch of docs, I just want to find a sound approach before getting into that.


teozkr:
This would be used something like this:

```rust
    let response = client.request(...).await?;
    if response.status() == StatusCode::SERVICE_UNAVAILABLE {
        // Server is dying, try to reconnect
        response.extensions().get::<PoolHandle>().unwrap().drain();
    }
```

seanmonstar:
Figuring out a way to drain the pool would be cool. I think #2605 is similar. We've moving the pool as part of 1.0 to `hyper-util`, so it would be more acceptable to experiment there.

Issue #2310, opened 2020-10-24
Only one HTTP2 Ping is sent with keep alive enabled


pdcalado:
I believe there is a bug in the HTTP2 keep alive implementation, more concretely in `src/proto/h2/ping.rs`.

This issue was originally detected in https://github.com/hyperium/tonic/issues/474, credits to @alce.

It seems that when enabling HTTP2 server keep alive, only the first Ping is sent after the configured interval.
Afterwards, no more Pings seem to be sent.

## How to reproduce

I'm applying the following changes to the `echo` example:

```diff
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -14,7 +14,10 @@ async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Er>
         ))),
 
         // Simply echo the body back to the client.
-        (&Method::POST, "/echo") => Ok(Response::new(req.into_body())),
+        (&Method::POST, "/echo") => {
+           tokio::time::delay_for(std::time::Duration::from_secs(60)).await;
+           Ok(Response::new(req.into_body()))
+       },
 
         // Convert to uppercase before sending back to client using a stream.
         (&Method::POST, "/echo/uppercase") => {
@@ -55,7 +58,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>
 
     let service = make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(echo)) }>
 
-    let server = Server::bind(&addr).serve(service);
+    let server = Server::bind(&addr)
+       .http2_keep_alive_interval(std::time::Duration::from_secs(10))
+       .http2_keep_alive_timeout(std::time::Duration::from_secs(5))
+       .serve(service);
 
     println!("Listening on http://{}", addr);
```

Running the example with `cargo run --example echo`, sniffing with `tcpdump -i lo port 3000`, and making a request with `curl -v --http2-prior-knowledge http://localhost:3000/echo --data "foo"`, I get the following:

```tcpdump
20:03:21.621105 IP localhost.hbci > localhost.46506: Flags [P.], seq 22:44, ack 160, win 512, options [nop,nop,TS val 2305866698 ecr 2305866696], length 22
20:03:21.621120 IP localhost.46506 > localhost.hbci: Flags [.], ack 44, win 512, options [nop,nop,TS val 2305866698 ecr 2305866698], length 0

[...]

PR #2315, opened 2020-10-29
Fix server keep alive to send more than one ping

pdcalado:
When enabling HTTP2 server keep alive, only the first ping is sent after the configured interval.
Afterwards, no more pings are sent.

More details on this issue: #2310 

The suggested solution checks if `ping_sent_at` has already been cleared by `Ponger::poll` when `KeepAliveState::PingSent` state is active. A test is added to make sure the server produces more than one ping when configured to do so.


seanmonstar (approved):
Excellent!

PR #2316, opened 2020-10-29
server: Remove high-level server API in-favor of lower-level API

LucioFranco:
(note: draft being open for initial feedback)


seanmonstar:
I'm going to close finally, since the #2806 proposes to do this for real, and the merge conflicts are dire. Maybe this branch wouldn't take much to rebase, and we could re-open, I dunno.

PR #2309, opened 2020-10-24
Bump tokio from 0.2.11 to 0.3

Urhengulas:
This PR shall update tokio to version `0.3`.

**Changes**
* adapt `AsyncRead`, `AsyncWrite` implementations
* rename `delay_for` -> `sleep`, `Delay` -> `Sleep`
* ...

Fixes #2302 

---

There are still quite some errors left to resolve, so I am happy to get some input there 😃

PR #2317, opened 2020-10-29
feat: update to Tokio 0.3

hawkw:
This branch updates Hyper's Tokio dependency to v0.3.x. This continues
the work started by @Urhengulas in PR #2309, updating additional parts
of the codebase and fixing some bugs introduced in that branch.

There are a few important notes:

* `h2` is currently a Git dependency, so this will block releasing a new
  Hyper version until a new `h2` is published.

* Tokio 0.3 doesn't currently have vectored IO APIs, so Hyper has
  (hopefully temporarily) lost its vectored IO support

* `AsyncRead::poll_read_buf` and `AsyncWrite::poll_write_buf` are being
  moved to `tokio-util`. Currently, they're not available, so we fall
  back to slicing and advancing `Bytes` and `BytesMut` buffers
  ourselves. When support lands upstream, we should change back.

* Ideally, Tokio's new `TcpSocket` type should gain support for more of
  the socket options hyper needs to set. If this happens, we should be
  able to remove some messy code involving `AsRawFd`/`FromRawFd` and
  `socket2`.

* This also bumps Hyper's minimum supported Rust version to 1.45.2,
  which is Tokio's minimum supported version. Tokio 0.3 doesn't build
  on 1.39.0

Closes #2309

Co-authored-by: Urhengulas <johann.hemmann@code.berlin>

### DIFF
--- a/src/common/drain.rs
+++ b/src/common/drain.rs
@@ -1,7 +1,9 @@
 use std::mem;
 
 use pin_project::pin_project;
-use tokio::sync::{mpsc, watch};
+use tokio::stream::Stream;
+use tokio::sync::mpsc;
+use tokio02::sync::watch;
 
 use super::{task, Future, Never, Pin, Poll};
 
@@ -30,7 +32,9 @@ pub struct Signal {
     _tx: watch::Sender<Action>,
 }
 
+#[pin_project::pin_project]
 pub struct Draining {
+    #[pin]
     drained_rx: mpsc::Receiver<Never>,
 }
 
@@ -66,8 +70,8 @@ impl Signal {
 impl Future for Draining {
     type Output = ();
 
-    fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
-        match ready!(self.drained_rx.poll_recv(cx)) {
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+        match ready!(self.project().drained_rx.poll_next(cx)) {
             Some(never) => match never {},
             None => Poll::Ready(()),
         }
--- a/src/common/drain.rs
+++ b/src/common/drain.rs
@@ -2,21 +2,12 @@ use std::mem;
 
 use pin_project::pin_project;
 use tokio::stream::Stream;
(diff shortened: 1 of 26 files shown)


src/common/drain.rs
```diff
@@ -1,7 +1,9 @@
 use std::mem;
 
 use pin_project::pin_project;
-use tokio::sync::{mpsc, watch};
+use tokio::stream::Stream;
+use tokio::sync::mpsc;
+use tokio02::sync::watch;
```
seanmonstar, line 6: :(

I can't remember, is the way forward here to change so just holding a `Pin<Box<dyn Future>>`, that we build by using the async fn in a loop until draining is signaled?

seanmonstar (commented):
Some of the benchmarks are still unhappy, they're guarded by the `nightly` cargo feature.

src/common/drain.rs
hawkw: Unfortunately, the current `watch` API in tokio 0.3 doesn't provide a (safe) way of doing that, since the `changed()` function borrows the `Watch` (https://docs.rs/tokio/0.3.3/tokio/sync/watch/struct.Receiver.html#method.changed).

hawkw: fixed! 7fd643a

src/server/tcp.rs
```diff
@@ -98,9 +102,46 @@ impl AddrIncoming {
...
                             trace!("error trying to set TCP keepalive: {}", e);
                         }
+
+                        // Take ownershop of the fd/socket back from the socket2
+                        // `Socket`, so that socket2 doesn't try to close it
+                        // when it's dropped.
+                        #[cfg(unix)]
+                        drop(std::os::unix::io::IntoRawFd::into_raw_fd(socket));
+                        #[cfg(windows)]
+                        drop(std::os::windows::io::IntoRawFd::into_raw_fd(socket));
```
seanmonstar, line 144: Looks like this should `IntoRawSocket`.

hawkw: Whoops, good catch!
```suggestion
                        drop(std::os::windows::io::IntoRawSocket::into_raw_socket(socket));
```

seanmonstar:
Finished in #2319

PR #2318, opened 2020-11-04
fix(server): fix graceful_shutdown on freshly opened connections

### LINKED ISSUE #1885: Graceful shutdown never finishes when there is opened connection
This will never exit. Something similar is doing google chrome. It creates two connections but sends request only to one of them. Fortunately chrome has some timeout so it closes that unused connection after some time and only after that finishes also hyper graceful shutdown.

When is this program run with argument then after one second it creates tcp connection and waits indefinitely without closing or dropping TcpStream. When is called without arguments then hyper server with graceful shutdown is created, it will run itself with argument so it will create child client process and never waits for child termination, then after 10 seconds is initiated graceful shutdown.

``` rust
extern crate hyper;
extern crate futures;
extern crate tokio_tcp;
extern crate log;
extern crate pretty_env_logger;

use hyper::{Body, Response, Server};
use hyper::rt::Future;
use hyper::service::service_fn_ok;
use hyper::server::conn::AddrStream;
use log::info;

fn main()
{
	let addr = ([127, 0, 0, 1], 3000).into();

	if std::env::args().nth(1).is_some()
	{
		std::thread::sleep_ms(1000);
		std::mem::forget(tokio_tcp::TcpStream::connect(&addr).wait().unwrap());
		loop {std::thread::sleep_ms(1000);}
	} 

	std::env::set_var("RUST_LOG", "debug");
	pretty_env_logger::init_timed();

	let (tx, rx) = futures::sync::oneshot::channel::<()>();

	let server = Server::bind(&addr)
		.serve(hyper::service::make_service_fn(|addr: &AddrStream|
		{
			let addr = addr.remote_addr();
			info!("connection {}", addr);
			service_fn_ok(move |_|
			{
				info!("request {}", addr);
				Response::new(Body::from("Hello, World!"))
			})
		}))
		.with_graceful_shutdown(rx)
		.map_err(|e| eprintln!("server error: {}", e));

	std::process::Command::new("cargo").args(&["run", "client"]).spawn().unwrap();

	std::thread::spawn(move ||
	{
		std::thread::sleep_ms(10000);
		tx.send(()).unwrap();
		info!("shutting down");
	});

	hyper::rt::run(server);
	info!("quiting");
}
```

Possible output:
```
    Finished dev [unoptimized + debuginfo] target(s) in 5.10s
     Running `target/debug/test`
 2019-07-28 21:30:00 DEBUG tokio_reactor > adding I/O source: 0
 2019-07-28 21:30:00 DEBUG tokio_reactor::registration > scheduling Read for: 0
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/test client`
 2019-07-28 21:30:02 INFO  test                        > connection 127.0.0.1:65114
 2019-07-28 21:30:02 DEBUG tokio_reactor::registration > scheduling Read for: 0
 2019-07-28 21:30:02 DEBUG tokio_reactor               > adding I/O source: 4194305
 2019-07-28 21:30:02 DEBUG tokio_reactor::registration > scheduling Read for: 1
 2019-07-28 21:30:10 INFO  test                        > shutting down
 2019-07-28 21:30:10 DEBUG hyper::server::shutdown     > signal received, starting graceful shutdown
 2019-07-28 21:30:10 DEBUG tokio_reactor               > dropping I/O source: 0
 2019-07-28 21:30:10 DEBUG tokio_reactor::registration > scheduling Read for: 1
 2019-07-28 21:30:10 DEBUG tokio_reactor::registration > scheduling Read for: 1
```

flumm:
When opening a http1 connection, the keep_alive state is set to Busy,
preventing a graceful_shutdown in that case. With this patch the
keep_alive disabling and closing of idle connections is decoupled and
should fix this issue.

Closes #1885

Although all tests go through (afaict), i am not entirely sure this is desired behaviour, though
we run into an issue where we use the graceful_shutdown for a daemon reload and an
open connection (with no data) can block our reload... (this issue fixes that)

Signed-off-by: Dominik Csapak <d.csapak@proxmox.com>

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -42,7 +42,7 @@ where
                 allow_half_close: false,
                 cached_headers: None,
                 error: None,
-                keep_alive: KA::Busy,
+                keep_alive: KA::Idle,
                 method: None,
                 title_case_headers: false,
                 notify_read: false,
@@ -686,7 +686,7 @@ where
         self.state.close_write();
     }
 
-    pub fn disable_keep_alive(&mut self) {
+    pub fn close_idle_connection(&mut self) {
         if self.state.is_idle() {
             trace!("disable_keep_alive; closing idle connection");
             self.state.close();
@@ -694,6 +694,10 @@ where
             trace!("disable_keep_alive; in-progress connection");
             self.state.disable_keep_alive();
         }
+    }
+
+    pub fn disable_keep_alive(&mut self) {
+        self.state.disable_keep_alive();
     }
 
     pub fn take_error(&mut self) -> crate::Result<()> {
--- a/src/proto/h1/dispatch.rs
+++ b/src/proto/h1/dispatch.rs
@@ -71,8 +71,8 @@ where
         }
     }
 
-    pub fn disable_keep_alive(&mut self) {
-        self.conn.disable_keep_alive();
+    pub fn close_idle_connection(&mut self) {
+        self.conn.close_idle_connection();
         if self.conn.is_write_closed() {
             self.close();
         }
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -568,7 +568,7 @@ where
     pub fn graceful_shutdown(self: Pin<&mut Self>) {
         match self.project().conn {
             Some(ProtoServer::H1(ref mut h1)) => {
-                h1.disable_keep_alive();
+                h1.close_idle_connection();
             }
             Some(ProtoServer::H2(ref mut h2)) => {
                 h2.graceful_shutdown();


flumm:
seems most of the ci tests time out, and the one test on macos that fails i cannot reproduce here on linux (i have no access to a mac)

seanmonstar:
Thanks for the PR!

I'm not sure that accepting a connection should immediately put in an idle state, ready to shutdown. We'd usually assume that a new connection means a request is coming soon. It also affects what a parsing error means.

I think the issue this is linking to should instead just be dealt with by a user selecting on a timeout after triggering a graceful shutdown. It may be that the current graceful shutdown API doesn't easily allow for that, I haven't thought too hard about that part.

flumm:
> Thanks for the PR!
> 
> I'm not sure that accepting a connection should immediately put in an idle state, ready to shutdown. We'd usually assume that a new connection means a request is coming soon.

ok no problem

>  It also affects what a parsing error means.

in what way exactly? (just to understand)

> 
> I think the issue this is linking to should instead just be dealt with by a user selecting on a timeout after triggering a graceful shutdown. It may be that the current graceful shutdown API doesn't easily allow for that, I haven't thought too hard about that part.

we solved a part of the issue another way, but is there any mechanism to have a timeout on an incoming connection until the service is called? i did not really find an example of this

misos1:
> I'm not sure that accepting a connection should immediately put in an idle state, ready to shutdown. We'd usually assume that a new connection means a request is coming soon.

@seanmonstar It was already a long time ago when I was thinking about this but it is not true that a request can be coming soon on any already idle keep-alive connection? Why would it be ok to close such a connection but not a freshly opened connection on which still did not arrive first request?

PR #2323, opened 2020-11-07
Update bytes to v0.6

olix0r:
Depends on:

- [ ] hyperium/h2#497
- [ ] hyperium/http-body#29
- [ ] hyperium/http#440

### DIFF
--- a/src/proto/h1/io.rs
+++ b/src/proto/h1/io.rs
@@ -190,6 +190,8 @@ where
             self.read_buf.reserve(next);
         }
         let dst = self.read_buf.bytes_mut();
+        // Safety: `dst` may include uninitialized memory, but `ReadBuf` will
+        // never uninitialize memory that has already been initialized.
         let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
         let mut buf = ReadBuf::uninit(dst);
         match Pin::new(&mut self.io).poll_read(cx, &mut buf) {
--- a/src/proto/h1/io.rs
+++ b/src/proto/h1/io.rs
@@ -2,6 +2,7 @@ use std::cell::Cell;
 use std::cmp;
 use std::fmt;
 use std::io::{self, IoSlice};
+use std::mem::MaybeUninit;
 
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
@@ -188,7 +189,9 @@ where
         if self.read_buf_remaining_mut() < next {
             self.read_buf.reserve(next);
         }
-        let mut buf = ReadBuf::uninit(&mut self.read_buf.bytes_mut()[..]);
+        let dst = self.read_buf.bytes_mut();
+        let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
+        let mut buf = ReadBuf::uninit(dst);
         match Pin::new(&mut self.io).poll_read(cx, &mut buf) {
             Poll::Ready(Ok(_)) => {
                 let n = buf.filled().len();
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ include = [
 ]
 
 [dependencies]
-bytes = "0.5"
+bytes = "0.6"
 futures-core = { version = "0.3", default-features = false }
 futures-channel = "0.3"
 futures-util = { version = "0.3", default-features = false }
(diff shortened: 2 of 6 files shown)


Urhengulas:
> This doesn't currently compile due to the tokio::ReadBuf API.

@olix0r: Thank you for taking ownership for this :) Can you please elaborate what's holding the PR back?

olix0r:
@Urhengulas Sorry, that comment is outdated (addressed in 34e60f7). It looks like there are some remaining changes needed in examples that I can try to fix. I won't be offended if someone else takes this branch and gets it over the line, though ;)

olix0r:
Also, this is blocked on a few other PRs:

* https://github.com/hyperium/h2/pull/498 -- not currently passing CI
* https://github.com/hyperium/http-body/pull/30
* https://github.com/hyperium/http/pull/441

paolobarbolini:
My h2 PR does pass CI https://github.com/hyperium/h2/pull/497

olix0r:
@paolobarbolini whoops! didn't mean to double up on those PRs, I just didn't look before branching. Closed my open PRs in favor of yours.

paolobarbolini:
> It looks like there are some remaining changes needed in examples that I can try to fix

It probably just needs `bytes::buf::BufExt` to be changed into `bytes::Buf`

olix0r:
OK, build passes. gitcop won't tell me what it doesn't like(?!), so i'm going to rebase and force push.

src/proto/h1/io.rs
```diff
@@ -188,7 +189,9 @@ where
         if self.read_buf_remaining_mut() < next {
             self.read_buf.reserve(next);
         }
-        let mut buf = ReadBuf::uninit(&mut self.read_buf.bytes_mut()[..]);
+        let dst = self.read_buf.bytes_mut();
+        let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
```
Urhengulas, line 193: That line looks threatening..

Can you please add a comment saying what it is doing?

seanmonstar:
Don't worry about gitcop (it's not "required"), it's just a reminder to me to fix the commit message format.

src/proto/h1/io.rs
olix0r: I don't actually know what it's doing ;)

I borrowed this from https://github.com/tokio-rs/tokio/blob/d78655337a68bded305782a8a8b4ac7be42aa6a7/tokio/src/io/util/read_buf.rs#L53-L55, which was the only updated use of `ReadBuf::uninit` I could find.

I would appreciate feedback who understand these APIs better than I do.

Urhengulas: @seanmonstar ?

Urhengulas: But let's see who introduced it 🙂

Urhengulas: https://github.com/tokio-rs/tokio/discussions/3121

olix0r: Thanks, added a comment summarizing that discussion.

src/proto/h1/io.rs
```diff
@@ -188,7 +189,11 @@ where
         if self.read_buf_remaining_mut() < next {
             self.read_buf.reserve(next);
         }
-        let mut buf = ReadBuf::uninit(&mut self.read_buf.bytes_mut()[..]);
+        let dst = self.read_buf.bytes_mut();
+        // Safety: `dst` may include uninitialized memory, but `ReadBuf` will
+        // never uninitialize memory that has already been initialized.
+        let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
+        let mut buf = ReadBuf::uninit(dst);
         match Pin::new(&mut self.io).poll_read(cx, &mut buf) {
```
hawkw, line 197: An alternative to all of this (which avoids the scary-looking pointer cast that @Urhengulas was concerned about) is to just use `tokio_util::io::poll_read_buf` from the latest `tokio-util` here:
```suggestion
        match tokio_util::io::poll_read_buf(Pin::new(&mut self.io), cx, &mut self.read_buf) {
```

That would mean adding a `tokio-util` dependency, but it's already a transitive dependency via `h2`.

src/proto/h1/io.rs
Urhengulas: Perfect 😄

src/proto/h1/io.rs
Urhengulas: That seems like a good solution!

And this is a useful function which can't be found in the docs... (code: https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/lib.rs#L67-L134).

hawkw:
I believe #2339 closes this.

PR #2331, opened 2020-11-17
chore(ci): check all feature combinations in CI

seanmonstar:


.github/workflows/CI.yml
```diff
@@ -101,14 +104,14 @@ jobs:
...
-      - name: Test
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: ${{ matrix.features }}
+      - name: Install cargo-hack
+        run: cargo install cargo-hack
+
+      - name: check --each-feature
+        run: cargo hack check --each-feature -Z avoid-dev-dependencies
```
taiki-e, line 114:
```suggestion
        run: cargo hack check --each-feature -Z avoid-dev-deps
```

PR #2335, opened 2020-11-18
refac: use tokio's TcpSocket for more sockopts

hawkw:
We still need `socket2` specifically for TCP keepalive, but we may just
end up removing that soon?

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -569,10 +569,11 @@ fn connect(
     connect_timeout: Option<Duration>,
 ) -> Result<impl Future<Output = Result<TcpStream, ConnectError>>, ConnectError> {
     // TODO(eliza): if Tokio's `TcpSocket` gains support for setting the
-    // keepalive timeout and send/recv buffer size, it would be nice to use that
-    // instead of socket2, and avoid the unsafe `into_raw_fd`/`from_raw_fd`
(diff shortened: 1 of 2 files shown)


src/client/connect/http.rs
```diff
@@ -569,29 +569,30 @@ fn connect(
...
     socket
         .set_nonblocking(true)
         .map_err(ConnectError::m("tcp set_nonblocking error"))?;
 
+    if let Some(dur) = config.keep_alive_timeout {
+        socket
+            .set_keepalive(Some(dur))
+            .map_err(ConnectError::m("tcp set_keepalive error"))?;
+    }
+
```
hawkw, line 595: @seanmonstar should I just remove keepalive from the hyper API instead? That way, we could just drop the socket2 dep entirely.

seanmonstar: Humm.... *maybe*. It is a confusing property, but people do enable it to notice when a connection has died... :shrug:

hawkw: Yeah...it's looking like Tokio will probably not add support for more sophisticated configuration (e.g. keepalive retries/probe interval/etc). But it would be nice if it at least had an API for enabling/disabling keepalive and maybe setting the initial timeout before keepalive probes are sent, which should be more portable.

PR #2342, opened 2020-11-24
remove dependency on `tracing`'s `log` feature

### LINKED ISSUE #2326: Remove log feature of tracing
The `log` feature of `tracing` should not be enabled by default in hyper (or h2). Since people likely are depending on the `log` records, this is a breaking change, and should be done in 0.14 (with a note in the release notes).

mikeyhew:
fixes #2326

I wasn't sure what the right way to update the changelog was, so I just used my best guess. Let me know if I should do something differently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ httpdate = "0.3"
 httparse = "1.0"
 h2 = { git = "https://github.com/hyperium/h2", optional = true }
 itoa = "0.4.1"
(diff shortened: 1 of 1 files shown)


mikeyhew:
how do I see the gitcop error message? The link just goes to https://gitcop.com/

seanmonstar:
Thanks! You can leave the changelog update out, those are automatically construct with a script at release time.

mikeyhew:
OK, I've removed the changelog update.

seanmonstar:
Thanks!

PR #2350, opened 2020-11-30
docs(client): add example of client conn usage

campbellC:
Add basic, module level example for the Builder performing a handshake,
spawning a task to run the Connection and sending a single request and
receiving the response.

Addresses #2272

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -7,6 +7,44 @@
 //!
 //! If don't have need to manage connections yourself, consider using the
 //! higher-level [Client](super) API.
+//!
+//! ## Example
+//! A simple example that uses the `SendRequest` struct to talk HTTP over a Tokio TCP stream
+//! ```no_run
+//! # #[cfg(all(feature = "client", feature = "http1", feature = "runtime"))]
+//! # mod rt {
+//! use http::{Request, StatusCode};
+//! use hyper::{client::conn::Builder, Body};
+//! use tokio::net::TcpStream;
+//!
+//! #[tokio::main]
+//! async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//!     let target_stream = TcpStream::connect("example.com:80").await?;
+//!
+//!     let (mut request_sender, connection) = Builder::new()
+//!         .handshake::<TcpStream, Body>(target_stream)
+//!         .await?;
+//!
+//!     // spawn a task to poll the connection and drive the HTTP state
+//!     tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


src/client/conn.rs
```diff
@@ -7,6 +7,41 @@
...
+//! A simple example that uses the `SendRequest` struct to talk HTTP over a Tokio TCP stream
+//! ```no_run
+//! # #[cfg(all(feature = "client", feature = "http1", feature = "runtime"))]
+//! # mod rt {
+//! use http::{Request, StatusCode};
+//! use hyper::{client::conn::Builder, Body};
+//! use tokio::net::TcpStream;
+//!
+//! #[tokio::main]
+//! async fn main() {
```
seanmonstar, line 21: Most of the docs follow the style of using `?` instead of `.unwrap()`, so I'd change this to `async fn main() -> Result<(), Box<dyn std::error::Error>> {` (and replace all the `unwrap`s with `?`).

src/client/conn.rs
```diff
@@ -7,6 +7,41 @@
...
+//!     let target_stream = TcpStream::connect("example.com:80").await.unwrap();
+//!
+//!     let handshake = Builder::new()
```
seanmonstar, line 24: This could be written as `let (request_sender, connection) = ...`.

src/client/conn.rs
```diff
@@ -7,6 +7,41 @@
...
+//!         .handshake::<TcpStream, Body>(target_stream)
+//!         .await.unwrap();
+//!
+//!     // spawn a task to poll the connection and drive the HTTP state
+//!     let connection = handshake.1;
+//!     tokio::spawn(connection.without_shutdown());
```
seanmonstar, line 30: I don't think the `without_shutdown` is required for the example. Perhaps if just spawning into the background, we could show printing an error if it occurs, whatcha think?

seanmonstar (changes requested):
Thanks for the PR! This looks pretty clean, I left just a couple thoughts inline.

campbellC:
I made the changes you requested - I believe I understood you correctly on how to handle the connection instead of using without_shutdown but correct me if I'm wrong there. 

I did this in a ammended commit which perhaps is not the cleanest way in a PR (this is my first Github PR so sorry if this makes inter-commit changes harder to see)

seanmonstar (approved):
Yep, perfect!

And don't worry about having amended a commit, the diff is small enough that it's easiest to just look at the full change.

PR #2353, opened 2020-12-06
Chore: Move OnUpgrade state from Body to Request/Response

### LINKED ISSUE #2340: Refactor HTTP upgrades into Extensions
In #2086, it was outlined how the HTTP upgrades should be moved off the `Body` type (and into `Extensions`). The public facing part was done, now we just need to update the internal `proto::h1` stuff to stash them into the `Extensions` instead of on a private `Body` field.

In order to do so, the `proto::MessageHead` will need to gain an `extensions` field, so we can place them there.

aknuds1:
Move OnUpgrade state from Body to http::Request/http::Response extensions.

Fixes #2340.

### DIFF
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -317,26 +317,34 @@ mod sealed {
     }
 
     impl CanUpgrade for http::Request<crate::Body> {
-        fn on_upgrade(self) -> OnUpgrade {
-            self.into_body().take_upgrade()
+        fn on_upgrade(mut self) -> OnUpgrade {
+            self.extensions_mut()
+                .remove::<OnUpgrade>()
+                .unwrap_or_else(OnUpgrade::none)
         }
     }
 
     impl CanUpgrade for &'_ mut http::Request<crate::Body> {
         fn on_upgrade(self) -> OnUpgrade {
-            self.body_mut().take_upgrade()
+            self.extensions_mut()
+                .remove::<OnUpgrade>()
+                .unwrap_or_else(OnUpgrade::none)
         }
     }
 
     impl CanUpgrade for http::Response<crate::Body> {
-        fn on_upgrade(self) -> OnUpgrade {
-            self.into_body().take_upgrade()
+        fn on_upgrade(mut self) -> OnUpgrade {
+            self.extensions_mut()
+                .remove::<OnUpgrade>()
+                .unwrap_or_else(OnUpgrade::none)
         }
     }
 
     impl CanUpgrade for &'_ mut http::Response<crate::Body> {
         fn on_upgrade(self) -> OnUpgrade {
-            self.body_mut().take_upgrade()
+            self.extensions_mut()
+                .remove::<OnUpgrade>()
+                .unwrap_or_else(OnUpgrade::none)
         }
     }
 }
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -23,7 +23,6 @@ use crate::common::{task, watch, Pin, Poll};
 use crate::common::{Future, Never};
 #[cfg(all(feature = "http2", any(feature = "client", feature = "server")))]
 use crate::proto::h2::ping;
-use crate::upgrade::OnUpgrade;
 
 type BodySender = mpsc::Sender<Result<Bytes, crate::Error>>;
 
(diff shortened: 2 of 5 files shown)


src/upgrade.rs
```diff
@@ -316,27 +316,39 @@ mod sealed {
         fn on_upgrade(self) -> OnUpgrade;
     }
 
+    /*
```
seanmonstar, line 319: We should be able to keep this implementation (in fact, removing it is a "breaking change", as you noticed by having to fix up the examples).

aknuds1: @seanmonstar Yes, I don't intend to remove this in the end. This is just a temporary kludge in order to solve the cases I could understand how to make compile, since I'm not sure how to work around the immutability. Before finalizing the PR, these cases should definitely be fixed as well.

Are you able to provide guidance on how to implement `CanUpgrade` for `http::Request<crate::Body>` and `http::Response<crate::Body>` though?

seanmonstar: It'd be the same as for the `&mut _` version, since taking an argument by-value still means you can mutate them (thus calling `extensions_mut().remove::<Thingy>()`.

aknuds1: @seanmonstar Thanks for the tip - I tried that strategy before and couldn't hack it, but will try again.

aknuds1: @seanmonstar I think it actually worked! Not sure why I struggled so much with it before.

aknuds1:
@seanmonstar I came up with a scheme that I think makes sense, storing OnUpgrade state in MessageHead::extensions and moving the state into Response::extensions and Request::extensions. I get a compilation error due to missing trait implementations for Extensions however, could you help me understand how to solve them please?

```
✗ cargo build --all-features
   Compiling hyper v0.14.0-dev (/home/arve/Projects/hyper)
error[E0277]: the trait bound `Extensions: Clone` is not satisfied
  --> src/proto/mod.rs:29:5
   |
29 |     extensions: http::Extensions,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `Extensions`
   |
   = note: required by `clone`
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0369]: binary operation `==` cannot be applied to type `Extensions`
  --> src/proto/mod.rs:29:5
   |
29 |     extensions: http::Extensions,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0369]: binary operation `!=` cannot be applied to type `Extensions`
  --> src/proto/mod.rs:29:5
   |
29 |     extensions: http::Extensions,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

seanmonstar:
Those are probably from `derive`s on `MessageHead`. You can likely remove the ones that are causing a problem (`Clone`, `PartialEq`), I don't think they're used any more.

aknuds1:
@seanmonstar Thanks, I thought those traits were necessary. Removed them, fixed the compile :)

aknuds1:
I'll rebase to fix commit messages, since everything else looks fine.

src/upgrade.rs
```diff
@@ -317,26 +317,34 @@ mod sealed {
     }
 
     impl CanUpgrade for http::Request<crate::Body> {
-        fn on_upgrade(self) -> OnUpgrade {
-            self.into_body().take_upgrade()
+        fn on_upgrade(mut self) -> OnUpgrade {
+            self.extensions_mut()
+                .remove::<OnUpgrade>()
+                .unwrap_or(OnUpgrade::none())
```
seanmonstar, line 323: You can save a function call by writing this as `.unwrap_or_else(OnUpgrade::none)` (and in the lines below).

seanmonstar (commented):
Fantastic work! This is exactly what I had in mind. Just one comment inline.

aknuds1:
Thanks @seanmonstar! I applied your suggestion. BTW, I amended into the original commit, and rebased. Do you prefer this way of working on PRs, or would you prefer if I avoid rewriting the commit history (i.e., just merge, add new commits)?

seanmonstar (approved):

PR #2354, opened 2020-12-10
fix(client): drop HTTP_2 -> HTTP_11 coerce warning

bnoordhuis:
We use hyper in a proxy-like way where an incoming HTTP/2 request is
proxied to a HTTP/1 backend and it was reported that the warnings
were a limiting factor in the total throughput of the system.

While it can be worked around by explicitly setting the request version
to HTTP_11, it seems likely other people will hit this performance snag,
and I don't see much value in warning for this particular condition, so
I suggest removing them.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -333,7 +333,7 @@ impl Http1Transaction for Server {
                 Version::HTTP_10 => extend(dst, b"HTTP/1.0 "),
                 Version::HTTP_11 => extend(dst, b"HTTP/1.1 "),
                 Version::HTTP_2 => {
-                    warn!("response with HTTP2 version coerced to HTTP/1.1");
+                    debug!("response with HTTP2 version coerced to HTTP/1.1");
                     extend(dst, b"HTTP/1.1 ");
                 }
                 other => panic!("unexpected response version: {:?}", other),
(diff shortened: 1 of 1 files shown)


bnoordhuis:
> I don't see much value in warning for this particular condition

I didn't really go into detail but here's why I feel that way.

* One way of looking at them: they don't tell me anything I didn't already know - I know I'm proxying HTTP/2 to HTTP/1. They're superfluous.

* Another perspective: the warnings tell me neither _why_ it's wrong what I'm doing nor _what_ I should be doing instead. They're unhelpful.

seanmonstar:
Yea, I think I added this warning when previously it was hitting an `unreachable!()`. It was in support of a proxy, similar to what you mentioned (linkerd2). I had fixed the panic because panic bad, but figured not updating to HTTP/1.1 before proxing on was a mistake, so I left the warning in.

I guess one point is that with an http1 connection, you can either use 1.0 or 1.1 semantics, and by having the version be `HTTP_2`, you might not be aware that it's assuming 1.1 for you. Though, probably if you're already supporting 2, you want to use 1.1 instead of 1.0 if possible, so I guess that coercion isn't itself worth a warning. Perhaps just worth a debug/trace level?

bnoordhuis:
That sounds reasonable. I've changed them from `warn!` to `debug!`, PTAL.

seanmonstar (approved):

PR #2357, opened 2020-12-13
fix(h1): ignore chunked trailers

### LINKED ISSUE #2171: Decoding response from gRPC REST proxy results in "Invalid chunk end CR" error
I am integrating with a go application that uses the gRPC REST proxy to provide an HTTP API.
My HTTP client is `reqwest` which uses `hyper:0.13.3` at this stage.

The go application sends the following response:

```
HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Trailer: Grpc-Trailer-Content-Type
Date: Wed, 01 Apr 2020 00:16:52 GMT
Transfer-Encoding: chunked

5e
{"error":"there are no existing invoices","message":"there are no existing invoices","code":2}
0
Grpc-Trailer-Content-Type: application/grpc

{"error":"there are no existing invoices","message":"there are no existing invoices","code":2}%
```

I've added a test to the hyper test suite that reproduces the issue: https://github.com/thomaseizinger/hyper/commit/a608b28dd9d93a0b0754653c859f262c14d9fd59

I don't know whether this is a valid HTTP response. Reading through the HTTP spec, it seems like this grpc trailer stuff is violating the `Transfer-Encoding: chunked` spec because the chunked is advertised as 0 bytes which from what I understand means the response is over, yet there are more bytes following.

Interestingly, `curl` is able to handle the response just fine but that could just be an artifact of curl being very robust to these kind of mistakes.

Even if this is an invalid response, it would be nice if `hyper` could be robust here and just ignore everything that is coming after the `0` chunk size indicator.

alpire:
Previously, hyper returned an "Invalid chunk end CR" error on chunked
responses with trailers, as described in RFC 7230 Section 4.1.2. This
commit adds code to ignore the trailers.

Closes #2171

### DIFF
--- a/src/proto/h1/decode.rs
+++ b/src/proto/h1/decode.rs
@@ -55,6 +55,8 @@ enum ChunkedState {
     Body,
     BodyCr,
     BodyLf,
+    Trailer,
+    TrailerLf,
     EndCr,
     EndLf,
     End,
@@ -196,6 +198,8 @@ impl ChunkedState {
             Body => ChunkedState::read_body(cx, body, size, buf),
             BodyCr => ChunkedState::read_body_cr(cx, body),
             BodyLf => ChunkedState::read_body_lf(cx, body),
+            Trailer => ChunkedState::read_trailer(cx, body),
+            TrailerLf => ChunkedState::read_trailer_lf(cx, body),
             EndCr => ChunkedState::read_end_cr(cx, body),
             EndLf => ChunkedState::read_end_lf(cx, body),
             End => Poll::Ready(Ok(ChunkedState::End)),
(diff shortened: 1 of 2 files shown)


seanmonstar (commented):
Thanks so much for writing this up! I really appreciate you including a test too!

For completeness, would you mind adding a couple more tests (either in `tests/`, or just testing the decoder in `proto::h1::decode::tests`)? It'd be good to check that this will handle multiple trailer liners (it looks like it does correctly), and perhaps some error cases.

alpire:
Thanks for the review and feedback! I just added the additional tests you mentioned:
- A test with multiple trailers (in `tests/`)
- A test checking the new error case I introduced, occurring when the trailer is missing a new line (in `proto::h1::decode::tests`)

seanmonstar (approved):
Awesome work, thanks!

PR #2358, opened 2020-12-14
fix(http2): Ignore request body when the method has no defined payload semantics

### LINKED ISSUE #2279: HEAD request using http/2 fails on bodies that don't implement is_end_stream
Trying to do a HEAD request on an http/2 server that uses response bodies that don't implement `HttpBody::is_end_stream` (so that it sometimes return `true`) fail with an http2 protocol error.

I noticed this while wrapping a `hyper::Body` but not passing though `is_end_stream`

oguzbilgener:
Fixes #2279

- Assume end of stream when the request method is GET, HEAD, CONNECT or DELETE.
- Also noticed that the `http2_parallel_10` integration test case sends a request body in a GET request as it started to fail after my change. So I split it into two test cases, a GET without body and a POST with body.

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -221,7 +221,13 @@ where
                             headers::set_content_length_if_missing(req.headers_mut(), len);
                         }
                     }
-                    let eos = body.is_end_stream();
+                    // for methods that cannot have a body, assume it's end of
+                    // stream, in case the `HttpBody` implementation is incorrect
+                    let eos = if headers::method_has_defined_payload_semantics(req.method()) {
+                        body.is_end_stream()
+                    } else {
+                        true
+                    };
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for the PR!

I wrote a comment on the related issue, since reading through this made me think that a request technically *could* still send trailers even though there's normally no payload. Or maybe I'm wrong and missed a tiny detail in one of the RFCs...

oguzbilgener:
No worries, feel free to close the PR if it's no longer relevant. I'm going to read up on what you described. :slightly_smiling_face:

seanmonstar:
So sorry about that, based on what was mentioned in #2279, I think it's best to close this and that issue too. I believe it overlooked some things.

PR #2361, opened 2020-12-16
refactor(client): log socket option errors instead of returning error (#2359)

### LINKED ISSUE #2359: HttpConnector should not error if a socket option fails
An error trying to set a socket option should not error the entire connect attempt. Instead, they should just be logged (perhaps at `warn` level).

aeryz:
Socket option errors are converted to warnings. This closes #2359. But I am not sure if it is better to return error if "set_keepalive" fails in line 595.

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -338,8 +338,9 @@ where
 
         let sock = c.connect().await?;
 
-        sock.set_nodelay(config.nodelay)
-            .map_err(ConnectError::m("tcp set_nodelay error"))?;
+        if sock.set_nodelay(config.nodelay).is_err() {
+            warn!("tcp set_nodelay error");
+        }
 
         Ok(sock)
     }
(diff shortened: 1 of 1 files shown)


src/client/connect/http.rs
```diff
@@ -338,8 +338,9 @@ where
 
         let sock = c.connect().await?;
 
-        sock.set_nodelay(config.nodelay)
-            .map_err(ConnectError::m("tcp set_nodelay error"))?;
+        if sock.set_nodelay(config.nodelay).is_err() {
```
seanmonstar, line 341: Probably worth logging the actual error too, by making this `if let Err(err) = ...`, and then having `warn!("tcp set_nodelay error: {}", err);`.

src/client/connect/http.rs
```diff
@@ -592,8 +592,8 @@ fn connect(
         .map_err(ConnectError::m("tcp set_nonblocking error"))?;
 
     if let Some(dur) = config.keep_alive_timeout {
-        if socket.set_keepalive(Some(dur)).is_err() {
-            warn!("tcp set_keepalive error");
+        if let Err(e) = socket.set_keepalive(Some(dur)) {
+            warn!("tcp set_keepalive error: {}", e);
         }
```
aeryz: Hey @seanmonstar. Thanks for the review. As I mentioned earlier, I am not sure if not returning an error when "set_keepalive" fails is the right thing to do. Isn't it might be critical to the client?

seanmonstar: Hm, I don't think it should error... do you have a more detailed argument/reason to support that it should?

aeryz: I thought for a second that it might be critial to keep keepalive low to probe the connection more frequently but I think you are right. Warning is enough.

seanmonstar (approved):

PR #2377, opened 2020-12-24
Expose client connection types without proto feature

### LINKED ISSUE #2376: Connector implementation crates have to pick http1/http2
The low-level connection APIs like `Connected`/`Connection`/`HttpConnector` all require one of `http1` or `http2` to be set. This means that external crates like `hyper-openssl` have to pick one of those features to enable, which is kind of out of scope for what they actually need to do. It would probably make sense for those bits of the client API to only require `client` and not `http1` or `http2` to make things less weird.

benesch:
Make it possible to refer to Connected, Connection, HttpConnector, etc.
without enabling either of the http1/http2 features. This makes feature
selection work better for downstream libraries like hyper-openssl, which
don't want to commit to any particular protocol.

Fix #2376.


src/client/connect/http.rs
```diff
@@ -27,6 +27,7 @@ use super::{Connected, Connection};
 ///
 /// Sets the [`HttpInfo`](HttpInfo) value on responses, which includes
 /// transport information such as the remote socket address used.
+#[cfg_attr(docsrs, doc(cfg(feature = "tcp")))]
```
seanmonstar, line 30: I believe this attribute is inherited from it being on the `client::connect::http` module, which you also added.

src/client/mod.rs
```diff
@@ -48,1243 +48,22 @@
...
-    retry_canceled_requests: bool,
-    set_host: bool,
-    ver: Ver,
-}
-
-/// A `Future` that will resolve to an HTTP Response.
-///
-/// This is returned by `Client::request` (and `Client::get`).
-#[must_use = "futures do nothing unless polled"]
-pub struct ResponseFuture {
```
seanmonstar, line 99: This type needs to still be accessible at `hyper::client::ResponseFuture`, to prevent a breaking change.

seanmonstar (commented):
Thanks so much for working on this! Just a couple thoughts inline.

src/client/mod.rs
benesch: Eep! Great catch. Fixed!

src/client/connect/http.rs
benesch: I agree that it should be! But for some reason this is what happens if you don't manually add this attribute:

image

The "requires `tcp` feature" annotation gets added to all the methods, but not the type itself. I didn't dig into why, since adding the feature manually gave the results I expected:

image

seanmonstar: Weird, but thanks for checking!

src/client/mod.rs
seanmonstar: Did you push the fix? I can't seem to see it...

benesch: 🤦🏽‍♂️

benesch: Should be good now!

PR #2381, opened 2020-12-28
docs(client): Change docs to reflect HttpConnector relies on tcp feature

### LINKED ISSUE #2374: Documentation: HttpConnector requires the "tcp" feature
The docs.rs [documentation for HttpConnector](https://docs.rs/hyper/0.14.1/hyper/client/struct.HttpConnector.html) with hyper 0.14.1 says:

>This is supported on crate feature client and (crate features http1 or http2) only.

As I understand it this seems based on of the following in hyper's lib.rs:

```
cfg_feature! {
    #![all(feature = "client", any(feature = "http1", feature = "http2"))]

    pub mod client;
    pub use crate::client::Client;
}
```

But it seems the HttpConnector struct [also depends on the `tcp` feature](https://github.com/hyperium/hyper/commit/5b348b821c3f43d8dd71179862190932fcca6a1c#diff-1b84118fef4e05f6f4c73c9ad3c5d469ec065f0ef62e20448683dadf6f281c6dL78-R78).

```
#[cfg(feature = "tcp")]
pub use self::connect::HttpConnector;
```

I'm not sure how much control you have over this, but would it be possible to fix the doc to show the extra required feature? =)

SkamDart:
Resolves #2374 (I think)

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -65,8 +65,10 @@ use crate::body::{Body, HttpBody};
 use crate::common::{exec::BoxSendFuture, lazy as hyper_lazy, task, Future, Lazy, Pin, Poll};
 use crate::rt::Executor;
 
-#[cfg(feature = "tcp")]
+cfg_feature! {
+#![cfg(feature = "tcp")]
 pub use self::connect::HttpConnector;
+}
 
 pub mod conn;
 pub mod connect;


seanmonstar:
Thanks for the PR! It turns out this was address in #2377 also.

PR #2382, opened 2020-12-29
Expose server traits without httpX features

fanatid:
Same as https://github.com/hyperium/hyper/pull/2377 but for server traits


src/server/mod.rs
```diff
@@ -52,452 +52,466 @@
...
-    /// to being out-of-memory, and simply error (and shutdown). Setting this
-    /// option to false will allow that.
-    ///
-    /// For more details see [`AddrIncoming::set_sleep_on_errors`]
-    pub fn tcp_sleep_on_accept_errors(mut self, val: bool) -> Self {
-        self.incoming.set_sleep_on_errors(val);
-        self
-    }
-}
+// pub mod accept;
```
seanmonstar, line 68: Did you mean to delete this?

seanmonstar (commented):
Thanks for doing this, good idea!

src/server/mod.rs
fanatid: Yes, I thought I removed this :laughing: 
Fixed.

PR #2387, opened 2021-01-06
feat(body): add trailers to Body channel (#2260)

### LINKED ISSUE #2260: Add sending trailers on Body channel
The `Body::channel()` variant is a common simple way to send data as a `Body`. It would be useful to also be able to send trailers on it.

### Implementation

The code to change is on the `hyper::body::Sender`, and the relevant `Kind` variant. I could imagine two possible ways to implement this, and I'm not sure which is better yet.

1. Change the internal mpsc channel to send `enum Msg { Data(Bytes), Trailers(HeaderMap) }`. However, the `Kind::Chan` variant would need to grow some cache slots, since `poll_data` could pop a `Msg::Trailers`, and then that would need to cached so that `poll_trailers` returns its.
2. Add a `oneshot` channel to send the trailers. The downside here is the increased size of `Sender` and `Kind::Channel` (they both are carrying an extra pointer), and the extra allocation.

I think the 2nd option is probably going to be better, but measurements are king. We should compare the changes to the `benches/end_to_end` results to see if the added feature makes much of a difference. If it does, we could consider making an internal "skinnier channel" that is used by the h1 dispatcher.

aeryz:
Closes #2260.


aeryz:
@seanmonstar Should I also add "try_send_trailers" to the Sender?

seanmonstar:
Thanks for the PR!

> Should I also add "try_send_trailers" to the Sender?

I don't think that's necessary, unless you have a good reason for it?

aeryz:
> Thanks for the PR!
> 
> 
> 
> > Should I also add "try_send_trailers" to the Sender?
> 
> 
> 
> I don't think that's necessary, unless you have a good reason for it?

`try_send_data` is implemented for threads that doesn't have an async context. By using `try_send_trailers`, they can send trailers as well.

src/body/body.rs
```diff
@@ -628,7 +651,7 @@ mod tests {
 
         assert_eq!(
             mem::size_of::<Sender>(),
-            mem::size_of::<usize>() * 4,
+            mem::size_of::<usize>() * 5,
```
seanmonstar, line 654: This was the main thing I was a little worried about when filing the issue... Can you tell in benchmarks how much the increased size (and perhaps the creation of a `oneshot` for each request) effects things? If it's nothing, cool. But it'd be good to know for sure.

aeryz:
@seanmonstar I ran `end_to_end` benchmarks. 
No trailers -> https://pasteboard.co/JJfrwqV.png
Trailers  -> https://pasteboard.co/JJfsdsf.png

I ran each one multiple times to make sure that everything works fine and I saw that adding trailers didn't cause much of a difference.

src/body/body.rs
```diff
@@ -545,14 +559,23 @@ impl Sender {
...
     pub async fn send_data(&mut self, chunk: Bytes) -> crate::Result<()> {
         self.ready().await?;
-        self.tx
+        self.data_tx
             .try_send(Ok(chunk))
             .map_err(|_| crate::Error::new_closed())
     }
 
+    /// Send trailers on trailers channel.
+    pub fn send_trailers(&mut self, trailers: HeaderMap) -> crate::Result<()> {
```
seanmonstar, line 571: Even though not specifically required due to the internal implementation, I'm thinking this should be an `async fn`, so that we could adjust the implementation later if need be. But I'm open to hearing why it shouldn't.

src/client/conn.rs
```diff
@@ -63,7 +63,10 @@ use tower_service::Service;
 
 use super::dispatch;
 use crate::body::HttpBody;
-use crate::common::{task, exec::{BoxSendFuture, Exec}, Future, Pin, Poll};
+use crate::common::{
```
seanmonstar, line 66: Seems unrelated to the change?

seanmonstar (commented):
Alright, I'm fine with the performance implications (doesn't seem to be any). I've left a few final thoughts inline, but in general I'm happy with how this turned out, thanks!

src/client/conn.rs
aeryz: I accidentally format it while debugging I guess.

src/body/body.rs
```diff
@@ -382,7 +382,7 @@ impl HttpBody for Body {
                 ..
             } => match ready!(Pin::new(trailers_rx).poll(cx)) {
                 Ok(t) => Poll::Ready(Ok(Some(t))),
-                Err(_) => Poll::Ready(Err(crate::Error::new_closed())),
+                Err(_) => Poll::Ready(Ok(None)),
```
aeryz, line 385: @seanmonstar If we don't send any trailers, polling the oneshot receiver returns `Err(Canceled)` so if we return `Err(..)` in this case, receive process will result in error. That's why I changed this to `Ok(None)`. This way we don't have to send empty trailers. Is it ok though?

seanmonstar: Oh yes, this change is the correct behavior. Otherwise, when people upgrade hyper then suddenly in H2 when the trailers are polled they'd error the stream since they weren't sending any before.

seanmonstar (approved):
Thanks! Looks good to me.

PR #2393, opened 2021-01-11
Switch to pin-project-lite

### LINKED ISSUE #2388: Switch from pin-project to pin-project-lite
futures-rs [just switched](https://github.com/rust-lang/futures-rs/pull/2273) and tokio has been using the `-lite` one since 0.2.0. I haven't used either so I don't actually know how hard switching is, but it seems like a good idea since it greatly improves compile times for people who don't depend on syn & quote otherwise.

jplatte:
Will fix #2388.

~~Blocked on taiki-e/pin-project-lite#3.~~ ([workaround](https://github.com/hyperium/hyper/issues/2388#issuecomment-758304097))
~~Blocked on taiki-e/pin-project-lite#25.~~ (see comment below)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ httparse = "1.0"
 h2 = { version = "0.3", optional = true }
 itoa = "0.4.1"
 tracing = { version = "0.1", default-features = false, features = ["std"] }
-pin-project = "1.0"
+pin-project-lite = "0.2.4"
 tower-service = "0.3"
 tokio = { version = "1", features = ["sync"] }
 want = "0.3"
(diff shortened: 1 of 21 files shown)


paolobarbolini:
After looking back at the issue I was able to remove the use of PinnedDrop in #2405, so we won't have to wait for PinnedDrop to land in pin-project-lite

Cargo.toml
```diff
@@ -35,6 +35,7 @@ h2 = { version = "0.3", optional = true }
 itoa = "0.4.1"
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project = "1.0"
```
paolobarbolini, line 37: You forgot this :smiley: 

```suggestion
```

seanmonstar (approved):
Awesome work, thanks so much!

(Weird that Github doesn't notify me that your marked it ready to review 🤷 )

PR #2399, opened 2021-01-13
Document requirements for features

elpiel:
client, server & runtime features all require at least one of the features `http1` or `http2`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,10 +47,10 @@
 //!
 //! - `http1`: Enables HTTP/1 support.
 //! - `http2`: Enables HTTP/2 support.
-//! - `client`: Enables the HTTP `client`.
-//! - `server`: Enables the HTTP `server`.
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! While you do need a combination of features enabled to do many things, I believe what this addition is describing isn't the full story: It is possible *now* to enable `client` without `http1` or `http2`, to have access the `connect` stuff, for instance. Same with `server`.

elpiel:
@seanmonstar thank for your time taking a look at this!
How can we make it more clear then? I'll be happy to make changes based on your suggestions.

seanmonstar:
I think the blue boxes that appear in the documentation are probably pretty clear. We could say there on the main page something like "To fully utilize this feature, others like a protocol version may also need to be enabled." Would that help?

elpiel:
Makes sense to leave it like this as well. If I have better ideas I will open an issue to discuss them again.
Thanks!

PR #2400, opened 2021-01-13
Apply unreachable_pub lint

### LINKED ISSUE #2390: Lint for unreachable_pub
This is a good clean up to help detect any types or methods that seem public but aren't really. By adjusting their privacy, we can reduce accidentally making things public in case we later on made some module the re-exports them public.

To do this, start by adding `#![deny(unreachable_pub)]`, and then fixing all failures by making them `pub(crate)` or `pub(super)` as appropriate (prefer `super` to `crate` when possible).

taiki-e:
Closes #2390

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -272,7 +272,7 @@ where
         ResponseFuture { inner }
     }
 
-    pub(crate) fn send_request_retryable(
+    pub(super) fn send_request_retryable(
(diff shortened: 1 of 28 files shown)


taiki-e:
Oh, visibility depends on the features...

seanmonstar (approved):
Thanks!

PR #2401, opened 2021-01-14
Set doc_cfg attributes for `tcp` feature gated Server methods

nickelc:
The docs on `Server` are missing the required `tcp` feature.

### DIFF
--- a/src/server/server.rs
+++ b/src/server/server.rs
@@ -60,6 +60,7 @@ impl Server<AddrIncoming, ()> {
     ///
     /// This method will panic if binding to the address fails. For a method
     /// to bind to an address and return a `Result`, see `Server::try_bind`.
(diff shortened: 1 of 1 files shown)


src/server/server.rs
```diff
@@ -60,6 +60,7 @@ impl Server<AddrIncoming, ()> {
     ///
     /// This method will panic if binding to the address fails. For a method
     /// to bind to an address and return a `Result`, see `Server::try_bind`.
+    #[cfg_attr(docsrs, doc(cfg(feature = "tcp")))]
```
seanmonstar, line 63: Hm, do you know if we put this on the `impl Server` above, right next to the actual `cfg` attribute, will it still show nicely in the docs?

nickelc: oh yeah, the `cfg_feature` macro works fine.

seanmonstar (approved):

PR #2404, opened 2021-01-17
feat(body): reexport http_body::SizeHint

erikdesjardins:
Fix a minor papercut: you can't implement `hyper::body::HttpBody::size_hint` without importing `SizeHint` from another crate.


erikdesjardins:
I can't tell what it doesn't like about the commit message

seanmonstar (approved):
Thanks, great idea!

Don't worry about the message, I can tweak it when merging. For reference, I think it's complaining about a line being longer than 80 (100?) characters.

PR #2405, opened 2021-01-17
Remove useless uses of Pin

paolobarbolini:
While looking further into #2388 I noticed `UnboundedReceiver` was getting pinned even though [`UnboundedReceiver::poll_recv`](https://docs.rs/tokio/1.0.2/tokio/sync/mpsc/struct.UnboundedReceiver.html#method.poll_recv) takes `&mut self`. This detail was probably missed in the migration from tokio 0.3, which [previously used `Stream::poll_next`](https://github.com/hyperium/hyper/pull/2369/files#diff-414def8886b02bb26f69e6e93e08fa35e230e7faf2b37cf1614cdf58fe987a5dL153-R154), but can now be applied, unblocking #2393.

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -138,25 +138,22 @@ impl<T, U> Clone for UnboundedSender<T, U> {
     }
 }
 
-#[pin_project::pin_project(PinnedDrop)]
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Nice catch!

PR #2406, opened 2021-01-19
docs(client): fix typo in httpbin.org domain

mikeroelens:
Saving the next dev that copy pastes this example from banging their head trying to figure out what's wrong with their DNS 😁


seanmonstar (approved):
Heh, woops! :D

Issue #2421, opened 2021-02-05
Build failure with v0.14.3


clehner:
Hi,

I get some build errors in CI with the newly released hyper v0.14.3:

In https://github.com/spruceid/didkit/runs/1842461977#step:6:437

```
error[E0433]: failed to resolve: could not find `h2` in `proto`
  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.3/src/client/conn.rs:92:24
   |
92 |             h2: proto::h2::ClientTask<B>, _phantom: PhantomData<fn(T)>,
   |                        ^^ could not find `h2` in `proto`

error[E0433]: failed to resolve: could not find `h2` in `proto`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.3/src/server/conn.rs:187:24
    |
187 |             h2: proto::h2::Server<Rewind<T>, S, B, E>,
    |                        ^^ could not find `h2` in `proto`

error[E0412]: cannot find type `PhantomData` in this scope
  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.3/src/client/conn.rs:92:53
   |
92 |             h2: proto::h2::ClientTask<B>, _phantom: PhantomData<fn(T)>,
   |                                                     ^^^^^^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
49 | use core::marker::PhantomData;
   |
49 | use crate::client::service::PhantomData;
   |
49 | use pin_project_lite::__private::PhantomData;
   |
49 | use std::marker::PhantomData;
   |

error[E0412]: cannot find type `Rewind` in this scope
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.3/src/server/conn.rs:187:35
    |
187 |             h2: proto::h2::Server<Rewind<T>, S, B, E>,
    |                                   ^^^^^^ not found in this scope
    |
help: consider importing this struct
    |
46  | use crate::common::io::Rewind;
    |

error[E0599]: no variant named `H2` found for enum `server::conn::ProtoServer<T, B, S, E>`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.3/src/server/conn.rs:185:9
    |
166 | / pin_project! {
167 | |     #[project = ProtoServerProj]
[...]

PR #2423, opened 2021-02-06
chore(ci): use --feature-powerset --depth 2 in features check

taiki-e:
This allows catching the issue like https://github.com/hyperium/hyper/issues/2421.

> > _also_ our CI didn't notice the failure of combining `--features http1,client,server`.
>
> This can be fixed by using `cargo hack --feature-powerset` ([all combination of features](https://github.com/taiki-e/cargo-hack#--feature-powerset)) instead of `cargo hack --each-feature` ([each feature](https://github.com/taiki-e/cargo-hack#--each-feature)) in CI. 
> However, if we use `--feature-powerset`, there will be more than 700 feature combinations, so we may need to use `--depth` ([specify a max number of simultaneous feature flags of --feature-powerset](https://github.com/taiki-e/cargo-hack/issues/58)) flag to limit the number of combinations.

We have too many features to check the full powerset (776), so limit the max number of simultaneous feature flags to 2 by [`--depth` option](https://github.com/taiki-e/cargo-hack/pull/59). I think this should be sufficient in most cases as carllerche said in https://github.com/taiki-e/cargo-hack/issues/58.

> Tokio has too many features to test the power set. However, **most** issues can be reproduced by testing at most 2 features simultaneously.

In fact, #2421 can be reproduced with a combination of two feature flags. (--features client,http1)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -111,8 +111,8 @@ jobs:
       - name: Install cargo-hack
         run: cargo install cargo-hack
 
-      - name: check --each-feature
-        run: cargo hack check --each-feature --skip ffi -Z avoid-dev-deps
+      - name: check --feature-powerset
+        run: cargo hack check --feature-powerset --depth 2 --skip ffi -Z avoid-dev-deps
 
   ffi:
     name: Test C API (FFI)


taiki-e:
CI failure is unrelated and will be fixed in #2424.

seanmonstar:
This seems right. There may be a couple extra combinations that are interesting, but perhaps they should just be explicitly added in CI instead of blowing up the depth.

seanmonstar (approved):

PR #2437, opened 2021-02-16
refactor: Use async/await more

Marwes:
I thought the `Unpin` relaxation in `Lazy` could be used to remove some more combinators in `connect_to`, but it seems that `future::select` is to restrictive, and won't work with `pin_mut!`. Left it in anyway, in case it ends up useful elsewhere

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -681,7 +681,7 @@ where
 
     /// Prevent shutdown of the underlying IO object at the end of service the request,
     /// instead run `into_parts`. This is a convenience wrapper over `poll_without_shutdown`.
-    pub fn without_shutdown(self) -> impl Future<Output = crate::Result<Parts<I, S>>>
+    pub async fn without_shutdown(self) -> crate::Result<Parts<I, S>>
     where
         S: Unpin,
         S::Future: Unpin,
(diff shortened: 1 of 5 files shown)


src/server/conn.rs
```diff
@@ -681,7 +681,7 @@ where
 
     /// Prevent shutdown of the underlying IO object at the end of service the request,
     /// instead run `into_parts`. This is a convenience wrapper over `poll_without_shutdown`.
-    pub fn without_shutdown(self) -> impl Future<Output = crate::Result<Parts<I, S>>>
+    pub async fn without_shutdown(self) -> crate::Result<Parts<I, S>>
```
seanmonstar, line 684: I think technically this a breaking change, since an async fn is automatically `!Unpin`, right?

seanmonstar (changes requested):
Looks like a nice clean up, thanks!

J-Vin (approved):

src/server/conn.rs
Marwes: Forgot that the auto trait propagation made the earlier future `Unpin`, despite it not being explicitly set. Reverted

seanmonstar (approved):

PR #2439, opened 2021-02-18
Creating AddrIncoming from an existed TcpListener

zonyitoo:


src/server/tcp.rs
```diff
@@ -51,6 +51,19 @@ impl AddrIncoming {
     pub fn bind(addr: &SocketAddr) -> crate::Result<Self> {
         AddrIncoming::new(addr)
     }
+    
+    /// Creates a new `AddrIncoming` from an existed `TcpListener`
+    pub fn from_listener(listener: TcpListener) -> crate::Result<Self> {
```
seanmonstar, line 56: With this addition, `from_std` can be written to call it directly, ya?

zonyitoo: Yes, of course.

src/server/tcp.rs
```diff
@@ -36,6 +36,16 @@ impl AddrIncoming {
...
         let listener = TcpListener::from_std(std_listener).map_err(crate::Error::new_listen)?;
+        AddrIncoming::from_listener(listener)
+    }
+
+    /// Creates a new `AddrIncoming` binding to provided socket address.
+    pub fn bind(addr: &SocketAddr) -> crate::Result<Self> {
+        AddrIncoming::new(addr)
+    }
+
+    /// Creates a new `AddrIncoming` from an existed `TcpListener`
```
seanmonstar, line 47:
```suggestion
    /// Creates a new `AddrIncoming` from an existing `tokio::net::TcpListener`.
```

zonyitoo:
Great. When will be the next release?

PR #2440, opened 2021-02-19
docs(server): add server example using `tower::make::Shared`

davidpdrsn:
`tower` 0.4.5 introduced `Shared` which is a `MakeService` that produces
services by cloning an inner service. This works quite well with `hyper`
if your service doesn't need the incoming connection and implements
`Clone`.

However that might not be entirely obvious so I thought it made sense to
add an example to the docs.

I wasn't quite sure if the example should go in the server or service
module docs but since there already is an example using
`make_service_fn` in the server docs I opted to add it there. Let me
know if you'd rather have it somewhere else.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ tokio = { version = "1", features = [
 ] }
 tokio-test = "0.4"
 tokio-util = { version = "0.6", features = ["codec"] }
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
I like it, thanks!

Issue #2407, opened 2021-01-25
Client auto Host header should omit the port if it is default


andyHa:
The HTTP spec states that the "host" header should only include the port if it isn't the standard port.

However, if an URL like `https://my-lovely-bucket.s3-eu-central-1.amazonaws.com:443` is accessed (as generated by the AWS SDK), a port header like `host: s3-eu-central-1.amazonaws.com:443` instead of `host: s3-eu-central-1.amazonaws.com` is submitted. This will break the computed signature of the S3 request.

Note that both (modern browsers - checked with latest FF and chrome) as well as good old curl both strip the port number for standard ports.

I'd say this has to be fixed in [client.rs:240](https://github.com/hyperium/hyper/blob/master/src/client/client.rs#L240) - I'd be happy to provide a PR if this helps.

andyHa:
It turns out, the AWS client only appends the port if you force it to do so - still I'd argue, the behavior should be along the sepc and other implementations...

seanmonstar:
Yea, I believe the intent is to just use the value the user provided as-is. I agree that the client should repair some things, but is this one that is worth changing for a user?

andyHa:
> Yea, I believe the intent is to just use the value the user provided as-is. I agree that the client should repair some things, but is this one that is worth changing for a user?

Let's put the AWS client aside for a second - my point is "using the value to user provided" is wrong. At least curl, FF and chrome *remove the value even if provided by the user*. Still, the spec itself (https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23) is a bit vaque. Hoewever, I'd recommend maintianing compatibility to the big players ;-)

seanmonstar:
Alright, seems fair. Since this is an automatic host header generation, we can try following the pack, and if the user already provided a `host` header, we'd leave it alone, since that's explicit. Seem right?

PR #2441, opened 2021-02-20
omitting default port from Host header

CfirTsabari:
Fixes hyperium/hyper#2407


seanmonstar (approved):
Excellent, thank you!

Issue #2428, opened 2021-02-09
Remove need for cbindgen type renames


seanmonstar:
The `cbindgen.toml` file contains some renaming rules, but I think just making all the public C API types using their C API name would be the clearest. So, this would mean replacing the name in `src/ffi`, and then removing the rule from `cbindgen.toml`.

PR #2442, opened 2021-02-21
refactor(ffi): Removed need for cbindgen type renames

CfirTsabari:
Fixes hyperium/hyper#2428


seanmonstar (approved):
Excellent work, thanks so much! ❤️

PR #2458, opened 2021-03-10
feat(tcp): add support for custom socket options

roman-kashitsyn:
This change extends to AddrIncoming API to support setting custom
options before binding the server socket.  This change introduces one
such option: `reuse_port`, which enables the `SO_REUSEPORT` option on
the socket on UNIX platforms.

Example:

    use hyper::server::conn::{AddrIncoming, SocketOptions};
    let incoming = AddrIncoming::bind_with_option(
        "0.0.0.0:80".parse().unwrap(),
        SocketOptions::new().reuse_port(true)
    );

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ want = "0.3"
 # Optional
 
 libc = { version = "0.2", optional = true }
-socket2 = { version = "0.3.16", optional = true }
+socket2 = { version = "0.3.16", optional = true, features = ["reuseport"] }
 
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the PR! Is the motivation that you want to make these options more convenient in hyper? Since it's possible to construct a `TcpListener` and then use `AddrIncoming::from_std`, it should already be _possible_ to be done outside of hyper, right?

roman-kashitsyn:
> Since it's possible to construct a `TcpListener` and then use `AddrIncoming::from_std`, it should already be _possible_ to be done outside of hyper, right?

I don't think so, `from_std` is private `pub(super)`. But if you're OK with making it public, I'll be totally happy with this solution.

seanmonstar:
Oh sorry, there's `from_listener` which is publicly available. It just requires you to convert to a `tokio::net::TcpListener` first.

PR #2459, opened 2021-03-10
feat(h2): log negotiated concurrency limits at DEBUG

hawkw:
Upstream PRs hyperium/h2#513 and hyperium/h2#516 added functions for
exposing the connection concurrency limits negotiated by a remote peer.
Recording these settings can be useful for debugging purposes.

In the higher-level `Client` and `Server` APIs, it's not easily possible to
provide a way for users to directly access settings on a particular
connection, since the client abstracts over a connection pool and the
server abstracts over a connect loop. Therefore, the user can't access
these settings directly. Instead, this branch adds new log messages at
the DEBUG level for recording negotiated concurrency limits.

In a follow-up, we can also add accessors to the `Connection` types in
the lower-level `client::conn` and `server::conn` APIs.

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -7,6 +7,7 @@ use futures_util::future::{self, Either, FutureExt as _, TryFutureExt as _};
 use futures_util::stream::StreamExt as _;
 use h2::client::{Builder, SendRequest};
 use tokio::io::{AsyncRead, AsyncWrite};
+use tracing::debug;
 
 use super::{decode_content_length, ping, PipeToSendStream, SendBuf};
 use crate::body::HttpBody;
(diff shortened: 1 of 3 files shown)


src/proto/h2/client.rs
```diff
@@ -105,6 +106,11 @@ where
         .await
         .map_err(crate::Error::new_h2)?;
 
+    debug!(
+        server.max_concurrent_send_streams = conn.max_concurrent_send_streams(),
```
seanmonstar, line 110: So I just wanted to talk through the naming here, cause it twisted me a little. This is in he `client` module, and logs this as the *server's* max send streams... But is this option actually the max streams the *server* can send? I thought `conn.max_concurrent_send_streams()` was how many this connection is allowed to send (so in this module, the client's send limit). Or am I all twisted up?

hawkw: it's the number of streams that the server _told us_ we can send. i wanted to make it clear that the setting was _coming_ from the server, but yeah, after thinking about it, maybe this does not actually suggest the correct meaning. i'll change that.

seanmonstar:
Always happy to add in more tracing support. This one seems stalled (and now conflicts), so I'm going to close.

PR #2464, opened 2021-03-13
Body docs: Link to to_bytes and aggregate

dbrgn:
Since these two functions are not methods on the `Body`, they aren't very discoverable. So a note in the docs would definitely be helpful.

2021-03-13-191948_605x192_scrot


dbrgn:
Hm, the `commit-message-check/gitcop` task fails, but clicking on "Details" results in a "We're sorry, but something went wrong" error message.

I guess I should have looked at the contribution guidelines first though 🙂

seanmonstar (approved):
Thanks!

PR #2467, opened 2021-03-15
Introduce Error::is_header_section_too_large (fixes #2462)

### LINKED ISSUE #2462: hyper::Error should distinguish http1 TooLarge from other parse errors
Currently if an http1 message head exceeds the `http1_max_buf_size`, a `hyper::Error` is returned where `err.is_parse() == true`. This is indistinguishable from other errors that `httparse` might bubble up, which poses a problem for deciding how to respond to these errors. A `TooLarge` can occur just because of the choices made when setting up the max buffer size (many implementations default to 1MB vs the ~400KB default for `hyper`), vs. an actually malformed HTTP message which probably indicates a misbehaving server.

Could we add a new method to `hyper::Error` that distinguishes these? `hyper::Error::is_buffer_size_exceeded()` as a strawman?

nox:
We rename Parse::TooLarge into Parse::HeaderSectionTooLarge while at it.

### DIFF
--- a/src/body/length.rs
+++ b/src/body/length.rs
@@ -43,7 +43,7 @@ impl DecodedLength {
             Ok(DecodedLength(len))
         } else {
             warn!("content-length bigger than maximum: {} > {}", len, MAX_LEN);
-            Err(crate::error::Parse::TooLarge)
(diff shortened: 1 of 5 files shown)


nox:
Superseded by #2504.

PR #2471, opened 2021-03-17
Support multiple Content-Length values on same line

### LINKED ISSUE #2470: Support multiple Content-Length values
Hi,

I try to work on some [failing tests](https://github.com/servo/servo/blob/master/tests/wpt/web-platform-tests/fetch/content-length/resources/content-lengths.json) of servo. To summarize the issue : the `fetch` specification has been recently updated to handle multiple values for the [`Content-Length` header](https://fetch.spec.whatwg.org/#content-length-header). I found Hyper client does not handle the header as expected by the specification. To be more precise, hyper does not handle the multiple values into only one header entry : `Content-Length: 42,42`. I don't know if we want to update hyper to follow this specification or not.

I reproduced the error with the [example client](https://github.com/hyperium/hyper/blob/master/examples/client.rs) and the following node http server code:
```js
var http = require('http');

http.createServer(function (req, res) {
    res.writeHead(200, {'Content-Length': '42,42'});
    res.write('012345678901234567890123456789012345678901');
    res.end();
}).listen(8080);
```

ghostd:
Closes #2470

### DIFF
--- a/src/headers.rs
+++ b/src/headers.rs
@@ -42,26 +42,26 @@ pub(super) fn content_length_parse_all_values(values: ValueIter<'_, HeaderValue>
     // be alright if they all contain the same value, and all parse
     // correctly. If not, then it's an error.
 
-    let folded = values.fold(None, |prev, line| match prev {
-        Some(Ok(prev)) => Some(
-            line.to_str()
-                .map_err(|_| ())
-                .and_then(|s| s.parse().map_err(|_| ()))
-                .and_then(|n| if prev == n { Ok(n) } else { Err(()) }),
-        ),
-        None => Some(
-            line.to_str()
-                .map_err(|_| ())
-                .and_then(|s| s.parse().map_err(|_| ())),
-        ),
-        Some(Err(())) => Some(Err(())),
-    });
-
-    if let Some(Ok(n)) = folded {
-        Some(n)
-    } else {
-        None
+    let mut content_length: Option<u64> = None;
+    for h in values {
+        if let Ok(line) = h.to_str() {
+            for v in line.split(',') {
+                if let Some(n) = v.trim().parse().ok() {
+                    if content_length.is_none() {
+                        content_length = Some(n)
+                    } else if content_length != Some(n) {
+                        return None;
+                    }
+                } else {
+                    return None
+                }
+            }
+        } else {
+            return None
+        }
     }
+
+    return content_length
 }
 
 #[cfg(all(feature = "http2", feature = "client"))]
--- a/tests/client.rs
+++ b/tests/client.rs
@@ -1043,6 +1043,30 @@ test! {
         error: |err| err.to_string() == "request has unsupported HTTP version",
 }
 
+test! {
+    name: client_handles_contentlength_values_on_same_line,
+
+    server:
+        expected: "GET /foo HTTP/1.1\r\nhost: {addr}\r\n\r\n",
+        reply: "\
+            HTTP/1.1 200 OK\r\n\
+            Content-Length: 3,3\r\n\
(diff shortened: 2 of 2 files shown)


tests/client.rs
```diff
@@ -1043,6 +1043,29 @@ test! {
         error: |err| err.to_string() == "request has unsupported HTTP version",
 }
 
+test! {
+    name: client_handles_contentlength_values_on_same_line,
```
seanmonstar, line 1047: Woot!

src/headers.rs
```diff
@@ -64,6 +58,13 @@ pub(super) fn content_length_parse_all_values(values: ValueIter<'_, HeaderValue>
     }
 }
 
+fn content_length_parse_line(value: &HeaderValue) -> Vec<Result<u64, ()>> {
```
seanmonstar, line 61: The creation of a `Vec` here will mean a heap allocation, which will slow things down a bit more. I believe it should be possible to still do this without that. It might end up making the combinators a nightmare though, so it may eventually be worth just sprinkling a change to a `for ...` and an early return instead... If you get stuck, I can see what I can come up with.

seanmonstar (changes requested):
Thanks for the PR! <3

src/headers.rs
ghostd: Hi,

Here is an other version without Vec:
```rust
pub(super) fn content_length_parse_all_values(values: ValueIter<'_, HeaderValue>) -> Option<u64> {
    // If multiple Content-Length headers were sent, everything can still
    // be alright if they all contain the same value, and all parse
    // correctly. If not, then it's an error.

    let fold_values = |prev: Option<Result<u64, ()>>, value: Result<u64, ()>| match prev {
        Some(Ok(prev)) => Some(
            value.and_then(|n| if prev == n { Ok(n) } else { Err(()) }),
        ),
        None => Some(value),
        Some(Err(())) => Some(Err(())),
    };

    let folded = values.fold(None, |prev, line| {
        return match line.to_str() {
            Ok(line) => {
                return line.split(',')
                    .map(|a| a.trim().parse().map_err(|_| ()))
                    .fold(None, fold_values)
                    .map(|v| fold_values(prev, v))
                    .unwrap_or(None);
            },
            Err(_) => Some(Err(()))
        };
    });

    if let Some(Ok(n)) = folded {
        Some(n)
    } else {
        None
    }
}
```
What do you think about this?

seanmonstar: Nice, that looks like it doesn't need to allocate, awesome!

Looking at this code, it does seem rather complicated trying to keep using `fold`. I'll see if I can come up with a version that uses a `for` loop that's less complicated...

ghostd: Here is a version with nested loops and early returns:
```rust
pub(super) fn content_length_parse_all_values(values: ValueIter<'_, HeaderValue>) -> Option<u64> {
    // If multiple Content-Length headers were sent, everything can still
    // be alright if they all contain the same value, and all parse
    // correctly. If not, then it's an error.

    let mut content_length: Option<u64> = None;
    for h in values {
        if let Ok(line) = h.to_str() {
            for v in line.split(',') {
                if let Some(n) = v.trim().parse().ok() {
                    if content_length.is_none() {
                        content_length = Some(n)
                    } else if content_length != Some(n) {
                        return None;
                    }
                } else {
                    return None
                }
            }
        } else {
            return None
        }
    }

    return content_length
}
```

seanmonstar: Oh awesome! Yea, this looks so much simpler than the `fold` that was there. And with the extra fold. Thanks, let's use this one!

seanmonstar (approved):

PR #2473, opened 2021-03-19
Allow HTTP/0.9 responses behind a flag (fixes #2468)

### LINKED ISSUE #2468: Support HTTP/0.9 responses in the client
There are servers out there that sometimes reply with HTTP/0.9 responses, even modern-ish stack (we saw such responses from MSFT servers for example).

I'm filing this here and not on the httparse repo because this is a substantial change worth discussing in the open, and because we need to think about how to support it. Firefox for example has some logic to content-sniff the start of an actual `HTTP/1.*` response in the first 4KB of received data on the first request made on the connection, and this is obviously not code that should live in httparse.

As a start, I feel like Hyper should check whether the response starts with "HTTP" at the beginning of the data it received, and if not, maybe it should pretend it saw `HTTP/1.0 200 OK\r\n\r\n`.

I say `HTTP/1.0` and not `HTTP/0.9` because it is literally impossible for `httparse::Response` to represent `0.9` in its version field.

nox:
The two first commits work around some issues I had with VS Code and rust-analyzer.

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -972,6 +972,14 @@ impl Builder {
         self
     }
 
+    /// Set whether HTTP/0.9 responses should be tolerated.
+    ///
+    /// Default is false.
+    pub fn http09_responses(&mut self, val: bool) -> &mut Self {
+        self.conn_builder.h09_responses(val);
+        self
+    }
+
     /// Set whether the connection **must** use HTTP/2.
     ///
     /// The destination must either allow HTTP2 Prior Knowledge, or the
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -122,6 +122,7 @@ where
 #[derive(Clone, Debug)]
 pub struct Builder {
     pub(super) exec: Exec,
+    h09_responses: bool,
     h1_title_case_headers: bool,
     h1_read_buf_exact_size: Option<usize>,
     h1_max_buf_size: Option<usize>,
@@ -493,6 +494,7 @@ impl Builder {
     pub fn new() -> Builder {
         Builder {
             exec: Exec::Default,
+            h09_responses: false,
             h1_read_buf_exact_size: None,
             h1_title_case_headers: false,
             h1_max_buf_size: None,
@@ -511,6 +513,11 @@ impl Builder {
         E: Executor<BoxSendFuture> + Send + Sync + 'static,
     {
         self.exec = Exec::Executor(Arc::new(exec));
+        self
+    }
+
+    pub(super) fn h09_responses(&mut self, enabled: bool) -> &mut Builder {
+        self.h09_responses = enabled;
         self
     }
 
@@ -699,6 +706,9 @@ impl Builder {
                     let mut conn = proto::Conn::new(io);
                     if opts.h1_title_case_headers {
                         conn.set_title_case_headers();
+                    }
+                    if opts.h09_responses {
+                        conn.set_h09_responses();
                     }
                     if let Some(sz) = opts.h1_read_buf_exact_size {
                         conn.set_read_buf_exact_size(sz);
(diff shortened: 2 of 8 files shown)


nox:
I guess someone forgot to run tests and that someone was me.

nox:
@seanmonstar Please tell me which kind of test you would want and in which file.

nox:
I've filed [an issue](https://github.com/rust-analyzer/rust-analyzer/issues/8120) against rust-analyzer for that macro_rules indirection problem.

nox:
I have removed the rust-analyzer `macro_rules` work around as the bug has been fixed upstream.

nox:
We probably want this option to apply only to initial HTTP/1 requests and not those happening after the first one on a keep-alive connection.

seanmonstar:
Nice work, thanks!

> Please tell me which kind of test you would want and in which file.

The tests in hyper are a little messy, but for this, there's probably 2 relevant places: in `proto/h1/role.rs`, perhaps a couple simple decode tests that it works. A test in `tests/client.rs` (excuse the mess) would be useful to make sure the rest of the stuff is hooked up properly (basically the config, and that you got an `http::Response` with the correct `Version` returned and the body).

nox:
In addition to probably not wanting to interpret 0.9 responses on non-initial requests, what do we do about the HTTP/2 parsing fallback too?

nox:
I added 3 tests and prevented accepting an HTTP/0.9 response after the first one we receive (though I didn't write a test for that but I will).

seanmonstar:
Just checking, this has everything you meant to include? I don't see anything missing, but you mentioned a few questions that I see are now struck out, so just making sure.

nox:
Yeah, the struck-out thought is that we don't need to care about h2 fallback given this is about *responses*, a detail I forgot when I wrote that :)

As for the rest, it feels complete to me, but maybe we should have a test to check against HTTP/0.9 responses on a keep-alive connection after the first response? You decide.

seanmonstar:
> maybe we should have a test to check against HTTP/0.9 responses on a keep-alive connection after the first response?

Hm, so, if we're reusing an HTTP/1.x connection, would it ever make sense for the server to have downgraded to HTTP/0.9? I want to say that the first response kind of solidifies the version of the whole connection, but maybe I'm wrong. If we assume it cannot downgrade, then that'd mean that any server that, after sending an HTTP/1.x response, decides to just start sending jibberish, we would assume it's busted...

You know, thinking more, probably it's worth to enforce that, since I think a more likely scenario is a server accidentally sending a bigger body than it declared, and we would want to say it was an error, not another response in HTTP/0.9. Whatcha think?

nox:
I think my code [already](https://github.com/hyperium/hyper/pull/2473/commits/2fe85ca9623f47f4efab2f8b25bc0ef9ccbe307c#diff-0861a40bcc9fa01e6f9ee19266e6c4e5e5378cd7bd92a887c2a4f745391ecd5cR167-R168) refuses to ever downgrade to HTTP/0.9 after we received a first response, so I think we agree we should enforce that, yes, hah.

Are there tests for keep-alive stuff already that I can take inspiration from?

seanmonstar:
> Are there tests for keep-alive stuff already that I can take inspiration from?

There are a few tests for keep-alive, but I think the way they are written is pretty bad. I've wanted to re-do hyper's test suite for a while now (it was some of the first Rust code I ever wrote, and instead of improving it, I've just adding junk on top). I'd probably rewrite those using [`tokio-test`](https://docs.rs/tokio-test/0.4.1/tokio_test/io/struct.Builder.html), which allows us to define an IO type and what expected reads and writes are.

I'm fine merging this without that specific test, to get it released sooner, unless you'd prefer I hold off.

seanmonstar (approved):

PR #2474, opened 2021-03-21
Prevent a panic if the channel used by Body is terminated

meh:
I don't know if this is the right way, but I got a panic during `Drop`, and this makes it werk.

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -6,6 +6,7 @@ use std::fmt;
 use bytes::Bytes;
 use futures_channel::mpsc;
 use futures_channel::oneshot;
+use futures_core::stream::FusedStream;
 use futures_core::Stream; // for mpsc::Receiver
 #[cfg(feature = "stream")]
 use futures_util::TryStreamExt;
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! What was the panic, and why did it happen during `Drop`? I'd like to better understand, since without knowing, this looks like potentially an extra method check on every iteration.

meh:
last

This is the stack trace I got (I also posted this on the Discord, I'm also there if you wanna ping me), it was happening after everything was done in the code, so during the `Drop`ping of stuff.

At fault might also be `rusoto_s3`, but I'm not sure if it's `rusoto_s3` using the `Body` wrong, or if it's the `Body` that is doing something wrong.

Definitely a weird edge case.

seanmonstar:
Interesting stack trace... It seems the panic is from polling for another chunk after a previous poll already returned `Poll::Ready(None)`. The code in `Body` doesn't poll it extra times, so it seems that at some level above, something is polling the body after it "finished". I'd lean towards the bug being up there, and not something to fix in hyper itself.

meh:
Suspected so, I will look into fixing it from the rusoto side, will link the PR/issue here once it's up.

seanmonstar:
For now, I'm going to close this. Curious how it ends, though!

PR #2475, opened 2021-03-21
add deps.rs security tracker tag to README

bionicles:
helps remind us to update dependencies regularly to patch security issues


seanmonstar:
Thanks for the PR!

I do think it'd be a good idea to add an audit step to CI, to make sure *new* CI runs don't download dependencies that need fixing.

However, I don't want to add a badge about it. Rust and Cargo allows libraries to state the minimum version of a dependency they need, but they don't need to be constantly updating that minimum version. Any new projects will automatically use the newest allowed version of all the dependencies, and existing projects can run `cargo audit` and then `cargo update -p whatever` to update. For example, the deps.rs page linked suggests hyper is depending on vulnerable code in `futures-util`, but it isn't. hyper just sets `0.3.x` as a valid dependency, and any user can update in their own project.

seanmonstar:
So I'm going to close this, to not add a badge. I still think it'd be good to add an `audit` step in CI.

Issue #2397, opened 2021-01-12
Return an error instead of aborting on panics in C API


seanmonstar:
The C API will currently abort after catching a panic, for safety reasons. However, we shouldn't abort, but rather catch it and return an error code.

seanmonstar:
A tricky part to this is that the various FFI functions have different return types, so we can't just always return some `HYPERE_INTERNAL_BUG` code.

One way I considered to do this was to have some trait implemented for return types. However, it's slightly scary to me to execute some indirect function outside of the `catch_unwind`, since it could technically panic also. We could just promise to be careful not to do anything too fancy in them... Another issue is that sometimes the type returned has different meanings per function, and returning `-1` or `0` might make sense in different functions.

Kixunil:
Wouldn't `unsafe trait` with contract requiring absence of panic solve this?

seanmonstar:
The trait would *need* to be `unsafe`, yes. However, there's a difference between "literally can't happen" and "I promise I won't". 🤷

Kixunil:
Here's a hack to at least force abort in case it goes wrong:

```rust
/// Calls the closure and guarantees to abort if the closure panics
pub fn force_abort_on_panic<R, F: FnOnce() ->R>(fun: F) -> R {
    let mut output = MaybeUninit::uninit();
    // drop impl of RunOnDrop is guaranteed to write the result into output
    // and any panic inside it would turn to abort, so uninitialized value can't be observed
    unsafe {
        drop(RunOnDrop(&mut output, ManuallyDrop::new(fun)));
        output.assume_init()
    }
}

struct RunOnDrop<'a, R, F: FnOnce() ->R>(&'a mut MaybeUninit<R>, ManuallyDrop<F>);

impl<'a, R, F: FnOnce() ->R> Drop for RunOnDrop<'a, R, F> {
    fn drop(&mut self) {
        // The struct is private, and the caller guarantees to initialize it
        // panic in drop always causes abort
        unsafe {
            *self.1 = MaybeUninit::new(self.1.take()());
        }
    }
}
```

Didn't try to compile it but I guess you get the idea.

nylanderdev:
Seeing as we might want different error handlers and values for different functions I suggest we create a new `ffi_fn!` macro rule instead of using a trait. Something that will allows us to write ffi functions like this:

```
ffi_fn! {
    /// Create a new "empty" body.
    ///
    /// If not configured, this body acts as an empty payload.
[...]

PR #2478, opened 2021-03-24
Return null instead of aborting in C API by extending ffi_fn! macro

nylanderdev:
Make C API functions that return pointers return null in case of a
panic, instead of aborting.
Add ffi_fn! macro rules that enable error values to be returned by
writing "?= <value>" or "?: <closure>" after an ffi function's body.

Incremental to #2397

### DIFF
--- a/src/ffi/macros.rs
+++ b/src/ffi/macros.rs
@@ -1,5 +1,5 @@
 macro_rules! ffi_fn {
-    ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) -> $ret:ty $body:block) => {
+    ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) -> $ret:ty $body:block ?= $default:expr) => {
         $(#[$doc])*
         #[no_mangle]
         pub extern fn $name($($arg: $arg_ty),*) -> $ret {
@@ -8,13 +8,21 @@ macro_rules! ffi_fn {
             match panic::catch_unwind(AssertUnwindSafe(move || $body)) {
                 Ok(v) => v,
                 Err(_) => {
-                    // TODO: We shouldn't abort, but rather figure out how to
-                    // convert into the return type that the function errored.
-                    eprintln!("panic unwind caught, aborting");
-                    std::process::abort();
+                    $default
                 }
             }
         }
+    };
+
+    ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) -> $ret:ty $body:block) => {
+        ffi_fn!($(#[$doc])* fn $name($($arg: $arg_ty),*) -> $ret $body ?= {
+            eprintln!("panic unwind caught, aborting");
+            std::process::abort()
+        });
+    };
+
+    ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) $body:block ?= $default:expr) => {
+        ffi_fn!($(#[$doc])* fn $name($($arg: $arg_ty),*) -> () $body ?= $default);
     };
 
     ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) $body:block) => {
--- a/src/ffi/body.rs
+++ b/src/ffi/body.rs
@@ -34,7 +34,7 @@ ffi_fn! {
     /// If not configured, this body acts as an empty payload.
     fn hyper_body_new() -> *mut hyper_body {
         Box::into_raw(Box::new(hyper_body(Body::empty())))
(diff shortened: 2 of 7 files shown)


nylanderdev:
Not sure if refactor but issue was tagged as such.

nylanderdev:
Maybe we shouldn't print to stderr unless aborting.

src/ffi/macros.rs
```diff
@@ -17,7 +17,43 @@ macro_rules! ffi_fn {
...
+                Ok(v) => v,
+                Err(_) => {
+                    eprintln!("panic unwind caught, returning default");
+                    $default
+                }
+            }
+        }
+    };
+
+    ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) -> $ret:ty $body:block ?: $on_err:expr) => {
```
seanmonstar, line 36: I think, now seeing the `?=` option that has us just return a static expression, that I might prefer that as much as possible, versus passing another function that might panic too... Do you have an example use case in mind? If not, maybe we just stick the `?=` to start with?

src/ffi/macros.rs
```diff
@@ -17,7 +17,43 @@ macro_rules! ffi_fn {
         }
     };
 
+    ($(#[$doc:meta])* fn $name:ident($($arg:ident: $arg_ty:ty),*) -> $ret:ty $body:block ?= $default:expr) => {
```
seanmonstar, line 20: I think it could be possible to not need to repeat ourselves, by having the the `?=` the top variant, and have a macro rule variant that doesn't match `?=`, and those would just call the macro again with `ffi_fn!($(#[$doc])* fn $name($($arg: $arg_ty),*) -> $ret $body ?= crate::ffi::ffi_panic());` or so, kinda like the others that have the return type auto-added. Right?

src/ffi/macros.rs
```diff
@@ -17,7 +17,43 @@ macro_rules! ffi_fn {
...
+        $(#[$doc])*
+        #[no_mangle]
+        pub extern fn $name($($arg: $arg_ty),*) -> $ret {
+            use std::panic::{self, AssertUnwindSafe};
+
+            match panic::catch_unwind(AssertUnwindSafe(move || $body)) {
+                Ok(v) => v,
+                Err(_) => {
+                    eprintln!("panic unwind caught, returning default");
```
seanmonstar, line 29: Yea, you may be right. I'm kinda thinking that if we've caught and properly returned a value, perhaps we shouldn't be trying to print to stdio...

seanmonstar (commented):
Awesome start! I left some thoughts inline.

src/ffi/macros.rs
nylanderdev: I'll fix that. Do you want us to still eprint before a panic like before?

src/ffi/macros.rs
nylanderdev: I'll remove it.

src/ffi/macros.rs
nylanderdev: No. Let's stick to `?=`.

nylanderdev: Though technically `?=` allows for function calls (that might panic), as long as they return the right type. I don't know how to fix that, as we're using `expr` for default values.

seanmonstar: Oh yea, it could allow that... Oh well. Still, I don't yet know if we'd want a function to inspect the panic and return a value based on it... Well, maybe to check for OOM, but that can't be done in stable yet anyways.

src/ffi/macros.rs
seanmonstar: Yea, probably worth still printing before the abort. Though, now that I see it's just an expression, it probably doesn't need to be an extra function call, just a block defined in the macro.

nylanderdev:
How do we want to deal with functions that return `()`? Right now they don't support default values (since the only valid one is `()`) but that also makes them default to aborting. We could either add `?=` support to them, and write `?= ()`, or we could use some shorthand, like simply `?` to mark those that won't abort. We could also make not aborting the default, and add a `!` or a `!!` to explicitly signify an aborting function.

seanmonstar:
Can we mark them as `?= ()`? I'd rather the explicitness than accidentally forgetting whether the value should have returned an error if it panicked. I assume that those that don't return a value, in practice wouldn't ever fail or panic, but we catch anyways just in case.

nylanderdev:
Supports `?= ()` now.

seanmonstar (approved):
Awesome work, thanks for starting this! I think we can do follow up PRs to add this to more of the functions (and then eventually remove the ability to have forgotten).

PR #2479, opened 2021-03-25
Using the hyper provided const POST, to match the example in https://hyper.rs/guides/client/advanced/

Vagelis-Prokopiou:


seanmonstar (approved):
Sure!

PR #2480, opened 2021-03-26
Decouple preserving header case from FFI (fixes #2313)

### LINKED ISSUE #2313: Option to preserve header's title case
I am building a HTTP proxy with hyper, and I realize that headers are all converted to lower case for HTTP/1.x.

There was an issue about adding `http1_title_case_headers` for clients: https://github.com/hyperium/hyper/issues/1492 , but we don't have a way to preserve the **response headers'** title case from remote servers.

Some client applications are expecting to receive header with name that is exactly the same as the server returns, for example:

- Nintendo Switch's connection checking will sent a GET request to `http://ctest.cdn.nintendo.net/` and expecting a response with header `X-Organization: Nintendo`.

nox:


nox:
This probably needs more tests, but I don't know where you want me to write those and which kind of tests you want.

nox:
It helps if I actually commit the damn file… Now `HeaderCaseMap` lives in `crate::ext`.

nox:
Mmh this is incomplete, I just realised `Server::encode` hardcodes some lowercased header names.

nox:
Ah, I just realised out of nowhere I didn't implement preserving the original case of received requests in the server role.

zonyitoo:
Proxy developers are looking forward the day that this PR merged into master.

nox:
> Ah, I just realised out of nowhere I didn't implement preserving the original case of received requests in the server role.

I implemented that and added a test.

src/client/client.rs
```diff
@@ -972,6 +972,17 @@ impl Builder {
         self
     }
 
+    /// Set whether HTTP/1 connections will write header names as provided
+    /// at the socket level.
+    ///
+    /// Note that this setting does not affect HTTP/2.
+    ///
+    /// Default is false.
+    pub fn http1_preserve_header_case(&mut self, val: bool) -> &mut Self {
```
seanmonstar, line 981: Ah yes, so the original type in FFI I think just assumes, if you included the extension, even if empty, then you probably wanted the same returned in the `Response`. It's probably right to have this be an option too, just wanted to call out the difference in the review.

seanmonstar: Oh yea, I think the option is also *needed* for the server side, since you get the request before you could return a response with the extension set.

src/ext.rs
```diff
@@ -0,0 +1,51 @@
+//! HTTP extensions
+
+use bytes::Bytes;
+#[cfg(feature = "http1")]
+use http::header::{GetAll, HeaderName, IntoHeaderName};
+use http::HeaderMap;
+
+/// A map from header names to their original casing as received in an HTTP response.
```
seanmonstar, line 8: This extension is used for both requests and responses.

Maybe after a brief description of what it is, would it make sense to have request and response examples? And/or maybe how it's used for clients and servers? Even more docs isn't required for merging, just rambling here...

src/ext.rs
```diff
@@ -0,0 +1,51 @@
...
+///
+/// [`http1_preserve_header_case`]: /client/struct.Client.html#method.http1_preserve_header_case
+#[derive(Clone, Debug, Default)]
+pub struct HeaderCaseMap(HeaderMap<Bytes>);
+
+#[cfg(feature = "http1")]
+impl HeaderCaseMap {
+    /// Returns a view of all spellings associated with that header name,
+    /// in the order they were found.
+    pub fn get_all(&self, name: &HeaderName) -> GetAll<'_, Bytes> {
```
seanmonstar, line 36: What do you think, should we have the values be `Bytes`? Some other newtype that implies some validation has happened? Something that would allow us to change to a different representation if we wanted (`Bytes` kinda has that when we make the vtable public...)?

src/proto/h1/conn.rs
```diff
@@ -74,11 +73,15 @@ where
         self.io.set_read_buf_exact_size(sz);
     }
 
-    #[cfg(feature = "client")]
     pub(crate) fn set_title_case_headers(&mut self) {
         self.state.title_case_headers = true;
     }
 
+    #[cfg(feature = "client")]
```
seanmonstar, line 80: Is this only needed for the client? Or maybe I got ahead of myself, and adding support for the received request header cases in the server side is a future addition?

src/proto/h1/role.rs
```diff
@@ -359,6 +361,217 @@ impl Http1Transaction for Server {
             extend(dst, b"\r\n");
         }
 
+        let orig_headers;
+        let extensions = mem::take(&mut msg.head.extensions);
```
seanmonstar, line 365: Instead of `take` here, you could do `msg.head.extensions.remove::<HeaderCaseMap>()`, yea?

src/proto/h1/role.rs
```diff
@@ -190,6 +191,7 @@ impl Http1Transaction for Server {
         let mut is_te = false;
         let mut is_te_chunked = false;
         let mut wants_upgrade = subject.0 == Method::CONNECT;
+        let mut header_case_map = HeaderCaseMap::default();
```
seanmonstar, line 194: Do you know if creating this even if not needed has any affect on server peformance? Might be easiest to test that with `cargo bench --bench pipeline` (that's the easiest, though I find it's slightly noisier than actually starting up a pipeline server and running `wrk` at it, but that's also more tedious...).

src/proto/h1/role.rs
```diff
@@ -592,72 +833,32 @@ impl Http1Transaction for Server {
...
-        if status.is_informational() || method == &Some(Method::CONNECT) && status.is_success() {
-            false
-        } else {
-            match status {
-                StatusCode::NO_CONTENT | StatusCode::NOT_MODIFIED => false,
-                _ => true,
-            }
-        }
-    }
+trait HeaderNameWriter {
```
seanmonstar, line 848: <3

seanmonstar (commented):
Wonderful work, thanks for all the care you put into it!

src/ext.rs
nox: I think `RawHeaderName` would be a nice addition. We can think about moving it to the `http` crate later. Or we can do it now.

src/proto/h1/role.rs
nox: I didn't want to use `Extensions::remove` because I didn't want to mutate that for no reason given we just need a reference to the header case map, so `take` felt less heavy to me. I can change it anyway.

src/proto/h1/role.rs
nox: I didnt bother checking this because that's just like what `Client::parse` does :)

src/proto/h1/conn.rs
nox: It's used only in `client::Builder::handshake`.

src/proto/h1/role.rs
seanmonstar: But `take` is mutating the `msg.head` anyways, right?

src/proto/h1/role.rs
seanmonstar: Yea, fair. Though also, (while I grumble about it all the time) the server benchmarks are watched and generally used as PR for the project, so it'd probably be good to be aware, at least.

src/ext.rs
seanmonstar: What do you propose the newtype would be able to do? Would it just be safer to start by returning some `impl Iterator<Item = impl AsRef<[u8]>>`?

src/proto/h1/role.rs
nox: Yeah yeah, I mean doing what amounts to a simple `memcpy` (moving `msg.head.extensions` out) vs mutating a glorified hashmap (calling `Extensions::remove`).

src/ext.rs
nox: I made it return `RawHeaderName`, which only implements `Clone` and `Debug`.

src/proto/h1/role.rs
nox: I've put both `HeaderCaseMap::default` calls in their own `if ctx.preserve_header_case` block just to be sure.

src/ext.rs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        self.0.append(name, RawHeaderName(orig));
+    }
+}
+
+/// Represents a raw header name, which may not in lower case.
+///
+/// Exclusively used by the `HeaderCaseMap` extension.
+#[derive(Clone, Debug)]
+pub struct RawHeaderName(Bytes);
```
seanmonstar, line 57: So before, with a `Bytes` a user could do quite a bit of things with it, not least of all view the actual slice of bytes. I do think there's value in us separating the type some, but probably need an `AsRef<[u8]>` at least.

Do you think returning an `impl AsRef<[u8]>` would be horrible, to start?

src/ext.rs
```diff
@@ -0,0 +1,64 @@
...
+///
+/// [`http1_preserve_header_case`]: /client/struct.Client.html#method.http1_preserve_header_case
+#[derive(Clone, Debug, Default)]
+pub struct HeaderCaseMap(HeaderMap<RawHeaderName>);
+
+#[cfg(feature = "http1")]
+impl HeaderCaseMap {
+    /// Returns a view of all spellings associated with that header name,
+    /// in the order they were found.
+    pub fn get_all(&self, name: &HeaderName) -> GetAll<'_, RawHeaderName> {
```
seanmonstar, line 36: Just realized that in a way, this reveals that the internal type is a `HeaderMap`. Maybe that's fine eventually, but to not leak that, we could change the return type to `-> impl Iterator<Item = Blah>`. Any concern with that?

seanmonstar (commented):
Very cool, just some minor API points left. I'll try to grab the benchmarks of `cargo bench --bench pipeline` for before and after, just to include as a data point in the PR.

src/ext.rs
nox: Nah that's a good idea.

src/ext.rs
nox: Nah that's a good idea.

nox:
This is done, I also removed `Default` for `HeaderCaseMap` while at it.

src/proto/h1/role.rs
nox: I tried just not moving out anything by splitting `msg` into more arguments to `encode_headers`, but that was way too many arguments for me to be confortable committing that.

src/ext.rs
seanmonstar: Or, I also just had this idea, in case we don't know yet. I thought, what do you want to do with this first. I assume you just want to receive the extension from the server request, and pass it on to the client request, and then same with the client response passed back to the server response. I assume you don't actually need to edit the map really.

If that's so, we could just ditch making this type and `get_all` public, and make that a subject of another PR. :shrug:

nox: We could ditch the entire `ext` module even.

nox: `ext` and `HeaderCaseMap` are not public items anymore. In my use case we pass requests and responses between client and upstream mostly as is so I don't need anything more.

seanmonstar: Oh you're right, we don't have to make them public _just yet_, and it still works. Seems like good plan to me. Making them public can be a more detailed proposal :)

seanmonstar (approved):

seanmonstar:
Well crap, didn't realize merging the other would cause conflicts. If you could resolve them, I can merge right after and get a release out.

nox:
@seanmonstar Rebased!

simpleearn (approved):

PR #2484, opened 2021-03-30
feat(client): http_connector supports `set_interface`

kolapapa:
In my scenario, I need to _**detect the network stability of multiple interfaces at the same time**_, including **ICMP**, **HTTP**, etc. 

The `bind_device()` of `socket2`'s crate meets the requirement of specific interface in `ICMP`(`ping`). But `HTTP` related, I use the `reqwest` crate, but I can't use `reqwest` to _bind a specific interface_ to achieve the flow in and out through a specific exit. 

And now, I need to do **upload** and **download**, **HTTP quality detection** and other functions at present. If I want to achieve my goal by supporting `bind_device()`, please help review it. I can't help you.

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -79,6 +79,7 @@ struct Config {
     reuse_address: bool,
     send_buffer_size: Option<usize>,
     recv_buffer_size: Option<usize>,
+    interface: Option<String>,
 }
 
 // ===== impl HttpConnector =====
@@ -119,6 +120,7 @@ impl<R> HttpConnector<R> {
                 reuse_address: false,
                 send_buffer_size: None,
                 recv_buffer_size: None,
+                interface: None,
             }),
             resolver,
         }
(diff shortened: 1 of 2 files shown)


src/client/connect/http.rs
```diff
@@ -228,6 +230,25 @@ impl<R> HttpConnector<R> {
...
+    ///
+    /// On Linux it can be used to specify a [VRF], but the binary needs
+    /// to either have `CAP_NET_RAW` or to be run as root.
+    ///
+    /// This function is only available on Android、Fuchsia and Linux.
+    ///
+    /// [VRF]: https://www.kernel.org/doc/Documentation/networking/vrf.txt
+    #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+    #[inline]
+    pub fn bind_device<S: ToString>(&mut self, interface: S) -> &mut Self {
```
seanmonstar, line 247: How about the name `set_interface`?

seanmonstar: Also, I'd probably say the generic should be `S: Into<String>`, since otherwise `set_interface(format("eth{}", num))` will make a string with `format!`, and then copy thanks to `ToString`.

src/client/connect/http.rs
```diff
@@ -612,6 +633,14 @@ fn connect(
     )
     .map_err(ConnectError::m("tcp bind local error"))?;
 
+    #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+    // That this only works for some socket types, particularly AF_INET sockets.
+    if config.local_address_ipv6.is_none() && config.interface.is_some() {
```
seanmonstar, line 638: Why only if the ipv6 address is none? What if the ipv4 address is some? What if ipv6 is some?

seanmonstar (changes requested):
Thanks for the PR! Some suggestions inline.

src/client/connect/http.rs
kolapapa: I agree with you

src/client/connect/http.rs
kolapapa: Sorry, I mistakenly thought that IPv6 does not support `SO_BINDTODEVICE` operation. Later,  I tried, is feasible.

hottea773:
@kolapapa is there a reason this was closed?

I'd like to make use of this change, so would be keen to get this merged.

hottea773:
(I've opened https://github.com/hyperium/hyper/pull/3076 which duplicates this, but would be happy for either to be merged.)

PR #2485, opened 2021-03-31
Update httpdate to v1.0

pyfisch:
Previous version would return wrong results for `Ord`.


pyfisch:
What's the correct commit message for this change? I am unsure about scope and type.

seanmonstar (approved):
Probably just `chore(dependencies)`, but no worries, I can fix it when squashing. Thanks!

PR #2488, opened 2021-04-01
.github: run gen_header.sh in CI environment

### LINKED ISSUE #2483: Errors running gen_header.sh
I'm trying to fix some of the errors you get when you run the curl test suite with the Hyper backend. I tried to pick an easy one to start - Curl expects you to return CURLE_UNSUPPORTED_PROTOCOL when a server returns HTTP/1.2, but Hyper/c-hyper.c currently doesn't. 

So I wanted to add an extra constant to ffi/error.rs, which would let me read that constant in hyper-c.c in the Curl code and return the appropriate error.

```diff
--- a/src/ffi/error.rs
+++ b/src/ffi/error.rs
@@ -24,6 +24,8 @@ pub enum hyper_code {
     HYPERE_FEATURE_NOT_ENABLED,
     /// The peer sent an HTTP message that could not be parsed.
     HYPERE_INVALID_PEER_MESSAGE,
+    /// The peer sent an HTTP version header that we cannot parse
+    HYPERE_INVALID_HTTP_VERSION,
 }

 // ===== impl hyper_error =====
@@ -31,9 +33,11 @@ pub enum hyper_code {
 impl hyper_error {
     fn code(&self) -> hyper_code {
         use crate::error::Kind as ErrorKind;
+        use crate::error::Parse;
         use crate::error::User;

         match self.0.kind() {
+            ErrorKind::Parse(Parse::Version) => hyper_code::HYPERE_INVALID_HTTP_VERSION,
             ErrorKind::Parse(_) => hyper_code::HYPERE_INVALID_PEER_MESSAGE,
             ErrorKind::IncompleteMessage => hyper_code::HYPERE_UNEXPECTED_EOF,
             ErrorKind::User(User::AbortedByCallback) => hyper_code::HYPERE_ABORTED_BY_CALLBACK,
```

However, in order to pick this up I need to run gen_header.sh. I ran into some errors running this script.

First I wasn't using the nightly but because `cargo rustc -- -Z unstable-options --pretty=expanded` swallows errors I didn't see that that was the issue - I had to modify the script in order to get it to work. 

Now I'm running into this issue: 

```
+ cargo rustc -- -Z unstable-options --pretty=expanded
   Compiling hyper v0.0.0 (/private/var/folders/s1/909yt58s4wj8h_7v7frr8rkm0000gn/T/tmp.O3fSWUSrBt)
error: The `ffi` feature currently requires the `client` and `http1` features.
  --> /Users/kevin/src/github.com/hyperium/hyper/capi/../src/ffi/mod.rs:37:1
   |
37 | compile_error!("The `ffi` feature currently requires the `client` and `http1` features.");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Which I think is due to a poor interaction between the "fake" module and the "real" ffi module - I've been unsuccessful at modifying the `cargo rustc` to add those features. I can keep trying but my guess is that this is something that would take me an hour to fix and someone else two minutes to tell me the right place to look.

kevinburke:
Clean up the script so that any unexpected error terminates the
script, and stop suppressing errors that may contain useful
information (for example, that you are using the stable version but
need to use the nightly).

This is useful because if hyper.h is not up to date going forward the
CI should flag it. As is, there are a bunch of changes to hyper.h that
have not been checked in (or were generated by a newer version of the
cbindgen script.)

Fixes #2483.


capi/include/hyper.h
```diff
@@ -83,41 +127,86 @@ typedef enum {
...
+ */
+typedef struct hyper_task hyper_task;
+
+/*
+ A waker that is saved and used to waken a pending task.
+ */
 typedef struct hyper_waker hyper_waker;
 
-typedef int (*hyper_body_foreach_callback)(void*, const hyper_buf*);
+typedef int (*hyper_body_foreach_callback)(void*, const struct hyper_buf*);
```
seanmonstar, line 201: Hm, is this change from an updated version of cbindgen? What's the repercussions of this change? Will users (like curl) need to change their type annotations? Or is it just the same thing before and after?

seanmonstar (commented):
Thanks so much for fixing this! I noticed a change in the output that I'd like to figure out before merging, commented inline.

capi/include/hyper.h
kevinburke: Oh gosh, I have no idea. Let me ask some folks who understand C better than me

kevinburke: I generated that using the latest `cbindgen` and same `cbindgen.toml` as previous. I can look through the cbindgen docs to see if they mention changes.

For what it's worth, the curl maintainer said he's OK treating Hyper support as bleeding edge, which implies he's not super concerned about backwards compatibility.

seanmonstar: Cool, thanks for looking! I'm not annoyed by the change or anything, I just want to better understand before we merge :)

kevinburke: My coworker - who has written a lot of C - says they're equivalent. I also tested compiling Curl with this new hyper.h and it generated with no issues. See https://github.com/meterup/homebrew-safe for details or `brew install meterup/safe/hyper meterup/safe/curl`.

kevinburke:
Note I just pushed some additional changes to make the diff clearer when cbindgen fails and to add a header to the top of hyper.h.

capi/include/hyper.h
seanmonstar: Cool, thanks for checking :D

seanmonstar (approved):
Superb, thanks so much!

capi/gen_header.sh
```diff
@@ -44,10 +47,14 @@ cp "$CAPI_DIR/include/hyper.h" "$header_file_backup"
...
 
-cd $WORK_DIR
+cd "${WORK_DIR}" || exit 2
 
 # Expand just the ffi module
-cargo rustc -- -Z unstable-options --pretty=expanded > expanded.rs 2>/dev/null
+if ! output=$(cargo rustc -- -Z unstable-options --pretty=expanded 2>&1 > expanded.rs); then
+    # As of April 2021 the script above prints a lot of warnings/errors, and
+    # exits with a nonzero return code, but hyper.h still gets generated.
+    echo "$output"
```
seanmonstar, line 56: I've noticed that all PRs now get Github Actions Annotations with the warnings from this script, and I suspect it's automatically doing it from this output. Do you think it'd be horrible if we silence this if `--verify` was passed?

kevinburke: No that's alright

PR #2492, opened 2021-04-07
chore(ffi): fix compile errors and warnings

kevinburke:
As I understand it, "cargo rustc" in gen_header.sh generates a ton of
errors, but still manages to generate an object that can be used by
cbindgen to generate hyper.h.

However, I tried to make a separate change to add more fields to
hyper.h, and learned that "cargo rustc" stops if it reaches 50 errors,
which I reached. I was able to buy some headroom and fix a number of
the compilation errors by adding imports to the fake Cargo.toml we
generate in gen_header.sh.

I wasn't sure how to resolve imports like "crate::Result" which appear
to reference the top-level src/error.rs, and print an error when they
are compiled in gen_header.sh. But I only need to buy headroom under
the 50 error count for now, which I was able to do by adding the
imports.

It is possible that someone more familiar with Rust than me could look
at this and know what to change to get the total number of errors to
zero.

### DIFF
--- a/capi/gen_header.sh
+++ b/capi/gen_header.sh
@@ -41,6 +41,27 @@ edition = "2018"
 publish = false
 
 [dependencies]
+# Determined which dependencies we need by running the "cargo rustc" command
+# below and watching the compile error output for references to unknown imports,
+# until we didn't get any errors.
+bytes = "1"
+futures-channel = "0.3"
+futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
+libc = { version = "0.2", optional = true }
+http = "0.2"
+http-body = "0.4"
+tokio = { version = "1", features = ["rt"] }
+
+[features]
+default = [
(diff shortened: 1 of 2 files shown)


kevinburke:
Here's an example error that I'm unsure how to fix, since I'm not sure how to reference `src/client/*.rs` from the fake Cargo.toml in gen_header.sh. I can change the path to `crate::ffi::client` but then building FFI as part of `cargo build --features ffi` at the top level stops working.

```
error[E0432]: unresolved import `crate::client`
Error:  --> /home/runner/work/hyper/hyper/capi/../src/ffi/client.rs:5:12
  |
5 | use crate::client::conn;
  |            ^^^^^^
  |            |
  |            unresolved import
  |            help: a similar path exists: `crate::ffi::client`
```

seanmonstar (approved):
😆

PR #2495, opened 2021-04-08
ffi: add new error constant

kevinburke:
The curl library wants to return a specific error code if a server
returns "HTTP/1.2". We distinguish this case with a special error in
Hyper's rust code, but it's difficult to distinguish this specific
error in the C code because the errors are lumped together.

Instead add a new constant for this specific error so that we can
distinguish it in the C code.

### DIFF
--- a/capi/include/hyper.h
+++ b/capi/include/hyper.h
@@ -104,6 +104,10 @@ typedef enum hyper_code {
    The peer sent an HTTP message that could not be parsed.
    */
   HYPERE_INVALID_PEER_MESSAGE,
+  /*
+   The peer sent an HTTP version header that we cannot parse
+   */
+  HYPERE_INVALID_HTTP_VERSION,
 } hyper_code;
 
 /*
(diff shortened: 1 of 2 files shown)


kevinburke:
Here's the patch that uses this constant to pass another test in the Curl test suite. https://github.com/curl/curl/commit/4aeb804a1a8ec53f87da7db5f04dda3a8a73a42a

kevinburke:
Hey, just following up. Do you have a different idea for how to approach this?

seanmonstar:
Thanks for reminding me Kevin! So, this is something where since I'm not as used to C APIs, I could use some help.

In Rust, it's possible to classify an error as both something more general, like a parse error, and also indicate exactly what kind of parse error, since you can easily nest enums. In C, I think I'm pretty much stuff with returning an integer. I realize curl wants to notice specifically when the parse error is about a weird `HTTP/1.2` version, but wouldn't it also (besides curl) be a breaking change to make change the error code returned? 

Is there a common way to identify classes of errors and still allowing them to become more specific? I guess the first thing I think of is having ranges be a certain general value (like, 100-199 is parse errors), but is that common? It seems it could be fragile, so I wonder if there's a better pattern.

PR #2504, opened 2021-04-15
Expose error enums (fixes #2462)

### LINKED ISSUE #2462: hyper::Error should distinguish http1 TooLarge from other parse errors

nox:
There are probably a ton of details to sort out but at least we can start discussing.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,6 +2,8 @@
 use std::error::Error as StdError;
 use std::fmt;
 
+use self::sealed::Sealed;
+
 /// Result type often returned from methods that can have hyper `Error`s.
 pub type Result<T> = std::result::Result<T, Error>;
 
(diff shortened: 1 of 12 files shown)


nox:
Ah yeah, in this I also renamed `VersionH2` to `H2Preface`.

nox:
And I also just changed `Parse::TooLarge` to `Parse::SectionHeaderTooLarge` like I did in #2467.

nox:
Rebased!

nox:
I feel like if we do this, there should be no variant that is `cfg`-gated, for simplicity.

src/error.rs
```diff
@@ -17,109 +19,127 @@ struct ErrorImpl {
...
+    Status(Sealed),
 }
 
-#[derive(Debug, PartialEq)]
-pub(super) enum User {
+/// Represents the kind of the user error.
+///
+/// This enum is non-exhaustive.
+#[non_exhaustive]
+pub enum User {
```
seanmonstar, line 99: I kinda think I wouldn't want to expose the user error types. At least not yet, I think they grew without too much planning. And some are specifically to give a better error message, but I'd want to be able to change the kind to something better if they exist.

src/error.rs
```diff
@@ -17,109 +19,127 @@ struct ErrorImpl {
...
+    Listen(Sealed),
     /// Error accepting on an Incoming stream.
     #[cfg(any(feature = "http1", feature = "http2"))]
     #[cfg(feature = "server")]
-    Accept,
+    Accept(Sealed),
     /// Error while reading a body from connection.
     #[cfg(any(feature = "http1", feature = "http2", feature = "stream"))]
-    Body,
+    Body(Sealed),
```
seanmonstar, line 58: I kinda think these various body errors probably could be organized into a single enum. The variant names by themselves right now aren't super clear, I think...

seanmonstar (commented):
Yea, this _is_ largely the direction I imagined. I still don't know if it's actually a smart thing to do XD

I'd be especially interested if we could mark some variants as private, so we don't have to settle on the hierarchy of some kinds just yet. But I haven't thought of a clever way to do that in Rust...

src/error.rs
seanmonstar: Now that I've mentioned wanting private variants, this is an error where I wish we had them. I'm not sure if all user errors should be at a top level `Kind::User(_)`, or if they should be split up. But with `Kind::User(_)` catching what's currently here, it'd be a behavior change to move them... That's my conundrum.

nox: If you have thoughts on how to rework `User` errors, I'm all ears cause I don't like them very much.

For example, I don't see what "the user" has to do with errors returned by `HttpBody::poll_data`. To me the user of Hyper is the author of the product that makes use of Hyper, not the eyeball who is using our product and sending us stuff that makes `poll_data` return an error.

For this reason, I very much want the various user variants to be exposed.

seanmonstar: My perception of user is a user of hyper. So the application developer who used hyper to do a thing.

It may also be true that some of the things I have classified as user errors aren't exactly. It's a little fuzzy.

seanmonstar:
So I just remembered something else related to this. Many of the variants in here are not something I'd want to make public. I was simply using variants so I wouldn't need to write a full descriptive string at the usage location.

I think some of this gets back to what can a developer really hope to do with an error. They likely care about certain classes of errors, like a timeout or something, to retry with a different hostname. But many others, the only thing to really be done is report an error to a human.

Do you know which variants you specifically would hope to react to, versus which you would just log? Maybe there's simpler classifications we can expose.

seanmonstar:
I don't think we'll do this. The design of `Error` was mentioned in the 1.0 roadmap, and is tracked in #2845.

PR #2506, opened 2021-04-17
Permit allowing misplaced spaces in responses

nox:


src/client/client.rs
```diff
@@ -961,6 +961,18 @@ impl Builder {
         self
     }
 
+    /// Set whether HTTP/1 connections will accept spaces between header names
+    /// and the colon that follow them in responses.
+    ///
+    /// Note that this setting does not affect HTTP/2.
+    ///
+    /// Default is false.
```
seanmonstar, line 969: What do you think of including a quote from the spec that most people don't want to enable this, as it's supposed to be illegal, but proxies _may_ want to.

nox: I have added a comment citing the RFC.

seanmonstar (approved):

PR #2510, opened 2021-04-22
fix(http1): http1_title_case_headers should move Builder

zonyitoo:
ref #2480 

@seanmonstar It was released in v0.14.6. Consider making a v0.14.7 and yank it. 😅


nox:
Oh dang, I did not notice that. I did notice #2511 though 😅

nox:
I've cherry-picked your commit in my own PR to make use of the right setter in the proxy example.

seanmonstar:
Thanks, I merged #2511 which included this commit too.

PR #2511, opened 2021-04-22
Implement forgotten settings for case preserving

nox:
Turns out I didn't implement the server-side public setters for #2480, oopsie.

### DIFF
--- a/examples/http_proxy.rs
+++ b/examples/http_proxy.rs
@@ -23,14 +23,21 @@ type HttpClient = Client<hyper::client::HttpConnector>;
 #[tokio::main]
 async fn main() {
     let addr = SocketAddr::from(([127, 0, 0, 1], 8100));
-    let client = HttpClient::new();
+
+    let client = Client::builder()
+        .http1_title_case_headers(true)
+        .http1_preserve_header_case(true)
+        .build_http();
 
     let make_service = make_service_fn(move |_| {
         let client = client.clone();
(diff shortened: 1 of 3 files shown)


zonyitoo:
What's the differences between `http1_title_case_headers` and `http1_preserve_header_case`?

nox:
`http1_title_case_headers`: `Sec-WebSocket-Key` becomes `Sec-Websocket-Key`.
`http1_preserve_header_case`: `Sec-WebSocket-Key` is sent as is.

Note that you can combine the two, so that existing headers in a given request will preserve their original case, but headers added by Hyper itself (say, `date`) will use the title case.

zonyitoo:
So `http1_title_case_headers` won't make `date` become `Date`?

nox:
It will, but `http1_preserve_header_case` won't on its own, that's why you need the two. *Preserve* means there was an original casing to begin with (i.e. the casing that was found when on the wire when parsing the HTTP message), that's not the case for headers emitted by Hyper itself.

nox:
Also, note that Hyper still doesn't let you pass arbitrary header case, as `HeaderCaseMap` is still private, so `http1_preserve_header_case` is only useful for proxies right now, where you receive a request from one end, and send it with the client on the other end.

nox:
This now includes @zonyitoo's fix from #2510.

seanmonstar:
Thanks for the quick fix!

PR #2513, opened 2021-04-24
Support requets to domain names in example http_server

telotic:
A `String` authority will be converted to `SocketAddr` (https://docs.rs/tokio/1.5.0/tokio/net/trait.ToSocketAddrs.html) via under-the-hood DNS lookups. The proxy only needs to check whether the authority part exists (the uri is an absolute uri) to support requests to domain names. This would make the example even more useful.

Another change I packed into this PR is to slightly simplify the `tunnel` method using `tokio::io::copy_bidirectional`. I'm open for objections if this introduces some side effects that I'm unware of.


seanmonstar (approved):
Nice clean up, thanks!

PR #2522, opened 2021-04-28
Don't keep-alive after an upgrading request

nox:


nox:
Oh right, I guess if I want to do that I need to explicitly send `Connection: close` in the server response so that clients don't mistakenly think the thing was kept alive. Will rework this later.

PR #2523, opened 2021-04-28
Implement CONNECT support over HTTP/2

nox:
There are quite a few details to sort out, grep for `FIXME(nox)`.

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -2,17 +2,21 @@ use std::error::Error as StdError;
 #[cfg(feature = "runtime")]
 use std::time::Duration;
 
+use bytes::Bytes;
 use futures_channel::{mpsc, oneshot};
 use futures_util::future::{self, Either, FutureExt as _, TryFutureExt as _};
 use futures_util::stream::StreamExt as _;
 use h2::client::{Builder, SendRequest};
+use http::{Method, StatusCode};
 use tokio::io::{AsyncRead, AsyncWrite};
 
-use super::{decode_content_length, ping, PipeToSendStream, SendBuf};
+use super::{ping, H2Upgraded, PipeToSendStream, SendBuf};
 use crate::body::HttpBody;
 use crate::common::{exec::Exec, task, Future, Never, Pin, Poll};
 use crate::headers;
+use crate::proto::h2::UpgradedSendStream;
 use crate::proto::Dispatched;
+use crate::upgrade::Upgraded;
 use crate::{Body, Request, Response};
 
 type ClientRx<B> = crate::client::dispatch::Receiver<Request<B>, Response<Body>>;
@@ -233,8 +237,25 @@ where
                             headers::set_content_length_if_missing(req.headers_mut(), len);
                         }
                     }
+
+                    let is_connect = req.method() == Method::CONNECT;
                     let eos = body.is_end_stream();
-                    let (fut, body_tx) = match self.h2_tx.send_request(req, eos) {
+                    let ping = self.ping.clone();
+
+                    if is_connect {
+                        if headers::content_length_parse_all(req.headers())
+                            .map_or(false, |len| len != 0)
+                        {
+                            warn!("h2 connect request with non-zero body not supported");
+                            cb.send(Err((
+                                crate::Error::new_h2(h2::Reason::INTERNAL_ERROR.into()),
+                                None,
+                            )));
+                            continue;
+                        }
+                    }
+
+                    let (fut, body_tx) = match self.h2_tx.send_request(req, !is_connect && eos) {
                         Ok(ok) => ok,
                         Err(err) => {
                             debug!("client send request error: {}", err);
@@ -243,45 +264,81 @@ where
                         }
                     };
 
-                    let ping = self.ping.clone();
-                    if !eos {
-                        let mut pipe = Box::pin(PipeToSendStream::new(body, body_tx)).map(|res| {
-                            if let Err(e) = res {
-                                debug!("client request body error: {}", e);
-                            }
-                        });
-
-                        // eagerly see if the body pipe is ready and
-                        // can thus skip allocating in the executor
-                        match Pin::new(&mut pipe).poll(cx) {
-                            Poll::Ready(_) => (),
-                            Poll::Pending => {
-                                let conn_drop_ref = self.conn_drop_ref.clone();
-                                // keep the ping recorder's knowledge of an
-                                // "open stream" alive while this body is
-                                // still sending...
-                                let ping = ping.clone();
-                                let pipe = pipe.map(move |x| {
-                                    drop(conn_drop_ref);
-                                    drop(ping);
-                                    x
+                    let send_stream = if !is_connect {
+                        if !eos {
+                            let mut pipe =
+                                Box::pin(PipeToSendStream::new(body, body_tx)).map(|res| {
+                                    if let Err(e) = res {
+                                        debug!("client request body error: {}", e);
+                                    }
                                 });
-                                self.executor.execute(pipe);
+
+                            // eagerly see if the body pipe is ready and
+                            // can thus skip allocating in the executor
+                            match Pin::new(&mut pipe).poll(cx) {
+                                Poll::Ready(_) => (),
+                                Poll::Pending => {
+                                    let conn_drop_ref = self.conn_drop_ref.clone();
+                                    // keep the ping recorder's knowledge of an
+                                    // "open stream" alive while this body is
+                                    // still sending...
+                                    let ping = ping.clone();
+                                    let pipe = pipe.map(move |x| {
+                                        drop(conn_drop_ref);
+                                        drop(ping);
+                                        x
+                                    });
+                                    self.executor.execute(pipe);
+                                }
                             }
                         }
-                    }
+
+                        None
+                    } else {
+                        Some(body_tx)
+                    };
 
                     let fut = fut.map(move |result| match result {
                         Ok(res) => {
                             // record that we got the response headers
                             ping.record_non_data();
 
-                            let content_length = decode_content_length(res.headers());
-                            let res = res.map(|stream| {
-                                let ping = ping.for_stream(&stream);
-                                crate::Body::h2(stream, content_length, ping)
-                            });
-                            Ok(res)
+                            let content_length = headers::content_length_parse_all(res.headers());
+                            if let (Some(mut send_stream), StatusCode::OK) =
+                                (send_stream, res.status())
+                            {
+                                if content_length.map_or(false, |len| len != 0) {
+                                    warn!("h2 connect response with non-zero body not supported");
+
+                                    send_stream.send_reset(h2::Reason::INTERNAL_ERROR);
+                                    return Err((
+                                        crate::Error::new_h2(h2::Reason::INTERNAL_ERROR.into()),
+                                        None,
+                                    ));
+                                }
+                                let (parts, recv_stream) = res.into_parts();
+                                let mut res = Response::from_parts(parts, Body::empty());
+
+                                let (pending, on_upgrade) = crate::upgrade::pending();
+                                let io = H2Upgraded {
+                                    ping,
+                                    send_stream: unsafe { UpgradedSendStream::new(send_stream) },
+                                    recv_stream,
+                                    buf: Bytes::new(),
+                                };
+                                let upgraded = Upgraded::new(io, Bytes::new());
+
+                                pending.fulfill(upgraded);
+                                res.extensions_mut().insert(on_upgrade);
+
+                                Ok(res)
+                            } else {
+                                let res = res.map(|stream| {
+                                    let ping = ping.for_stream(&stream);
+                                    crate::Body::h2(stream, content_length.into(), ping)
+                                });
+                                Ok(res)
+                            }
                         }
                         Err(err) => {
                             ping.ensure_not_timed_out().map_err(|e| (e, None))?;
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -1,5 +1,5 @@
-use bytes::Buf;
-use h2::SendStream;
+use bytes::{Buf, Bytes};
+use h2::{RecvStream, SendStream};
 use http::header::{
     HeaderName, CONNECTION, PROXY_AUTHENTICATE, PROXY_AUTHORIZATION, TE, TRAILER,
     TRANSFER_ENCODING, UPGRADE,
@@ -7,11 +7,14 @@ use http::header::{
 use http::HeaderMap;
 use pin_project::pin_project;
 use std::error::Error as StdError;
-use std::io::IoSlice;
+use std::io::{self, Cursor, IoSlice};
+use std::mem;
+use std::task::Context;
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
-use crate::body::{DecodedLength, HttpBody};
+use crate::body::HttpBody;
 use crate::common::{task, Future, Pin, Poll};
-use crate::headers::content_length_parse_all;
+use crate::proto::h2::ping::Recorder;
 
 pub(crate) mod ping;
 
@@ -80,15 +83,6 @@ fn strip_connection_headers(headers: &mut HeaderMap, is_request: bool) {
             let name = name.trim();
             headers.remove(name);
         }
-    }
-}
-
-fn decode_content_length(headers: &HeaderMap) -> DecodedLength {
-    if let Some(len) = content_length_parse_all(headers) {
-        // If the length is u64::MAX, oh well, just reported chunked.
-        DecodedLength::checked_new(len).unwrap_or_else(|_| DecodedLength::CHUNKED)
-    } else {
-        DecodedLength::CHUNKED
     }
 }
 
@@ -172,7 +166,7 @@ where
                             is_eos,
                         );
 
-                        let buf = SendBuf(Some(chunk));
+                        let buf = SendBuf::Buf(chunk);
                         me.body_tx
                             .send_data(buf, is_eos)
                             .map_err(crate::Error::new_body_write)?;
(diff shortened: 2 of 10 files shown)


tests/server.rs
```diff
@@ -1482,6 +1482,72 @@ async fn http_connect_new() {
...
+    use tokio::io::{AsyncReadExt, AsyncWriteExt};
+    let _ = pretty_env_logger::try_init();
+    let listener = tcp_bind(&"127.0.0.1:0".parse().unwrap()).unwrap();
+    let addr = listener.local_addr().unwrap();
+    let (done_tx, mut done_rx) = tokio::sync::mpsc::channel(10);
+
+    let conn = connect_async(addr).await;
+    let (h2, connection) = h2::client::handshake(conn).await.unwrap();
+    tokio::spawn(async move {
+        dbg!(connection.await).unwrap();
```
inikulin, line 1496: `dbg!` leftover

src/proto/h2/mod.rs
```diff
@@ -243,32 +246,155 @@ impl<B: Buf> SendStreamExt for SendStream<SendBuf<B>> {
...
-        self.0.as_ref().map(|b| b.chunks_vectored(dst)).unwrap_or(0)
+        match *self {
+            Self::Buf(ref b) => b.chunks_vectored(dst),
+            Self::Cursor(ref c) => c.chunks_vectored(dst),
+            Self::None => 0,
+        }
+    }
+}
+
+// FIXME(nox): Should this type be public? I'm asking this because
```
cbranch, line 297: This doesn't have to be public. The only valid use case I can think of is to call `SendStream::send_reset` with the appropriate error condition. You could achieve that with public function e.g. `h2_reset(&mut Upgraded, h2::Reason)` which for h2 streams (i.e. they're downcastable to `H2Upgraded`) sends a RST_STREAM.

src/proto/h2/server.rs
```diff
@@ -267,8 +274,28 @@ where
...
+                        let (req, connect_parts) = if !is_connect {
+                            (
+                                Request::from_parts(
+                                    parts,
+                                    crate::Body::h2(stream, content_length, ping),
+                                ),
+                                None,
+                            )
+                        } else {
+                            // FIXME(nox): What happens to the request body? Should we check `content_length`?
```
cbranch, line 288: iirc in h1 the body is returned as usual, and should be here too: the case is where you get a response that is *not* upgradable (eg you expected 101 and got 4xx) and you decide not to take the upgrade.

nox: Ack, I'll change the code accordingly.

nox: Actually, Hyper does return an error if you attempt to respond with a successful status code and a body to a CONNECT request:

https://github.com/hyperium/hyper/blob/dbea7716f157896bf7d2d417be7b4e382e7dc34f/src/proto/h1/role.rs#L322-L327

I'll do the same for h2.

nox: Wait never mind, that's about the response, not the request body.

cbranch: No, I think this is my fault - as a client you want to be able to read the response body as a body. For a server reading a CONNECT request, any "body" should be present in the upgraded stream rather than the request, even if you choose not to upgrade the request.

nox: I made non-zero request and response bodies be an error condition for upgrades, for now. We can always make Hyper support bodies on those requests later, and h1 already reject those bodies in responses on the server-side.

src/proto/h2/mod.rs
nox: I just removed the comment, we can always file a follow-up for implementing that `h2_reset` call. I don't even think we could make use of such a method in our current setup at work.

tests/server.rs
```diff
@@ -1482,6 +1482,72 @@ async fn http_connect_new() {
     assert_eq!(s(&vec), "bar=foo");
 }
 
+#[tokio::test]
+async fn h2_connect() {
```
inikulin, line 1486: it would be interesting to have a test for CONNECTs multiplexed over a single h2 connection

tests/server.rs
```diff
@@ -1482,6 +1482,72 @@ async fn http_connect_new() {
...
+        assert_eq!(response.status(), StatusCode::OK);
+
+        let bytes = response.body_mut().data().await.unwrap().unwrap();
+        assert_eq!(&bytes[..], b"Bread?");
+        let _ = response
+            .body_mut()
+            .flow_control()
+            .release_capacity(bytes.len());
+
+        send_stream.send_data("Baguette!".into(), true).unwrap();
```
inikulin, line 1514: From previous experience it's worth testing quite long bodies that don't fit in MTU as a whole to ensure that pending reads/write are waken up correctly. And 0 length bodies as a special case.

Also, for the multiplexing case: if one of the streams error does it terminate the rest of the tunnels along with the connection?

inikulin (changes requested):

tests/server.rs
nox: I'll write additional tests, staying away from libstd channels this time 🤡

nox: There is now a second stream that connects twice and makes one of the stream reset while the other one does its thing successfully.

tests/server.rs
nox: I added a test with a very large body. I'm not sure what you want me to test for 0 length bodies, as receiving 0 bytes usually signal EOF?

src/proto/h2/mod.rs
```diff
@@ -243,32 +241,149 @@ impl<B: Buf> SendStreamExt for SendStream<SendBuf<B>> {
...
+        cx: &mut Context<'_>,
+        read_buf: &mut ReadBuf<'_>,
+    ) -> Poll<Result<(), io::Error>> {
+        if self.buf.is_empty() {
+            self.buf = match ready!(self.recv_stream.poll_data(cx)) {
+                None => return Poll::Ready(Ok(())),
+                Some(Ok(buf)) => {
+                    self.ping.record_data(buf.len());
+                    buf
+                }
```
nox, line 317: I guess that if we receive an empty buffer and `self.recv_stream.is_end_stream()` is false, then we should probably poll again to be sure we don't prematurely return `Ok(())` without putting anything in `read_buf`.

nox: Made the change and added a test for that.

tests/server.rs
nox: I added a test that checks that 0-length non-EOS data frames don't make the AsyncRead impl bail out.

tests/server.rs
```diff
@@ -1482,6 +1485,303 @@ async fn http_connect_new() {
...
+    }
+
+    tokio::spawn(async move {
+        let (mut recv_stream_1, mut send_stream_1) = connect_and_recv_bread(&mut h2).await;
+
+        let (_, mut send_stream_2) = connect_and_recv_bread(&mut h2).await;
+
+        send_stream_1.send_data("Baguette!".into(), true).unwrap();
+
+        send_stream_2.send_reset(h2::Reason::PROTOCOL_ERROR);
```
inikulin, line 1596: I would like more streams :) Let's say 50 transfering data and 50 error.

inikulin: Also, in case of error, is it propagated to the other peer's Upgraded "socket"?

inikulin: let's cover that in the test too (if not yet, sorry, didn't have a chance to give it a full round of review yet)

nox: Added some assertion to check the error kind and the error payload, with 80 clients stopping at 4 different points.

inikulin:
@seanmonstar may I ask you to take a look as well?

src/proto/h2/mod.rs
```diff
@@ -243,32 +241,152 @@ impl<B: Buf> SendStreamExt for SendStream<SendBuf<B>> {
...
-        self.send_data(SendBuf(None), true)
+        self.send_data(SendBuf::None, true)
             .map_err(crate::Error::new_body_write)
     }
 }
 
-struct SendBuf<B>(Option<B>);
+enum SendBuf<B> {
+    Buf(B),
+    Cursor(Cursor<Box<[u8]>>),
```
seanmonstar, line 251: Whatcha think of this being `Bytes`, and `H2Upgraded::write` pushing into a `BytesMut`? I can't say if it's actually better, it'd only be better in these situations:

- The `BytesMut` reserves more space and the `split_off` means the other space is used without allocating.
- The sliced off `Bytes` are flushed on the connection and dropped before another `poll_write`, which means the `BytesMut` would know it could reuse its buffer instead of allocating a new one.

Not a strong desire here, just curious what you think.

src/proto/h2/mod.rs
```diff
@@ -243,32 +241,152 @@ impl<B: Buf> SendStreamExt for SendStream<SendBuf<B>> {
...
+    fn poll_write(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        if buf.is_empty() {
+            return Poll::Ready(Ok(0));
+        }
+        // FIXME(nox): PipeToSendStream does some weird stuff, first reserving
+        // one byte and then polling reset if the capacity is 0, should we do
```
seanmonstar, line 347: Well, the reserving 1 byte is that we don't yet know how big the next data chunk might be. Here, you know how big `buf` is, so that different is fine.

The polling of reset is because `poll_capacity` won't return stream errors, only "user errors". So in the `PipeToSendStream`, it's registering the task to listen for a reset while waiting for the user's stream to produce an item too. It looks like `send_data` _also_ doesn't return resets, just "user errors". So maybe it _does_ need to poll for reset in here too?

src/proto/h2/mod.rs
```diff
@@ -243,32 +241,152 @@ impl<B: Buf> SendStreamExt for SendStream<SendBuf<B>> {
...
+        // that here too? Should we poll reset somewhere?
+        self.send_stream.reserve_capacity(buf.len());
+        Poll::Ready(match ready!(self.send_stream.poll_capacity(cx)) {
+            None => Ok(0),
+            Some(Ok(cnt)) => self.write(&buf[..cnt], false).map(|()| cnt),
+            Some(Err(e)) => {
+                // FIXME(nox): Should all H2 errors be returned as is with a
```
seanmonstar, line 354: What's your recommendation for this FIXME? I think what you have so far is probably fine to start.

src/proto/h2/server.rs
```diff
@@ -331,18 +362,22 @@ enum H2StreamState<F, B>
 where
     B: HttpBody,
 {
-    Service(#[pin] F),
+    Service(#[pin] F, Option<(Pending, Recorder, RecvStream)>),
```
seanmonstar, line 365: What do you think of turning this into a struct variant? Mostly just to give a label to what this `Option` is for?

seanmonstar: Oh, does that play well with pin projection? Maybe it's an annoying change, and just a doc comment would be enough to remind the next spelunker.

src/server/conn.rs
```diff
@@ -615,6 +615,7 @@ where
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     I: AsyncRead + AsyncWrite + Unpin,
     B: HttpBody + 'static,
+    B::Data: Send,
```
seanmonstar, line 618: Is there something that was already requiring this bound? If not, I think it could be a breaking change, no?

seanmonstar (commented):
This is fantastic, excellent work! Love the detailed tests too!

src/proto/h2/mod.rs
nox: I'm sorry I don't understand either of your bullet points 😅

What would the `BytesMut` do, and when would we actually send the buffer to the h2 stream? Do you mean that `poll_write` would never send a partial buffer anymore? Wouldn't that be bad for flow control?

src/proto/h2/mod.rs
nox: When should we call `poll_reset`? Before `reserve_capacity`? Before `poll_capacity`? When `poll_capacity` returns `None`?

src/proto/h2/mod.rs
nox: I think it's fine too, though maybe we want the cause to be a `hyper::Error` and not a `h2::Error`?

src/proto/h2/server.rs
nox: I think pin projection handles struct variants just fine, I'll try to name the fields.

src/server/conn.rs
nox: I'll carefully compare the rustdoc outputs before and after that change, that's why I made it a separate commit in the first place heh.

src/proto/h2/server.rs
```diff
@@ -162,6 +162,7 @@ where
     S: HttpService<Body, ResBody = B>,
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     B: HttpBody + 'static,
+    B::Data: Send,
```
nox, line 170: This bound is on an impl for a `pub(crate)` item.

src/proto/h2/server.rs
```diff
@@ -206,6 +207,7 @@ impl<T, B> Serving<T, B>
 where
     T: AsyncRead + AsyncWrite + Unpin,
     B: HttpBody + 'static,
+    B::Data: Send,
```
nox, line 215: This bound is on an impl for a private item.

src/proto/h2/server.rs
```diff
@@ -364,6 +366,7 @@ impl<F, B, E> H2Stream<F, B>
 where
     F: Future<Output = Result<Response<B>, E>>,
     B: HttpBody,
+    B::Data: Send + 'static,
```
nox, line 402: This bound is on an impl for a `pub` item. That item is `H2Stream<F, B>`, which is used in `ConnStreamExec<F, B>`, which is also `pub` but not reachable, and used to bind `E` in the various `Connection<I, S, E>` impl blocks.

src/proto/h2/server.rs
```diff
@@ -428,6 +431,7 @@ impl<F, B, E> Future for H2Stream<F, B>
...
```
nox, line 488: Same as above.

src/server/conn.rs
nox: This is indeed a breaking change. I can limit its scope to the two methods `poll_without_shutdown` and `without_shutdown`.

src/server/conn.rs
```diff
@@ -772,6 +773,7 @@ where
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     I: AsyncRead + AsyncWrite + Unpin + 'static,
     B: HttpBody + 'static,
+    B::Data: Send,
```
nox, line 776: Breaking change.

src/server/conn.rs
```diff
@@ -978,6 +980,7 @@ where
     S: HttpService<Body, ResBody = B>,
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     B: HttpBody + 'static,
+    B::Data: Send,
```
nox, line 983: `ProtoServer<T, B, S, E>` is not `pub`.

src/server/conn.rs
```diff
@@ -1028,6 +1031,7 @@ pub(crate) mod spawn_all {
         S: HttpService<Body>,
         E: ConnStreamExec<S::Future, S::ResBody>,
         S::ResBody: 'static,
+        <S::ResBody as HttpBody>::Data: Send,
```
nox, line 1034: This is on `NoopWatcher`. It is a `pub` item but it is not actually reachable from outside the crate. It is used in the `Future` impl for `Server<I, S, E>`, in a bound on `E`.

src/server/conn.rs
```diff
@@ -1075,6 +1079,7 @@ pub(crate) mod spawn_all {
         NE: Into<Box<dyn StdError + Send + Sync>>,
         S: HttpService<Body, ResBody = B>,
         B: HttpBody + 'static,
+        B::Data: Send,
```
nox, line 1082: This is on `NewSvcTask<I, N, S, E, W>`, like `NoopWatcher` it's not actually reachable, but it is used in the `NewSvcExec<I, N, S, E, W>` trait.

src/server/conn.rs
```diff
@@ -1141,6 +1146,7 @@ mod upgrades {
         S::Error: Into<Box<dyn StdError + Send + Sync>>,
         I: AsyncRead + AsyncWrite + Unpin,
         B: HttpBody + 'static,
+        B::Data: Send,
```
nox, line 1149: Breaking change.

src/server/conn.rs
```diff
@@ -1159,6 +1165,7 @@ mod upgrades {
         S::Error: Into<Box<dyn StdError + Send + Sync>>,
         I: AsyncRead + AsyncWrite + Unpin + Send + 'static,
         B: HttpBody + 'static,
+        B::Data: Send,
```
nox, line 1168: Breaking change.

src/server/server.rs
```diff
@@ -100,6 +100,7 @@ where
     S: MakeServiceRef<IO, Body, ResBody = B>,
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     B: HttpBody + Send + Sync + 'static,
+    B::Data: Send,
```
nox, line 103: Breaking change.

src/server/server.rs
```diff
@@ -156,6 +157,7 @@ where
     S: MakeServiceRef<IO, Body, ResBody = B>,
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     B: HttpBody + 'static,
+    B::Data: Send,
```
nox, line 160: Breaking change.

src/server/server.rs
```diff
@@ -426,6 +428,7 @@ impl<I, E> Builder<I, E> {
         S: MakeServiceRef<I::Conn, Body, ResBody = B>,
         S::Error: Into<Box<dyn StdError + Send + Sync>>,
         B: HttpBody + 'static,
+        B::Data: Send,
```
nox, line 431: Breaking change.

src/server/shutdown.rs
```diff
@@ -51,6 +51,7 @@ where
     S: MakeServiceRef<IO, Body, ResBody = B>,
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     B: HttpBody + Send + Sync + 'static,
+    B::Data: Send,
```
nox, line 54: Breaking change.

src/server/shutdown.rs
```diff
@@ -98,6 +99,7 @@ where
     S: HttpService<Body>,
     E: ConnStreamExec<S::Future, S::ResBody>,
     S::ResBody: Send + Sync + 'static,
+    <S::ResBody as HttpBody>::Data: Send,
```
nox, line 102: Breaking change.

src/server/shutdown.rs
```diff
@@ -114,6 +116,7 @@ where
     S::Error: Into<Box<dyn StdError + Send + Sync>>,
     I: AsyncRead + AsyncWrite + Unpin,
     S::ResBody: HttpBody + Send + 'static,
+    <S::ResBody as HttpBody>::Data: Send,
```
nox, line 119: Breaking change.

nox (commented):
Made an exhaustive review of the additional bounds, they are unfortunately indeed breaking changes.

src/proto/h2/mod.rs
nox: I removed the comment. Feel free to ask me to change the inner payload to `hyper::Error` but IMO that's not every important.

src/proto/h2/mod.rs
nox: I now call `poll_reset` at the very beginning of the function.

src/proto/h2/server.rs
nox: Rather than naming the fields of the enum variant, I introduced a new structure for the triplet.

nox:
I have an idea to remove the bound on `B::Data`.

src/proto/h2/server.rs
```diff
@@ -364,6 +364,7 @@ impl<F, B, E> H2Stream<F, B>
 where
     F: Future<Output = Result<Response<B>, E>>,
     B: HttpBody,
+    B::Data: 'static,
```
nox, line 410: `H2Stream` is private, that is not a breaking change.

src/proto/h2/server.rs
```diff
@@ -428,6 +429,7 @@ impl<F, B, E> Future for H2Stream<F, B>
...
```
nox, line 498: `H2Stream` is private, that is not a breaking change.

src/proto/h2/mod.rs
```diff
@@ -243,32 +242,201 @@ impl<B: Buf> SendStreamExt for SendStream<SendBuf<B>> {
...
+
+#[repr(transparent)]
+struct Neutered<B> {
+    _inner: B,
+    impossible: Impossible,
+}
+
+enum Impossible {}
+
+unsafe impl<B> Send for Neutered<B> {}
```
nox, line 428: This is my cursed idea to remove the `Send` bounds on `B::Data`. We know we are not ever actually going to send a `B::Data` at any point over that stream once it is upgraded, so AFAIK it doesn't matter whether it is `Send` or not.

So instead of introducing that `Send` bound, we transmute `SendStream<SendBuf<B>>` to `SendStream<SendBuf<Neutered<B>>>` where `Neutered<B>` forcefully implements `Send` because it cannot actually be fully initialised, as one of its fields is an empty type.

`Neutered<B>` though still needs to implement `Buf` because the definition of `SendStream<B>` itself requires `B: Buf` (that could probably be relaxed in h2, now that I think about it, I personally dislike unnecessary trait bounds on type definitions, they just make the code more verbose).

Note that we don't actually use the `SendStream<SendBuf<Neutered<B>>>`, instead we use `UpgradedSendStream::as_inner_unchecked` to retrieve the `SendStream<SendBuf<B>>` to minimize the amounts of monomorphisations for `SendStream<_>` methods.

nox (commented):
I have implemented my cursed idea, tell me what you think about it.

src/proto/h2/mod.rs
seanmonstar: It was more just a passing thought, but I don't have data that in this instance it would be an improvement. Maybe something to try out separately and see if there is actually any performance improvement, or if I'm spouting nonsense.

src/proto/h2/mod.rs
seanmonstar: This is quite the monster. I do think it's clear what it does (maybe a comment on the `Neutered` type and/or `impossible` field would be good). It's clever, I think I like it!

My one question is if the transmute is safe. I know the empty enum has no size, but could optimizations in the compiler take advantage of it being impossible to create that type, and decide "it's impossible to create `SendBuf::Buf(Neutered)`, so we'll just chop that variant out completely"? I'd prefer if we find the answer, we could include a code comment pointing to that guarantee.

nox: > My one question is if the transmute is safe. I know the empty enum has no size, but could optimizations in the compiler take advantage of it being impossible to create that type, and decide "it's impossible to create SendBuf::Buf(Neutered), so we'll just chop that variant out completely"? I'd prefer if we find the answer, we could include a code comment pointing to that guarantee.

I'll add an assertion, but AFAIK rustc never kills variants that have *some* impossible fields but have others that are not zero-sized.

nox: Wait I already check this, that's what the assertion in `UpgradedSendStream::new` verifies.

nox: I've added `#[repr(usize)]` on `SendBuf<B>` to be extra sure rustc doesn't get fancy about the representation of the enum.

nox:
I added preliminary client-side support.

src/proto/h2/client.rs
```diff
@@ -243,45 +259,76 @@ where
...
-                            let content_length = decode_content_length(res.headers());
-                            let res = res.map(|stream| {
-                                let ping = ping.for_stream(&stream);
-                                crate::Body::h2(stream, content_length, ping)
-                            });
-                            Ok(res)
+                            let content_length = headers::content_length_parse_all(res.headers());
+                            if let Some(mut send_stream) = send_stream {
+                                if content_length.map_or(false, |len| len != 0) {
+                                    warn!("h2 connect response with non-zero body not supported");
```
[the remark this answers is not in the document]
nox: No it doesn't, good catch.

src/proto/h2/server.rs
```diff
@@ -267,8 +272,36 @@ where
...
+                            )
+                        } else {
+                            if content_length.map_or(false, |len| len != 0) {
+                                warn!("h2 connect request with non-zero body not supported");
+                                respond.send_reset(h2::Reason::INTERNAL_ERROR);
+                                return Poll::Ready(Ok(()));
+                            }
+                            let (pending, upgrade) = crate::upgrade::pending();
+                            debug_assert!(parts.extensions.get::<OnUpgrade>().is_none());
+                            parts.extensions.insert(upgrade);
```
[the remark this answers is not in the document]
nox: `h2` itself will fail if passed a non-authority URI for a CONNECT request, on the Hyper side I just did like what already exists for HTTP/1.

src/proto/h2/server.rs
```diff
@@ -267,8 +272,36 @@ where
...
+                                Request::from_parts(
+                                    parts,
+                                    crate::Body::h2(stream, content_length.into(), ping),
+                                ),
+                                None,
+                            )
+                        } else {
+                            if content_length.map_or(false, |len| len != 0) {
+                                warn!("h2 connect request with non-zero body not supported");
+                                respond.send_reset(h2::Reason::INTERNAL_ERROR);
```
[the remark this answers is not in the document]
nox: Does it truly break the rules? I don't remember seeing prose explicitly rejecting this.

nox (commented):
Replied to Lucas' comments, I'll make sure we don't reject bodies on non-200 responses.

src/proto/h2/server.rs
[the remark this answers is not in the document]
LPardue: SGTM, thanks for explaining.

src/proto/h2/server.rs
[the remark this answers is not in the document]
LPardue: So the most current HTTP specs (close to being RFCs) say https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.9.3.6.p.12

> A server MUST NOT send any Transfer-Encoding or Content-Length header fields in a 2xx (Successful) response to CONNECT. A client MUST ignore any Content-Length or Transfer-Encoding header fields received in a successful response to CONNECT.

> A CONNECT request message does not have content. The interpretation of and allowability of data sent after the header section of the CONNECT request message is specific to the version of HTTP in use.

And the HTTP/2 spec doesn't seem to clearly state things either way. No point splitting hairs over an error code so lets leave that. In the meantime I opened an issue on the HTTP/2 spec in the hope that RFC 7540 CONNECT definition could get tightened up. See https://github.com/httpwg/http2-spec/issues/849

seanmonstar (approved):
Stupendous!

PR #2527, opened 2021-04-29
feat(http2): allow HTTP/2 requests by ALPN when http2_only is unset

goffrie:
Previously, Hyper would unconditionally reject requests with `Version::HTTP_2` if `http2_only(true)` was not provided when building the client. This change allows such requests to succeed as long as the underlying connection is actually negotiated as HTTP/2.

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -167,11 +167,7 @@ where
                     )));
                 }
             }
-            other_h2 @ Version::HTTP_2 => {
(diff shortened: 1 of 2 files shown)


seanmonstar:
@sfackler This looks right to me, but want to look as well, since you've done a bit with ALPN?

seanmonstar (approved):

PR #2529, opened 2021-04-30
feat: Join headers from http/2 requests on http/1 connections

### LINKED ISSUE #2528: Hyper does not re-join cookie headers split by http/2
As per https://tools.ietf.org/html/rfc7540#section-8.1.2.5 http/2 allows splitting cookie headers into multiple parts if it improves compression efficiency. The spec mandates that the header needs to be rejoined if the associated request is proxied to a http/1 server, though.

> To allow for better compression efficiency, the Cookie header field
> MAY be split into separate header fields, each with one or more
> cookie-pairs.  If there are multiple Cookie header fields after
> decompression, these MUST be concatenated into a single octet string
> using the two-octet delimiter of 0x3B, 0x20 (the ASCII string "; ")
> before being passed into a non-HTTP/2 context, such as an HTTP/1.1
> connection, or a generic HTTP server application.

Currently, the Hyper server correctly (?) exposes the split cookie header as separate entries in the `HeaderMap`, but if the request is then proxied to a h1 server, the `Client` does not re-join them again (although it should).

I do not think it's desirable to _always_ and _forcibly_ re-join cookie headers before sending them out on http/1 connections, given that hyper is a low-level library and users might want to experiment with that. It might also be a breaking change for users of this library.

Therefore I propose to only re-join headers iff the incoming request was h2 and the outgoing one is h1. Hyper neatly exposes the HTTP version on the request, which should let one do that. I'd be happy to send a PR implementing that.

NeoLegends:
Closes #2528.

Added a test for the function re-joining the headers, and extended the coercion integration test!

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -2,15 +2,16 @@
 // can't upgrade yet
 #![allow(deprecated)]
 
+use std::convert::TryFrom;
 use std::fmt::{self, Write};
 use std::mem;
 
 #[cfg(any(test, feature = "server", feature = "ffi"))]
 use bytes::Bytes;
 use bytes::BytesMut;
-use http::header::{self, Entry, HeaderName, HeaderValue};
 #[cfg(feature = "server")]
 use http::header::ValueIter;
+use http::header::{self, Entry, HeaderName, HeaderValue};
 use http::{HeaderMap, Method, StatusCode, Version};
 
 use crate::body::DecodedLength;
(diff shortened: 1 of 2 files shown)


src/proto/h1/role.rs
```diff
@@ -1033,6 +1033,16 @@ impl Http1Transaction for Client {
...
+
+                // http/2 can sometimes split cookie headers into multiple haeder
+                // values, which is not compatible with http/1.
+                //
+                // We re-join them on behalf of the client.
+                //
+                // https://tools.ietf.org/html/rfc7540#section-8.1.2.5
+                if let Some(joined_cookies) = unsplit_cookie_headers(&msg.head.headers) {
+                    msg.head.headers.insert(header::COOKIE, joined_cookies);
+                }
```
nox, line 1045: Actually we probably just want to set a boolean value here, and use it later in `encode_headers` with a new `header::COOKIE` arm that would prevent writing the header name multiple times, joining them instead, to avoid looking up into the header map and mutating it.

NeoLegends:
I'm not super happy about 58a2930 yet, it feels like alot of repetition.

PR #2535, opened 2021-05-04
Allow to config h2 max concurrent reset streams.

boxdot:
Setting streams to 0 makes h2 work on wasm platforms without `Instant::now` implementation.

Related to:

* https://github.com/hyperium/h2/pull/511
* https://github.com/hyperium/h2/pull/537

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -10,7 +10,7 @@ use tokio::io::{AsyncRead, AsyncWrite};
 
 use super::{decode_content_length, ping, PipeToSendStream, SendBuf};
 use crate::body::HttpBody;
-use crate::common::{task, exec::Exec, Future, Never, Pin, Poll};
+use crate::common::{exec::Exec, task, Future, Never, Pin, Poll};
 use crate::headers;
 use crate::proto::Dispatched;
 use crate::{Body, Request, Response};
@@ -44,6 +44,7 @@ pub(crate) struct Config {
     pub(crate) keep_alive_timeout: Duration,
     #[cfg(feature = "runtime")]
     pub(crate) keep_alive_while_idle: bool,
+    pub(crate) max_concurrent_reset_streams: Option<usize>,
 }
 
 impl Default for Config {
@@ -59,6 +60,7 @@ impl Default for Config {
             keep_alive_timeout: Duration::from_secs(20),
             #[cfg(feature = "runtime")]
             keep_alive_while_idle: false,
+            max_concurrent_reset_streams: None,
         }
     }
 }
@@ -70,6 +72,9 @@ fn new_builder(config: &Config) -> Builder {
         .initial_connection_window_size(config.initial_conn_window_size)
         .max_frame_size(config.max_frame_size)
         .enable_push(false);
+    if let Some(max) = config.max_concurrent_reset_streams {
+        builder.max_concurrent_reset_streams(max);
+    }
     builder
 }
 
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -1151,6 +1151,21 @@ impl Builder {
         self
     }
 
+    /// Sets the maximum number of HTTP2 concurrent locally reset streams.
+    ///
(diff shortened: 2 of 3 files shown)


src/proto/h2/client.rs
```diff
@@ -44,6 +44,7 @@ pub(crate) struct Config {
     pub(crate) keep_alive_timeout: Duration,
     #[cfg(feature = "runtime")]
     pub(crate) keep_alive_while_idle: bool,
+    pub(crate) max_concurrent_reset_streams: Option<usize>,
```
pickfire, line 47: Can this be zero in the first place?

boxdot: If this is zero, then it will override the default value in `h2`, which 10. We could copy the default value from `h2` here, what do you think?

pickfire: Err, that seemed very implicit which is probably a bad idea. Can it be `NonZeroUsize` or so if it's `None` we can default to 10 but when it's `Some(n)` it will be the same? I am asking about `Some(0)` for the case but not sure if we really want that.

boxdot: I am not entirely sure what you are proposing, maybe we can discuss it again? Just to understand: there are 3 possible config options that we try to make available to the user:

1. Not specified => the default configuration value from `h2` will be used,
2. Specified `> 0` => `h2` will use reset streams,
3. Specified `0` => `h2` will not use reset streams at all.

To make it less implicit, maybe we should export the default configuration from `h2` and use it here?

pickfire: That seemed a bit confusing but when I look at it again. I think it may be better if we just make it `usize`.

1. Not specified => default to `proto::DEFAULT_RESET_STREAM_MAX`
2. Specified => `h2` used specified (`0` will not use reset streams at all)

So we already have the default value set by default, I don't see why we need `Option` since it will always have a value.

boxdot: Perfect. This definitely works.

src/proto/h2/client.rs
```diff
@@ -32,6 +32,10 @@ const DEFAULT_CONN_WINDOW: u32 = 1024 * 1024 * 5; // 5mb
 const DEFAULT_STREAM_WINDOW: u32 = 1024 * 1024 * 2; // 2mb
 const DEFAULT_MAX_FRAME_SIZE: u32 = 1024 * 16; // 16kb
 
+// Same as the default value in `h2`:
+// https://docs.rs/h2/0.3.3/h2/client/struct.Builder.html#method.max_concurrent_reset_streams
+const DEFAULT_MAX_CONCURRENT_RESET_STREAMS: usize = 10;
```
pickfire, line 37: Hard-coding it here might make it different in the future, ideally we just want to use a re-export of the original value?

boxdot: Fair point. I created a PR there: https://github.com/hyperium/h2/pull/543

src/proto/h2/client.rs
seanmonstar: I actually _would_ suggest that this be an `Option<usize>`, and if `None`, then the option just isn't set on the `h2` builder. That'd mean less of `h2` needs to be modified/exposed. We can just document "_a_ reasonable default is used", without saying what it is.

boxdot: @pickfire Considering the last comment of @seanmonstar, I would restore my previous commit if it is fine with you.

pickfire: @boxdot sure, @seanmonstar is the maintainer and I didn't do much here, reverting it is the best choice.

seanmonstar (approved):

src/proto/h2/client.rs
boxdot: Decided in another comment to go with an option.

seanmonstar:
I've got https://github.com/hyperium/hyper/pull/2548 going to fix the weird failure on nightly.

Issue #2532, opened 2021-05-01
Replace usage of `mem::uninitialized` with `MaybeUninit`


memoryruins:
When the following comment was added in #1845, hyper's minimum supported rust version was `1.27.0`

https://github.com/hyperium/hyper/blob/4e9a006498c7bdb5bb2ccb76a4c877f6da7e23b2/src/proto/h1/role.rs#L1-L3

which I assume is why [`std::mem::MaybeUninit`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html), stabilized in `1.36.0`, could not be used yet. Today, hyper's MSRV is `1.46.0`, so it should now be possible to investigate removing the `#![allow(deprecated)]`.

seanmonstar:
Good call!

mohsen-alizadeh:
Refactored the code.

Soveu:
One problem with refactoring the code is that [httparse::Request::new](https://docs.rs/httparse/1.4.0/httparse/struct.Request.html#method.new) accepts `&mut [Header]`, so it is assumed that they are initialized. It would need to accept `&mut [MaybeUninit<Header>]`

Otherwise `MaybeUninit::uninit().assume_init()` (like it is done above) gives the same problems as `mem::uninitialized`

RalfJung:
FYI, this seems to be one of the major sources of regressions in https://github.com/rust-lang/rust/pull/87041.

PR #2536, opened 2021-05-04
refactor(http1): Replace usage of mem::uninitialized with MaybeUninit

mohsen-alizadeh:
https://github.com/hyperium/hyper/issues/2532


Soveu:
Problem is, `MaybeUninit::uninit().assume_init()` is UB in the same way `mem::uninitialized` is

PR #2541, opened 2021-05-06
feat(server): add server::Builder::serve_service

### LINKED ISSUE #2154: Provide  convenience around serving `ServiceFn<fn(http::request::Request<hyper::body::body::Body>) -> impl std::future::Future {handler}, hyper::body::body::Body>`
When making a service to pass to `server` I've never used `make_service_fn` in any other way than: 
```rust
let server = Server::bind(&addr).serve(make_service_fn(|_| async move {
        Ok::<_, Infallible>(service_fn(handler))
}));
```

It would be nice to be able to write:
```rust
let server = Server::bind(&addr).serve(service_fn(handler));
```
or at least:
```rust
let server = Server::bind(&addr).serve(serve_service_fn(service_fn(handler)));
```

Thoughts?

davidpdrsn:
This adds `server::Builder::serve_service` which is a convenience for doing

```rust
Server::bind(&addr).serve(make_service_fn(|_| async move {
    Ok::<_, Infallible>(service_fn(handler))
}));
```

That now becomes

```rust
Server::bind(&addr).serve_service(service_fn(handler));
```

It basically copies [`tower::make::Shared`][] into Hyper so users don't need
to add another dependency to do this.

Also adjusted the docs for `hyper::server` a bit to make `server_service`
the top example as I believe its the more common use-case.

Fixes #2154

[`tower::make::Shared`]: https://docs.rs/tower/0.4.7/tower/make/struct.Shared.html

### DIFF
--- a/src/server/server.rs
+++ b/src/server/server.rs
@@ -13,7 +13,7 @@ use super::accept::Accept;
 use crate::body::{Body, HttpBody};
 use crate::common::exec::{ConnStreamExec, Exec, NewSvcExec};
 use crate::common::{task, Future, Pin, Poll, Unpin};
-use crate::service::{HttpService, MakeServiceRef};
+use crate::service::{HttpService, MakeServiceRef, Shared, SharedFuture};
 // Renamed `Http` as `Http_` for now so that people upgrading don't see an
 // error that `hyper::server::Http` is private...
 use super::conn::{Http as Http_, NoopWatcher, SpawnAll};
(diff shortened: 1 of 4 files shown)


src/server/server.rs
```diff
@@ -418,22 +418,65 @@ impl<I, E> Builder<I, E> {
...
+    /// // Then bind and serve...
+    /// let server = Server::bind(&addr).serve_service(service);
+    ///
+    /// // Run forever-ish...
+    /// if let Err(err) = server.await {
+    ///     eprintln!("server error: {}", err);
+    /// }
+    /// # }
+    /// ```
+    pub fn serve_service<S, B>(self, service: S) -> Server<I, Shared<S>, E>
```
davidpdrsn, line 465: I don't think the name `serve_service` is particularly good but couldn't come up with anything else.

davidbarsky: `serve`?

seanmonstar: What if instead of a separate method, it's part of `hyper::service`?

```rust
Server::bind(addr)
    .serve(hyper::service::shared(svc))
```

davidpdrsn: > `serve`?

`serve` is already taken https://docs.rs/hyper/0.14.7/hyper/server/struct.Builder.html#method.serve 😞 This method is actually not on `Server` but on `server::Builder`. Fixing description now.

> What if instead of a separate method, it's part of `hyper::service`?

I was thinking that since its such a common use case it made sense to have a dedicated method for it.

davidpdrsn:
We discussed this in Discord and decided to wait as a bigger change to Server is planned as part of either 0.15 or a potential 1.0.

Issue #2534, opened 2021-05-03
illegal header name from httparse


let4be:
Why do we have this defined as a panic?
https://github.com/hyperium/hyper/blob/dbea7716f157896bf7d2d417be7b4e382e7dc34f/src/proto/h1/role.rs#L36
https://github.com/hyperium/hyper/blob/dbea7716f157896bf7d2d417be7b4e382e7dc34f/src/proto/h1/role.rs#L47
https://github.com/hyperium/hyper/blob/dbea7716f157896bf7d2d417be7b4e382e7dc34f/src/proto/h1/role.rs#L60

Even if under `debug_assertions` this library has no right to panic as it's completely normal for it to be used on invalid input(I use it in a broad web crawler and web is full of surprises - good and bad). It should always simply return an error back to the user if anything bad happened

as I understand it `panics` are reserved for unrecoverable errors, and because this is a library it has no right to panic and crash client code... Error "unrecoverability" is completely at library's user discretion

let4be:
I'm also not quite sure how to disable this, I'm still getting 
`thread 'tokio-runtime-worker' panicked at 'illegal header name from httparse: [51, 48, 50, 10, 83, 116, 97, 116, 117, 115]', /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.7/src/proto/h1/role.rs:951:28`

after putting 
```
[profile.dev]
panic = "abort"
debug-assertions = false

[profile.release]
panic = "abort"
debug-assertions = false
opt-level = 3
```
into my `Cargo.toml`

any advice?

let4be:
Hm... having a problem with fork + clone approach
My steps
1. `git clone https://github.com/hyperium/hyper` (or fork)
2. `cd hyper && git checkout tags/v0.14.7`
3. do edits
4. add dependency to a local version of hyper via `path`

Now I have weird compillation errors because it seems like some other libraries I use depend on hyper too and there is some sort of conflict, but I do not understand why
before doing a specific version checkout via git I looked at `Cargo.lock` and I checkout the exact version my project was using...

Any advice on how to properly fork and fix locally?

let4be:
Okay, adding 
```
[patch.crates-io]
hyper = { path = "../hyper" }
```
to the my program's `Cargo.toml` and bumping version in the cloned hyper seems to fixed an issue and I now able to compile using local Hyper, yay

[...]

PR #2544, opened 2021-05-10
Return Parse::Internal error if there's an illegal header name or value

bensadiku:
Initial implementation that aims to solve #2534 by returning a `Parse::Internal` error instead of panic.  

I intentionally added a new Parse error instead of a new Kind since i think this is still relevant and the diff is much cleaner.
cc @nox  for [this message](https://github.com/hyperium/hyper/issues/2534#issuecomment-835436759), if this is very important to add here, let me know.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -73,6 +73,7 @@ pub(super) enum Parse {
     Header,
     TooLarge,
     Status,
+    #[cfg_attr(debug_assertions, allow(unused))]
     Internal,
 }
 
@@ -487,18 +488,6 @@ impl From<httparse::Error> for Parse {
 impl From<http::method::InvalidMethod> for Parse {
     fn from(_: http::method::InvalidMethod) -> Parse {
         Parse::Method
-    }
-}
-
-impl From<http::header::InvalidHeaderName> for Parse {
-    fn from(_: http::header::InvalidHeaderName) -> Parse {
-        Parse::Internal
-    }
-}
-
-impl From<http::header::InvalidHeaderValue> for Parse {
-    fn from(_: http::header::InvalidHeaderValue) -> Parse {
-        Parse::Internal
     }
 }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -73,6 +73,7 @@ pub(super) enum Parse {
     Header,
     TooLarge,
     Status,
+    Internal,
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -477,6 +479,18 @@ impl From<http::method::InvalidMethod> for Parse {
     }
 }
 
+impl From<http::header::InvalidHeaderName> for Parse {
```
seanmonstar, line 482: I'd say these `From` implementations are perhaps a little too implicit. I think we should likely be sure to be explicit when labeling that an error in an internal bug.

src/proto/h1/role.rs
```diff
@@ -205,8 +198,8 @@ impl Http1Transaction for Server {
         headers.reserve(headers_len);
 
         for header in &headers_indices[..headers_len] {
-            let name = header_name!(&slice[header.name.0..header.name.1]);
-            let value = header_value!(slice.slice(header.value.0..header.value.1));
+            let name = header_name!(&slice[header.name.0..header.name.1])?;
```
seanmonstar, line 201: I think the macros could be written to just `return Err(whatever)` inside them, and then these lines don't really need to change, ya?

seanmonstar (changes requested):
Thanks for taking this on! <3

src/proto/h1/role.rs
bensadiku: Not sure i follow, `header_name` and `header_value` macros returns a `Result`, from the new `maybe_panic` macro.

src/error.rs
bensadiku: I agree they are implicit, i initially did not have it but then the lint complains the _variant is never constructed_ (even though it is) for the Parse::Internal error

seanmonstar: That's ok, it's probably because the variants are only "used" in release mode. It's possible to add an `#[cfg_attr(debug_assertions, allow(unused))]` on the variant. Then we don't need these impls.

src/proto/h1/role.rs
seanmonstar: In the macro, you make it so it doesn't wrap the value back in `Ok`, and on error `return Err(err)`. Then at the use site, we don't need to add a question mark.

src/error.rs
bensadiku: Sweet, i wasn't sure if supressing the lint was the right approach. Thanks for the help!

bensadiku:
@seanmonstar can you approve the ci?

src/proto/h1/role.rs
```diff
@@ -29,44 +29,37 @@ const AVERAGE_HEADER_SIZE: usize = 30; // totally scientific
...
+                Err(e) => maybe_panic!(e),
             }
         }
+    }};
+}
 
-        #[cfg(not(debug_assertions))]
-        {
-            // Unsafe: httparse already validated header value
-            unsafe { HeaderValue::from_maybe_shared_unchecked($bytes) }
```
seanmonstar, line 67: I just realized this change, which will makes things a little slower even when not using debug_assertions. Can we keep this code?

bensadiku: Absolutely, seems like `from_maybe_shared_unchecked` internally only panics in debug builds so that should be good enough?

seanmonstar:
Thanks for carrying this through!

PR #2545, opened 2021-05-11
Use new methods for parsing with uninitialized headers from httparse

### LINKED ISSUE #2532: Replace usage of `mem::uninitialized` with `MaybeUninit`

Soveu:
This patch fixes #2532
It uses methods implemented ~~[in this PR](https://github.com/seanmonstar/httparse/pull/97)~~ [in this PR](https://github.com/seanmonstar/httparse/pull/98), so CI might fail before it gets merged

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -1,16 +1,14 @@
-// `mem::uninitialized` replaced with `mem::MaybeUninit`,
-// can't upgrade yet
 #![allow(deprecated)]
 
 use std::fmt::{self, Write};
-use std::mem;
+use std::mem::{self, MaybeUninit};
 
 #[cfg(any(test, feature = "server", feature = "ffi"))]
 use bytes::Bytes;
 use bytes::BytesMut;
-use http::header::{self, Entry, HeaderName, HeaderValue};
 #[cfg(feature = "server")]
 use http::header::ValueIter;
+use http::header::{self, Entry, HeaderName, HeaderValue};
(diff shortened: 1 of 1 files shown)


src/proto/h1/role.rs
```diff
@@ -1,9 +1,7 @@
-// `mem::uninitialized` replaced with `mem::MaybeUninit`,
-// can't upgrade yet
 #![allow(deprecated)]
```
seanmonstar, line 1:
```suggestion
```

seanmonstar:
Alright, I've released `httparse` v1.5.0, so we can use it in this branch, and then merge 'er in!

Soveu:
`Some commit messages are invalid` :pensive:

Soveu:
How am I supposed to change commit messages now?

seanmonstar:
Don't worry about the commit messages, I can adjust it when squashing/merging.

seanmonstar (approved):

Issue #2526, opened 2021-04-29
HTTP/2 Adaptive Window sometimes triggers ENHANCE_YOUR_CALM


blyxxyz:
Downloading https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin tends to fail midway through if rustls and `http2_adaptive_window` are used.

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let url = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin".parse()?;
    let https = hyper_rustls::HttpsConnector::with_native_roots();
    let client: hyper::Client<_, hyper::Body> = hyper::Client::builder()
        .http2_adaptive_window(true)
        .build(https);
    let mut response = client.get(url).await?;
    hyper::body::to_bytes(response.body_mut()).await?;
    Ok(())
}
```

```toml
[dependencies]
hyper = { version = "0.14.7", features = ["http2"] }
hyper-rustls = "0.22.1"
tokio = { version = "1.5.0", features = ["macros", "rt-multi-thread"] }
```
Outputs:
```
Error: hyper::Error(Body, Error { kind: Io(Kind(BrokenPipe)) })
```

It fails consistently on my Debian machine. It fails inconsistently on @austinbutler's macOS and Arch Linux machines, see ducaale/xh#135.

I can't reproduce it when using reqwest's `native-tls` feature, so it's possible the real problem is somewhere in hyper-rustls or rustls itself.

I don't know if `http2_adaptive_window` is really vital.

seanmonstar:
Hm, interesting! I don't think the `native-tls` backend uses ALPN, so it may be using the HTTP/1 protocol then.

If you turn on the tracing logs, does it reveal anything interesting shortly before the error occurs? Particularly the trace/debug logs for `h2`...

sfackler:
native-tls does support ALPN now, though I'm not sure if the hyper wrapper uses it.

seanmonstar:
Good clarification, but I don't think reqwest currently is setup to use it.

blyxxyz:
Aha:
```
 TRACE h2::codec::framed_read           > read.bytes=17 
 TRACE h2::codec::framed_read           > FramedRead::decode_frame; offset=17
 TRACE h2::codec::framed_read           > decoding frame from 17B 
 TRACE h2::codec::framed_read           > frame.kind=GoAway 
[...]

PR #2550, opened 2021-05-15
perf(http2): reduce adaptive window pings as the BDP stabilizes

seanmonstar:
This introduces a delay to sending a ping to calculate the BDP that
becomes shorter as the BDP is changing, to improve throughput quickly,
but then also becomes longer as the BDP stabilizes, to reduce the amount
of pings sent. This improved the performance of the adaptive window
end_to_end benchmark.

It should also reduce the amount of pings the remote has to deal with,
hopefully preventing hyper from triggering ENHANCE_YOUR_CALM errors.

cc #2526

### DIFF
--- a/src/proto/h2/ping.rs
+++ b/src/proto/h2/ping.rs
@@ -51,9 +51,15 @@ pub(super) fn channel(ping_pong: PingPong, config: Config) -> (Recorder, Ponger)
         bdp: wnd,
         max_bandwidth: 0.0,
         rtt: 0.0,
+        ping_delay: Duration::from_millis(100),
+        stable_count: 0,
     });
 
-    let bytes = bdp.as_ref().map(|_| 0);
+    let (bytes, next_bdp_at) = if bdp.is_some() {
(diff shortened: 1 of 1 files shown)


src/proto/h2/ping.rs
```diff
@@ -282,12 +310,11 @@ impl Ponger {
...
 
-        let (bytes, rtt) = match locked.ping_pong.poll_pong(cx) {
+        match locked.ping_pong.poll_pong(cx) {
             Poll::Ready(Ok(_pong)) => {
-                let rtt = locked
+                let rtt = now - locked
                     .ping_sent_at
-                    .expect("pong received implies ping_sent_at")
-                    .elapsed();
+                    .expect("pong received implies ping_sent_at");
```
nox, line 317: Nit: binding `locked.ping_sent_at.expect(…)` to its own variable before computing `rtt` would probably help readability here, the line feeds mixed with the subtraction feels hard to follow to me.

nox (commented):
Looks nice, just one nit.

PR #2562, opened 2021-05-26
fix(http1): reduce memory used with flatten write strategy

seanmonstar:
If the write buffer was filled with large bufs from the user, such that
it couldn't be fully written to the transport, the write buffer could
start to grow significantly as it moved its cursor without shifting over
the unwritten bytes.

This will now try to shift over the unwritten bytes if the next buf
wouldn't fit in the already allocated space.

### DIFF
--- a/src/proto/h1/io.rs
+++ b/src/proto/h1/io.rs
@@ -56,7 +56,12 @@ where
     B: Buf,
 {
     pub(crate) fn new(io: T) -> Buffered<T, B> {
-        let write_buf = WriteBuf::new(&io);
+        let strategy = if io.is_write_vectored() {
+            WriteStrategy::Queue
+        } else {
+            WriteStrategy::Flatten
+        };
+        let write_buf = WriteBuf::new(strategy);
         Buffered {
             flush_pipeline: false,
             io,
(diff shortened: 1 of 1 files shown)


src/proto/h1/io.rs
```diff
@@ -419,6 +424,24 @@ impl<T: AsRef<[u8]>> Cursor<T> {
 }
 
 impl Cursor<Vec<u8>> {
+    /// If we've advanced the position a bit in this cursor, and wish to
+    /// extend the underlying vector, we may wish to unshift the "read" bytes
+    /// off, and move everything else over.
+    fn maybe_unshift(&mut self, additional: usize) {
```
nox, line 430: I'm not sure this method name really fits the method, but I have no alternative to offer.

It took me a hot second to understand that `additional` is the threshold we check to decide whether or not to drain the leading bytes of the buffer.

nox (approved):
LGTM

src/proto/h1/io.rs
seanmonstar: I agree, coming up with a name was hard.

PR #2566, opened 2021-06-02
refactor(lib): Switch from pin-project to pin-project-lite (again)

jplatte:
Second try, turned out to require a bunch more changes in `src/client/conn.rs` and `src/server/conn.rs` due to the `ProtoClient` and `ProtoServer` enums which previously had `cfg`-conditional variants (which aren't supported by pin-project-lite). Now the variants that previously were disabled by `cfg(not(feature = "http1"))` and `cfg(not(feature = "http2"))` are instead made to be uninhabited.

The code would be a decent amount simpler if the non-exhaustive variants could be omitted in `match`es (https://github.com/rust-lang/rust/issues/51085).

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -48,7 +48,7 @@
 
 use std::error::Error as StdError;
 use std::fmt;
-#[cfg(feature = "http2")]
+#[cfg(not(all(feature = "http1", feature = "http2")))]
 use std::marker::PhantomData;
 use std::sync::Arc;
 #[cfg(all(feature = "runtime", feature = "http2"))]
@@ -57,12 +57,14 @@ use std::time::Duration;
 use bytes::Bytes;
 use futures_util::future::{self, Either, FutureExt as _};
 use httparse::ParserConfig;
-use pin_project::pin_project;
+use pin_project_lite::pin_project;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tower_service::Service;
 
 use super::dispatch;
 use crate::body::HttpBody;
+#[cfg(not(all(feature = "http1", feature = "http2")))]
+use crate::common::Never;
 use crate::common::{
     exec::{BoxSendFuture, Exec},
     task, Future, Pin, Poll,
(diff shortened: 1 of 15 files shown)


src/client/conn.rs
```diff
@@ -74,17 +74,33 @@ use crate::upgrade::Upgraded;
...
-enum ProtoClient<T, B>
-where
-    B: HttpBody,
-{
-    #[cfg(feature = "http1")]
-    H1(#[pin] Http1Dispatcher<T, B, proto::h1::ClientTransaction>),
-    #[cfg(feature = "http2")]
-    H2(#[pin] proto::h2::ClientTask<B>, PhantomData<fn(T)>),
+#[cfg(not(feature = "http1"))]
+type Http1Dispatcher<T, B> = (Never, T, Pin<Box<B>>);
```
seanmonstar, line 81: It seems worth it to put the extra generics inside a `PhantomData`. Even though `Never` is uninhabitable, the other types affect the `mem::size_of` anyways.

src/server/conn.rs
```diff
@@ -111,77 +113,90 @@ enum ConnectionMode {
...
-        PhantomData<E>,
-    ),
-    #[cfg(feature = "http2")]
-    H2(#[pin] proto::h2::Server<Rewind<T>, S, B, E>),
+#[cfg(feature = "http1")]
+type Http1Dispatcher<T, B, S> =
+    proto::h1::Dispatcher<proto::h1::dispatch::Server<S, Body>, B, T, proto::ServerTransaction>;
+
+#[cfg(not(feature = "http1"))]
+type Http1Dispatcher<T, B, S> = (Never, T, Box<Pin<B>>, Box<Pin<S>>);
```
seanmonstar, line 176: Same comment here and a couple lines down about `PhantomData`.

seanmonstar (commented):
Thanks so much, excellent work!

seanmonstar (approved):
<3

seanmonstar:
Note the practical affects of this PR:

- Dependency count with `--features full` dropped from 65 to 55.
- Time to compile after a clean dropped from 48s to 35s (on a pretty underpowered VM).

🎉

jplatte:
I'm actually a bit surprised about the dependency diff. pin-project should "only" pull in pin-project-internal > syn > quote > proc-macro2 > unicode-xid AFAICT.

seanmonstar:
Hm, I think you're right. I just based that number off making a new crate, adding hyper as a dependency (to not get any dev-dependencies), and then recording how it says `Building [    ] 1/55` and then `Building [    ] 1/65`. Probably 1 of those is the empty new lib, but otherwise I'm not sure...

seanmonstar:
Ah, paying more attention, it's because the `build.rs` step of a dependency is counted in there too. And maybe another step related to proc macro libs?

jplatte:
Will there be a patch release with this soon?

seanmonstar:
Yep, just released v0.14.9.

PR #2567, opened 2021-06-02
server: implement set_sleep_on_errors() for generic Accept implementations

vavrusa:
This adds AcceptWithSleep struct which implements Accept trait and adds a functionality to sleep on errors similar to the AddrIncoming for TcpListener. The reason for this is that this feature is missing for UnixSocket listeners,
and other custom listeners implementing Accept, but it's not really specific to TcpListener. I'm not sure if/how you would prefer to have this implemented, so opening this PR to get the ball rolling.

### DIFF
--- a/src/server/accept.rs
+++ b/src/server/accept.rs
@@ -6,10 +6,16 @@
 //!   connections.
 //! - Utilities like `poll_fn` to ease creating a custom `Accept`.
 
+use std::fmt;
+use std::time::Duration;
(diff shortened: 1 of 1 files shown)


seanmonstar:
We've known `Accept` had some design problems for a while, and with #2849, we're removing it from hyper entirely. A utility that makes this pattern easy to apply for any listener could probably be a good addition to `hyper-util`.

PR #2568, opened 2021-06-03
Document the typical value for an `io` param

jsha:
Per https://github.com/rust-lang/rust/issues/84824, it can be hard to figure out
from the trait bounds what types meet those bounds.

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -87,7 +87,8 @@ where
     H2(#[pin] proto::h2::ClientTask<B>, PhantomData<fn(T)>),
 }
 
-/// Returns a handshake future over some IO.
+/// Returns a handshake future over some IO. The io parameter will usually be a
+/// [tokio::net::TcpStream](https://docs.rs/tokio/1.6.1/tokio/net/struct.TcpStream.html).
 ///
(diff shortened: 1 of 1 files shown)


src/client/conn.rs
```diff
@@ -87,7 +87,8 @@ where
     H2(#[pin] proto::h2::ClientTask<B>, PhantomData<fn(T)>),
 }
 
-/// Returns a handshake future over some IO.
+/// Returns a handshake future over some IO. The io parameter will usually be a
+/// [tokio::net::TcpStream](https://docs.rs/tokio/1.6.1/tokio/net/struct.TcpStream.html).
```
seanmonstar, line 91: I wonder, so as to not encode the version directly into the source, can we just point at `(tokio::net::TcpStream)`? Also, I'd put the second sentence in a new line / new paragraph.

seanmonstar (commented):
Thanks for the suggestion!

seanmonstar:
As part of hyper 1.0, we're trying to de-couple the Tokio preference from hyper directly, but I do think we should have more guides telling people the most likely way they will do a thing.

PR #2571, opened 2021-06-07
refactor(error): remove PartialEq derives for error kind enums

seanmonstar:
Replaced the comparisons with `matches!`. This should reduce a bit of
code generation that isn't really needed.


davidpdrsn (approved):
Nice!

PR #2573, opened 2021-06-08
refactor(error): add header parse error details in `hyper::Error`

### LINKED ISSUE #2569: Which header is invalid?
Closely related to #2083: when we use curl+hyper to get a HTTP response the only error message hyper gives is `invalid HTTP header parsed` - which is really hostile in the cases we got maybe 20 different headers back and maybe we're not up to date on the details of the RFC.

We need a way for hyper to be more specific. Which exact header did it find a problem with? Example response to prove my point. This has an uppercase o instead of a zero in the Content-Length header.
~~~
HTTP/1.1 200 OK swsclose
Date: Tue, 09 Nov 2010 14:49:00 GMT
Server: test-server/fake
Last-Modified: Tue, 13 Jun 2000 12:10:00 GMT
ETag: "21025-dc7-39462498"
Accept-Ranges: bytes
Content-Length: 5O
Connection: close
Content-Type: text/html
Funny-head: yesyes
~~~

seanmonstar:
When a header parse error is because of content-length or
transfer-encoding semantics, include a better error message in the
`hyper::Error`.

Closes #2569

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -72,11 +72,22 @@ pub(super) enum Parse {
     #[cfg(feature = "http1")]
     VersionH2,
     Uri,
-    Header,
+    Header(Header),
     TooLarge,
     Status,
     #[cfg_attr(debug_assertions, allow(unused))]
     Internal,
+}
+
+#[derive(Debug)]
+pub(super) enum Header {
+    Token,
+    #[cfg(feature = "http1")]
+    ContentLengthInvalid,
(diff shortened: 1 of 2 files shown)


davidpdrsn:
Sorry for the delay here! I'll make sure to review this tomorrow.

davidpdrsn (approved):
Nice!

PR #2585, opened 2021-06-24
fix(client): retry when check out returns closed connection with h2 only

nox:


nox:
@seanmonstar Any idea how to test this?

seanmonstar:
Um, testing is probably annoyingly tricky... What's the exact steps that could happen to cause this problem in the first place? Maybe we can mock one of them...

seanmonstar (approved):

PR #2597, opened 2021-07-16
fix(h2): preserve proxy authentication headers

nox:
That Proxy-Authenticate and Proxy-Authorization are forbidden over h2
is not actually specified anywhere, plus h2 also supports CONNECT
requests, which are specifically made to do requests over a proxy,
and those proxies may require authentication, sometimes through
Proxy-Authorization.

Note that there is an openwebdocs project that just started to clear
up any MDN-induced confusion in implementations:

> https://github.com/openwebdocs/project/issues/43

Also, Daniel Stenberg (curl) also recently commented about
Proxy-Connection not being stripped out by curl, Hyper may want to
also remove Proxy-Connection from the list for that reason:

> https://twitter.com/bagder/status/1415967315817082880

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -1,9 +1,6 @@
 use bytes::{Buf, Bytes};
 use h2::{RecvStream, SendStream};
-use http::header::{
-    HeaderName, CONNECTION, PROXY_AUTHENTICATE, PROXY_AUTHORIZATION, TE, TRAILER,
-    TRANSFER_ENCODING, UPGRADE,
-};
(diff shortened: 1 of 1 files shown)


nox:
Cc @bagder

bagder:
`Proxy-Connection` is certainly required for h1 connections over proxy, I doubt h2 proxies require it but that's just me guessing.

seanmonstar (approved):
Yea, it does seem hyper was a little aggressive here, motivated by an MDN article. Thanks for the fix!

The HTTP/2 spec does directly spell out SHOULD remove `Proxy-Connection`. I imagine proxies that have been updated to use h2 will have to have dealt with that. If we get reports of it being a problem, we can consider then.

PR #2598, opened 2021-07-19
fix(h2): improve I/O errors emitted by H2Upgraded

nox:


src/proto/h2/mod.rs
```diff
@@ -335,21 +339,34 @@ where
...
-            Some(Ok(cnt)) => self.send_stream.write(&buf[..cnt], false).map(|()| cnt),
-            Some(Err(e)) => Err(h2_to_io_error(e)),
-        })
+
+        let cnt = match ready!(self.send_stream.poll_capacity(cx)) {
+            None => Some(0),
+            Some(Ok(cnt)) => self
+                .send_stream
+                .write(&buf[..cnt], false)
+                .ok()
```
seanmonstar, line 352: This will ignore the error, but that's on purpose, right?

seanmonstar (approved):
I think this looks right.

(Have you been able to try out this in your app to see it improves it for you?)

src/proto/h2/mod.rs
nox: Yeah, that's the intent, as we will get the right error from `poll_reset` anyway.

seanmonstar: Ok got it. Perhaps a comment to that effect will help us when looking in this file a few months or years later 😅

nox: Done.

PR #2604, opened 2021-07-24
docs(ffi): copy-paste error

divergentdave:
The docs for `hyper_waker_wake` were accidentally copied over from `hyper_waker_free`, so I rewrote it.

### DIFF
--- a/src/ffi/task.rs
+++ b/src/ffi/task.rs
@@ -418,7 +418,9 @@ ffi_fn! {
 }
 
 ffi_fn! {
-    /// Free a waker that hasn't been woken.
+    /// Wake up the task associated with a waker.
+    ///
+    /// NOTE: This consumes the waker. You should not use or free the waker afterwards.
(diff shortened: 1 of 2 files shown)


src/ffi/task.rs
```diff
@@ -418,7 +418,7 @@ ffi_fn! {
 }
 
 ffi_fn! {
-    /// Free a waker that hasn't been woken.
+    /// Wake up the task associated with a waker.
```
seanmonstar, line 421:
```suggestion
    /// Wake up the task associated with a waker.
    ///
    /// NOTE: This consumes the waker. You should not use or free the waker afterwards.
```

seanmonstar (commented):
Oh, thanks for catching this! I added a note suggestion too.

seanmonstar:
Ah, this will also need to update the `hyper.h` file to make CI happy... I can check it out locally and run the script.

divergentdave:
I got it, I added the note and re-generated the C header.

PR #2609, opened 2021-08-01
Add an accessor for the `Connect` implementation inside a Client to allow bypassing the connection pool.

khuey:
I use hyper to (among other things) talk to Docker over TCP. Docker's API is primarily HTTP, however, the attach command hijacks the HTTP connection (see https://docs.docker.com/engine/api/v1.41/#operation/ContainerAttach).

I added this accessor to get the connection outside of the pool, and then use into_parts to break it into the raw i/o object to do the necessary communication.

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -392,6 +392,11 @@ where
         }
     }
 
+    /// Access the `Connect` implementation directly, bypassing the connection pool.
+    pub fn connector(&self) -> &C {
+        &self.connector
+    }
+
     fn connect_to(
         &self,
         pool_key: PoolKey,


src/client/client.rs
```diff
@@ -392,6 +392,11 @@ where
         }
     }
 
+    /// Access the `Connect` implementation directly, bypassing the connection pool.
```
seanmonstar, line 395:
```suggestion
    /// Get a reference to the `Connect` value of this `Client`.
```

I don't think it needs to mention bypassing the pool. When I read that, it made me think "wait, do I have to be careful when bypassing something?"

seanmonstar (commented):
I imagine the `Connect` type is usually cheap to clone, so you could keep a copy around when constructing a `Client`... This seems like a fine addition. But, can you do much with just a shared reference?

khuey:
Oh, heh, yeah, it already is in Arc. So it's easy to keep another copy outside of the Client.

PR #2611, opened 2021-08-03
Expose http09 and http1 options on low-level client connection builders.

### LINKED ISSUE #2461: http1 options are exposed through client::Builder but not lower-level client::conn::Builder interface
`http1_read_buf_exact_size`, `http1_max_buf_size`, and `http1_title_case_headers` are all configuration methods on `hyper::client::Builder`, but not `hyper::client::conn::Builder`. I notice that the methods are present on the latter type but are `pub(super)` rather than `pub`. 

Currently, I believe users of the `conn` interface are stuck with the default values of these settings. Are there any reasons not to make these `pub` with similar docstrings to the higher-level API? I'd be happy to provide a PR if not.

aturon:
These options are currently available on the high-level builder only.

Along the way, rename the setters to follow the public API conventions and add docs.

Closes #2461

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -971,7 +971,7 @@ impl Builder {
     ///
     /// Default is an adaptive read buffer.
     pub fn http1_read_buf_exact_size(&mut self, sz: usize) -> &mut Self {
-        self.conn_builder.h1_read_buf_exact_size(Some(sz));
+        self.conn_builder.http1_read_buf_exact_size(Some(sz));
         self
     }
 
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Cool, thanks! Looks great to me.

aturon:
It looks like some tests are failing for (I think?) unrelated reasons.

The trailing semicolon issue on nightly seems to stem from https://github.com/rust-lang/rust/pull/87385

I don't know enough about the `hyper.h` check to comment on that.

aturon:
Passing now, ty!

PR #2613, opened 2021-08-04
fix(headers): apply header title case for consecutive dashes

blyxxyz:
Fix the header title-casing to work with consecutive dashes. Previously with two dashes in a row the first dash would uppercase the second dash which would then not count, so `weird--header` would be cased as `Weird--header` instead of `Weird--Header`.


seanmonstar (approved):
Looks good, and I like that the code as a whole is simpler, thanks!

PR #2616, opened 2021-08-10
fix: missing import in examples/http_proxy.rs

wperron:
The HTTP Proxy example references `hyper::http` directly, but it is not
used directly. This change adds the missing `use` statement for it.


seanmonstar:
Thanks for the PR! I believe the example works before because it's using finding it as the `http` crate, as opposed to a re-export from hyper.

wperron:
that's possible! I just copy/pasted in a separate project and had this compile time error :shrug:

seanmonstar:
I'm going to close, since I find it better to suggest using `http` crate directly, instead of import from `hyper::http`. But it is an option.

PR #2618, opened 2021-08-13
fix (header): rustsec-2021-0078 deny plus sign in content-length

KatsuoRyuu:
change of the content-length header parsing to use a parser that expects
the content-length content to be of type digits (u64)

Since Rust Rocket v0.4.x is still using the old hyper library, this is
important applications to parse the security audit.

BREAKING CHANGE: None, unless some systems rely on a security flaw in
hyper.


seanmonstar:
Thanks for the PR! However, the 0.10.x branch is no longer maintained, and doesn't do new releases. v0.11 was released in June 2017, and the final bug fix release for 0.10.x was 2019. I would recommend either using the newer async Rocket, or switching to something that is using a newer hyper.

PR #2619, opened 2021-08-13
0.13.x security fixes

polachok:
We have a large codebase utilizing tokio 0.2 and hyper 0.13. It would not be possible for us to promptly upgrade everything to tokio 1 and hyper 0.14. So here's the backport of security fixes for 0.13.x branch.


seanmonstar:
Thanks for the PR! Assuming CI is still happy with this, we can do a 0.13.x release.

However, it should be noted that 0.13.x is no longer maintained, and people should take the time to upgrade.

PR #2622, opened 2021-08-18
fix(h2): improve errors emitted by H2 stream shutdown

nox:


src/proto/h2/mod.rs
```diff
@@ -377,9 +377,21 @@ where
...
     ) -> Poll<Result<(), io::Error>> {
-        Poll::Ready(self.send_stream.write(&[], true))
+        if self.send_stream.write(&[], true).is_ok() {
+            return Poll::Ready(Ok(()))
+        }
+
+        Poll::Ready(Err(h2_to_io_error(
+            match ready!(self.send_stream.poll_reset(cx)) {
+                Ok(Reason::NO_ERROR) | Ok(Reason::CANCEL) | Ok(Reason::STREAM_CLOSED) => {
+                    return Poll::Ready(Ok(()))
```
nox, line 389: If there was no error or the stream was cancelled or closed, we just return `Ok(())`, does that make sense to you @seanmonstar?

seanmonstar: [Originally](https://httpwg.org/specs/rfc7540.html#ErrorCodes), `STREAM_CLOSED` is meant to signal that the sender did something wrong, sending a frame after it said it wouldn't. So that seems like it should be an error.

I kind of think `CANCEL` should also be returned, at least so the sender knows. The user could chose to ignore those, right?

nox: Done

seanmonstar (approved):

PR #2639, opened 2021-08-29
Export more things with Cargo features [server, !http1, !http2]

jplatte:
I discovered in https://github.com/tokio-rs/axum/pull/286 that two items from hyper that axum uses (`Server`, `server::conn::AddrStream`) are not available without one of `http1`, `http2` being activated even though it seemed like they could be made available. This PR makes them available (among some other refactorings), albeit with `Server` not actually being usable.

### DIFF
--- a/src/server/server.rs
+++ b/src/server/server.rs
@@ -1,26 +1,33 @@
-use std::error::Error as StdError;
 use std::fmt;
 #[cfg(feature = "tcp")]
 use std::net::{SocketAddr, TcpListener as StdTcpListener};
-
 #[cfg(feature = "tcp")]
 use std::time::Duration;
 
-use pin_project_lite::pin_project;
-use tokio::io::{AsyncRead, AsyncWrite};
-
-use super::accept::Accept;
-use crate::body::{Body, HttpBody};
-use crate::common::exec::{ConnStreamExec, Exec, NewSvcExec};
-use crate::common::{task, Future, Pin, Poll, Unpin};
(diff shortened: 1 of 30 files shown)


jplatte:
Found that this isn't complete yet...

src/server/server.rs
```diff
@@ -34,15 +41,27 @@ pin_project! {
     }
 }
 
+/// A listening HTTP server that accepts connections in both HTTP1 and HTTP2 by default.
+///
+/// Needs at least one of the `http1` and `http2` features to be activated to actually be useful.
+#[cfg(not(any(feature = "http1", feature = "http2")))]
+pub struct Server<I, S, E = Exec> {
```
seanmonstar, line 48: Why should we expose a type that can't do anything if the features aren't enabled?

jplatte: This allows axum to unconditionally re-export the type. It could also conditionally re-export I suppose¹, but it seemed logical that hyper would always define `hyper::Server` when the equally-named feature is enabled 🙂

¹ which would mean having to enable at least one of its "feature re-exports" for http1 / http2 in order to use the re-export, as opposed to enabling one or both directly; certainly not such a bad thing in the grand scheme of things but potentially a small papercut

seanmonstar: Oh I get it now. I should have better read the description and linked issue, my bad!

seanmonstar (approved):
Looks good, thanks!

PR #2655, opened 2021-09-24
Defend against slow request header read attacks in H1 scenarios.

silence-coding:
I tried to implement the H1 request header slow attack, but for the H2 scenario, I still don't know how to implement it. In addition, through slowhttptest, I found that the realization of client_read_header_timout is still beneficial, and the connection can be released quickly. Please give me some advice


src/server/conn.rs
```diff
@@ -103,6 +103,7 @@ pub struct Http<E = Exec> {
     h1_keep_alive: bool,
     h1_title_case_headers: bool,
     h1_preserve_header_case: bool,
+    h1_header_read_timeout: Option<Duration>,
```
paolobarbolini, line 106: [I can't leave the comment on an unchanged section of the file so I'll leave it here]:

`Duration` gets imported at the beginning of the file only if the `runtime` and `http2` features are on, so it doesn't work when just `http1` is on

src/proto/h1/dispatch.rs
```diff
@@ -180,12 +186,29 @@ where
...
+            if let Some(read_timeout_fut) = &mut self.header_read_timeout_fut {
+                if Pin::new(read_timeout_fut).poll(cx).is_ready() {
+                    warn!("read header from client timeout");
+                    return Err(crate::Error::new_header_timeout());
+                }
+            } else {
+                self.header_read_timeout_fut = Some(Box::pin(sleep(self.header_read_timeout.unwrap())))
+            }
+        }
+        Ok(())
```
paolobarbolini, line 200: You can remove the use of unwrap by using `if let`

```suggestion
        if let Some(header_read_timeout) = self.header_read_timeout {
            if let Some(read_timeout_fut) = &mut self.header_read_timeout_fut {
                if Pin::new(read_timeout_fut).poll(cx).is_ready() {
                    warn!("read header from client timeout");
                    return Err(crate::Error::new_header_timeout());
                }
            } else {
                self.header_read_timeout_fut = Some(Box::pin(sleep(header_read_timeout)))
            }
        }

        Ok(())
```

src/proto/h1/dispatch.rs
```diff
@@ -180,12 +186,29 @@ where
...
+
     fn poll_read(&mut self, cx: &mut task::Context<'_>) -> Poll<crate::Result<()>> {
         loop {
             if self.is_closing {
                 return Poll::Ready(Ok(()));
             } else if self.conn.can_read_head() {
-                ready!(self.poll_read_head(cx))?;
+                self.check_header_read_timeout(cx)?;
+                let read_head_rt = ready!(self.poll_read_head(cx));
+                self.header_read_timeout_fut.take();
```
paolobarbolini, line 210: This is a bit clearer

```suggestion
                self.header_read_timeout_fut = None;
```

src/proto/h1/dispatch.rs
```diff
@@ -180,12 +186,29 @@ where
...
+    }
+
     fn poll_read(&mut self, cx: &mut task::Context<'_>) -> Poll<crate::Result<()>> {
         loop {
             if self.is_closing {
                 return Poll::Ready(Ok(()));
             } else if self.conn.can_read_head() {
-                ready!(self.poll_read_head(cx))?;
+                self.check_header_read_timeout(cx)?;
+                let read_head_rt = ready!(self.poll_read_head(cx));
```
paolobarbolini, line 209: Nitpick: since both the `check_header_read_timeout` and `poll_read_head`  wakers could be waken before the runtime gets a chance to poll this function again, it might be clever to first poll `poll_read_head` and then poll `check_header_read_timeout` if `poll_read_head` returned `Poll::Pending`. In fact that's how [it's implemented in `tokio::time::Timeout`](https://docs.rs/tokio/1.12.0/src/tokio/time/timeout.rs.html#148-157)

paolobarbolini (commented):
I'm not an official reviewer but I'm really interested in this feature so here's my review anyway.

silence-coding:
Thank you very much for your advice, I have optimised @paolobarbolini

src/proto/h1/dispatch.rs
```diff
@@ -180,12 +186,31 @@ where
...
         loop {
             if self.is_closing {
                 return Poll::Ready(Ok(()));
             } else if self.conn.can_read_head() {
-                ready!(self.poll_read_head(cx))?;
+                if let Poll::Ready(read_head_rt) = self.poll_read_head(cx) {
+                    read_head_rt?;
+                }
+                self.check_header_read_timeout(cx)?;
+                self.header_read_timeout_fut = None;
```
paolobarbolini, line 213: I believe your last change invalidated the timeout :smiley: 

```suggestion
                if let Poll::Ready(read_head_rt) = self.poll_read_head(cx) {
                    self.header_read_timeout_fut = None;
                    read_head_rt?;
                }
                self.check_header_read_timeout(cx)?;
```

src/proto/h1/dispatch.rs
```diff
@@ -180,12 +186,31 @@ where
         task::yield_now(cx).map(|never| match never {})
     }
 
+    fn check_header_read_timeout(&mut self, cx: &mut task::Context<'_>) -> crate::Result<()> {
+        if let Some(header_read_timeout) = self.header_read_timeout {
+            if let Some(read_timeout_fut) = &mut self.header_read_timeout_fut {
+                if Pin::new(read_timeout_fut).poll(cx).is_ready() {
+                    warn!("read header from client timeout");
+                    return Err(crate::Error::new_header_timeout());
```
paolobarbolini, line 194: This line returns an error when trying to compile with just the `http1` and `client` features. You can check it by running `cargo check --no-default-features --features client,http1`

PR #2670, opened 2021-10-21
refactor(http1): Quite trace in opentelemetry contexts

taqtiqa-mark:
Setting up tracing+opentelemetry+jaeger, with only a root span, produces
no trace data except for output produced by these two `trace_span!` usages.
The test stack was Tokio+Hyper so there either should have been a lot
or no trace output.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -67,8 +67,7 @@ where
         return Ok(None);
     }
 
-    let span = trace_span!("parse_headers");
-    let _s = span.enter();
+    let span = span!("parse_headers").entered();
     T::parse(bytes, ctx)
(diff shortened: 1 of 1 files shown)


taqtiqa-mark:
I should note I've proposed this change only because it appears to be the only place where the `*_span!` flavor of these macros is used.

However, it would IMO be preferable to use the `*_span!` macros.

The reason is that there is then a seamless experience when migrating from tracing via std/files to tracing via opentelemetry+jaeger etc.  Right now you go from getting the full trace data from Hyper in your console to getting nothing in Jaeger.

Unless there is a trick I'm missing in setting up tracing+opentelemetry ?

PR #2671, opened 2021-10-21
 refactor(http1): Quite trace in opentelemetry contexts

taqtiqa-mark:
Setting up tracing+opentelemetry+jaeger, with only a root span, produces
no trace data except for output produced by these two `trace_span!` usages.
The test stack was an application using Tokio+Hyper so there either should
have been a lot or no trace output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+### v0.14.14 (2021-10-22)
+
+
+#### Bug Fixes
+
+* **client:**
+  * make ResponseFuture implement Sync ([bd6c35b9](https://github.com/hyperium/hyper/commit/bd6c35b98f9513f14ed9cecad933bc7fdb8635ea))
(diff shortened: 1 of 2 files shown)


taqtiqa-mark:
@seanmonstar I found this comment from eliza weis intetesting on discord

> if a span is below the max level, instrument just won't generate any of the code for recording the span if a lot of them are debug or trace, and you enable info by default, even without using the max-level features, the overhead should be minimal-to-nonexistent i guess a rule of thumb is just, "if some event is logged, would i want to know that it was in this function when it was logged?"

Perhaps a better avenue to explore is instument?

taqtiqa-mark:
two other examples from eliza on discord regarding spans across await points
```rust
let span = debug_span!("request", ?request).entered();
debug!("sending request");
client.send_request(request)
    .instrument(span.exit())
    .await; 
```

And

```rust
you could also write
let span = debug_span!("request", ?request).;
debug!(parent: &span, "sending request");
client.send_request(request)
    .instrument(span)
    .await; 
```

Apologies for formatting sent from my phone sans glasses :)

taqtiqa-mark:
Also possibly of interest going forward 
https://github.com/tokio-rs/tracing/issues/1668

PR #2675, opened 2021-10-26
feat(h1): add server header read timeout

### LINKED ISSUE #2457: Defends against slow HTTP attacks.
I think hyper needs to add capabilities such as client_header_timeout and client_body_timeout to defend against slow attacks.

The Nginx defense method is as follows:
https://www.nginx.com/blog/mitigating-ddos-attacks-with-nginx-and-nginx-plus/#client_header_timeout

paolobarbolini:
Implements a timeout for reading the request header. The timer starts either when the connection is established (first request) or when the first byte is read from the header (subsequent requests reusing the same connection). The entire header must be read before the specified timeout is elapsed.

This is the first step towards protecting from Slowloris types of attacks

Supersedes #2661 by fixing https://github.com/hyperium/hyper/pull/2661#discussion_r728809455

Partially fixes #2457

Closes #2661

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,7 +86,7 @@ full = [
 ]
 
 # HTTP versions
-http1 = []
+http1 = ["tokio/time"]
 http2 = ["h2"]
 
 # Client/Server
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,7 +86,7 @@ full = [
 ]
 
 # HTTP versions
-http1 = ["tokio/time"]
+http1 = []
 http2 = ["h2"]
 
 # Client/Server
@@ -100,6 +100,7 @@ stream = []
 runtime = [
     "tcp",
     "tokio/rt",
+    "tokio/time",
 ]
 tcp = [
     "socket2",
--- a/src/error.rs
+++ b/src/error.rs
@@ -44,6 +44,9 @@ pub(super) enum Kind {
     #[cfg(any(feature = "http1", feature = "http2"))]
     #[cfg(feature = "server")]
     Accept,
+    /// User took too long to send headers
+    #[cfg(all(feature = "http1", feature = "server", feature = "runtime"))]
+    HeaderTimeout,
     /// Error while reading a body from connection.
     #[cfg(any(feature = "http1", feature = "http2", feature = "stream"))]
     Body,
@@ -104,9 +107,6 @@ pub(super) enum User {
     #[cfg(any(feature = "http1", feature = "http2"))]
     #[cfg(feature = "server")]
     UnexpectedHeader,
-    /// User took too long to send headers
-    #[cfg(all(feature = "http1", feature = "runtime"))]
-    HeaderTimeout,
     /// User tried to create a Request with bad version.
     #[cfg(any(feature = "http1", feature = "http2"))]
     #[cfg(feature = "client")]
@@ -313,9 +313,9 @@ impl Error {
         Error::new_user(User::UnexpectedHeader)
     }
 
-    #[cfg(all(feature = "http1", feature = "runtime"))]
+    #[cfg(all(feature = "http1", feature = "server", feature = "runtime"))]
     pub(super) fn new_header_timeout() -> Error {
-        Error::new_user(User::HeaderTimeout)
+        Error::new(Kind::HeaderTimeout)
     }
 
     #[cfg(any(feature = "http1", feature = "http2"))]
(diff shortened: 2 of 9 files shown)


paolobarbolini:
Resolved the merge conflicts after the last PR was merged.

Ping @seanmonstar. We're waiting for a code review :smiley:

silence-coding:
@paolobarbolini  From the header_read_timeout_no_write test case, it seems that once the connection is established, the timeout mechanism is started. Is that correct? I think this implementation is lacking, which means that the client's connection pool cannot establish connections in advance, and the client's keepAlive configuration is broken and does not keep alive as expected.

paolobarbolini:
> @paolobarbolini From the header_read_timeout_no_write test case, it seems that once the connection is established, the timeout mechanism is started. Is that correct? I think this implementation is lacking, which means that the client's connection pool cannot establish connections in advance, and the client's keepAlive configuration is broken and does not keep alive as expected.

Right, I didn't consider connection pools starting connections in advance. Fixed

Cargo.toml
```diff
@@ -86,7 +86,7 @@ full = [
 ]
 
 # HTTP versions
-http1 = []
+http1 = ["tokio/time"]
```
seanmonstar, line 89: In order to keep the runtime stuff optional, I think this should be part of the `runtime` feature, similar to the http2 keep-alive code as well.

src/error.rs
```diff
@@ -104,6 +104,9 @@ pub(super) enum User {
     #[cfg(any(feature = "http1", feature = "http2"))]
     #[cfg(feature = "server")]
     UnexpectedHeader,
+    /// User took too long to send headers
+    #[cfg(feature = "http1")]
+    HeaderTimeout,
```
seanmonstar, line 109: The classification of "user" in hyper has so far meant "user of the library", so the programmer writing some code that uses hyper. I wouldn't consider this a "user error", just probably a variant on the main enum.

src/proto/h1/role.rs
```diff
@@ -69,6 +69,14 @@ where
 
     let span = trace_span!("parse_headers");
     let _s = span.enter();
+
+    if let Some(h1_header_read_timeout) = ctx.h1_header_read_timeout {
+        if ctx.h1_header_read_timeout_fut.is_none() {
+            debug!("setting h1 header read timeout timer");
+            *ctx.h1_header_read_timeout_fut = Some(Box::pin(tokio::time::sleep(h1_header_read_timeout)));
```
seanmonstar, line 76: Is there a way we can either:

- Reuse the `Sleep` future, by calling `reset` with a new deadline.
- Reuse the `Box`, by setting a new one instead of clearing the `Box`.

Either of those would cut down on having to make an extra allocation per request, and we can just have 1 per connection.

src/server/server.rs
```diff
@@ -302,6 +302,17 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Set a timeout for reading client request headers. If a client does not 
+    /// transmit the entire header withing this time, the connection is closed.
+    ///
+    /// Default is None.
+    #[cfg(feature = "http1")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http1")))]
+    pub fn http1_header_read_timeout(mut self, read_timeout: Duration) -> Self {
```
seanmonstar, line 311: Should we make the argument be an `impl Into<Option<Duration>>`, so people can explicitly turn it off if they want? Is that ever a good idea?

seanmonstar (changes requested):
Sorry for the lag, thanks for writing this! It's very thorough, and I especially appreciate that there's tests so we don't accidentally break it in the future ❤️ 

I left a few comments inline.

src/proto/h1/role.rs
paolobarbolini: I ended up reusing the same `Sleep`

src/server/server.rs
paolobarbolini: I'm not sure. The default value is already `None`, so staying with the style of the other methods never setting it should be the way of disabling it?

src/server/conn.rs
```diff
@@ -372,6 +375,17 @@ impl<E> Http<E> {
         self
     }
 
+    /// Set a timeout for reading client request headers. If a client does not 
+    /// transmit the entire header withing this time, the connection is closed.
```
seanmonstar, line 379:
```suggestion
    /// transmit the entire header within this time, the connection is closed.
```

src/server/server.rs
```diff
@@ -309,6 +309,17 @@ impl<I, E> Builder<I, E> {
         self
     }
 
+    /// Set a timeout for reading client request headers. If a client does not 
+    /// transmit the entire header withing this time, the connection is closed.
```
seanmonstar, line 313:
```suggestion
    /// transmit the entire header within this time, the connection is closed.
```

seanmonstar (approved):

NeoLegends:
Great work here! This PR leaves me wondering though, won't hyper also need this for h2 and h3, or are there other mechanisms in place there that prevent slow loris attacks?

DFINITYManu:
Valiant effort, but still there is no way to set a timeout for clients that connect but send no data (by experimentation I discovered that the timeout does not fire if I just telnet into the TCP port and hang on there).  Which means it's trivial to gum up a hyper server just by sending a lot of connections and nothing else.

PR #2677, opened 2021-10-27
Restore a way to force queue writing strategy (fixes #2676)

### LINKED ISSUE #2676: Provide a way to force the queue write strategy, again
Once upon a time, there was [a way](https://github.com/hyperium/hyper/issues/2282) to force Hyper to use the queue write strategy.

This functionality was then [removed](https://github.com/hyperium/hyper/pull/2338) from Hyper 0.14.0.

Unfortunately, the heuristic to decide whether to use flatten or queue strategy has its limits. In my case, I need to buffer the output of an egress `HttpBody` implementation until a few MBs, and then I pass that big buffer to Hyper to write it over a h1 stream. When I pass it, Hyper is using the flatten strategy, so [it ends up copying the entire big buffer again](https://github.com/hyperium/hyper/blob/b5022f3854d1f9ed3e76233dd63f319efc3b8f47/src/proto/h1/io.rs#L548).

I'll make a PR to restore the builder setter to force queue strategy.

nox:


src/client/client.rs
```diff
@@ -1022,6 +1022,23 @@ impl Builder {
...
+    /// Note that setting this to false may mean more copies of body data,
+    /// but may also improve performance when an IO transport doesn't
+    /// support vectored writes well, such as most TLS implementations.
+    ///
+    /// Setting this to true will force hyper to use queued strategy
+    /// which may eliminate unnecessary cloning on some TLS backends
+    ///
+    /// Default is `auto`. In this mode hyper will try to guess which
+    /// mode to use
+    pub fn http1_writev(&mut self, enabled: bool) -> &mut Builder {
```
seanmonstar, line 1037: I don't have a strong feeling one way or the other, just wanted to hear what you think: should it be possible for a user to set it back to the default? Such as if another library wrapping hyper also wants to expose this option, should they be able to start with a default that doesn't overwrite the default? Or should the manage that themselves, and only call this function if configured to do so?

nox: I guess I can make `enabled` be a `impl Into<Option<bool>>`, but that feels like slightly overkill to me.

seanmonstar: It feels overkill to me too. I thought it, but didn't want to request it. Just felt it was worth talking about 😅

nox: I don't mind either way, if you think just a bool is not good enough, I can make the change.

seanmonstar (approved):

Issue #2086, opened 2019-12-20
Move HTTP upgrades API off the Body type


seanmonstar:
Introduced in https://github.com/hyperium/hyper/pull/1563, the HTTP Upgrade/`CONNECT` API currently is accessed via `hyper::Body::on_upgrade`. This has some downsides that I'd like to address with a modified API:

- It makes the `Body` type fatter.
- It requires people keep the `Body` around, meaning they can't use stream combinators to read the body first, *and then* wait for the upgrade to finish. This also makes it more annoying for users who may wish adjust their `http::Request<Body>` into some `http::Request<Doodad>`.

## Proposed API

Similar to the proposal in #1586, the change would be to pass the `http::Request` or `http::Response` to a free function.

- `hyper::upgrade::on(req_or_resp) -> OnUpgrade`
  - We can also allow passing just a mutable reference, thus not consuming the request or response.
- Optional other functions could be added, but not necessary: `has_upgrade`, etc.

## Implementation Plan

The types would be inserted into the `http::Extensions` of the request/response (though with a private wrapper type to hide the exact details).

- Add the new `hyper::upgrade::on` function.
- The `Body::on_upgrade` method would become `#[deprecated]`.
- So as to not be a behavioral breaking change, the `OnUpgrade` would need to be placed in both the `Body` *and* the `Extensions`, wrapped in some `Arc<Lock>` that only allows taking it once.

I'd like to in a future release be able to *not* put it in an `Arc<Lock>` and place it in the `Body`, but I'm not sure how that could really be done in 0.13.x. (I knew I'd find something I forgot to change after finally releasing XD).

dekellum:
Just my 2¢, and I don't know how you or other would feel about it, but at this early stage it should be fine IMO to just release a 0.14 with it changed to the way you believe it should be.  I suspect that only a few of us users have gone far upgrading to 0.13 yet and and its just another version to bump in our Cargo.toml.  Said from the opposite direction a quick 0.14 turnover is less concerning than breaking changes in 0.13.4 or whatever.

Maybe 0.13 just wasn't your lucky number?

If I had managed to release anything with an "^0.13" dependency already I might feel a little different, but not much.

dekellum:
A related question: should any of this updated API actually go in the _http_ crate?

VecrsIssues:
Should support for intermediate 1xx responses also be added?

https://tools.ietf.org/html/rfc7231#section-6.2

This could be useful for 103: Early Hints, or something else added in the future.

PR #2680, opened 2021-10-30
feat(upgrade): allow http upgrades with any body type

luqmana:
Allow using `Request<T>`/`Response<T>` for any given T with `upgrade::on` instead of just restricting it to `hyper::Body`.

#2086 mentioned moving the upgrades API off of the Body type with part of the motivation being:
>This also makes it more annoying for users who may wish adjust their `http::Request<Body>` into some `http::Request<Doodad>`.

But the current API is still restricted to just the `hyper::Body` body type which seems more strict than it needs to be.

### DIFF
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -258,7 +258,7 @@ mod sealed {
         fn on_upgrade(self) -> OnUpgrade;
     }
 
-    impl CanUpgrade for http::Request<crate::Body> {
+    impl<B> CanUpgrade for http::Request<B> {
         fn on_upgrade(mut self) -> OnUpgrade {
             self.extensions_mut()
                 .remove::<OnUpgrade>()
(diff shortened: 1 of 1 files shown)


src/upgrade.rs
```diff
@@ -258,31 +258,31 @@ mod sealed {
         fn on_upgrade(self) -> OnUpgrade;
     }
 
-    impl CanUpgrade for http::Request<crate::Body> {
+    impl<T> CanUpgrade for http::Request<T> {
```
nox, line 261: This parameter is usually called `B` in most of Hyper.

luqmana: Thanks, updated.

seanmonstar (approved):
Nice, thanks!

luqmana:
Np, thanks for merging! What's the policy on getting a release on crates.io?

PR #2682, opened 2021-11-03
feat(h2): implement the extended CONNECT protocol from RFC 8441

nox:
Needs https://github.com/hyperium/h2/pull/565 first.

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -1,9 +1,67 @@
-//! HTTP extensions
+//! HTTP extensions.
 
 use bytes::Bytes;
 #[cfg(feature = "http1")]
 use http::header::{HeaderName, IntoHeaderName, ValueIter};
 use http::HeaderMap;
+#[cfg(feature = "http2")]
+use std::fmt;
+
+#[cfg(feature = "http2")]
+/// Represents the `:protocol` pseudo-header used by
+/// the [Extended CONNECT Protocol].
+///
+/// [Extended CONNECT Protocol]: https://datatracker.ietf.org/doc/html/rfc8441#section-4
+#[derive(Clone, Eq, PartialEq)]
+pub struct Protocol {
+    inner: h2::ext::Protocol,
+}
+
+#[cfg(feature = "http2")]
+impl Protocol {
+    /// Converts a static string to a protocol name.
+    pub const fn from_static(value: &'static str) -> Self {
+        Self {
+            inner: h2::ext::Protocol::from_static(value),
+        }
+    }
+
+    /// Returns a str representation of the header.
+    pub fn as_str(&self) -> &str {
+        self.inner.as_str()
+    }
+
+    pub(crate) fn from_inner(inner: h2::ext::Protocol) -> Self {
+        Self { inner }
+    }
+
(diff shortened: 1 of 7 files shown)


nox:
`hyper::ext` is now public and it reexports `h2::ext::Protocol`.

src/ext.rs
```diff
@@ -1,10 +1,13 @@
-//! HTTP extensions
+//! HTTP extensions.
 
 use bytes::Bytes;
 #[cfg(feature = "http1")]
 use http::header::{HeaderName, IntoHeaderName, ValueIter};
 use http::HeaderMap;
 
+#[cfg(feature = "http2")]
+pub use h2::ext::Protocol;
```
seanmonstar, line 9: This makes the `h2` dependency version public. So we can't upgrade any h2 "breaking changes", which I'd rather we be able to do.

We _could_ expose a function to add `Protocol`, kind like how there's `hyper::upgrade::on(req)`. WDYT?

nox: > We _could_ expose a function to add `Protocol`, kind like how there's `hyper::upgrade::on(req)`. WDYT?

AFAIK that would still be a breaking change, as `Protocol` would still be exposed through the request's extensions.

Should I try duplicating `Protocol` here, and replacing it by h2's when actually passing the request to h2?

nox: Actually, the `h2` dependency is already public, through the error types.

nox: Anyway I added a commit to make `hyper::ext::Protocol` be a wrapper of `h2::ext::Protocol`.

seanmonstar: > Protocol would still be exposed through the request's extensions.

That's true... I have one regret around `hyper::upgrade::OnUpgrade`, that I made it the actual type put in extensions. If I had wrapped it in a private internal type, then it wouldn't be possible to depend on the extension type being in there.

We _could_ do similar in `h2`. Make the extension type actually private, and you have to call a function instead. Then in hyper, we just wrap calling that function. It's more of a dance, but perhaps stability is worth it. I'm actively pushing towards a 1.0, so this is more on my mind.

> Actually, the h2 dependency is already public, through the error types.

You mean through the `source`? That's an interesting point. It seems it's still an [open question](https://github.com/rust-lang/api-guidelines/discussions/96), whether that should be considered part of the public API. People seem to lean towards saying it's not, but it _would_ still break people at runtime if they depended on it... I could see an argument for us removing those connections as part of a hyper 1.0. Do you think we should consider that?

nox: I think rewrapping is fine, I actually never use `hyper::upgrade::on`, I don't like it, I prefer to reach in the extensions myself.

If you want to hide the type, you end up with weird traits like `CanUpgrade` and I don't like that at all.

For `h2`,  I think we should actually just bite the bullet and make it a public dependency, I've often wished I could reach into the `h2` parts of Hyper from Hyper's API.

Why do you want to push for 1.0 btw, any specific reason?

nox:
Let's land this!

nox:
Ping on this?

seanmonstar (approved):

PR #2688, opened 2021-11-06
docs(client, body): fix doctests failing when certain features are disabled

### LINKED ISSUE #2687: `body::to_bytes` doctest fails with default feature set
**Version**

Current git master aka [8062714](https://github.com/hyperium/hyper/commit/80627141ede3cdaf37b833ee91d08be0af0f5c84)

**Platform**

Linux 5.14.0-4-amd64 #1 SMP Debian 5.14.16-1 (2021-11-03) x86_64 GNU/Linux

**Description**

I cloned the repo in order to hack on the code, and I started out by making sure I could run the tests.
That is, I did:
```sh
git clone https://github.com/hyperium/hyper
cd hyper
cargo test
```

However, one of the doctests fails to compile:

```
   Doc-tests hyper

running 6 tests
test src/ext.rs - ext::HeaderCaseMap (line 14) ... ignored
test src/ext.rs - ext::HeaderCaseMap (line 22) ... ignored
test src/body/to_bytes.rs - body::to_bytes::to_bytes (line 19) ... FAILED
test src/service/make.rs - service::make::make_service_fn (line 106) ... ok
test src/service/util.rs - service::util::service_fn (line 13) ... ok
test src/body/body.rs - body::body::Body::empty (line 123) ... ok

failures:

---- src/body/to_bytes.rs - body::to_bytes::to_bytes (line 19) stdout ----
error[E0433]: failed to resolve: could not find `Client` in `hyper`
  --> src/body/to_bytes.rs:28:21
   |
12 | let client = hyper::Client::new();
   |                     ^^^^^^ could not find `Client` in `hyper`

For more information about this error, try `rustc --explain E0433`.
Couldn't compile the test.

failures:
    src/body/to_bytes.rs - body::to_bytes::to_bytes (line 19)

test result: FAILED. 3 passed; 1 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.27s
```

It looks like the test depends on some feature(s) outside of the default set.
I successfully fixed the issue locally by adding a `#[cfg(...)]` guard to the test, but I'm not sure if the real bug is in fact that the test depends on extra features.

Also, there seems to be a [CI job](https://github.com/hyperium/hyper/blob/master/.github/workflows/CI.yml#L145) that runs the test against many feature sets. For some reason it did not catch this. Maybe I'm looking at the wrong thing, or the job is buggy, or it's just really bad luck. In any case, might be worth investigating.

whentze:
The first commit fixes #2687, the second one fixes a very similar issue regarding a client doctest.

### DIFF
--- a/src/body/to_bytes.rs
+++ b/src/body/to_bytes.rs
@@ -17,6 +17,7 @@ use super::HttpBody;
 /// # Example
 ///
 /// ```
+/// # #[cfg(all(feature = "client", any(feature = "http1", feature = "http2")))]
 /// # async fn doc() -> hyper::Result<()> {
 /// use hyper::{body::HttpBody};
 ///
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -27,10 +27,10 @@
 //! [full client example](https://github.com/hyperium/hyper/blob/master/examples/client.rs).
 //!
 //! ```
+//! # #[cfg(all(feature = "tcp", feature = "client", any(feature = "http1", feature = "http2")))]
+//! # async fn fetch_httpbin() -> hyper::Result<()> {
 //! use hyper::{body::HttpBody as _, Client, Uri};
 //!
-//! # #[cfg(feature = "tcp")]
-//! # async fn fetch_httpbin() -> hyper::Result<()> {
 //! let client = Client::new();
 //!
 //! // Make a GET /ip to 'http://httpbin.org'


whentze:
Huh, gitcop complained about the commit message subject being too long so I reworded it, but it looks like it's not re-running the check for some reason 😐

seanmonstar:
It's OK, gitcop isn't marked required. It's just there as a guide to either nudge people to make a commit message that follows the pattern, or to me to clean up the message when squashing.

seanmonstar (approved):

PR #2689, opened 2021-11-08
feat(client): cancel blocking DNS lookup if GaiFuture dropped

ahmedsobeh:


src/client/connect/dns.rs
```diff
@@ -159,6 +159,10 @@ impl fmt::Debug for GaiFuture {
     }
 }
 
+impl Drop for GaiFuture {
+    fn drop(&mut self) { JoinHandle::abort(self) }
```
seanmonstar, line 163: You probably want `self.inner.abort()`.

src/client/connect/dns.rs
```diff
@@ -159,6 +159,10 @@ impl fmt::Debug for GaiFuture {
     }
 }
 
+impl Drop for GaiFuture {
+    fn drop(&mut self) { self.inner.abort() }
```
seanmonstar, line 163:
```suggestion
    fn drop(&mut self) {
        self.inner.abort();
    }
```

seanmonstar (approved):

Issue #2678, opened 2021-10-27
Meta: Tracing


seanmonstar:
hyper uses [tracing](https://crates.io/crates/tracing) for logging and events. This issue is meant to be a meta or tracking issue around the overall usage of `tracing` inside hyper. Eventually, users should be able to depend on certain spans and events. Which exactly, we need to figure out.

## Relevant issues

- [x] #2498 
- [ ] Design spans specific to the server
- [ ] Design spans specific to the client
- [x] By hyper 1.0, make `tracing` support "unstable" if not also 1.0

taqtiqa-mark:
@seanmonstar do you want to ignore events that are not in the context of a span?

Using the current code (still early), and the example `client_json`, these spanless events currently look like:

```log
TBC
```

taqtiqa-mark:
Recording some issues related to OpenTelemetry, these do not appear to impact other layers (json, print, jounald).
Context: With OTel none of the data shown above is currently visible.  We are not alone, and existing issues are still open.  As they are resolved the OTel tracing story in Hyper will improve:

- https://github.com/open-telemetry/opentelemetry-specification/issues/1682
- https://github.com/open-telemetry/opentelemetry-specification/issues/2125
- https://github.com/open-telemetry/opentelemetry-rust/issues/625
- https://github.com/tokio-rs/tracing/issues/1531
- https://github.com/tokio-rs/tracing/issues/1481

Geal:
Adding some context on https://github.com/seanmonstar/warp/issues/923: from a server point of view, it would be useful to have a span that is valid from the first byte received to the last byte sent. Otherwise, when using the `Service` trait with a chunked response, a span created in the service will not include any event sent when writing the body. Currently they appear under their own trace id, different from the request's trace id.

Such a span would need to be changed for each new request on an existing connection (in keep alive) for integration with opentelemetry.

I have looked a bit at how it could be implemented, at least for HTTP1, that would require adding a span member to `proto::h1::conn::Conn` and enter the span on every operation. I can make an attempt at it if there is no ongoing work in it right now

davidpdrsn:
@Geal this is exactly how tower-http's [`Trace` middleware](https://docs.rs/tower-http/0.2.0/tower_http/trace/index.html) works. The span it creates lasts from request is received until the entire response body is sent (or earlier if the client doesn't consume the entire body).

Geal:
@davidpdrsn Thanks, that worked perfectly

[...]

Issue #2498, opened 2021-04-08
Document that hyper is using tracing


seanmonstar:
The main `lib.rs` page would benefit from a section explaining how to view logs in hyper.

nox:
Would it be best to do that while carrying #2261 to completion?

seanmonstar:
I think they can happen separately. Really, a larger tracing effort would be good to organize for hyper, as there's several metrics people would like to know, and some simple placement of spans and proper key-value designations in some events in hyper could allow them to more easily get that info.

taqtiqa-mark:
@seanmonstar not sure if the docs [here](https://github.com/hyperium/hyper/pull/2697/files?authenticity_token=QGM7ib6nf7liG6WrZAmg%2BCGdlgERiX0%2FN%2Fm24OaG7k76wriSEm07gDh3mvkLsAgxXzTqW68HZq44qTswA4Qeig%3D%3D&file-filters%5B%5D=.rs#diff-8e6412c29b9579988c7b6c7620cc1c903e312f50482ee8687281cd20aaf8d314) are sufficient to close this? 

 They'll expand as the implementation roles out, but right now have a 10,00ft view of what tracing in Hyper means (OTel semantic conventions).

seanmonstar:
The 1.0 docs now mention that `tracing` is available as an unstable feature. I think that's enough for now.

PR #2697, opened 2021-11-18
Issue 2678: Span PoC - direction/feedback

taqtiqa-mark:
Seeking feedback on the approach:

TLDR;

```bash
podman run -p6831:6831/udp -p6832:6832/udp -p16686:16686 jaegertracing/all-in-one:latest
cargo run --features="full tracing/max_level_trace" --example client_json_tracing_otel
firefox http://localhost:16686
# See also stdout
```

- ~~Use of a feature flag `layers`, adding the flag to `full` and placing tracing logic behind that flag.~~ See issue #2698 
- Use OpenTelemetry semantic conventions as the starting point.
- The initial documentation is in `src/common/layers/mod.rs`
- [src/proto/h1/role.rs]:  `Server::encode` span (328-470) is the most complete and serves point of reference.  We start here because these are the functions the the subject of spans in the current code base:

  - The `cfg_attr` is a template I propose to use elsewhere: Acceptable?
  - The use of `tracing::Span::current().record(...)`: Acceptable?
  - The `cfg!(feature = ~~"layers"~~ TBD)` guard: Acceptable?
  - I suggest using the `tracing-error` crate, so I don't propose to do too much around the errors: Acceptable?
  - Function arguments that require a tracing `Value` trait are skipped. The proposed workaround until the Value trait is unsealed is to `impl Debug for ...` and `impl Display for ...` in `src/common/layers/values/*.rs`: Acceptable?
  - Move all `debug!(..)` and `trace!(..)` either into a span field, or if not appropriate, behind a `cfg!(feature = "layers")` guard, leave all other levels alone: Acceptable?
  
Once these four spans (`Client:encode`, `Client:parse`, `Server:encode`, `Server:parse`) are  stabilized and merged:

- Propose additional functions to add spans for, per @seanmonstar 's suggestion [finangle guide](https://twitter.github.io/finagle/guide/Tracing.html#standard-annotations)
- Once all spans are in place, rationalize the fields (and their updating) across/between spans
- Add `tracing-error`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ httpdate = "1.0"
 httparse = "1.5.1"
 h2 = { version = "0.3.3", optional = true }
 itoa = "0.4.1"
-tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project-lite = "0.2.4"
 tower-service = "0.3"
 tokio = { version = "1", features = ["sync"] }
@@ -42,7 +41,11 @@ want = "0.3"
 # Optional
 
 libc = { version = "0.2", optional = true }
+log = { version = "0.4.14", optional = true }
+serde_json = { version = "1.0", optional = true }
 socket2 = { version = "0.4", optional = true }
+tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
+tracing-subscriber = { version = "0.3", optional = true }
 
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
@@ -80,9 +83,10 @@ full = [
     "client",
     "http1",
     "http2",
+    "layers",
+    "runtime",
     "server",
     "stream",
-    "runtime",
 ]
 
 # HTTP versions
@@ -107,6 +111,9 @@ tcp = [
     "tokio/rt",
     "tokio/time",
 ]
+
+# Tracing layers
+layers = ["serde_json", "tracing", "tracing-subscriber"]
 
 # C-API support (currently unstable (no semver))
 ffi = ["libc"]
@@ -139,6 +146,21 @@ required-features = ["full"]
 name = "client_json"
 path = "examples/client_json.rs"
 required-features = ["full"]
+
+[[example]]
+name = "client_json_tracing"
+path = "examples/client_json_tracing.rs"
+required-features = ["full",
+    "tracing/max_level_info",
+    ]
+
+[[example]]
+name = "client_json_tracing_off"
+path = "examples/client_json_tracing_off.rs"
+required-features = ["tracing",
+    "tracing/max_level_off",
+    "log/max_level_off",
+    ]
 
 [[example]]
 name = "echo"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,10 +47,12 @@ opentelemetry-jaeger = { version = "0.15.0", features = ["reqwest_collector_clie
 serde_json = { version = "1.0", optional = true }
 socket2 = { version = "0.4", optional = true }
 reqwest = {version = "0.11", optional = true }
-tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
+tracing = { version = "0.1", default-features = false, features = ["std", "log"], optional = true }
 tracing-core = { version = "0.1", default-features = false, optional = true }
+tracing-log = { version = "0.1", default-features = false, optional = true }
 tracing-opentelemetry = { version = "0.16", optional = true }
-tracing-subscriber = { version = "0.3", features = ["std"], optional = true }
+tracing-subscriber = { version = "0.3", features = ["std", "env-filter"], optional = true }
+tracing-tree = { version = "0.1", default-features = false, optional = true }
 
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
@@ -119,14 +121,18 @@ tcp = [
 
 # Tracing layers
 layers = [
+    "log",
     "opentelemetry",
     "opentelemetry-jaeger",
     "serde_json",
     "reqwest",
     "tracing",
     "tracing-core",
+    "tracing-log",
     "tracing-opentelemetry",
-    "tracing-subscriber"]
+    "tracing-subscriber",
+    "tracing-tree",
+    ]
 
 # C-API support (currently unstable (no semver))
 ffi = ["libc"]
@@ -176,8 +182,8 @@ required-features = ["full",
     ]
 
 [[example]]
-name = "client_otel_tracing"
-path = "examples/client_otel_tracing.rs"
+name = "client_json_tracing_otel"
+path = "examples/client_json_tracing_otel.rs"
 required-features = ["full",
     "tracing/max_level_trace",
     ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,10 +42,15 @@ want = "0.3"
 
 libc = { version = "0.2", optional = true }
 log = { version = "0.4.14", optional = true }
+opentelemetry = { version = "0.16.0", features = ["rt-tokio", "trace"], optional = true }
+opentelemetry-jaeger = { version = "0.15.0", features = ["reqwest_collector_client","rt-tokio"], optional = true }
 serde_json = { version = "1.0", optional = true }
 socket2 = { version = "0.4", optional = true }
+reqwest = {version = "0.11", optional = true }
 tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
-tracing-subscriber = { version = "0.3", optional = true }
+tracing-core = { version = "0.1", default-features = false, optional = true }
+tracing-opentelemetry = { version = "0.16", optional = true }
+tracing-subscriber = { version = "0.3", features = ["std"], optional = true }
 
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
@@ -113,7 +118,15 @@ tcp = [
 ]
 
 # Tracing layers
-layers = ["serde_json", "tracing", "tracing-subscriber"]
+layers = [
+    "opentelemetry",
+    "opentelemetry-jaeger",
+    "serde_json",
+    "reqwest",
+    "tracing",
+    "tracing-core",
+    "tracing-opentelemetry",
+    "tracing-subscriber"]
 
 # C-API support (currently unstable (no semver))
 ffi = ["libc"]
@@ -148,24 +161,38 @@ path = "examples/client_json.rs"
 required-features = ["full"]
 
 [[example]]
-name = "client_json_tracing"
-path = "examples/client_json_tracing.rs"
+name = "client_json_tracing_json"
+path = "examples/client_json_tracing_json.rs"
 required-features = ["full",
-    "tracing/max_level_info",
+    "tracing/max_level_trace",
     ]
 
 [[example]]
 name = "client_json_tracing_off"
 path = "examples/client_json_tracing_off.rs"
-required-features = ["tracing",
+required-features = ["full",
     "tracing/max_level_off",
     "log/max_level_off",
+    ]
+
+[[example]]
+name = "client_otel_tracing"
+path = "examples/client_otel_tracing.rs"
+required-features = ["full",
+    "tracing/max_level_trace",
     ]
 
 [[example]]
 name = "echo"
 path = "examples/echo.rs"
 required-features = ["full"]
+
+[[example]]
+name = "echo_tracing_otel"
+path = "examples/echo_tracing_otel.rs"
+required-features = ["full",
+    "tracing/max_level_trace",
+    ]
 
 [[example]]
 name = "gateway"
--- a/examples/client_json_tracing_otel.rs
+++ b/examples/client_json_tracing_otel.rs
@@ -19,7 +19,7 @@
 
 use hyper::body::Buf;
 use hyper::Client;
-// use hyper::OtelLayer;
+//use hyper::OtelLayer;
 use serde::Deserialize;
 use tracing::info;
 use tracing_subscriber::layer::SubscriberExt;
@@ -32,42 +32,30 @@ type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    // // Set up `tracing-subscriber` to process tracing data.
-    // // Create a jaeger exporter pipeline for a `trace_demo` service.
-    // let addr = std::net::SocketAddr::from(([127, 0, 0, 1], 6831));
-    // // Build a Jaeger batch span processor
-    // let jaeger_processor = opentelemetry::sdk::trace::BatchSpanProcessor::builder(
-    //     opentelemetry_jaeger::new_pipeline()
-    //         .with_service_name("mre-jaeger")
-    //         .with_agent_endpoint(addr)
-    //         .with_trace_config(opentelemetry::sdk::trace::config().with_resource(
-    //             opentelemetry::sdk::Resource::new(vec![
-    //                 opentelemetry::KeyValue::new("service.name", "client"),
-    //                 opentelemetry::KeyValue::new("service.namespace", "client-namespace"),
-    //             ]),
-    //         ))
-    //         .init_async_exporter(opentelemetry::runtime::Tokio)
-    //         .expect("Jaeger Tokio async exporter"),
-    //     opentelemetry::runtime::Tokio,
-    // )
-    // .build();
-    // Start an otel jaeger trace pipeline
-    opentelemetry::global::set_text_map_propagator(opentelemetry::sdk::propagation::TraceContextPropagator::new());
-    //global::set_text_map_propagator(opentelemetry_jaeger::Propagator::new());
-    // Then create a composite propagator
-    // let composite_propagator =  opentelemetry::sdk::propagation::TextMapCompositePropagator::new(vec![
-    //     Box::new(baggage_propagator),
-    //     Box::new(trace_context_propagator),
-    // ]);
+    // First create propagators
+    let baggage_propagator = opentelemetry::sdk::propagation::BaggagePropagator::new();
+    let trace_context_propagator = opentelemetry::sdk::propagation::TraceContextPropagator::new();
+    let jaeger_propagator = opentelemetry_jaeger::Propagator::new();
+    opentelemetry::global::set_text_map_propagator(
+        opentelemetry::sdk::propagation::TraceContextPropagator::new(),
+    );
+
+    // Second compose propagators
+    let _composite_propagator =  opentelemetry::sdk::propagation::TextMapCompositePropagator::new(vec![
+        Box::new(baggage_propagator),
+        Box::new(trace_context_propagator),
+        Box::new(jaeger_propagator),
+    ]);
     let tracer = opentelemetry_jaeger::new_pipeline()
         .with_service_name("client_json2")
         .install_batch(opentelemetry::runtime::Tokio)
         .unwrap();
     // Initialize `tracing` using `opentelemetry-tracing` and configure logging
     tracing_subscriber::Registry::default()
-        //.with(tracing_subscriber::EnvFilter::new("TRACE"))
+        .with(tracing_subscriber::EnvFilter::new("TRACE"))
+        .with(hyper::OtelLayer::new(tracer))
         .with(tracing_subscriber::fmt::layer())
-        .with(tracing_opentelemetry::layer().with_tracer(tracer))
+        //.with(tracing_tree::HierarchicalLayer::new(2))
         .init();
 
     // // Setup Tracer Provider
@@ -89,30 +77,34 @@ async fn main() -> Result<()> {
     //let subscriber = tracing_subscriber::Registry::default().with(telemetry);
 
     // Trace executed (async) code
-    //tracing::subscriber::with_default(subscriber, || async {
-    // Create a span and enter it, returning a guard....
+    // tracing::subscriber::with_default(subscriber, || async {
+        // Create a span and enter it, returning a guard....
     let root_span = tracing::span!(tracing::Level::INFO, "root_span_echo").entered();
+    root_span.in_scope(|| async {
+        // After setting up OpenTelemetry/Jaeger, redirect all `log`'s events to our subscriber/layer
+        // let logger = tracing_log::LogTracer::new();
+        // log::set_boxed_logger(Box::new(logger))?;
 
-    // We are now inside the span! Like `enter()`, the guard returned by
-    // `entered()` will exit the span when it is dropped...
+        // We are now inside the span! Like `enter()`, the guard returned by
+        // `entered()` will exit the span when it is dropped...
 
-    // Log a `tracing` "event".
-    info!(status = true, answer = 42, message = "first event");
+        // Log a `tracing` "event".
+        info!(status = true, answer = 42, message = "first event");
 
-    let url = "http://jsonplaceholder.typicode.com/users".parse().unwrap();
-    let users = fetch_json(url).await.expect("Vector of user data");
-    // print users
-    println!("users: {:#?}", users);
+        let url = "http://jsonplaceholder.typicode.com/users".parse().unwrap();
+        let users = fetch_json(url).await.expect("Vector of user data");
+        // print users
+        println!("users: {:#?}", users);
 
-    // print the sum of ids
-    let sum = users.iter().fold(0, |acc, user| acc + user.id);
-    println!("sum of ids: {}", sum);
+        // print the sum of ids
+        let sum = users.iter().fold(0, |acc, user| acc + user.id);
+        println!("sum of ids: {}", sum);
 
-    // ...but, it can also be exited explicitly, returning the `Span`
-    // struct:
-    let _root_span = root_span.exit();
+        // ...but, it can also be exited explicitly, returning the `Span`
+        // struct:
+        //let _root_span = root_span.exit();
+    }).await;
     opentelemetry::global::shutdown_tracer_provider();
-    //}).await;
     Ok(())
 }
 
--- a/examples/client_json_tracing_otel.rs
+++ b/examples/client_json_tracing_otel.rs
@@ -46,11 +46,12 @@ async fn main() -> Result<()> {
         Box::new(trace_context_propagator),
         Box::new(jaeger_propagator),
     ]);
+    // Third create Jaeger pipeline
     let tracer = opentelemetry_jaeger::new_pipeline()
         .with_service_name("client_json2")
         .install_batch(opentelemetry::runtime::Tokio)
         .unwrap();
-    // Initialize `tracing` using `opentelemetry-tracing` and configure logging
+    // Initialize `tracing` using `opentelemetry-tracing` and configure stdout logging
     tracing_subscriber::Registry::default()
         .with(tracing_subscriber::EnvFilter::new("TRACE"))
         .with(hyper::OtelLayer::new(tracer))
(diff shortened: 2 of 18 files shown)


Cargo.toml
```diff
@@ -108,6 +119,21 @@ tcp = [
     "tokio/time",
 ]
 
+# Tracing layers
+layers = [
```
hawkw, line 123: why is the feature flag to enable `tracing` support called `layers`? i would expect it to be called `tracing` or something...

Cargo.toml
```diff
@@ -42,7 +41,18 @@ want = "0.3"
 # Optional
 
 libc = { version = "0.2", optional = true }
+log = { version = "0.4.14", optional = true }
```
hawkw, line 44: why is there now a dependency on the `log` crate?

Cargo.toml
```diff
@@ -108,6 +119,21 @@ tcp = [
...
+    "opentelemetry",
+    "opentelemetry-jaeger",
+    "serde_json",
+    "reqwest",
+    "tracing",
+    "tracing-core",
+    "tracing-log",
+    "tracing-opentelemetry",
+    "tracing-subscriber",
+    "tracing-tree",
```
hawkw, line 134: i don't think that the feature that enables `tracing` instrumentation in Hyper should enable all of these dependencies. suppose I want to log `tracing` spans from hyper using `tracing-subscriber`, in the plaintext format. in that case, i would want to enable `tracing` instrumentation in `hyper`...but i would not want `hyper` to suddenly depend on `log`, `serde_json`, `reqwest`, `tracing_opentelemetry`, `opentelemetry`, `opentelemetry-jaeger`, and `tracing-tree`.

Cargo.toml
```diff
@@ -140,11 +166,40 @@ name = "client_json"
...
+    "log/max_level_off",
+    ]
+
+[[example]]
+name = "client_json_tracing_otel"
+path = "examples/client_json_tracing_otel.rs"
+required-features = ["full",
+    "tracing/max_level_trace",
+    ]
+
```
hawkw, line 190: take it or leave it: i think it might be worth putting all the `tracing` examples in a `examples/tracing` directory, so it's clear they're all related to various `tracing`-related configurations?

examples/client_json_tracing_otel.rs
```diff
@@ -0,0 +1,107 @@
...
+        println!("users: {:#?}", users);
+
+        // print the sum of ids
+        let sum = users.iter().fold(0, |acc, user| acc + user.id);
+        println!("sum of ids: {}", sum);
+
+        // ...but, it can also be exited explicitly, returning the `Span`
+        // struct:
+        //let _root_span = root_span.exit();
+    }).await;
```
hawkw, line 82: this is incorrect. `Span::in_scope` here is not instrumenting the `async` block, but the _closure_ that returns it...which, in this case, doesn't contain any `tracing` events, as it immediately returns the `async` block. So, calling `in_scope` here is _not_ doing anything. instead, these events occur within `root_span`'s scope because of the `Span::entered` call prior to the closure...but `entered` should not be used in asynchronous code. see [the documentation](https://docs.rs/tracing/0.1.29/tracing/span/struct.Span.html#in-asynchronous-code) for details on why.

instead, this should be:
```suggestion
    use tracing::instrument::Instrument;
    
    let root_span = tracing::span!(tracing::Level::INFO, "root_span_echo);
    async {
        // Log a `tracing` "event".
        info!(status = true, answer = 42, message = "first event");

        let url = "http://jsonplaceholder.typicode.com/users".parse().unwrap();
        let users = fetch_json(url).await.expect("Vector of user data");
        // print users
        println!("users: {:#?}", users);

        // print the sum of ids
        let sum = users.iter().fold(0, |acc, user| acc + user.id);
        println!("sum of ids: {}", sum);
    }
         .instrument(root_span)
        .await;
```

examples/client_json_tracing_otel1.rs
```diff
@@ -0,0 +1,119 @@
...
+    println!("users: {:#?}", users);
+
+    // print the sum of ids
+    let sum = users.iter().fold(0, |acc, user| acc + user.id);
+    println!("sum of ids: {}", sum);
+
+    // ...but, it can also be exited explicitly, returning the `Span`
+    // struct:
+    let _root_span = root_span.exit();
+    //}).await;
```
hawkw, line 95: again, `Span::entered` should not be used in asynchronous code (such as `async fn main`):
https://docs.rs/tracing/0.1.29/tracing/span/struct.Span.html#in-asynchronous-code

this should probably be:
```suggestion
    // Create a span
    let root_span = tracing::span!(tracing::Level::INFO, "root_span_echo");
    async move {
            // We are now inside the span!
        
            // Log a `tracing` "event".
            info!(status = true, answer = 42, message = "first event");
    
            let url = "http://jsonplaceholder.typicode.com/users".parse().unwrap();
            let users = fetch_json(url).await.expect("Vector of user data");
            // print users
            println!("users: {:#?}", users);
    
            // print the sum of ids
            let sum = users.iter().fold(0, |acc, user| acc + user.id);
            println!("sum of ids: {}", sum);
    }
         .instrument(root_span)
         .await;
```

examples/client_json_tracing_otel.rs
```diff
@@ -0,0 +1,107 @@
...
+        info!(status = true, answer = 42, message = "first event");
+
+        let url = "http://jsonplaceholder.typicode.com/users".parse().unwrap();
+        let users = fetch_json(url).await.expect("Vector of user data");
+        // print users
+        println!("users: {:#?}", users);
+
+        // print the sum of ids
+        let sum = users.iter().fold(0, |acc, user| acc + user.id);
+        println!("sum of ids: {}", sum);
```
hawkw, line 77: were these `println!`s intended to be `tracing` events? otherwise, i'm not sure if i understand the motivation behind including them in the example...

examples/echo_tracing_otel.rs
```diff
@@ -0,0 +1,115 @@
...
+    let telemetry = OtelLayer::new(tracer);
+
+    // Use tracing subscriber `Registry`, or any other subscriber that `impl LookupSpan`
+    tracing_subscriber::registry()
+        .with(telemetry)
+        .try_init()
+        .expect("Default subscriber");
+
+        // Create a span and enter it, returning a guard....
+        let root_span = tracing::span!(tracing::Level::INFO, "root_span_echo").entered();
```
hawkw, line 97: again, `entered` should probably not be used in an `async fn`.

src/common/layers/json.rs
```diff
@@ -0,0 +1,79 @@
+use std::collections::BTreeMap;
+
+/// A layer for the tracing framework to record structured event and span data.
+#[derive(Debug)]
+pub struct HyperLayer;
```
hawkw, line 5: i'm not sure if i understand the motivation for `hyper` providing its own `tracing_subscriber::Layer` implementations? the core idea of `tracing` is that it serves as a _facade_ for connecting instrumentation in libraries with subscribers defined in _other_ libraries. right now, it looks like this `Layer` doesn't do any kind of `hyper`-specific behavior...so why is it part of `hyper` rather than a separate crate providing a general-purpose JSON layer?

if the intention is to eventually add hyper-specific behavior, i think a better approach is still probably to include that layer in a separate crate --- call it `hyper-tracing`, or something? that way, the application can plug in whatever `tracing` layer it needs, and `hyper`'s core functionality --- actually implementing HTTP clients and servers --- just exposes `tracing` instrumentation.

src/common/layers/otel.rs
```diff
@@ -0,0 +1,787 @@
+use opentelemetry::{
```
hawkw, line 1: as far as i can tell, this module appears to be...a vendored version of `tracing-opentelemetry`'s `OpenTelemetryLayer` type? why does this need to be part of `hyper`?

again, the purpose of `tracing` providing a facade layer is that the application can choose what `Layer`s to use to record the _instrumentation_ emitted by libraries like `hyper`. for example, if an application developer wanted to record traces, including `hyper`'s instrumentation, using opentelemetry, they would simply depend on the `tracing-opentelemetry` crate. this shouldn't have to be part of `hyper` itself in order to use the instrumentation in `hyper`.

src/proto/h1/role.rs
```diff
@@ -289,13 +325,41 @@ impl Http1Transaction for Server {
...
-            "Server::encode status={:?}, body={:?}, req_method={:?}",
-            msg.head.subject,
-            msg.body,
-            msg.req_method
-        );
+        let mut otel_status_messages: Vec<String> = vec![];
+        if cfg!(feature = "layers") {
+            tracing::Span::current()
+                .record("http.method", &format!("{:?}", msg.req_method).as_str());
+            tracing::Span::current().record(
```
hawkw, line 357: it's probably more efficient to avoid getting the current span multiple times ---`Span::current` performs a thread-local storage access. prefer:
```suggestion
            let span = tracing::Span::current();
            span.record("http.method", &format!("{:?}", msg.req_method).as_str());
            span.record(
```

also, we should record the method as a `Debug` trait object, in order to avoid allocating a string here:
```suggestion
            let span = tracing::Span::current();
            span.record("http.method", &msg.req_method as &dyn std::fmt::Debug));
            span.record(
```

src/proto/h1/role.rs
```diff
@@ -289,13 +325,41 @@ impl Http1Transaction for Server {
...
+                "http.status_code",
+                &format!("{:?}", msg.head.subject).as_str(),
```
hawkw, line 359:
```suggestion
                &msg.head.subject as &dyn std::fmt::Debug
```

src/proto/h1/role.rs
```diff
@@ -289,13 +325,41 @@ impl Http1Transaction for Server {
...
+            );
+            trace!("Server::encode body={:?}", msg.body);
```
hawkw, line 361: prefer structured fields:
```suggestion
            trace!(body = ?msg.body);
```

src/proto/h1/role.rs
```diff
@@ -328,16 +392,42 @@ impl Http1Transaction for Server {
...
+                            .push("response with HTTP2 version coerced to HTTP/1.1".to_string());
+                    }
                     extend(dst, b"HTTP/1.1 ");
                 }
-                other => panic!("unexpected response version: {:?}", other),
+                other => {
+                    if cfg!(feature = "layers") {
+                        // 0.0 is a placeholder until `impl Debug for Version`
+                        tracing::Span::current().record("http.flavor", &0.0);
+                        tracing::Span::current().record("otel.status", &format!("{:?}", opentelemetry::trace::StatusCode::Error).as_str());
```
hawkw, line 425: again, we probably want to avoid getting the current span twice if we don't have to...

src/proto/h1/role.rs
```diff
@@ -1011,6 +1127,30 @@ impl Http1Transaction for Client {
         }
     }
 
+    #[cfg_attr(feature = "layers",
+        instrument( level = "trace",
+                    skip(msg),
```
hawkw, line 1132: probably also don't want to record `dst` here...
```suggestion
                    skip(msg, dst),
```

hawkw (changes requested):
i have some concerns about the current approach here. i don't really understand why it's necessary for `hyper` itself to implement its own `Layer`s for recording `tracing` data. as far as i can tell, these don't implement any `hyper`-specific behaviors, and are better off implemented outside of `hyper`.

if there _is_ a compelling reason for `hyper` to provide `hyper`-specific `tracing` layers that implement behaviors tightly coupled with `hyper`, I think those probably ought to go in a separate crate, rather than within the `hyper` crate. since the `tracing` facade allows a _loose_ coupling between instrumentation and subscribers, there shouldn't be any reason for any `hyper` layers to need to access `hyper`'s internals.

also, if there _is_ some reason these layers need to be defined within `hyper` itself...they almost certainly should be enabled by a separate feature flag than the `tracing` spans and events exposed by `hyper`. if i want to use `tracing-subscriber::fmt` to record those spans and events, i might want to enable that feature flag...but i don't need any of the `hyper` layers and their (often heavyweight) dependencies. there should be a separate feature flag for _instrumentation_ in hyper.

seanmonstar:
Thanks for the PR! I appreciate your effort to improve tracing in hyper.

I agree with several of Eliza's points:

- We shouldn't need any new dependencies.
- I'd probably lean towards not having any new example files. Maybe one if the existing ones don't do a good job of letting you use tracing, but ideally they all do.
- Any layers should be possible in a separate crate. The point of a layer is to allow someone to define what information that is emitted they want to serialize somewhere. All hyper should be doing is emitting all the useful information.

This isn't to say that the layers here aren't useful, I'm sure they are from an open telemetry perspective. Likely, building them and the examples can help you come up with what exact events and spans you actually need to serialize. I'd say keep experimenting with the examples, but in the goal of doing as I mentioned in [#2678](https://github.com/hyperium/hyper/issues/2498#issuecomment-953383248):

Determine what spans, events, and fields that absolutely need to be inside hyper (as opposed to ones that can occur outside hyper, in a custom `impl Service`), and propose them. I'd start with a list in a new issue. Those from finagle might be useful, but also from your examples you might find what is missing.

The point is that hyper should emit spans and events that a user couldn't figure out from outside hyper. I assume this will most often be timing related. Most of the other information can be recorded at the start of an `impl Service`.

taqtiqa-mark:
> i have some concerns about the current approach here. i don't really understand why it's necessary for `hyper` itself to implement its own `Layer`s for recording `tracing` data. as far as i can tell, these don't implement any `hyper`-specific behaviors, and are better off implemented outside of `hyper`.
> 
> if there _is_ a compelling reason for `hyper` to provide `hyper`-specific `tracing` layers that implement behaviors tightly coupled with `hyper`, I think those probably ought to go in a separate crate, rather than within the `hyper` crate. since the `tracing` facade allows a _loose_ coupling between instrumentation and subscribers, there shouldn't be any reason for any `hyper` layers to need to access `hyper`'s internals.
> 
> also, if there _is_ some reason these layers need to be defined within `hyper` itself...they almost certainly should be enabled by a separate feature flag than the `tracing` spans and events exposed by `hyper`. if i want to use `tracing-subscriber::fmt` to record those spans and events, i might want to enable that feature flag...but i don't need any of the `hyper` layers and their (often heavyweight) dependencies. there should be a separate feature flag for _instrumentation_ in hyper.

Overall I agree.  I don't believe I proposed examples or layers in the questions I posed.  Would appreciate your feedback on the questions raised.

Cargo.toml
taqtiqa-mark: > take it or leave it: i think it might be worth putting all the `tracing` examples in a `examples/tracing` directory, so it's clear they're all related to various `tracing`-related configurations?

OT but I'll take it into account closer to the time a final PR is ready.

Cargo.toml
taqtiqa-mark: > i don't think that the feature that enables `tracing` instrumentation in Hyper should enable all of these dependencies. suppose I want to log `tracing` spans from hyper using `tracing-subscriber`, in the plaintext format. in that case, i would want to enable `tracing` instrumentation in `hyper`...but i would not want `hyper` to suddenly depend on `log`, `serde_json`, `reqwest`, `tracing_opentelemetry`, `opentelemetry`, `opentelemetry-jaeger`, and `tracing-tree`.

Summarizing - response to the question of introducing a feature: No objection, in fact please consider features that allow more fine grained control. Accurate summary?

Cargo.toml
taqtiqa-mark: > why is there now a dependency on the `log` crate?

Is required to turn off tracing in an app that uses hyper (the example).

Cargo.toml
taqtiqa-mark: > why is the feature flag to enable `tracing` support called `layers`? i would expect it to be called `tracing` or something...

I believe that is where I started - again from memory - you can't have a feature name that is congruent to a crate name - but I'm not a Rust expert

taqtiqa-mark:
> i have some concerns about the current approach here. i don't really understand why it's necessary for `hyper` itself to implement its own `Layer`s for recording `tracing` data. as far as i can tell, these don't implement any `hyper`-specific behaviors, and are better off implemented outside of `hyper`.
> 
> if there _is_ a compelling reason for `hyper` to provide `hyper`-specific `tracing` layers that implement behaviors tightly coupled with `hyper`, I think those probably ought to go in a separate crate, rather than within the `hyper` crate. since the `tracing` facade allows a _loose_ coupling between instrumentation and subscribers, there shouldn't be any reason for any `hyper` layers to need to access `hyper`'s internals.
> 
> also, if there _is_ some reason these layers need to be defined within `hyper` itself...they almost certainly should be enabled by a separate feature flag than the `tracing` spans and events exposed by `hyper`. if i want to use `tracing-subscriber::fmt` to record those spans and events, i might want to enable that feature flag...but i don't need any of the `hyper` layers and their (often heavyweight) dependencies. there should be a separate feature flag for _instrumentation_ in hyper.

Largely agree with these points.  I'm trying to check how these spans 'look' in different contexts and to decide whether any changes are required - as temporary workarounds.  Still exploring/experimenting.  any layers in hyper would be proposed separately.

Apologies if the questions I posed gave the impression I was wanting to bring these layers in.
That was not my intention.

src/proto/h1/role.rs
taqtiqa-mark: > prefer structured fields:

I'm pretty much leaving existing code unchanged where it is not fully replaced by some span/event structure.  If I was to propose changing this it would be to storing the byte count in one of the span fields.

examples/client_json_tracing_otel.rs
taqtiqa-mark: Thanks for the correction.  I think this makes the point that examples are required - ideally someone should be able to get tracing without having to become an expert in the tracing stack - 80/20 rule - some standard boiler plate setup should suffice for 80%+ of use cases.

src/proto/h1/role.rs
taqtiqa-mark: > also, we should record the method as a Debug trait object, in order to avoid allocating a string here:

Thanks, that is way beyond my ken at this point.

taqtiqa-mark:
> Thanks for the PR! I appreciate your effort to improve tracing in hyper.
> 
> I agree with several of Eliza's points:
> 
>     * We shouldn't need any new dependencies.

I understand the concern.  The issue is: getting debug/trace output in a usable form should be off-the-shelf functionality.  IMO, the Hyper instructions need not be more complex than this:

```bash
podman run -p6831:6831/udp -p6832:6832/udp -p16686:16686 jaegertracing/all-in-one:latest
cargo run --features="some thing tracing/max_level_trace" --example client_json_tracing_otel
firefox http://localhost:16686
```
or some such.
Can that be done without any dependencies?
I agree the feature flag is not right as it stands right now. See below for a proposal.

> 
>     * I'd probably lean towards not having any new example files. Maybe one if the existing ones don't do a good job of letting you use tracing, but ideally they all do.

See @hawkw 's earlier comment on the subtleties of setting up tracing for async.  IMO, debug output should just work and not require you to spend more than 2 minutes on setup.  

Adding spans/events, and nothing more, is just throwing the same problem onto every user to keep solving.  
Perhaps we can come up with some boiler plate that will solve 80% of the problems.

To make it concrete: 
In issues reporting bugs there should be a set of one/two commands that a user is asked to run and provides output that identifies the issue.

> 
>     * Any layers should be possible in a separate crate. The point of a layer is to allow someone to define what information that is emitted they want to serialize somewhere. All hyper should be doing is emitting all the useful information.

Right now it isn't clear to me that some layer is/isn't required - even as a workaround - to get the full output to solve the problems.
I think it is axiomatic that any layer in hyper would have to add value not available elsewhere.

> 
> 
> This isn't to say that the layers here aren't useful, I'm sure they are from an open telemetry perspective. Likely, building them and the examples can help you come up with what exact events and spans you actually need to serialize.

Correct. They've exposed several things that need to be addressed, but that is likely my ignorance, and may be non-issues.

> I'd say keep experimenting with the examples, but in the goal of doing as I mentioned in [#2678](https://github.com/hyperium/hyper/issues/2498#issuecomment-953383248):
> 
> Determine what spans, events, and fields that absolutely need to be inside hyper (as opposed to ones that can occur outside hyper, in a custom `impl Service`), and propose them.

Isn't this that?

> I'd start with a list in a new issue. 

Ditto.

Also bear in mind:

1) We can conditionally add spans/events/logs behind some feature flag 
2) Even if we add that feature flag as default we can compile out the same.

I think this means erring on the side of being liberal in extracting information - there is no runtime performance penalty and we need debug output that doesn't result in issues sitting open for >1 month (ideally)?

> Those from finagle might be useful, but also from your examples you might find what is missing.
> 
> The point is that hyper should emit spans and events that a user couldn't figure out from outside hyper. I assume this will most often be timing related. Most of the other information can be recorded at the start of an `impl Service`.

OK, the `impl Service` is a new idea - can you elaborate how you see it working here.  Is there prior art I can look at to get my head around what you have in mind.

I take your point about the finangle structure being proven prior-art.
It does still leave the question of what functions get spans/events/logs added, but this isn't abstract, we have something concrete guiding how we proceed - initially at least:

There are two reports of hang behavior that we are blind about.  I'm not even sure if they have the same root cause.

The spans/events required, as a first step, are those that expose the cause of the problem - the finangle guide is one useful way to organize them. However,

> Determine what spans, events, and fields that absolutely need to be inside hyper (

I don't want to get to the point of finding out what spans/events/logs are needed to identify the issue in my repro only to find out that the approach taken is totally unacceptable, e.g. think your `impl Service` but after everything is 'done' - being done in terms of solving the problems we have open issues about.  Where solving could mean establishing the problem lies outside of Hyper (i.e. in `want`, `mio` etc).

davidbarsky:
> OK, the impl Service is a new idea - can you elaborate how you see it working here. Is there prior art I can look at to get my head around what you have in mind.

I apologize, I should've been more clear when I mentioned `tower_http` on Discord. [tower_http's trace module](https://docs.rs/tower-http/0.1.2/tower_http/trace/index.html) allows you to instrument and trace a Hyper server and client by passing through an `impl tower::Service`. I think that's one of the reasons that there's been not-as-much pressure to instrument Hyper _directly_ with tracing, since `tower_http::trace` can handle a surprising amount.

hawkw:
@taqtiqa-mark, re:
> I understand the concern. The issue is: getting debug/trace output in a usable form should be off-the-shelf functionality. IMO, the Hyper instructions need not be more complex than this:
> 
> podman run -p6831:6831/udp -p6832:6832/udp -p16686:16686 jaegertracing/all-in-one:latest
> cargo run --features="some thing tracing/max_level_trace" --example client_json_tracing_otel
> firefox http://localhost:16686
> 
> or some such.
> Can that be done without any dependencies?
> I agree the feature flag is not right as it stands right now. See below for a proposal.
> 
>     * I'd probably lean towards not having any new example files. Maybe one if the existing ones don't do a good job of letting you use tracing, but ideally they all do.
> 
> See @hawkw 's earlier comment on the subtleties of setting up tracing for async. IMO, debug output should just work and not require you to spend more than 2 minutes on setup.
> 
> Adding spans/events, and nothing more, is just throwing the same problem onto every user to keep solving.
> Perhaps we can come up with some boiler plate that will solve 80% of the problems.

I understand your motivation here. The point I've been trying to make, though, is that enabling logging, and determining how it's formatted, is an application-level concern. The general consensus in the Rust library ecosystem is that library crates (almost) _never_ print anything to standard out by default, and don't have opinions about how logs should be formatted. Instead, the `log` and `tracing` crates provide facade layers that allow libraries to emit diagnostics that are collected by a `Logger` or `Subscriber` that's configured in application code. The idea is that the user of a library should always be the one who decides what diagnostics to opt in to, and how they should be recorded. If `hyper` provided its own opinions about how traces are recorded, or about what traces should be enabled by default, it would be diverging from what every other crate that uses the `tracing` and `log` facade layers does.

If users need to enable logging in a new application or in a reproduction of a bug report or something, they can just paste
```rust
tracing_subscriber::fmt::init();
```
into their `main` function to get traces output to stdout using a default format. 

If users developing a production application want to use OpenTelemetry, they can configure [`tracing-opentelemetry`](https://lib.rs/crates/tracing-opentelemetry) to suit their use case. If users want JSON output that's compatible with a particular log processing system, there's a [large](https://lib.rs/crates/tracing-bunyan-formatter) [number](https://lib.rs/crates/tracing-stackdriver) of [crates](https://lib.rs/crates/tracing-elastic-apm) that provide formatters for various JSON formats. If (for some reason) they want to send logs to a Slack channel...[they can do _that_, too](https://lib.rs/crates/tracing-layer-slack). And so on. It's way out of scope for `hyper` to implement any of that functionality. Rather, it's `hyper`'s job to play nice with the rest of the ecosystem and behave the way users expect, by providing opt-in diagnostics like every other Rust library.

Cargo.toml
hawkw: > you can't have a feature name that is congruent to a crate name - but I'm not a Rust expert

An [optional dependency](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies) is equivalent to a feature flag that only enables one dependency. I think that `tracing` should just be an optional dependency and the instrumentation points should all be flagged with `#[cfg(feature = "tracing")]`. If we do end up exposing other functionality, like the OpenTelemetry layer you've implemented, it should be under a separate feature flag.

taqtiqa-mark:
Thanks for engaging - I agree with all of that.  Also to my mind, providing a default does not necessarily constrain you outside of the default case.
However, I do wonder, in this context, about how bright the line is between opinionated and agnosticism on what hyper span output is and what the user does with that output. 

I have in mind only debug and trace level spans, etc. I also think they should be off by default, so not under the `full` feature - see issue #2698 

This comment is why I sought early feedback:

> If users developing a production application want to use OpenTelemetry, they can configure tracing-opentelemetry to suit their use case. 
...\<snip\>... 
Rather, it's hyper's job to play nice with the rest of the ecosystem and behave the way users expect, by providing opt-in diagnostics like every other Rust library.

Buying into OpenTelemetry means you can format for stdout, JSON, Jaeger, OTLP, Prometheus, etc.
However, you are constrained (as much as any convention is a constraint) to OpenTelemetry diagnostics, and its semantic conventions.

I just want to make sure whether you're suggesting, say, a `hyper-opentelemetry` crate?  

In particular how would that work w.r.t. empty fields being populated at runtime - it seems there the OTel details would still leak into the hyper crate?

Cargo.toml
taqtiqa-mark: > the instrumentation points should all be flagged with #[cfg(feature = "tracing")]

What do you think of using two flags `debug` and `trace` or some such.  Where the use cases are distinct, Use `debug` when if you are investigating the integration of your app and hyper.
Use `trace` when you are investigating hyper internals.
The debug spans would then be constrained to around only public functions.  

Not sure if that type of distinction can work out?


> If we do end up exposing other functionality, like the OpenTelemetry layer you've implemented, it should be under a separate feature flag.

Agreed, or probably better to have a separate crate

davidpdrsn:
For a while I've wanted to add support for [OpenTelemetry's conventional span field names](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md), to `tower-http`'s `Trace` middleware. I haven't read the code very closely but from the discussion it feels this might be related, as `Trace` can be used to instrument `hyper` compatible `tower::Service`s. I have opened a draft PR to `tower-http` https://github.com/tower-rs/tower-http/pull/180 if someone is interested.

taqtiqa-mark:
Thanks @davidpdrsn I'll take a look at that.  Right now I think I have enough feedback to make a meaningfully different second iteration to try address some of the issues raised.

src/proto/h1/role.rs
taqtiqa-mark: From memory... this is what led me to add layers, likely incorrectly. I'd come across some blogs that made me think any logic around value processing belonged in a layer.  

For what it is worth: I've removed all layers, and if they are appropriate, we'd need to discuss where they live - a proc-macro-attribute crate doesn't appear to allow modules to be exported - so they don't/can't live in the proc-macro crate I've created as part of the follow-up PR.

src/proto/h1/role.rs
taqtiqa-mark: On further reflection I've followed this suggestion - if the proposed approach in the PR gets rejected it won't be because of this :)

PR #2700, opened 2021-11-19
feat(h2): implement server::Builder::http2_max_send_buf_size

nox:


src/server/conn.rs
```diff
@@ -406,9 +406,13 @@ impl<E> Http<E> {
...
 
-    /// Sets whether HTTP2 is required.
+    /// Set the HTTP/2 maximum send buffer size.
     ///
-    /// Default is false
+    /// Default is ~ 400kb.
+    ///
+    /// # Panics
+    ///
+    /// The minimum value allowed is 8192. This method panics if the passed `max` is less than the minimum.
```
gabi-250, line 415: I think you need to keep the original documentation here.

nox: You are correct, and I have no damn clue how I ended up changing the wrong docs hah.

seanmonstar:
We went with a slightly different solution in `h2`, and just merged config support for hyper in #2718.

PR #2706, opened 2021-11-23
fix(http1): return 414 when URI contains more than 65534 characters

### LINKED ISSUE #2701: When the URL contains more than 65534 characters, the error code returned by Hyper is not 414.
The length of the from_shared method of the third-party library http::Uri is verified， The maximum length is (u16::MAX - 1).

rajing:
Previous behavior returned a 404 Bad Request. Conforms to HTTP 1.1 RFC.

Closes #2701

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -71,6 +71,7 @@ pub(super) enum Parse {
     #[cfg(feature = "http1")]
     VersionH2,
     Uri,
+    UriTooLong,
     Header(Header),
     TooLarge,
     Status,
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -71,6 +71,7 @@ pub(super) enum Parse {
     #[cfg(feature = "http1")]
     VersionH2,
     Uri,
+    UriTooLong,
```
seanmonstar, line 74: Perhaps the `Error::is_parse_too_large()` should return true for this variant as well?

rajing: Thanks for the feedback. Pull request has been updated with the changes.

rajing: Sorry, forgot to add the file. Pull request has now been updated.

src/error.rs
```diff
@@ -152,7 +153,7 @@ impl Error {
 
     /// Returns true if this was an HTTP parse error caused by a message that was too large.
     pub fn is_parse_too_large(&self) -> bool {
-        matches!(self.inner.kind, Kind::Parse(Parse::TooLarge))
+        matches!(self.inner.kind, Kind::Parse(Parse::TooLarge) | Kind::Parse(Parse::UriTooLong))
```
seanmonstar, line 156:
```suggestion
        matches!(
            self.inner.kind,
            Kind::Parse(Parse::TooLarge) | Kind::Parse(Parse::UriTooLong)
        )
```

seanmonstar (approved):

Issue #2708, opened 2021-11-30
"Expect: 100-continue" header is using case-sensitive comparison


JonathanMurray:
**Version**
0.14 (and master-branch 2021-11-30)

**Platform**
Darwin C02Y72T1JGH5 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64

**Description**
I have a service built with Hyper that I'm now trying to communicate with from a client written in Java. I noticed when doing this that Java's standard HttpClient sends 100-continue expectations as `Expect: 100-Continue` (https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java#L408), whereas Hyper compares the expectation header value against `100-continue` (case-sensitive) (https://github.com/hyperium/hyper/blob/5f938fffa64df23a2e4af81ed4e6d8bd760e2d05/src/proto/h1/role.rs#L274). This has the effect that the Java client never receives a `100 Continue` from the Rust service, which makes it seemingly unable to handle the response correctly. https://datatracker.ietf.org/doc/html/rfc2616#section-14.20 seems to state that the capitalisation of the `100-continue` token should not matter from the server's perspective:
```
Comparison of expectation values is case-insensitive for unquoted
   tokens (including the 100-continue token), and is case-sensitive for
   quoted-string expectation-extensions.
```

Trying this:

changing the `expect_continue_sends_100` test to send the request expectation with an upper-case `C`. 
```
diff --git a/tests/server.rs b/tests/server.rs
index 13530a36..0b991337 100644
--- a/tests/server.rs
+++ b/tests/server.rs
@@ -808,7 +808,7 @@ fn expect_continue_sends_100() {
         b"\
         POST /foo HTTP/1.1\r\n\
         Host: example.domain\r\n\
-        Expect: 100-continue\r\n\
+        Expect: 100-Continue\r\n\
         Content-Length: 5\r\n\
         Connection: Close\r\n\
         \r\n\
```
and running `cargo test --features full expect_continue_sends_100`

I expect to see this happen: 

test passes

Instead, this happens: 

The test fails with:
```
---- expect_continue_sends_100 stdout ----
[...]

PR #2709, opened 2021-11-30
fix(server): use case-insensitive comparison for Expect: 100-continue

JonathanMurray:
According to rfc2616#section-14.20 the header value is case-
insensitive. Certain clients (including Java's std lib
https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java#L408)
send the expectation as `100-Continue` and this should be handled
by the server.

(This is an attempt to address https://github.com/hyperium/hyper/issues/2708)

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -274,7 +274,7 @@ impl Http1Transaction for Server {
                     // According to https://datatracker.ietf.org/doc/html/rfc2616#section-14.20
                     // Comparison of expectation values is case-insensitive for unquoted tokens
                     // (including the 100-continue token)
-                    expect_continue = value.to_str().map(|s| s.eq_ignore_ascii_case("100-continue")).unwrap_or(false);
+                    expect_continue = value.as_bytes().eq_ignore_ascii_case(b"100-continue");
                 }
                 header::UPGRADE => {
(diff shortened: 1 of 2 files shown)


src/proto/h1/role.rs
```diff
@@ -271,7 +271,10 @@ impl Http1Transaction for Server {
                     }
                 }
                 header::EXPECT => {
-                    expect_continue = value.as_bytes() == b"100-continue";
+                    // According to https://datatracker.ietf.org/doc/html/rfc2616#section-14.20
+                    // Comparison of expectation values is case-insensitive for unquoted tokens
+                    // (including the 100-continue token)
+                    expect_continue = value.to_str().map(|s| s.eq_ignore_ascii_case("100-continue")).unwrap_or(false);
```
paolobarbolini, line 277: `eq_ignore_ascii_case` should also available with `.as_bytes()`
See: https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.eq_ignore_ascii_case

JonathanMurray: Ah, that's great! Using that instead now.

seanmonstar (approved):

seanmonstar:
Thank you!

PR #2710, opened 2021-11-30
refactor(lib): resolve unused warnings from server-specific code

seanmonstar:


src/proto/h1/mod.rs
```diff
@@ -1,4 +1,6 @@
+#[cfg(all(feature = "server", feature = "runtime"))]
 use std::pin::Pin;
+#[cfg(all(feature = "server", feature = "runtime"))]
 use std::time::Duration;
```
olix0r, line 4: tioli

```suggestion
#[cfg(all(feature = "server", feature = "runtime"))]
use std::{pin::Pin, time::Duration};
```

PR #2711, opened 2021-12-02
feat(error): add additional methods to determine kind

omjadas:
Add additional methods to determine error kind.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -41,10 +41,9 @@ pub(super) enum Kind {
     #[cfg(all(feature = "tcp", feature = "server"))]
     Listen,
     /// Error accepting on an Incoming stream.
-    #[cfg(any(feature = "http1", feature = "http2"))]
-    #[cfg(feature = "server")]
+    #[cfg(all(any(feature = "http1", feature = "http2"), feature = "server"))]
     Accept,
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! I'm generally in favor of having accessor methods to inspect the error more. However, I've also been thinking the current `Kind` enum variants are a little disorganized, and have thought it may warrant an internal cleanup. Are there any variants that you specifically needed to know, or were you just wanting to add methods for all variants?

omjadas:
> Thanks for the PR! I'm generally in favor of having accessor methods to inspect the error more. However, I've also been thinking the current `Kind` enum variants are a little disorganized, and have thought it may warrant an internal cleanup. Are there any variants that you specifically needed to know, or were you just wanting to add methods for all variants?

The ones that I am particularly interested are `Io` and `Shutdown`.

omjadas:
@seanmonstar Just coming back to this. I would be happy to update this PR to only add methods for the `Io` and `Shutdown` error kinds (assuming you are happy with those two) and potentially any others you think should be exposed as well.

seanmonstar:
We'll be going in a different direction, adding a [`Code`](#2845).

PR #2716, opened 2021-12-08
Add fuzzing harness to fuzz a single request parse

5225225:
This *might* be somewhat redundant seeing as both `h2` and `httparse` are fuzzed independently, but it's probably worth fuzzing the code here.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for the PR! I am generally in favor of fuzzing, but at the same time I've found fuzzing crates to be kind of volatile (they seem to break compiling somewhat frequently). I'd also prefer not to depend on `async-io` for this. I'm going to close here for now.

PR #2727, opened 2021-12-28
refactor(client) use Box<str> in ..::dns::Name

monoid:
Use Box<str> in hyper::client::connect::dns::Name, so
its size is 16 bytes, not 24 bytes.  As Name never
change its contents, read-only Box<str> is perfectly OK.

### DIFF
--- a/src/client/connect/dns.rs
+++ b/src/client/connect/dns.rs
@@ -38,7 +38,7 @@ pub(super) use self::sealed::Resolve;
 /// A domain name to resolve into IP addresses.
 #[derive(Clone, Hash, Eq, PartialEq)]
 pub struct Name {
-    host: String,
+    host: Box<str>,
 }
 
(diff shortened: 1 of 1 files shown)


src/client/connect/dns.rs
```diff
@@ -85,7 +85,7 @@ impl FromStr for Name {
 
     fn from_str(host: &str) -> Result<Self, Self::Err> {
         // Possibly add validation later
-        Ok(Name::new(host.to_owned()))
+        Ok(Name::new(host.to_boxed_str()))
```
seanmonstar, line 88:
```suggestion
        Ok(Name::new(host.into()))
```

monoid: Indeed. Updated.

seanmonstar (approved):
Thanks!

Issue #2732, opened 2022-01-07
`hyper::Error` should either print `source` in display xor return a `Some` from `Error::cause()`, not both


nagisa:
**Version**
f1b89c117cffebed4b2b8eb2d221fd9b25c1d3d1

**Platform**
Any

**Description**
Today the `std::fmt::Display` impl for `hyper::Error` has the following snippet:

https://github.com/hyperium/hyper/blob/f1b89c117cffebed4b2b8eb2d221fd9b25c1d3d1/src/error.rs#L500-L501

and the `std::error::Error::source` impl for `hyper::Error` reads as such:

https://github.com/hyperium/hyper/blob/f1b89c117cffebed4b2b8eb2d221fd9b25c1d3d1/src/error.rs#L510-L513

This is somewhat problematic for code that properly prints out the error causation chain. e.g. I have seen an error like this:

```
error: error trying to connect: tcp connect error: Connection refused (os error 111)
caused by: tcp connect error: Connection refused (os error 111)
caused by: Connection refused (os error 111)
```

this is pretty unreadable. Now one might argue that `hyper::Error::description()` serves this use-case, and while that's true, it seems pretty awkward to pull in hyper into the code that's implements error presentation layer just to downcast and check for `hyper`.

seanmonstar:
See also #2542

seanmonstar:
I should probably write up some thoughts as to why this is currently the way it is.

Wrapping an error (so that there's an inner source) frequently provides more context that is useful to users. Without showing the source, the extra context can lose some of its meaning. For example, seeing just "error trying to connect" is also too vague, just as "Connection refused" can be. Most users of hyper tend to just log the `Display` output of the error, without anything fancy. So the format for hyper is to default to be most useful to them.

I recognize that makes it more verbose for people using utilities to crawl the source chain. I currently chose to cater to the more common case.

I'm also aware that the "error working group" currently recommends against doing what hyper does. But I don't agree, since it means hurting the common case. And since I believe most users will generally just default to logging `{}` of an error, I don't think "more education" will fix that. Thus, I'd suggest that the WG consider instead that errors print their source chain by default, and allow advance users to opt-in to custom formatting. I explored this concept in https://github.com/seanmonstar/errors/issues/1, and have a not-published crate that does this, even...

nox:
@seanmonstar What do you think of making `{:#}` not print the source parts?

PR #2737, opened 2022-01-18
feat(error): add `Error::message`

davidpdrsn:
This is an alternate solution to https://github.com/hyperium/hyper/issues/2732 that doesn't change `hyper::Error`'s `Display` impl (like https://github.com/hyperium/hyper/pull/2542 does).

It adds `Error::message` which returns the message that is unique to the error, without the message from the source. That way users can create a newtype around `hyper::Error` and use this in the `Display` impl to work around https://github.com/hyperium/hyper/issues/2732.

Fixes https://github.com/hyperium/hyper/issues/2732

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -396,6 +396,11 @@ impl Error {
         }
     }
 
+    /// The error's standalone message, without the message from the source.
+    pub fn message(&self) -> impl fmt::Display + '_ {
+        self.description()
+    }
+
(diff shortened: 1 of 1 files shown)


nox (approved):

seanmonstar (approved):

davidpdrsn:
@seanmonstar any chance you wanna cut a release with this?

PR #2745, opened 2022-01-31
feat(errors): add Error::is_shutdown

nox:


nox:
I want to be able to not log shutdown-related errors, any objection @seanmonstar?

src/error.rs
```diff
@@ -56,7 +56,6 @@ pub(super) enum Kind {
     /// The body write was aborted.
     BodyWriteAborted,
     /// Error calling AsyncWrite::shutdown()
-    #[cfg(feature = "http1")]
```
olix0r, line 59: Why change the feature-gate here? Would it be better to clarify the helper?

src/error.rs
```diff
@@ -201,6 +200,11 @@ impl Error {
         self.find_source::<TimedOut>().is_some()
     }
 
+    /// Returns true if the error was caused by a shutdown.
+    pub fn is_shutdown(&self) -> bool {
```
olix0r, line 204: e.g.

```suggestion
    #[cfg(feature = "http1")]
    pub fn is_http1_shutdown(&self) -> bool {
```

Either that, or this should probably also handle h2 goaways?

seanmonstar: Hm, that does make me think the name could confuse people. It's not a graceful shutdown, it was an error that occurred when calling `AsyncWrite::shutdown`.

It originally was just meant to provide additional context on top of the IO error. Is that context useful on its own?

nox: I don't understand what you mean. There is nothing we can do about those errors in the wild (most often they are a "broken pipe" error), so we don't want to log them at all, so we need to discriminate them to ignore them.

seanmonstar: So, it does seem like the context is useful to you. Now the question is, is the name confusing, such that someone might think it's a "graceful shutdown" error received over HTTP/2? Is that a bad thing? Should we hijack the name to make both cases be a `is_shutdown` error?

nox: h2 errors currently are all represented as `Kind::H2` right? Maybe I should rename that method to `is_http1_shutdown` then.

seanmonstar:
I'm thinking we close this?

nox:
> I'm thinking we close this?

Is there a replacement for it?

joshka:
> Is there a replacement for it?

I just raised https://github.com/hyperium/hyper/pull/3863 as a possible replacement.

Issue #2385, opened 2021-01-03
instant.elapsed() will panic


silence-coding:
this method is panic.  hyper::client::pool::702
![image](https://user-images.githubusercontent.com/32766901/103480903-4c374300-4e12-11eb-942f-8e69ac939081.png)

![image](https://user-images.githubusercontent.com/32766901/103480922-70931f80-4e12-11eb-8187-6cdac9d5f52c.png)

silence-coding:
![image](https://user-images.githubusercontent.com/32766901/103481245-ae914300-4e14-11eb-939e-1be709ffe10b.png)

seanmonstar:
Hm. This should never happen, sounds like a bug in libstd.

silence-coding:
I don't think of how to trigger the scene yet, which seems unlikely to happen.

![image](https://user-images.githubusercontent.com/32766901/103497321-22b30180-4e7c-11eb-9c87-34708744275a.png)

silence-coding:
Is the obtained CLOCK_MONOTONIC in multi-core CPU scenarios incorrect?

malaire:
What OS?
`std::time::Instant` contains code to fix non-monotonic monotonic clocks but it's not used on some platforms which have so far worked without it: https://doc.rust-lang.org/src/std/time.rs.html#224-264

silence-coding:
Whether to replace elapsed with saturation_duration_since？

yonip23:
I came across a similar problem once, I think the solution is to use the [chrono](https://docs.rs/chrono/0.4.19/chrono/) crate instead.

silence-coding:
> Whether to replace elapsed with saturation_duration_since？

@seanmonstar Whether hyper will fix this issue, if it is waiting for std to fix it, then I will close the issue.

seanmonstar:
Fixed in #2746

PR #2746, opened 2022-01-31
Avoid panics in uses of `Instant`

olix0r:
We have reports of runtime panics (linkerd/linkerd2#7748) that sound a
lot like rust-lang/rust#86470. We don't have any evidence that these
panics originate in hyper, but #2385 reports a similar issue.

Even though this is almost definitely a bug in Rust, it seems most
prudent to actively avoid the uses of `Instant` that are prone to this
bug.

This change replaces uses of `Instant::elapsed` and `Instant::sub` with
calls to `Instant::saturating_duration_since` to prevent this class of
panic.

### DIFF
--- a/src/client/pool.rs
+++ b/src/client/pool.rs
@@ -458,7 +458,9 @@ impl<T: Poolable> PoolInner<T> {
                     trace!("idle interval evicting closed for {:?}", key);
                     return false;
                 }
-                if now - entry.idle_at > dur {
+
+                // Avoid `Instant::sub` to avoid issues like rust-lang/rust#86470.
+                if now.saturating_duration_since(entry.idle_at) > dur {
                     trace!("idle interval evicting expired for {:?}", key);
                     return false;
                 }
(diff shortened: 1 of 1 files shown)


src/client/pool.rs
```diff
@@ -721,7 +723,8 @@ impl Expiration {
 
     fn expires(&self, instant: Instant) -> bool {
         match self.0 {
-            Some(timeout) => instant.elapsed() > timeout,
+            // Avoid `Instant::elapsed` to avoid issues like rust-lang/rust#86470.
+            Some(timeout) => Instant::now().saturating_duration_since(instant) > timeout,
```
seanmonstar, line 727: It definitely feels like `Instant::elapsed()` should handle this, or otherwise it's not really safe to use. Looks like https://github.com/rust-lang/rust/pull/89926 agrees.

olix0r: Yeah, I agree. It will probably take at least a few months for those changes to be available in a stable release (the RFC still appears to be open). This change would eliminate these problems well in advance of that, even for projects that are not on the latest stable Rust version.

Is there something I can do to make the temporary nature of this hack/fix better documented?

PR #2749, opened 2022-02-04
Include connection info in Client::send_request errors

nox:


src/client/client.rs
```diff
@@ -242,7 +242,7 @@ where
             if req.version() == Version::HTTP_2 {
                 warn!("Connection is HTTP/1, but request requires HTTP/2");
                 return Err(ClientError::Normal(
-                    crate::Error::new_user_unsupported_version(),
+                    crate::Error::new_user_unsupported_version().with_client_connect_info(pooled.conn_info.clone()),
```
seanmonstar, line 245: What specific info would you want to collect? Would it be feasible if we included some of that info in the `warn!`, above? Or perhaps right next to it, as a `trace!`?

nox: No, we have our own logging directives from the errors we get and create, we never use the `warn`/`trace` stuff from Hyper itself because we don't control it. We need to be able to access that data from the error itself.

The specific info we need comes from the `Connected` value of the streams returned by our own connector, so logging in Hyper itself wouldn't help anywya.

taqtiqa-mark: > ... we never use the warn/trace stuff from Hyper itself because we don't control it.

I'm wondering if the current proposal to add a tracing instrumentation abstraction layer would be enough to bring Hyper spans and traces within your control.
To elaborate, the proposal is to add something like this (`client_send` is vapour-ware):

```rust
#[client_send(level=tracing::Level::Trace, name="Client::send_request", skip=[self,req])]
async fn send_request(
```

Hyper might default to using some proc-macro crate that covers the 80-20 rule. I wonder if, longer term, it wouldn't be possible for users to develop their own span/trace handling crates.

Admittedly, this would mean developing a bespoke proc-macro crate for your needs.  
Is this nuts or could it be useful?

nox: I don't think this would be useful. As stated before, even if we had access to the spans and traces, we also need access to the extra data stored in the connection's `Connected` value.

src/error.rs
```diff
@@ -206,9 +211,20 @@ impl Error {
         self.inner.cause
     }
 
+    /// Returns the info of the client connection on which this error occurred.
+    #[cfg(all(feature = "client", any(feature = "http1", feature = "http2")))]
+    pub fn client_connect_info(&self) -> Option<&Connected> {
```
seanmonstar, line 216: I realize they're not implemented yet, but this made me think of the errors RFCs about "generic context" access, and would seem like a good fit eventually. https://github.com/rust-lang/rfcs/pull/2895 and https://github.com/rust-lang/rfcs/pull/3192

seanmonstar (commented):
So, I don't know yet what the best outcome here is. I think including some info in the error is probably a good idea. I think there's some prior art in other libraries, and some potential new features in `libstd` that we may wish to consider, before refining what exactly we expose.

src/client/client.rs
taqtiqa-mark: > .... even if we had access to the spans and traces, we also need access to the extra data stored in the connection's `Connected` value

Understood, I believe you would.  

In your `client_send` you'd inspect for `send_request`, and then do as your PR here does replace:
```rust
crate::Error::new_user_unsupported_version(),
```

with
```rust                 crate::Error::new_user_unsupported_version().with_client_connect_info(pooled.conn_info.clone())
```

It might be possible for Hyper to aid this type of use with some place holder to search for , etc.  But even as it stands now this should be possible?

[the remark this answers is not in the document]
taqtiqa-mark: @seanmonstar, not familiar with finagle... Can you point to the library doc or impl. Also would be informative to know an example use you considered best practice or at least idiomatic.
Context: I'm refactoring the traceable errors in my tracing PR and this seems on-point.

nox: I'm sorry I don't understand what you are getting at, but I'll note we are not "waiting for a PR to get accepted", we are using this in production already.

nox:
@seanmonstar Ping on this?

seanmonstar:
I think including this info in the error somehow is a good goal. We should do it. I want to figure out _how_ exactly that info should be exposed, though. As part of 1.0, the pooling `Client` will likely get moved out and broken up, with most pieces probably in `hyper-util`. So, how do we expose this info?

- We could have a unique error type coming from the pooling client, which is essentially `enum SendError<E> { Connect(E), Http(hyper::Error) }`. In that case, the info could just be part of the generic "connect" error.
- Or we could reuse `hyper::Error`, having the connect variant be internal somehow. In that case, how do we attach this useful extra information to the `hyper::Error`? I could imagine we want to attach other useful things in the future, so what mechanism should that use?

nox:
IMO for now 2nd way is good, as the client still lives in Hyper. Would you mind we merge this PR as is and see how we do it for 1.0 later?

seanmonstar:
So, I *could* just click merge, but it will most likely be different with 1.0 (and that's starting now-ish). How do you feel now?

nox:
SGTM, that’s at least one less patch to maintain on my side until 1.0 :)

nox:
@seanmonstar Would you merge it if I rebase it again?

seanmonstar:
Sure, we can just merge it into 0.14, while noting that a more generic mechanism is probably better for hyper-util (in a hyper 1.0 world).

seanmonstar (approved):

PR #2751, opened 2022-02-10
feat(server): add an option to disable adding the `DATE` header

deatondg:
Add an option to disable the automatic addition of the `DATE` header for
situations when an accurate time is not available or the `DATE` header
is otherwise undesirable, for example communicating with a non-compliant
client.

According to RFC7231#7.1.1.2, 'An origin server MUST NOT send a Date
header field if it does not have a clock capable of providing a
reasonable approximation of the current instance in Coordinated
Universal Time.' Otherwise, sending the date is required.

Resolve the now closed issue hyperium#912.

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -59,6 +59,8 @@ where
                 h1_header_read_timeout_running: false,
                 preserve_header_case: false,
                 title_case_headers: false,
+                #[cfg(feature = "server")]
+                add_date: true,
                 h09_responses: false,
                 #[cfg(feature = "ffi")]
                 on_informational: None,
@@ -109,6 +111,11 @@ where
 
     pub(crate) fn set_preserve_header_case(&mut self) {
         self.state.preserve_header_case = true;
+    }
+
+    #[cfg(feature = "server")]
+    pub(crate) fn disable_add_date(&mut self) {
+        self.state.add_date = false;
     }
 
     #[cfg(feature = "client")]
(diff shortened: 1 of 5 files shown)


seanmonstar:
Thanks for the PR! I think providing this is probably a good idea... Do you have a use case that needs this now?

NateLing:
> Thanks for the PR! I think providing this is probably a good idea... Do you have a use case that needs this now?

@seanmonstar i need this. when i use http proxy server (like nginx), i need not response the `DATE` header from hyper.  The `date` header is not required for inter-application communication. This may improve performance.

diceride:
@NateLing Thank you for working on this. I added some comments. I tried this out and it improved the performance on low-end environments.

@seanmonstar What do you think?

diceride:
@NateLing Thank you, I was looking for this. After trying it out, the performance improved on low-end systems as expected. I added a comment to your changes, it is just an idea about the wording of the option. Let me know what you think.

Black-Hack:
Hello, I would really like for this option to be added. Is there anything blocking this? If so maybe I can help.

seanmonstar:
I think this just got forgotten. Looking it over, it seems like a fine addition. I think:

- Perhaps it could be publicly called `auto_date()`?
- Resolve conflicts.
- Add to 0.14.x and to master.

darren-fu:
any news on this?

tusharmath:
This is a must have configuration. For small payloads, having vs not-having the date header can make a massive difference in performance > 100%. Would love to have this merged 🙏

RyanAD:
This would also be useful when using as an HTTP Proxy. After a CONNECT request not all clients expect the Date header.

darren-fu:
> This would also be useful when using as an HTTP Proxy. After a CONNECT request not all clients expect the Date header.

agree

edwardwc:
@seanmonstar will do

Black-Hack:
I think this issue can be closed now, since #3644 is merged. Thank you @edwardwc and @seanmonstar

PR #2752, opened 2022-02-10
fix(http1): make `sec-websocket-` title case as `Sec-WebSocket-`

deatondg:
When using the `http1_title_case_headers` option, "correctly" title-case
the WebSocket headers `sec-websocket-*` to `Sec-WebSocket-*`, which is
the casing used in RFC6455. Previously, these headers were title-cased
to `Sec-Websocket-*`. It is likely that non-compliant case-sensitive
programs, such as the iOS Roku remote app, will expect the new behavior.
The ideal solution is to eventually expose `HeaderCaseMap`, but this is
a reasonable stopgap solution.

Arguably, this is not actually a fix but just a change. However, I thought
I might as well attempt to get this merged as this could prove useful. A 
search on [grep.app](https://grep.app) suggests that `Sec-WebSocket-` is 
about 5x more common than `Sec-Websocket-`, but this doesn't actually prove
anything about whether non-compliant programs are more likely to expect my
suggested behavior or the existing behavior.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -1354,11 +1354,27 @@ fn record_header_indices(
     Ok(())
 }
 
+// Special constants for title-casing sec-websocket-*
+// Maybe overkill, but we have compile-time length guarantees.
+const SEC_WEBSOCKET_LENGTH: usize = 14;
+const SEC_WEBSOCKET_LOWER_CASE: &[u8; SEC_WEBSOCKET_LENGTH] = b"sec-websocket-";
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! That is a sad situation indeed... I'm curious, are there currently situations that are not working because of this behavior in hyper?

nox:
I would rather prefer we expose `HeaderCaseMap` somehow.

seanmonstar:
I agree, I'd rather expose the ability to customize casing more directly. Closing for now.

PR #2755, opened 2022-02-10
make mod as pub

jkvargas:
So, the idea here is to be able to use MakeServiceRef and MakeServiceFn outside of the api scope.
:)


jkvargas:
A bit weird, my commit message does not have 72 characters.

PR #2758, opened 2022-02-13
feat(server): Port encode_headers trace_span

### LINKED ISSUE #2742: RFC Hyper tracing attributes: proc-macro-attributes crate
### Updates:
1. 3-Feb-2022:
  - OpenTelemetry semantic conventions for HTTP spans. 
  - Rationalize proposed `trace_*` attributes.
  - Update examples to current PoC implementation

### RFC

Hyper tracing, per issue #2678.

This RFC is concerning the semantics of setting up tracing in a Hyper application.

Specifically, out of scope are the particular values for `field(...)`, `skip(...)`, `name` and `target` parameters in the Tracing `instrument` attribute.

A particular goal is to flexibly compose the tracing `#[instrument(...)]` attribute

Built in tracing support that can be compiled out when not needed - zero runtime penalty when not used.
Yet is flexible enough to (hopefully) satisfy the users of Hyper, and the Hyper project itself.

See PR #2697.  Having abandoned several aspects such as layers.

**Additional context**

These Hyper tracing proc-macro-attributes would be behind a `cfg_attr` guard  but that is not relevant right now.

I have a proc-macro crate `hyper-tracing-attributes` that has the following functionality implemented to the point where I'm confident (code generation) the functionality proposed can be achieved.

The proposal is:

1. Default use of [OpenTelemetry semantic conventions for HTTP spans](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/) fields.
1. A ~~hyper~~`http-tracing-attributes` crate: I don't anticipate needing consumer/subscriber or layer specific attributes, but if they do end up being needed, one could use e.g. a ~hyper~`http-tracing-<layer>-attributes` crate etc.  In fact there is nothing Hyper specific (yet) about this crate and it could by used to generate Tracing `#[instrument(...)]` attributes for any HTTP oriented code
    - ~should it be named more generically as `http-tracing-attributes` crate?~
1. `server_send`, etc naming per the finagle [standard annotations](https://twitter.github.io/finagle/guide/Tracing.html#standard-annotations)
2. attributes to ease composition:
   - MVP: `trace_field`
   - ~`trace_name`~
   - Post-MVP: `trace_target`

````rust
    #[server_send(level = ~trace~ Level::INFO, name = "Server::Encode", skip = [dst,msg])]
    #[other]
    #[trace_field(some.thing = tracing::field::Empty)]
    #[trace_field(one.more = "thing")]
    #[another]
    ~#[trace_name("neater")]~
    #[trace_target("bulls-eye")]
    fn encode(msg: Encode<'_, Self::Outgoing>, dst: &mut Vec<u8>) -> crate::Result<Encoder> {
        ...
        if cfg!(feature = "tracing") {
            tracing::Span::current().record("some.thing", "else");
        }
    }
````

This would generate

````rust
    #[other]
    #[another]
    #[instrument( level = "trace",
                    name = "neater",
                    target = "bulls-eye"
                    skip(dst,msg),
                    fields(// Custom
                            some.thing = tracing::field::Empty,
                            one.more   = "thing",
                            // Standardized
                            otel.name           = "Server::Encode",
                            otel.kind           = ?SpanKind::Server,
                            otel.status_code    = ?opentelemetry::trace::StatusCode::Unset,
                            otel.status_message = tracing::field::Empty,
                            // OTel required (HTTP)
                            http.host     = tracing::field::Empty,
                            http.method   = tracing::field::Empty,
                            http.scheme   = tracing::field::Empty,
                            http.target   = tracing::field::Empty,
                            http.url      = tracing::field::Empty,
                            net.peer.ip   = tracing::field::Empty,
                            net.peer.name = tracing::field::Empty,
                            net.peer.port = tracing::field::Empty,
                            // OTel optional (HTTP)
                            http.flavor                 = tracing::field::Empty,
                            http.request_content_length = tracing::field::Empty,
                            // OTel optional (General)
                            net.transport = "IP.TCP",
                        )
                    )
                )]
    fn encode(msg: Encode<'_, Self::Outgoing>, dst: &mut Vec<u8>) -> crate::Result<Encoder> {
        ...
        if cfg!(feature = "tracing") {
            tracing::Span::current().record("some.thing", "else");
        }
    }
````

### PR DESCRIPTION
**Pull request by taqtiqa-mark**
Close #2742 (at least in terms of settling on a general approach)

TLDR;

```bash
podman run -p6831:6831/udp -p6832:6832/udp -p16686:16686 jaegertracing/all-in-one:latest
cargo run --features="full traceable tracing/max_level_trace" --example echo_traceable
```
In a separate console:
```bash
curl localhost:3000/echo -XPOST -d 'hello world'
curl localhost:3000/echo -XPOST -d 'hello world 10'
curl localhost:3000/echo -XPOST -d 'hello world 100'
firefox http://localhost:16686
```

Port encode_headers from `trace_span!(...)` to `#[instrument(...)]` attribute.

There are currently only two tracing spans in Hyper:
- parse_headers
- encode_headers

The current spans cannot be compiled out.
These changes allow tracing spans and events to be included/excluded
 at compile time.

The tracing span is created using the tracing `#[instrument(...)]` attribute.
The tracing instrument attribute is generated by a proc-macro-attribute,
`#[server_send(...)]` ([named per the finagle standard annotations](https://twitter.github.io/finagle/guide/Tracing.html#standard-annotations)), provided by the proc-macro crate [`tracing-attributes-http`](https://docs.rs/tracing-attributes-http/latest/tracing_attributes_http/).

The tracing span **field names** adopt the OpenTelemetry HTTP semantic conventions. Behaviors that rely on the OpenTelemetry crate are optional, behind the flag `otel`.
Otherwise, the proc-macro crate is generic.


taqtiqa-mark:
Recording here some objections voiced on Discord:

- **OpenTelemetry crate dependency:**  This arises dues to some OTel helper functions being used.  It would be possible to refactor such that the OpenTelemetry crate dependency is eliminated.
- **The use of the `otel.*` namespace in fields/tags:**  Somewhat related to the above.  These too can be eliminated by removing the tracing span fields:
  - `otel.name`
  - `otel.kind`
  - `otel.status_code`
  - `otel.status_message`
  - `otel.library.name`

Cargo.toml
```diff
@@ -109,6 +115,13 @@ tcp = [
     "tokio/time",
 ]
 
+# Trace functionality
+traceable = [
```
taqtiqa-mark, line 119: Is there appetite for "level" features, e.g. 'trace`, `debug`, `warn`. Such that one could conceivable build Hyper with zero tracing (runtime overhead), full tracing or anything in between.  The costs would be one off, at compile time, but they don't seem to be too large.

taqtiqa-mark: We currently have features such as `client` and `server`.  

Those feature flags could be leveraged here too.

Furthermore, if we have features `incoming` and `outgoing` it would be possible to get quite targeted tracing data - avoiding being swamped by trace data.  This would in turn allow us to be more liberal in what is traced.

Example:

```bash
cargo run --features="traceable server incoming" --example echo_traceable
```

taqtiqa-mark:
> * **OpenTelemetry crate dependency:**  ....
> * **The use of the `otel.*` namespace in fields/tags:**  ...

These will be opt-in via a feature: `otel` that is off by default.

Cargo.toml
```diff
@@ -42,13 +41,18 @@ want = "0.3"
 # Optional
 
 libc = { version = "0.2", optional = true }
+opentelemetry = { version = "0.17.0", features = ["rt-tokio", "trace"], optional = true }
```
taqtiqa-mark, line 44: this is now removed as a dependency.

Cargo.toml
```diff
@@ -42,13 +41,18 @@ want = "0.3"
...
+tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
+tracing-attributes-http = { version = "0.1", default-features = false, optional = true }
 
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
 matches = "0.1"
 num_cpus = "1.0"
+opentelemetry-jaeger = { version = "0.16", features = ["reqwest_collector_client","rt-tokio"]}
 pretty_env_logger = "0.4"
+reqwest = {version = "0.11"}
```
taqtiqa-mark, line 55: These are no longer dependencies.

Cargo.toml
```diff
@@ -66,6 +70,8 @@ tokio = { version = "1", features = [
 tokio-test = "0.4"
 tokio-util = { version = "0.6", features = ["codec"] }
 tower = { version = "0.4", features = ["make", "util"] }
+tracing-opentelemetry = "0.17.0"
+tracing-subscriber = { version = "0.3", features = ["std", "env-filter"]}
```
taqtiqa-mark, line 74: No longer dependencies.

Cargo.toml
```diff
@@ -109,6 +115,13 @@ tcp = [
     "tokio/time",
 ]
 
+# Trace functionality
+traceable = [
+    "opentelemetry",
```
taqtiqa-mark, line 120: This has been removed.

Cargo.toml
```diff
@@ -141,11 +154,26 @@ name = "client_json"
 path = "examples/client_json.rs"
 required-features = ["full"]
 
+[[example]]
```
taqtiqa-mark, line 157: Moved to the `tracing-attributes-http` crate.

src/lib.rs
```diff
@@ -86,11 +86,16 @@ pub mod upgrade;
 #[cfg(feature = "ffi")]
 pub mod ffi;
 
+#[cfg(feature = "traceable")]
+#[macro_use]
+extern crate tracing;
+#[cfg(feature = "traceable")]
+extern crate opentelemetry;
+
```
taqtiqa-mark, line 94: Dependency removed.

src/lib.rs
```diff
@@ -86,11 +86,16 @@ pub mod upgrade;
 #[cfg(feature = "ffi")]
 pub mod ffi;
 
+#[cfg(feature = "traceable")]
+#[macro_use]
+extern crate tracing;
```
taqtiqa-mark, line 91: Don`t use `extern crate`

src/proto/h1/role.rs
```diff
@@ -320,14 +320,28 @@ impl Http1Transaction for Server {
...
+            let s = span_setup(&span, &msg);
+            s.record(
+                "http.status_code",
+                &format!("{:?}", msg.head.subject).as_str(),
+            );
+            s.record(
+                "http.response_content_length_uncompressed",
+                &format!("{:?}", msg.body.as_ref().unwrap()).as_str(),
+            );
+        }
```
taqtiqa-mark, line 344: This should be behind a inner proc-macro.

src/proto/h1/role.rs
```diff
@@ -342,7 +356,18 @@ impl Http1Transaction for Server {
...
+                span.as_ref().unwrap().record(
+                    "otel.status_code",
+                    &format!("{:?}", opentelemetry::trace::StatusCode::Error).as_str(),
+                );
+                span.as_ref().unwrap().record(
+                    "otel.status_message",
+                    &format!("ERROR: response with 1xx status code not supported.").as_str(),
+                );
+            }
+
```
taqtiqa-mark, line 370: This will be replaced by a proc-macro-attribute or a proc-macro-function, such as `traceable_error(...)` or `traceable_error!(...)`.  Still working out a PoC.

@seanmonstar, any preference you have at this point?

src/proto/h1/role.rs
```diff
@@ -359,16 +384,51 @@ impl Http1Transaction for Server {
         let init_cap = 30 + msg.head.headers.len() * AVERAGE_HEADER_SIZE;
         dst.reserve(init_cap);
         if msg.head.version == Version::HTTP_11 && msg.head.subject == StatusCode::OK {
+            #[cfg(feature = "traceable")]
+            span.as_ref().unwrap().record("http.flavor", &1.1);
```
taqtiqa-mark, line 388: This will be replaced by a proc-macro-function, such as `record_field!("http.flavor", &1.1)`

src/proto/h1/role.rs
```diff
@@ -359,16 +384,51 @@ impl Http1Transaction for Server {
...
+                        // 0.0 is a placeholder for `other` until `impl Debug for Version`
+                        span.as_ref().unwrap().record("http.flavor", &0.0);
+                        span.as_ref().unwrap().record(
+                            "otel.status_code",
+                            &format!("{:?}", opentelemetry::trace::StatusCode::Error).as_str(),
+                        );
+                        span.as_ref().unwrap().record(
+                            "otel.status_message",
+                            &format!("PANIC: unexpected response version: {:?}", other).as_str(),
+                        );
```
taqtiqa-mark, line 428: Note to self (@taqtiqa-mark): Use this error to simultaneously try out the ergonomics of the alternative syntax for traceable errors.

Come to think of it - is there any reason not to provide both and let users adopt as they prefer? 
The implementation details should be shareable.

seanmonstar:
I really appreciate you caring about this topic. Figuring out tracing in hyper is indeed a goal for this year. However, I think this is not the direction we ultimately want to go. For instance:

- hyper shouldn't adopt OpenTelemetry directly. There are other standards that people may wish to use instead. We also don't want to tie the stability of hyper 1.0 with OpenTelemetry. That is to say, if they release a new standard, we shouldn't have to force hyper to 2.0 just to upgrade the spans/events.
  - Instead, we should determine the bare minimum spans/events that **must** be in hyper, so that another layer outside of hyper can translate them into whatever format they want. Some hypothetical `hyper-opentelemetry` crate could provide tracing `Subscriber`s that receive hyper's events, and then fill them out in the OpenTelemetry style and send them to the configured destination.
- hyper shouldn't rely on any new public dependencies.
- hyper shouldn't use proc-macros to do this, they significantly slow down compilation. If an end user wants to use proc-macros for their own helpers, hey, no problem. We just shouldn't force it into hyper itself.

----

I mentioned that our goal is to identify the bare minimum that hyper itself needs to emit. One way to do that is to try to make some separate crate, that hyper has zero knowledge of, which translates the events into the target format. You could make a `mark-hyper-otel` crate or similar, and see how far _that_ can go. It probably can use pieces from `tower_http::trace`. 

Once it's determine **what a user wants to measure in the first place**, the next step would be to propose those events/spans in a new issue. We can discuss the proposal before writing code, since combined minds may help determine something different. It's better that way, so we don't have to "throw away" code in a PR that hasn't been discussed yet.

Does that clarify things?

taqtiqa-mark:
> I really appreciate you caring about this topic. Figuring out tracing in hyper is indeed a goal for this year.

 No worries.  Apologies for this being 2-steps fwd, 1-back - this is early in my Rust journey.

> However, I think this is not the direction we ultimately want to go.

No worries.

> Does that clarify things?

It does.  Thanks for taking the time to set out those details.

Issue #2759, opened 2022-02-13
Client::new requirement not clear


usupaye:
**Version**
hyper 0.14(.17)

**Description**
In the documentation, it is not clear that the function `Client::new` needs the `tcp` feature enable, otherwise it would throw an function not found error

**Example**
Cargo.toml:
```toml
[package]
name = "hyper_test"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
hyper = { version = "0.14", features = ["client", "http1", "http2"] }
```

main.rs:
```rust
use hyper::client::Client;

fn main() {
    let client = Client::new();
}
```
output:
```
error[E0599]: no function or associated item named `new` found for struct `Client` in the current scope
 --> src/main.rs:4:26
  |
4 |     let client = Client::new();
  |                          ^^^ function or associated item not found in `Client<_, _>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `hyper_test` due to previous error
```

seanmonstar:
Ah ok, seems like we can add a `doc(cfg(tcp))` to that function.

PR #2760, opened 2022-02-15
docs(client): Add tcp feature for Client::new docs

ilyatrefilov:
added cfg attr to denote required tcp feature in docs for Client::new method

Closes: #2759


seanmonstar (approved):
Thanks!

Issue #2685, opened 2021-11-05
Make cdylib only build when C API required


seanmonstar:
There's some problems with compiling hyper as a `cdylib` (https://github.com/rust-lang/rust/issues/82151), and even without problems, it's extra compilation work if you don't need to build a C library. The goal is to make it so users can opt-in to building a `cdylib`, instead of it being always on. To that end, we've pushed a bit in the tooling to allow this, and here's the steps that need to be done:

- [x] Open an RFC to allow `cargo rustc --crate-type cdylib` (https://github.com/rust-lang/rfcs/pull/3180)
- [x] RFC is approved and merged
- [x] Feature is added to cargo ([One PR](https://github.com/rust-lang/cargo/pull/8789))
- [x] https://github.com/rust-lang/cargo/issues/10356
- [x] Remove `cdylib` from hyper's `Cargo.toml` manifest
- [ ] https://github.com/rust-lang/cargo/issues/10083

seanmonstar:
It's now possible in nightly to enable this at runtime:

```
cargo rustc -Z unstable-options --crate-type cdylib
```

Since the C API is "experimental", perhaps we can require that nightly cargo be used already...

seanmonstar:
@bagder how do you feel about using "nightly" cargo. If installing Rust with `rustup`, you can just run `rustup install nightly` and then run `cargo +nightly blah blah` (or `rustup default nightly` to skip needing to add `+nightly` for a directory).

The background here is that currently hyper sets in its manifest to build both a Rust lib (`rlib`) and a C lib (`cdylib`) at the same time. This is wasteful, and causes some problems for people building Rust in some environments. But it's the only "stable" way to specify library type. We've added the ability to cargo to specify this at command-line time, but it will take a while to become part of "stable" cargo. By using this now, we can a) fix some of those other Rust builds, and b) test the nightly feature so it can become stable.

bagder:
Whatever helps you go about this in a more convenient manner I will adapt to. Using nightly like this sounds totally doable in my end!

PR #2763, opened 2022-02-23
Only build cdylib when the FFI cfg is active

lqd:
Hi, 

as part of the [initiative to improve compile times](https://blog.rust-lang.org/inside-rust/2022/02/22/compiler-team-ambitions-2022.html#faster-builds-initiatives--%EF%B8%8F) we've been looking at rustc benchmarks over a number of crates in the ecosystem.

We've noticed `hyper` was preventing pipelining opportunities, both for:
- its own dependencies: compilation only starts when dependencies are done with codegen, not just when their metadata is available, e.g. for `tokio` or `h2`, depending on the chosen features. 
- its dependent crates: it seems crates depending on `hyper` also wait for codegen instead of its metadata, as we can see e.g. here in [`warp-0.3.2`](https://lqd.github.io/rustc-benchmarking-data/results/round-13-cargo-timing-opt-j8/cargo-timing-warp-0.3.2-opt-j8.html). 
 
In this cargo timing example, `hyper` could start building 6.6s earlier with `h2`'s metadata, and `warp` could start building as soon as `hyper`'s metadata is available, around 6s earlier (from different measurements, `hyper`'s metadata is not visible in the link above). 

Adding both of these together, there could be a sizable opportunity to improve compile times in the whole `hyper` ecosystem, in addition to `hyper` contributors working on the crate itself.

I've discussed this with @nox and they've directed me at https://github.com/hyperium/hyper/issues/2685 as the cause, with this [cargo issue as a blocker](https://github.com/rust-lang/cargo/issues/10356). Until that is fixed however, I wanted to discuss this hack as an alternative, to re-enable pipelining (without creating a whole different FFI-binding crate) and improve compile times in the meantime. 

The idea is to use `#![crate_type]` to build the cdylib only when the FFI cfg is active, in the source rather than via cargo. This works, I believe, on stable and on your MSRV of 1.46 (but I'm not sure I was able to fully test this, and hope CI will help). It has however been recently deprecated, and that deprecation will land on stable in the 1.59 release tomorrow. I don't think it'll go away before the cargo issue is fixed though, so it still seems pretty safe. (Also: I've had to allow the lint to build with deny warnings, and unknown lints to support 1.46+ which don't yet know about that new lint. Another note: maybe the `ffi` feature should be checked in addition to the `hyper_unstable_ffi` cfg ?).

I believe there's another cargo feature which could maybe help, to specify dependency crate types in the cargo manifest; all in all, it seems issues in cargo will be fixed and this is mostly a temporary measure to help things until those principled fixes are available.

Just to give you an idea of the impact this has, I've benchmarked building `hyper` (with the `client,http1,http2` features) from `-j2` to `-j12`: debug builds improve by 7-8%, and release builds by 18-29% (26% mean). This does translate to downstream crates. 

I can provide precise numbers and cargo timing graphs if needed.

(Opening as draft for discussion whether something like this could work for you ? and to test it out on CI)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,11 @@
 #![cfg_attr(test, deny(warnings))]
 #![cfg_attr(all(test, feature = "nightly"), feature(test))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(hyper_unstable_ffi, crate_type = "cdylib")]
+#![cfg_attr(
+    hyper_unstable_ffi,
+    allow(unknown_lints, deprecated_cfg_attr_crate_type_name)
+)]
 
 //! # hyper
 //!
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,6 @@ include = [
   #"build.rs",
 ]
 
-[lib]
-crate-type = ["lib", "staticlib", "cdylib"]
-
 [dependencies]
 bytes = "1"
 futures-core = { version = "0.3", default-features = false }


seanmonstar:
Oh interesting! This is a wonderful solution, assuming it works all the way through. rustc will know to compile its dependencies correctly? I assumed cargo needed to know.

src/lib.rs
```diff
@@ -5,6 +5,8 @@
 #![cfg_attr(test, deny(warnings))]
 #![cfg_attr(all(test, feature = "nightly"), feature(test))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(hyper_unstable_ffi, crate_type = "cdylib")]
+#![cfg_attr(hyper_unstable_ffi, allow(unknown_lints, deprecated_cfg_attr_crate_type_name))]
```
seanmonstar, line 9:
```suggestion
#![cfg_attr(
    hyper_unstable_ffi,
    allow(unknown_lints, deprecated_cfg_attr_crate_type_name)
)]
```

Looks like rustfmt wants this instead.

lqd: Ah sorry, I didn't notice that, it's fixed now.

lqd:
> assuming it works all the way through. rustc will know to compile its dependencies correctly?

I hope so but I'm not actually sure yet. 

The FFI tests done in the CI workflow do work locally for me, but I don't know exactly know the setup used in curl to test this. If you could point me at some documentation or have ideas on how to test it more thoroughly, it would be very helpful. 

Maybe running the capi examples and just not building them would be enough ? Maybe having a different crate depending on this, much like your cargo issue example ?

lqd:
After further testing I don't think this actually will work, CI will confirm that I'm sure. There were likely stray `.so`s locally that made the examples build. Re-doing the tests on a clean checkout now fails.

It'll at least provide good numbers for the cargo fix. And I'll test those when they land. I will also test the WIP PR for the fix.

lqd:
It's not much use keeping this open. I _will_ help fix this though.

Issue #2426, opened 2021-02-08
Ideas for supporting HTTP Early Hints server-side


acfoltzer:
An important customer of ours is very interested in using [HTTP 103 Early Hints](https://tools.ietf.org/html/rfc8297) to lower the perceived latency experienced by their end users. This requires the ability to return more than one response head during the message exchange. The basic example from the RFC is:

```
   Client request:

     GET / HTTP/1.1
     Host: example.com


   Server response:

     HTTP/1.1 103 Early Hints
     Link: </style.css>; rel=preload; as=style
     Link: </script.js>; rel=preload; as=script

     HTTP/1.1 200 OK
     Date: Fri, 26 May 2017 10:02:11 GMT
     Content-Length: 1234
     Content-Type: text/html; charset=utf-8
     Link: </style.css>; rel=preload; as=style
     Link: </script.js>; rel=preload; as=script

     <!doctype html>
     [... rest of the response body is omitted from the example ...]
```

With hyper's service function model, we currently only have the chance to return a single response head. Supporting multiple response heads may require some substantial API changes or additions, so I'm hoping we can discuss potential designs before diving into implementation. Please take any type signatures mentioned below to be a bit handwavy; I know things are more complex with the tower types and such.

## Channel/stream-based interfaces

These are my preferred approaches from a hyper user's perspective, but I admit I tend to get lost when trying to follow all the types involved with tower services under the hood, so I'm not sure if this is particularly feasible on the implementation side.

### Service function channel argument

We could add a variant of `hyper::service::service_fn()` whose closure takes an additional sender argument, i.e., `FnMut(Request<R>, mpsc::Sender<Response<()>>) -> S`. The service implementation would be free to send many interim responses, and would return the final response in the same way that existing service functions do. For example:

```rust
async fn my_service_fn(
    req: Request<Body>,
    mut interim_sender: mpsc::Sender<Response<()>>,
) -> Result<Response<Body>> {
    let early_hints = Response::builder()
        .status(103)
        .header("Link", "</style.css>; rel=preload; as=style")
        .header("Link", "</script.js>; rel=preload; as=script")
        .body(())?;
    interim_sender.send(early_hints).await?;
    let resp = todo!("build the final response")?;
    Ok(resp)
}
```

### Service function response stream return value

[...]

Issue #1586, opened 2018-06-28
Add HTTP/2 push support to Server


vbrandl:
What is the state of HTTP/2 server push in hyper? I couldn't find anything in the documentation. Are there any plans to implement it in the future?

seanmonstar:
First step is that the h2 dependency would need Push support: https://github.com/carllerche/h2/issues/291

seanmonstar:
With `h2` supporting push promises, I'd like to start proposing an API to support them in hyper, initially from the **server** side:

- We'd add a `hyper::push` module to contain these new pieces.
- Add a `pusher(req: &mut http::Request) -> hyper::Result<Pusher>` that can get a `Pusher` related the request's stream, or an `Error` explaining why a `Pusher` wasn't available. Error cases could include that its an HTTP/1 message, that the client disabled server push, and the like.
- The `Pusher` would have an `async fn push_request(&mut self, req: http::Request<()>) -> hyper::Result<()>` method that tries to send the `PUSH_PROMISE` frame to the client. Error cases could include that the request isn't a legal `PUSH_PROMISE`, that the clients max concurrent streams limit has been reached, the stream has closed, and the like.
- Once a push promise has been sent to the client, the same `http::Request` shall be passed to `Service::call`, and that future spawned in a new task, to generate the response for the `PUSH_PROMISE`.

An example using this proposal looks like this:

```rust
async fn handle(mut req: Request<Body>) -> Result<Response<Body>, E> {
    match hyper::push::pusher(&mut req) {
        Ok(mut pusher) => {
            let promise = Request::builder()
                .uri("/app.js")
                .body(())
                .unwrap();
            if let Err(e) = pusher.push_request(promise).await {
                eprintln!("push failed: {}", e);
            }
        },
        Err(e) => eprintln!("http2 pusher unavailable: {}", e),
    }

    Ok(Response::new(index_page_stream()))
}
```

djc:
@stammw would the proposed API make sense for HTTP 3?

stammw:
I was waiting to figure out a first implementation of push streams for `quinn-h3` before
answering this, but here is what I can say just by comparing the two specs and
looking at `hyper`'s code:

`H2` and `H3` share the same semantics on push promises / push streams, except:
1. `H3 draft` states that Promises are conveyed on response streams, in any part
   of it: before header, interleaved with DATA frames...
   
   In `RFC 7540`, I can't find an explicit statement saying which stream shall be
[...]

PR #2772, opened 2022-03-08
Define the VISION

seanmonstar:
This is an overview of what the shape of hyper looks like, but also somewhat zoomed out, so that the _vision_ can survive while the exact minute details might shift and change over time. It starts with what _is_ hyper, defines a framework for how we make decisions, includes use cases we've considered, and then shows how the library can ideally fulfill its purpose.

There's two documents here. The VISION is the main document, and includes a list of prioritized tenets inside it. The TENETS document is supplemental, and goes more in-depth on what the principles mean, and examples of how they've been used before.

This is the first step in the [hyper 1.0 timeline](https://seanmonstar.com/post/676912131372875776/hyper-10-timeline).

[This **blog post** includes background and motivation for defining a VISION for hyper.](https://seanmonstar.com/post/678179333918097408/hypers-vision)

Feedback very much welcome!


docs/TENETS.md
```diff
@@ -0,0 +1,60 @@
...
+
+[We prioritize kindness][CONDUCT], compassion and empathy towards all contributors. Technical skill is not a substitute for human decency.
+
+[CONDUCT]: https://github.com/hyperium/hyper/blob/master/docs/CODE_OF_CONDUCT.md
+
+### Examples
+
+It's not usually hard for an open source library to stay open and also meet its other priorities. Here's some instances where being **Open** would be more important than **Correct** or **Fast**:
+
+- Say an individual were to bring forward a contribution that makes hyper more correct, or faster, perhaps fixing some serious bug. But in doing so, they also insulted people, harrassed other contributors or users, or shamed everyone for the previous code. They felt their contribution was "invaluable". We would not accept such a contribution, instead banning the user and rewriting the code amongst the kind collaborators of the project.
```
joshtriplett, line 21: First of all, :+1: :+1: :+1: for this example, thank you.

Another example of prioritizing Open over Fast, in a different direction:

- Say someone brings a contribution that adds a new feature useful for performance or correctness, but their work accomplishes this by integrating hyper with a proprietary library. We would not accept such a contribution, because we don't want such a feature limited only to those users willing to compromise openness, and we don't want to bifurcate the ecosystem between those who make that compromise and those who don't.

docs/TENETS.md
```diff
@@ -0,0 +1,60 @@
...
+
+A fast experience delights users. A faster network library means a faster application, resulting in delighting our users’ users. Whether with one request, or millions.
+
+Being _fast_ means we improve throughput, drive down CPU usage, and improve sustainability.
+
+Fast _enough_. We don't sacrifice sanity for speed.
+
+## 3. HTTP/*
+
+hyper is specifically focused on HTTP. Supporting new HTTP versions is in scope, but supporting separate protocols is not.
```
joshtriplett, line 43:
```suggestion
hyper is specifically focused on HTTP. Supporting new HTTP versions is in scope, including HTTP over various transports (e.g. HTTP/3), but supporting separate protocols is not.
```

docs/VISION.md
```diff
@@ -0,0 +1,137 @@
...
+
+Those that stick around after that recommendation are users that wish both to learn at a lower level, and to pick and choose what batteries they plug in to hyper as they move along. While they do care about the other tenets, that hyper is **Understandable** is of extra importance to them.
+
+## The Library
+
+So with all that context in mind, what does hyper, the library, actually look like? This doesn't highlight what _is_ and _isn't_ present. What currently needs to change to reach this vision is left to individual version roadmaps.
+
+### Layers
+
+In all cases, a user brings their own runtime and IO to work with hyper. The IO is provided to hyper, and hyper acts on top of it. hyper returns `Future`s that the user then decides how to poll, likely involving their runtime options.
```
joshtriplett, line 60: Could you add something here, and possibly in the tenets, about the degree to which hyper provides optional runtime integration? For instance, hyper currently has an optional `runtime` feature; what criteria would hyper use to evaluate proposals for additional such features? (e.g. maintainability, not compromising higher-priority tenets) Similarly, what criteria would be used to evaluate (optional) integration with various different versions of traits such as `AsyncRead`/`AsyncWrite`?

These seem like questions people will reasonably ask, for which it'd be helpful to have a reference in the tenets and vision.

nicoburns:
I don't seem to be able to comment on the lines for some reason, so quoting here:

> If a message has a `content-length`, enforce only that many bytes are read or written.

Is this the sort of thing that might be able to be overridden? I've had experiences where I've had to interact with HTTP servers that would send incorrect content-length headers before (I believe if you just accepted all bytes until they closed the connection it would work).

seanmonstar:
> > If a message has a content-length, enforce only that many bytes are read or written.
>
> Is this the sort of thing that might be able to be overridden?

I suppose it's _possible_, but that's a really really big bug. Message length discrepancies are the basis of message smuggling vulnerabilities. But to the point of this document, hyper is **Correct** first, but if something is really needed, **Flexible** allows us to provide options to turn some things off.

docs/TENETS.md
seanmonstar: Good example. I originally did have a bullet point here to expand a situation along these lines, but never expanded it. Seems worth doing so <3

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+Who are the *users* of hyper? How would they use hyper?
+
+### Low-Level Client Library (curl, reqwest, aws-sdk)
+
+These client libraries care that hyper is **Flexible**, since they are
+expressing their own opinion on how a more-featured HTTP client should act.
+This includes opinions on connection establishment, management, pooling, HTTP
+version options, and even runtimes.
+
+curl's main reason for using hyper is that it is **Safe**.
```
bryangarza, line 41: Not sure, but maybe there is some subterm we can split them by, `featureful client libraries` vs `foundational client libraries` or something. Just a thought.

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+like? This doesn't highlight what _is_ and _isn't_ present. What currently
+needs to change to reach this vision is left to individual version roadmaps.
+
+### Layers
+
+In all cases, a user brings their own runtime and IO to work with hyper. The IO
+is provided to hyper, and hyper acts on top of it. hyper returns `Future`s that
+the user then decides how to poll, likely involving their runtime options.
+
+![](https://i.imgur.com/ly6g75q.png)
```
bryangarza, line 93: should this live in the repo itself?

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+A **Correct** implementation includes more than just enforcing certain
+characters when encoding and decoding. Order of frames, and flags in certain
+frames can affect the state of the connection. Some examples of things enforced
+at this layer:
+
+- If a message has a `content-length`, enforce only that many bytes are read or
+  written.
+- Reading a `Response` before a `Request` is even written implies a mismatched
+  reply that should be interpretted as an error.
+- The presense of some headers, such as `Connection: close`, or the absense of
```
bryangarza, line 117: nit: typos
```suggestion
- The presence of some headers, such as `Connection: close`, or the absence of
```

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+
+Beyond what is directly in the hyper crate, there are useful (utile) parts that
+may not meet hyper's stability promise. Developing, experimenting, and exposing
+those parts is the purpose of the `hyper-util` crate. That crate does not have
+the same stability level as hyper. However, the goal is that things that other
+libraries might want to expose as a public dependency do not live in
+`hyper-util` forever, but rather stabilize and get promoted into `hyper`.
+
+Exactly what gets put into `hyper-util` presently is kept in the roadmap
+documents.
```
bryangarza, line 150: Do we have a link to the roadmap we can add here?

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+
+hyper also establishes a Minimum Supported Rust Version (MSRV). hyper will
+support Rust versions at least 6 months old. If a new Rust version is released
+with a feature hyper wishes to use, we won't do so until at least 6 months
+afterwards. hyper will only ever require a new Rust version as a _minor_
+release (1.x), not as a patch (1.x.y).
+
+## Security
+
+The security of hyper is a large part of what makes hyper _protective_. We make
```
bryangarza, line 175: Other than in the charter, this is the only other usage of the word `protective`. Would be good to expound on it a bit. This line begs the question: "what other parts of hyper also make it protective?". I think that might already be answered in the doc but it would be good to make the association between the term `protective` and some of the other points on this doc more explicit.  And maybe also explicitly define what we mean by `protective` somewhere in the doc?

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+interface.
+
+By exposing this layer publicly, we take care of the **Correct** tenet, by not
+forcing the user to send the specific frames themselves. The API should be
+designed in a way that a user cannot easily (if at all) create an _incorrect_
+HTTP connection.
+
+Motivated by the **Flexible** tenet, there _are_ version-specific options that
+can be configured at this level, and version-specific functionality can usually
+be handled via `http::Extensions`.
```
acfoltzer, line 138: Is it worth discussing what the vision is for cases that fall outside of what can be handled with `Extensions`? To use the example we discussed in Discord recently, suppose we want to handle custom h2 frames that arrive interleaved with the data frames, which therefore can arrive after the head is received and `Extensions` can no longer be modified.

docs/VISION.md
```diff
@@ -0,0 +1,230 @@
...
+## Security
+
+The security of hyper is a large part of what makes hyper _protective_. We make
+hyper secure via the combined efforts of being **Correct**, focusing on
+**HTTP/\***, and making it all **Understandable**.
+
+### Memory Safety
+
+Being **Correct** requires that hyper be memory-safe. Using the Rust language
+gets us most of the way there. But it there is the ability to write `unsafe`
```
acfoltzer, line 182:
```suggestion
gets us most of the way there. But there is the ability to write `unsafe`
```

docs/VISION.md
acfoltzer: #1586 and #2426 are other possibly relevant cases to consider here

docs/VISION.md
seanmonstar: The roadmap to 1.0 is being written at this very moment, and will be a new PR soon.

LucioFranco (approved):

LucioFranco:
Exciting stuff! Thanks sean

docs/TENETS.md
```diff
@@ -0,0 +1,93 @@
...
+hyper is specifically focused on HTTP. Supporting new HTTP versions is in scope,
+but supporting separate protocols is not.
+
+This also defines what the abstraction layer is: the API is designed around
+sending and receiving HTTP messages.
+
+## 4. Flexible
+
+hyper enables as many usecases as possible. It has no opinion on application
+structure, and makes less assumptions about its environment. This includes being
```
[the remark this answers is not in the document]
seanmonstar: Nice catch, thanks! Should say "few".

docs/TENETS.md
```diff
@@ -0,0 +1,100 @@
...
+## 2. Fast
+
+A fast experience delights users. A faster network library means a faster
+application, resulting in delighting our users’ users. Whether with one request,
+or millions.
+
+Being _fast_ means we improve throughput, drive down CPU usage, and improve
+sustainability.
+
+Fast _enough_. We don't sacrifice sanity for speed.
```
Yneth, line 68: any ideas, considerations regarding io uring support which may come with new async interface with IO (owned buffers instead of borrowed ones)?

PR #2775, opened 2022-03-12
Move non-conn code out of conn.rs

SabrinaJewson:
The actual code for `Server` was previously organized very confusingly: it was thrice layered with `SpawnAll` and `Serve` which both appeared in conn.rs despite not having anything to do with the lower-level conn API. This PR changes that, removing all layering and having the code for the higher-level `Server` appear inside `server.rs` only.


seanmonstar:
Oh thanks for doing this! Yea, the server module didn't used to have separate files, and so some parts didn't get moved cleanly. I like this!

seanmonstar (approved):
Sorry it took so long to get back to review. Wow, this is quality stuff! And I appreciate that it likely was rather tedious, thanks for cleaning this up!

PR #2776, opened 2022-03-12
perf(body): have `to_bytes` collect with a `BytesMut`

SabrinaJewson:
By collecting into a `BytesMut` instead of a `Vec<u8>`, cloning the resulting bytes can be made O(1). The only disadvantage is it's a couple more bytes to store, but that's fine.

### DIFF
--- a/src/body/to_bytes.rs
+++ b/src/body/to_bytes.rs
@@ -1,4 +1,4 @@
-use bytes::{Buf, BufMut, Bytes};
+use bytes::{Buf, BufMut, Bytes, BytesMut};
 
 use super::HttpBody;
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! Have you checked the `benches/body.rs` if this has a difference? Looking at `BytesMut::freeze`, if it hasn't had any slices made, then it is equivalent to what a `Vec` does. So my gut is that this will be slightly slower...

SabrinaJewson:
Oh dear, looks like I misread `shallow_clone_vec` from `bytes` when I first read it - I had concluded that it copied over the bytes, when that wasn't the case at all. So actually cloning a `bytes`-`Vec` _is_ O(1) and this change isn't needed. I'm going to blame `bytes` for this though because if [it wasn't unsound](https://github.com/tokio-rs/bytes/issues/522) I wouldn't've assumed that :P

PR #2790, opened 2022-03-21
Send an error back to client when dispatch misbehaves (fixes #2649)

### LINKED ISSUE #2649: Consider removing "dispatch dropped without returning error"
Hyper panics with "dispatch dropped without returning error" if the background dispatch task goes away, either because the runtime was dropped or some user-provided code such as a HttpBody implementation panicked

Hyper should catch the panic and return a proper error instead of panicking with "dispatch dropped without returning error".

Cc @seanmonstar

nox:


nox:
There is now a test with a body implementation panicking. Should the error kind be "canceled", or something else?

seanmonstar:
> Should the error kind be "canceled", or something else?

Probably not canceled. I've mostly tried to keep that error kind to mean the request was stopped before any part of it made it to the IO transport, so it could be retried.

nox:
> Probably not canceled. I've mostly tried to keep that error kind to mean the request was stopped before any part of it made it to the IO transport, so it could be retried.

I introduced a new `User` variant for the error.

src/client/dispatch.rs
```diff
@@ -198,33 +198,59 @@ impl<T, U> Drop for Envelope<T, U> {
...
 pub(crate) enum Callback<T, U> {
-    Retry(oneshot::Sender<Result<U, (crate::Error, Option<T>)>>),
-    NoRetry(oneshot::Sender<Result<U, crate::Error>>),
+    Retry(Option<oneshot::Sender<Result<U, (crate::Error, Option<T>)>>>),
+    NoRetry(Option<oneshot::Sender<Result<U, crate::Error>>>),
+}
+
+impl<T, U> Drop for Callback<T, U> {
+    fn drop(&mut self) {
+        // FIXME(nox): What errors do we want here?
```
seanmonstar, line 207: Was this FIXME meant to change anything?

seanmonstar (approved):

PR #2792, opened 2022-03-23
feat(ext): support non-canonical HTTP/1 reason phrases

acfoltzer:
Add a new extension type `hyper::ext::ReasonPhrase` gated by a new feature flag `http1_reason_phrase`. When enabled, store any non-canonical reason phrases in this extension when parsing responses, and write this reason phrase instead of the canonical reason phrase when emitting responses.

Reason phrases are a disused corner of the spec that implementations ought to treat as opaque blobs of bytes. Unfortunately, real-world traffic sometimes does depend on being able to inspect and manipulate them.

My main outstanding question is: should this new feature flag be included in `full`? Excluding it means this patch also has to modify the CI configuration in order to run the `client` and `server` tests.

### DIFF
--- a/src/ext/h1_reason_phrase.rs
+++ b/src/ext/h1_reason_phrase.rs
@@ -1,0 +1,221 @@
+use std::convert::TryFrom;
+
+use bytes::Bytes;
+
+/// A reason phrase in an HTTP/1 response.
+///
+/// # Clients
+///
+/// For clients, a `ReasonPhrase` will be present in the extensions of the `http::Response` returned
+/// for a request if the reason phrase is different from the canonical reason phrase for the
+/// response's status code. For example, if a server returns `HTTP/1.1 200 Awesome`, the
+/// `ReasonPhrase` will be present and contain `Awesome`, but if a server returns `HTTP/1.1 200 OK`,
+/// the response will not contain a `ReasonPhrase`.
+///
+/// ```no_run
+/// # #[cfg(all(feature = "tcp", feature = "client", feature = "http1"))]
+/// # async fn fake_fetch() -> hyper::Result<()> {
+/// use hyper::{Client, Uri};
+/// use hyper::ext::ReasonPhrase;
+///
+/// let res = Client::new().get(Uri::from_static("http://example.com/non_canonical_reason")).await?;
+///
+/// // Print out the non-canonical reason phrase, if it has one...
+/// if let Some(reason) = res.extensions().get::<ReasonPhrase>() {
+///     println!("non-canonical reason: {}", std::str::from_utf8(reason.as_bytes()).unwrap());
+/// }
+/// # Ok(())
+/// # }
+/// ```
+///
+/// # Servers
+///
+/// When a `ReasonPhrase` is present in the extensions of the `http::Response` written by a server,
+/// its contents will be written in place of the canonical reason phrase when responding via HTTP/1.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ReasonPhrase(Bytes);
+
+impl ReasonPhrase {
+    /// Gets the reason phrase as bytes.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    /// Converts a static byte slice to a reason phrase.
+    pub fn from_static(reason: &'static [u8]) -> Self {
+        // TODO: this can be made const once MSRV is >= 1.57.0
+        if find_invalid_byte(reason).is_some() {
+            panic!("invalid byte in static reason phrase");
+        }
+        Self(Bytes::from_static(reason))
+    }
+
+    /// Converts a `Bytes` directly into a `ReasonPhrase` without validating.
+    ///
+    /// Use with care; invalid bytes in a reason phrase can cause serious security problems if
+    /// emitted in a response.
+    pub unsafe fn from_bytes_unchecked(reason: Bytes) -> Self {
+        Self(reason)
+    }
+}
+
+impl TryFrom<&[u8]> for ReasonPhrase {
+    type Error = InvalidReasonPhrase;
+
+    fn try_from(reason: &[u8]) -> Result<Self, Self::Error> {
+        if let Some(bad_byte) = find_invalid_byte(reason) {
+            Err(InvalidReasonPhrase { bad_byte })
+        } else {
+            Ok(Self(Bytes::copy_from_slice(reason)))
+        }
+    }
+}
+
+impl TryFrom<Vec<u8>> for ReasonPhrase {
+    type Error = InvalidReasonPhrase;
+
+    fn try_from(reason: Vec<u8>) -> Result<Self, Self::Error> {
+        if let Some(bad_byte) = find_invalid_byte(&reason) {
+            Err(InvalidReasonPhrase { bad_byte })
+        } else {
+            Ok(Self(Bytes::from(reason)))
+        }
+    }
+}
+
+impl TryFrom<String> for ReasonPhrase {
+    type Error = InvalidReasonPhrase;
+
+    fn try_from(reason: String) -> Result<Self, Self::Error> {
+        if let Some(bad_byte) = find_invalid_byte(reason.as_bytes()) {
+            Err(InvalidReasonPhrase { bad_byte })
+        } else {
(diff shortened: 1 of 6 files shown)


src/ext/h1_reason_phrase.rs
```diff
@@ -0,0 +1,101 @@
...
+/// # }
+/// ```
+///
+/// # Servers
+///
+/// When a `ReasonPhrase` is present in the extensions of the `http::Response` written by a server,
+/// its contents will be written in place of the canonical reason phrase when responding via HTTP/1.
+#[cfg(any(feature = "http1_reason_phrase", feature = "ffi"))]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ReasonPhrase(pub(crate) Bytes);
```
acfoltzer, line 37: :bike::derelict_house: Not particularly tied to this name, but tried to match the already-existing `ext` types which plainly state their purpose without mentioning their associated protocol version.

src/ffi/http_types.rs
```diff
@@ -24,9 +24,6 @@ pub struct hyper_headers {
     orig_casing: HeaderCaseMap,
 }
 
-#[derive(Debug)]
-pub(crate) struct ReasonPhrase(pub(crate) Bytes);
```
acfoltzer, line 28: I figured we may as well reuse the public type for the FFI bits.

src/proto/h1/role.rs
```diff
@@ -358,7 +358,17 @@ impl Http1Transaction for Server {
 
         let init_cap = 30 + msg.head.headers.len() * AVERAGE_HEADER_SIZE;
         dst.reserve(init_cap);
-        if msg.head.version == Version::HTTP_11 && msg.head.subject == StatusCode::OK {
+
+        #[cfg(feature = "http1_reason_phrase")]
+        let custom_reason_phrase = msg.head.extensions.get::<crate::ext::ReasonPhrase>();
+        #[cfg(not(feature = "http1_reason_phrase"))]
+        #[allow(non_upper_case_globals)]
+        const custom_reason_phrase: Option<()> = None;
```
acfoltzer, line 366: Here's where it starts getting a bit messy. To reduce the overall `cfg`ness of this code, I'm relying on `const` evaluation to ensure this does not have an effect on performance when the flag is not enabled.

acfoltzer (commented):
So the tricky question is: is this worth a feature flag? Here are the performance numbers for the `cargo bench` suite (red is with the new feature disabled, green is with it enabled): https://gist.github.com/acfoltzer/c64e92ac95c390cb1c08624ecf61bfdf

It looks to me like roughly a wash, though there's a fair amount of noise in the sample. The problem is that none of the existing benches use the custom reason phrase, so it's hard to see the overall effect that existing users of `"full"` might get if they're already receiving responses with non-canonical phrases.

From an API cleanliness perspective I would prefer rolling this extension into the existing `"http1"` flag, but I also don't want to be introducing undue surprises for users of `client` who are not currently having to bear the cost of either the comparison against the canonical reason phrase or the allocation to store a custom one when sighted.

Cargo.toml
```diff
@@ -86,6 +86,9 @@ full = [
 http1 = []
 http2 = ["h2"]
 
+# Support for non-canonical HTTP/1 reason phrases
+http1_reason_phrase = ["http1"]
```
seanmonstar, line 90: Yea, so let's talk about if we need a feature. My initial gut feeling is that it seems like such a small feature that it doesn't pull its weight, just include it. Especially if the benchmarks don't show a noticeable difference.

src/ext/h1_reason_phrase.rs
seanmonstar: Yea the name seems fine to me. I don't think `Http1ReasonPhrase` is better 👀 .

src/ext/h1_reason_phrase.rs
```diff
@@ -0,0 +1,101 @@
...
+        std::str::from_utf8(self.as_bytes())
+    }
+
+    /// Converts a static byte slice to a reason phrase.
+    pub const fn from_static(reason: &'static [u8]) -> Self {
+        Self(Bytes::from_static(reason))
+    }
+
+    /// Converts a static string to a reason phrase.
+    pub const fn from_static_str(reason: &'static str) -> Self {
```
seanmonstar, line 56: I've seen `from_static` in a lot of places, but never a `from_static_str`. Usually I'd expect people could either stick a `b` prefix on their string, or could call `as_bytes()`.

src/ext/h1_reason_phrase.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct ReasonPhrase(pub(crate) Bytes);
+
+impl ReasonPhrase {
+    /// Gets the reason phrase as bytes.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    /// Gets the reason phrase as a `&str` if the phrase is valid UTF-8.
+    pub fn to_str(&self) -> Result<&str, Utf8Error> {
```
seanmonstar, line 46: I'd leave this off for now, just getting the bytes and being able to use `std::str::from_utf8()` is what I'd expect most people to do.

src/ext/h1_reason_phrase.rs
```diff
@@ -0,0 +1,101 @@
...
+        std::str::from_utf8(self.as_bytes())
+    }
+
+    /// Converts a static byte slice to a reason phrase.
+    pub const fn from_static(reason: &'static [u8]) -> Self {
```
seanmonstar, line 51: Looking at this and the other constructors, I think we probably need to verify there's no illegal bytes in here that would mess up the HTTP/1 message.

seanmonstar (changes requested):
Very good start, I appreciate the thoroughness of the docs and tests 🤩

src/ext/h1_reason_phrase.rs
acfoltzer: Ack, that's a really good catch. I'll convert some of the existing `From` impls to `TryFrom`, and expose an `unsafe` constructor for unchecked creation.

seanmonstar: Is the unchecked one needed yet? A `from_static` can likely just panic, similar to `HeaderName::from_static`. The verification can happen at compile time for most static strings.

Cargo.toml
acfoltzer: I'd be in favor of including it. I think it'd take some very unusual traffic (like maxing out reason phrase length on every response) to cause a noticeable difference. It also helps that the cost is incurred on responses rather than on requests which are usually more concerning as a DoS vector.

I'll go ahead and fold it into `http1` but will keep it in a separate commit while this is under review.

src/ext/h1_reason_phrase.rs
acfoltzer: We need to check when constructing at runtime, because the server role will happily splat out whatever bytes are there into the status line. If someone put non-spec characters into the byte string this would lead to Problems.

The unchecked one is there for the client role to save the validation step; it already knows the characters are valid since it made it through parsing.

acfoltzer: Implemented these interfaces in 699c4417, added tests in 4ca1eee9

src/ext/h1_reason_phrase.rs
acfoltzer: Removed in 699c4417

src/ext/h1_reason_phrase.rs
acfoltzer: Removed in 699c4417

Cargo.toml
acfoltzer: Made this change in d621c741, which led to a pleasant reduction in the amount of awkwardly `cfg`ed code :sparkles:

seanmonstar:
@acfoltzer was this waiting on anything else, or just another review from me?

acfoltzer:
@seanmonstar yep, though it looks like I'll need to fix some merge conflicts first. I'll try to get to that soon

acfoltzer:
@seanmonstar rebased and squashed, incorporating the proposed change to fold this into the `http1` feature rather than introducing an entirely new feature just for the reason phrases. I think it's ready for a final review.

seanmonstar (approved):

Issue #2780, opened 2022-03-17
Hyper Curl test 580 and 580 Root cause


liamwarfield:
**Version**
hyper::ffi  current version

**Platform**
All?

**Description**
Found the RC of failing curl tests 580 and 581!

[short summary of the bug]
[hyper_headers_foreach](https://docs.rs/hyper/0.14.17/hyper/ffi/fn.hyper_headers_foreach.html#) iterates through headers in by header name, not by order recieved! The smoking gun for for curl tests 580 and 581 is [this](https://docs.rs/hyper/0.14.17/src/hyper/ffi/http_types.rs.html#410) line.

For test 581:
Raw response sent by server:
```
HTTP/1.1 200 OK
Content-Length: 18

WE ROOLZ: 118369
HTTP/1.1 200 all good!
Date: Tue, 09 Nov 2010 14:49:00 GMT
Server: test-server/fake
Content-Type: text/html
Content-Length: 0
Connection: close
Content-Type: changed/my/mind
```

Curl is expecting an iterator that returns headers in the order that they were received:
```
HTTP/1.1 200 all good![CR][LF]
Date: Tue, 09 Nov 2010 14:49:00 GMT[CR][LF]
Server: test-server/fake[CR][LF]
Content-Type: text/html[CR][LF]
Content-Length: 0[CR][LF]
Connection: close[CR][LF]
Content-Type: changed/my/mind[CR][LF]
[CR][LF]
```
but hyper is returning ALL of the `Content-Type` headers before moving onto the `Connection` header. Results in:
```
HTTP/1.1 200 all good![CR][LF]
Date: Tue, 09 Nov 2010 14:49:00 GMT[CR][LF]
Server: test-server/fake[CR][LF]
Content-Type: text/html[CR][LF]
Content-Type: changed/my/mind[CR][LF]  <------ comes to early!
Content-Length: 0[CR][LF]
Connection: close[CR][LF]
[CR][LF]
```

**Hyper Code Responcible**
```rust
hyper_headers_foreach_callback, userdata: *mut c_void) {
        let headers = non_null!(&*headers ?= ());
        for name in headers.headers.keys() {
            let mut names = headers.orig_casing.get_all(name);

            /// HERE is the bug, this iterates through all headers with the
            /// same name before moving on to the next header name.
            for value in headers.headers.get_all(name) {
                let (name_ptr, name_len) = if let Some(orig_name) = names.next() {
[...]

PR #2798, opened 2022-03-30
fix(ffi): Change hyper_headers_foreach to iterate in recieved order

liamwarfield:
Libcurl expects that headers are iterated in the same order that they
are recieved. Previously this caused curl tests 580 and 581 to fail.

SUMMARY OF CHANGES: Add a new data structure to HeaderCaseMap that
represents the order in which headers originally appear in a HTTP
message. This datastructure is `Vec<(Headername, multimap-index)>`.
This vector is ordered by the order which headers were recieved.
This new datastucture is used in hyper_headers_foreach to iterate
through headers in the correct order.

CLOSES ISSUE: #2780
BEAKING CHANGE: HeaderCaseMap::append now requires that names passed
into it impl Into<HeaderName> + Clone. Currently all types that impl
IntoHeaderName (in hyper and http) also impl these traits already.

### DIFF
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -560,6 +560,7 @@ impl Builder {
             h1_parser_config: Default::default(),
             h1_title_case_headers: false,
             h1_preserve_header_case: false,
+            #[cfg(feature = "ffi")]
             h1_preserve_header_order: false,
             h1_max_buf_size: None,
             #[cfg(feature = "ffi")]
@@ -966,9 +967,11 @@ impl Builder {
                     if opts.h1_title_case_headers {
                         conn.set_title_case_headers();
                     }
+                    #[cfg(feature = "ffi")]
                     if opts.h1_preserve_header_case {
                         conn.set_preserve_header_case();
                     }
+                    #[cfg(feature = "ffi")]
                     if opts.h1_preserve_header_order {
                         conn.set_preserve_header_order();
                     }
--- a/src/client/conn.rs
+++ b/src/client/conn.rs
@@ -156,6 +156,8 @@ pub struct Builder {
     h1_writev: Option<bool>,
     h1_title_case_headers: bool,
     h1_preserve_header_case: bool,
+    #[cfg(feature = "ffi")]
+    h1_preserve_header_order: bool,
     h1_read_buf_exact_size: Option<usize>,
     h1_max_buf_size: Option<usize>,
     #[cfg(feature = "ffi")]
@@ -558,6 +560,7 @@ impl Builder {
             h1_parser_config: Default::default(),
             h1_title_case_headers: false,
             h1_preserve_header_case: false,
+            h1_preserve_header_order: false,
             h1_max_buf_size: None,
             #[cfg(feature = "ffi")]
             h1_headers_raw: false,
@@ -701,6 +704,21 @@ impl Builder {
     /// Default is false.
     pub fn http1_preserve_header_case(&mut self, enabled: bool) -> &mut Builder {
         self.h1_preserve_header_case = enabled;
+        self
+    }
+
+    /// Set whether to support preserving original header order.
+    ///
+    /// Currently, this will record the order in which headers are received, and store this
+    /// ordering in a private extension on the `Response`. It will also look for and use
+    /// such an extension in any provided `Request`.
+    ///
+    /// Note that this setting does not affect HTTP/2.
+    ///
+    /// Default is false.
+    #[cfg(feature = "ffi")]
+    pub fn http1_preserve_header_order(&mut self, enabled: bool) -> &mut Builder {
+        self.h1_preserve_header_order = enabled;
         self
     }
 
@@ -950,6 +968,9 @@ impl Builder {
                     }
                     if opts.h1_preserve_header_case {
                         conn.set_preserve_header_case();
+                    }
+                    if opts.h1_preserve_header_order {
+                        conn.set_preserve_header_order();
                     }
                     if opts.h09_responses {
                         conn.set_h09_responses();
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -125,22 +125,34 @@ impl HeaderCaseMap {
 
 #[derive(Clone, Debug)]
 /// Hashmap<Headername, numheaders with that name>
-pub(crate) struct OriginalHeaderOrder(HashMap<HeaderName, usize>, Vec<(HeaderName, usize)>);
+pub(crate) struct OriginalHeaderOrder {
+    /// Stores how many entries a Headername maps to. This is used
+    /// for accounting.
+    num_entries: HashMap<HeaderName, usize>,
+    /// Stores the ordering of the headers. ex: `vec[i] = (headerName, idx)`,
+    /// The vector is ordered such that the ith element
+    /// represents the ith header that came in off the line.
+    /// The `HeaderName` and `idx` are then used elsewhere to index into
+    /// the multi map that stores the header values.
+    entry_order: Vec<(HeaderName, usize)>,
+}
 
 #[cfg(all(feature = "http1", feature = "ffi"))]
 impl OriginalHeaderOrder {
     pub(crate) fn default() -> Self {
-        Self(Default::default(), Default::default())
+        OriginalHeaderOrder {
+            num_entries: HashMap::new(),
+            entry_order: Vec::new(),
+        }
     }
 
-    #[cfg(any(test, feature = "ffi"))]
     pub(crate) fn insert(&mut self, name: HeaderName) {
-        if !self.0.contains_key(&name) {
+        if !self.num_entries.contains_key(&name) {
             let idx = 0;
-            self.0.insert(name.clone(), 1);
-            self.1.push((name, idx));
+            self.num_entries.insert(name.clone(), 1);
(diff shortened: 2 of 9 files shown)


liamwarfield:
There a few things that still need to be done here (like testing), but this is here so that I can get some comments.

seanmonstar:
Thanks for working on this! A couple of thoughts:

- Since the `HeaderCaseMap` is used by things that don't need this specific functionality, I think it's best to create a separate type.
- It's probably even a good idea to make this not the default in the C API, but make it a connection option.

I can explain more about either point if you want.

liamwarfield:
I think you're right about breaking things out into another type. As for how to separate this out in the c api, how does this sound:

1. add `HeaderOrderMap` type to a new function 
```rust
pub struct hyper_headers {
    pub(super) headers: HeaderMap,
    orig_casing: HeaderCaseMap,
    orig_order: OriginalHeaderOrder,
}
```
2. Add a new function called `hyper_headers_foreach_ordered`, and revert `hyper_headers_foreach` to its original implementation.

Could you elaborate on how connection options work?

liamwarfield:
Just pushed up a commit for adding a new connection option.

seanmonstar:
Awesome, I'll review this afternoon. The other uses so far are proxies that enable the option, and then pass the extension on to an inner service. For instance, @nox did some work to make that possible for a proxy he supports. There is an issue filed here about exposing the extension publicly to make it more available.

src/ext.rs
```diff
@@ -115,8 +117,54 @@ impl HeaderCaseMap {
 
     pub(crate) fn append<N>(&mut self, name: N, orig: Bytes)
     where
-        N: IntoHeaderName,
+        N: IntoHeaderName + std::fmt::Display,
```
seanmonstar, line 120: This can be removed now, right?

src/client/conn.rs
```diff
@@ -704,6 +707,20 @@ impl Builder {
...
+    /// Set whether to support preserving original header order.
+    ///
+    /// Currently, this will record the order in which headers are received, and store this
+    /// ordering in a private extension on the `Response`. It will also look for and use
+    /// such an extension in any provided `Request`.
+    ///
+    /// Note that this setting does not affect HTTP/2.
+    ///
+    /// Default is false.
+    pub fn http1_preserve_header_order(&mut self, enabled: bool) -> &mut Builder {
```
seanmonstar, line 719: Yea, nice, exactly what I meant, thanks! (We might find that `cfg(feature = "ffi")` needs to be sprinkled around, sorry about that pain.)

src/ffi/client.rs
```diff
@@ -103,6 +104,22 @@ ffi_fn! {
     } ?= std::ptr::null_mut()
 }
 
+ffi_fn! {
+    /// temp
+    fn hyper_clientconn_options_set_preserve_header_case(opts: *mut hyper_clientconn_options) {
```
seanmonstar, line 109: To follow the example of the other options that are boolean, perhaps we should have the user pass a `c_int` here as well?

src/ext.rs
```diff
@@ -115,8 +117,54 @@ impl HeaderCaseMap {
...
     {
         self.0.append(name, orig);
     }
 }
+
+#[derive(Clone, Debug)]
+/// Hashmap<Headername, numheaders with that name>
+pub(crate) struct OriginalHeaderOrder(HashMap<HeaderName, usize>, Vec<(HeaderName, usize)>);
```
seanmonstar, line 128: I'm immediately understanding the separate hashmap and vector. Do they need to be separate? Can the hashmap just contain the vector as the value, instead of the number?

seanmonstar (changes requested):
Hey, this is really good work. Really! ❤️ 

I left some thoughts inline.

src/ffi/client.rs
liamwarfield: On it!

liamwarfield: Quick question about FFI in general:

When I was playing around with the code on the `curl` side of things I noticed that `hyper_clientconn_options_set_preserve_header_case` would get optimized out if I didn't add a `println!` to the function  (ie: I would call the function from curl, but to boolean value wouldn't change) . Is there something like `#[no_mangle]`  that I should add here?

src/ext.rs
liamwarfield: I'm just using the hash map to keep track off how many headers of the same name have already been seen. This makes the bookkeeping around inserting and appending easier.
Side Note: if the number of header names is low, it might be faster to use a BTreeMap here to avoid string hashing.

The vector is what really stores the ordering of the headers. ex: `vec[i] = (headerName, idx)`, This vector is ordered such that the `i`th element represents the `i`th header that came in off the line. The `headerName` and `idx` are then used elsewhere to index into the multi map that stores the header values.
```rust
           for (name, idx) in headers.orig_order.get_in_order() {
            let orig_name = headers.orig_casing.get_all(&name).nth(*idx).unwrap();
            let value = headers.headers.get_all(name).iter().nth(*idx).unwrap();
```

src/ext.rs
liamwarfield: Yep, I had it in there for some printf debuging

src/ffi/client.rs
```diff
@@ -95,6 +95,7 @@ ffi_fn! {
     fn hyper_clientconn_options_new() -> *mut hyper_clientconn_options {
         let mut builder = conn::Builder::new();
         builder.http1_preserve_header_case(true);
+        builder.http1_preserve_header_order(true);
```
liamwarfield, line 98: Should I remove this line (making an ffi connection not preserve ordering by default)?

src/ffi/http_types.rs
```diff
@@ -428,6 +435,37 @@ ffi_fn! {
...
 }
 
+ffi_fn! {
+    /// Iterates the headers in the order the were recieved, passing each name and value pair to the callback.
+    ///
+    /// The `userdata` pointer is also passed to the callback.
+    ///
+    /// The callback should return `HYPER_ITER_CONTINUE` to keep iterating, or
+    /// `HYPER_ITER_BREAK` to stop.
+    fn hyper_headers_foreach_ordered(headers: *const hyper_headers, func: hyper_headers_foreach_callback, userdata: *mut c_void) {
```
liamwarfield, line 445: Should I add some logic here along the lines of:
```
If !header_order_preserved { // or headers.orig_order is default
   hyper_headers_foreach(headers, func, userdata);
   return;
}
```
Currently, if the connection does not have `preserve_header_order` set, this function will iterate over nothing.

liamwarfield (commented):
Did a quick review while squashing.

src/client/conn.rs
liamwarfield: Done

src/ext.rs
liamwarfield: Done

src/ffi/client.rs
seanmonstar: I'm thinking we'll eventually want to make both of these options default off. Just because curl does want these doesn't mean all users of the C API want to pay the cost...

src/ffi/http_types.rs
seanmonstar: I think probably it should all just be one function. If the order map is set, then use it, otherwise iterate normally.

src/ext.rs
seanmonstar: Hm, ok. What do you think of giving the fields names, and adding a comment above each field explaining how they relate? That should help Future Us make changes some day.

src/ffi/client.rs
liamwarfield: No that I've added the setters for both of these, I'll just yank it out now. I'll put up a PR to curl once this has merged.

src/ffi/http_types.rs
liamwarfield: done

liamwarfield:
Just pushed up a patch, addressing the last round of comments.

src/ffi/client.rs
```diff
@@ -94,7 +94,6 @@ ffi_fn! {
     /// Creates a new set of HTTP clientconn options to be used in a handshake.
     fn hyper_clientconn_options_new() -> *mut hyper_clientconn_options {
         let mut builder = conn::Builder::new();
-        builder.http1_preserve_header_case(true);
```
seanmonstar, line 97: Looks like this removal makes a lint trigger about no longer needing to be `let mut builder`.

seanmonstar (approved):
Awesome work, this looks really good. Thank you!

Just needs a tiny change to fix the unnused mut warning, and I think we can merge.

seanmonstar:
This isn't a requirement, but something that I'm slightly curious about: when this cfg is enabled (but without setting the builder option), does it make a performance difference? Such as running `cargo bench --bench end_to_end`, or `--bench pipeline`? 🤷

liamwarfield:
I'll look into the performance stuff when I get some time later.

seanmonstar:
I'm also happy to just merge with the simple lint fix. Whatever you like :)

seanmonstar:
Superb work, thank you thank you! ❤️

liamwarfield:
The Curl Commit to enable and fix tests 580 and 581 has been merged!

https://github.com/curl/curl/pull/8707

PR #2803, opened 2022-04-01
feat(client): add a new configuration option to filter connections

diceride:
Adds a new `set_connection_filter` option to define a callback function for filtering connections.

The argument of the callback function is the DNS resolved IPv4 or IPv6 address.

If the callback function returns `Ok` the connection will be established.
If the callback function returns an `Error` the connection will not be established.

Default is `None`.

**Example - Simple connection filter to protect against server-side request forgery (SSRF) requests**

```rust
fn filter_global(socket_addr: &SocketAddr) -> Result<(), ...> {
    match socket_addr {
        SocketAddr::V4(addr) => {
            if addr.ip().is_global() {
                Ok(())
            } else {
                Err(io::Error::new(io::ErrorKind::NotConnected, "not connected").into())
            }
        }
        SocketAddr::V6(addr) => {
            if addr.ip().is_global() {
                Ok(())
            } else {
                Err(io::Error::new(io::ErrorKind::NotConnected, "not connected").into())
            }
        }
    }
}

let mut connector = HttpConnector::new();
connector.set_connection_filter(filter_global);

let client = Client::builder().set_host(true).build(connector);
```

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -79,6 +79,7 @@ struct Config {
     local_address_ipv6: Option<Ipv6Addr>,
     nodelay: bool,
     reuse_address: bool,
+    connection_filter: Option<fn(&SocketAddr) -> Result<(), Box<dyn StdError + Send + Sync>>>,
     send_buffer_size: Option<usize>,
     recv_buffer_size: Option<usize>,
 }
@@ -119,6 +120,7 @@ impl<R> HttpConnector<R> {
                 local_address_ipv6: None,
                 nodelay: false,
                 reuse_address: false,
+                connection_filter: None,
                 send_buffer_size: None,
                 recv_buffer_size: None,
             }),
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the pull request! So, it looks like you want to filter at the DNS level, right? It should be possible already to take `hyper::client::connect::dns::GaiResolver`, and since it's a `Service<Name, Response = impl Iterator<Item = SocketAddr>>`, you can use [`tower::filter`](https://docs.rs/tower/latest/tower/filter/index.html) to do so. So, something like:

```rust
let gai = GaiResolver::new();
let dns = tower::filter::Filter::new(gai, |addrs| {
    // use Iterator::filter
    addrs.filter(|addr| addr.is_global())
});
```

Note, the code might not be perfect, I didn't try to compile it. But the concept is there.

PR #2806, opened 2022-04-06
Propose the 1.0 Roadmap

seanmonstar:
The VISION outlines a decision-making framework, use-cases, and general shape
of hyper. This roadmap describes the currently known problems with hyper, and
then shows what changes are needed to make hyper 1.0 look more like what is in
the VISION.

[Rendered](https://github.com/hyperium/hyper/blob/roadmap/docs/ROADMAP.md)

(Another step in the [hyper 1.0 timeline](https://seanmonstar.com/post/676912131372875776/hyper-10-timeline).)

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -1,0 +1,404 @@
+# hyper 1.0 Roadmap
+
+## Goal
+
+Align current hyper to the [hyper VISION][VISION].
+
+The VISION outlines a decision-making framework, use-cases, and general shape
+of hyper. This roadmap describes the currently known problems with hyper, and
+then shows what changes are needed to make hyper 1.0 look more like what is in
+the VISION.
+
+## Known Issues
+
+
+> **Note**: These known issues are as of hyper v0.14.x. After v1.0 is released,
+ideally these issues will have been solved. Keeping this history may be helpful
+to Future Us, though.
+
+### Higher-level Client and Server problems
+
+Both the higher-level `Client` and `Server` types have stability concerns.
+
+For the `hyper::Server`:
+
+- The `Accept` trait is complex, and too easy to get wrong. If used with TLS, a slow TLS handshake
+  can affect all other new connections waiting for it to finish.
+- The `MakeService<&IO>` is confusing. The bounds are an assault on the eyes.
+- The `MakeService` API doesn't allow to easily annotate the HTTP connection with `tracing`.
+- Graceful shutdown doesn't give enough control.
+
+
+It's more common for people to simply use `hyper::server::conn` at this point,
+than to bother with the `hyper::Server`.
+
+While the `hyper::Client` is much easier to use, problems still exist:
+
+- The whole `Connect` design isn't stable.
+  - ALPN and proxies can provide surprising extra configuration of connections.
+  - Some `Connect` implementations may wish to view the path, in addition to the scheme, host, and port.
+  - Wants `runtime` feature
+- The Pool could be made more general or composable. At the same time, more customization is
+  desired, and it's not clear
+how to expose it yet.
+
+
+### Runtime woes
+
+hyper has been able to support different runtimes, but it has sometimes awkward
+default support for Tokio.
+
+- The `runtime` cargo-feature isn't additive
+- Built-in Tokio support can be confusing
+- Executors and Timers
+  - The `runtime` feature currently enables a few options that require a timer, such as timeouts and
+    keepalive intervals. It implicitly relies on Tokio's timer context. This can be quite confusing.
+- IO traits
+  - Should we publicly depend on Tokio's traits?
+  - `futures-io`?
+    - Definitely nope.
+    - Not stable. (0.3?)
+    - No uninitialized memory.
+  - Eventual `std` traits?
+    - They've been in design for years.
+    - We cannot base our schedule on them.
+    - When they are stable, we can:
+      - Provide a bridge in `hyper-util`.
+      - Consider a 2.0 of hyper.
+  - Define our own traits, provide util wrappers?
+
+### Forwards-compatibility
+
+There's a concern about forwards-compatibility. We want to be able to add
+support for new HTTP features without needing a new major version. While most
+of `http` and `hyper` are prepared for that, there's two potential problems.
+
+- New frames on an HTTP stream (body)
+   - Receiving a new frame type would require a new trait method
+     - There's no way to implement a "receive unknown frame" that hyper doesn't know about.
+   - Sending an unknown frame type would be even harder.
+     - Besides being able to pass an "unknown" type through the trait, the user would need to be
+       able to describe how that frame is encoded in HTTP/2/3.
+- New HTTP versions
+  - HTTP/3 will require a new transport abstraction. It's not as simple as just using some
+    `impl AsyncRead + AsyncWrite`. While HTTP/2 bundled the concept of stream creation internally,
+    and thus could be managed wholy on top of a read-write transport, HTTP/3 is different. Stream
+    creation is shifted to the QUIC protocol, and HTTP/3 needs to be able to use that directly.
+  - This means the existing `Connection` types for both client and server will not be able to
+    accept a QUIC transport so we can add HTTP/3 support.
+
+### Errors
+
+It's not easy to match for specific errors.
+
+The `Error::source()` can leak an internal dependency. For example, a
+`hyper::Error` may wrap an `h2::Error`. Users can downcast the source at
+runtime, and hyper internally changing the version of its `h2` dependency can
+cause runtime breakage for users.
+
+Formatting errors is in conflict with the current expected norm. The
+`fmt::Display` implementation for `hyper::Error` currently prints its own
+message, and then prints the message of any wrapped source error. The Errors
+Working Group currently recommends that errors only print their own message
+(link?). This conflict means that error "reporters", which crawl a source chain
+and print each error, has a lot of duplicated information.
+
+```
+error fetching website: error trying to connect: tcp connect error: Connection refused (os error 61)
+tcp connect error: Connection refused (os error 61)
+Connection refused (os error 61)
+```
+
+While there is a good reason for why hyper's `Error` types do this, at the very
+least, it _is_ unfortunate.
+
+### You call hyper, or hyper calls you?
+
+> Note: this problem space, of who calls whom, will be explored more deeply in
+> a future article.
+
+At times, it's been wondered whether hyper should call user code, or if user
+code should call hyper. For instance, should a `Service` be called with a
+request when the connection receives one, or should the user always poll for
+the next request.
+
+There's a similar question around sending a message body. Should hyper ask the
+body for more data to write, or should the user call a `write` method directly?
+
+These both get at a root topic about [write
+observability](https://github.com/hyperium/hyper/issues/2181). How do you know
+when a response, or when body data, has been written successfully? This is
+desirable for metrics, or for triggering other side-effects. 
+
+The `Service` trait also has some other frequently mentioned issues. Does
+`poll_ready` pull its complexity weight for servers? What about returning
+errors, what does that mean? Ideally users would turn all errors into
+appropriate `http::Response`s. But in HTTP/2 and beyond, stream errors are
+different from HTTP Server Error responses. Could the `Service::Error` type do
+more to encourage best practices?
+
+## Design
+
+The goal is to get hyper closer to the [VISION][], using that to determine the
+best way to solve the known issues above. The main thrust of the proposed
+changes are to make hyper more **Flexible** and stable.
+
+In order to keep hyper **Understandable**, however, the proposed changes *must*
+be accompanied by providing utilities that solve the common usage patterns,
+documentation explaining how to use the more flexible pieces, and guides on how
+to reach for the `hyper-util`ity belt.
+
+The majority of the changes are smaller and can be contained to the *Public
+API* section, since they usually only apply to a single module or type. But the
+biggest changes are explained in detail here.
+
+### Split per HTTP version
+
+The existing `Connection` types, both for the client and server, abstract over
+HTTP version by requiring a generic `AsyncRead + AsyncWrite` transport type.
+But as we figure out HTTP/3, that needs to change. So to prepare now, the
+`Connection` types will be split up.
+
+For example, there will now be `hyper::server::conn::http1::Connection` and
+`hyper::server::conn::http2::Connection` types.
+
+These specific types will still have a very similar looking API that, as the
+VISION describes, provides **Correct** connection management as it pertains to
+HTTP.
+
+There will be still be a type to wrap the different versions. It will no longer
+be generic over the transport type, to prepare for being able to wrap HTTP/3
+connections. Exactly how it will wrap, either by using internal trait objects,
+or an `enum Either` style, or using a `trait Connection` that each type
+implements, is something to be determined. It's likely that this "auto" type
+will start in `hyper-util`.
+
+### Focus on the `Connection` level
+
+As mentioned in the *Known Issues*, the higher-level `Client` and `Server` have
+stability and complexity problems. Therefore, for hyper 1.0, the main API will
+focus on the "lower-level" connection types. The `Client` and `Server` helpers
+will be moved to `hyper-util`.
+
+## Public API
+
+### body
+
+The `Body` struct is removed. Its internal "variants" are [separated into
+distinct types](https://github.com/hyperium/hyper/issues/2345), and can start
+in either `hyper-util` or `http-body-util`.
+
+The exported trait `HttpBody` is renamed to `Body`.
+
+A single `Body` implementation in `hyper` is the one provided by receiving
+client responses and server requests. It has the name `Streaming`.
+
+> **Unresolved**: Other names can be considered during implementation. Another
+> option is to not publicly name the implementation, but return `Response<impl
+Body>`s.
+
+The `Body` trait will be experimented on to see about making it possible to
+return more frame types beyonds just data and trailers.
+
+> **Unresolved**: What exactly this looks like will only be known after
+> experimentation.
+
+### client
+
+The high-level `hyper::Client` will be removed, along with the
+`hyper::client::connect` module. They will be explored more in `hyper-util`.
+
+As described in *Design*, the `client::conn` module will gain `http1` and
+`http2` sub-modules, providing per-version `SendRequest`, `Connection`, and
+`Builder` structs. An `auto` version can be explored in `hyper-util`.
+
+### error
+
+The `hyper::Error` struct remains in place.
+
+All errors returned from `Error::source()` are made opaque. They are wrapped an
+internal `Opaque` newtype that still allows printing, but prevents downcasting
+to the internal dependency.
+
+A new `hyper::error::Code` struct is defined. It is an opaque struct, with
+associated constants defining various code variants.
+
+> Alternative: define a non-exhaustive enum. It's not clear that this is
+> definitely better, though. Keeping it an opaque struct means we can add
+> secondary parts to the code in the future, or add bit flags, or similar
+> extensions.
+
+The purpose of `Code` is to provide an abstraction over the kind of error that
+is encountered. The `Code` could be some behavior noticed inside hyper, such as
+an incomplete HTTP message. Or it can be "translated" from the underlying
+protocol, if it defines protocol level errors. For example, an
+`h2::Reason::CANCEL`.
+
+### rt
+
+The `Executor` trait stays in here.
+
+Define a new trait `Timer`, which describes a way for users to provide a source
+of sleeping/timeout futures. Similar to `Executor`, a new generic is added to
+connection builders to provide a `Timer`.
+
+### server
+
+The higher-level `hyper::Server` struct, its related `Builder`, and the
+`Accept` trait are all removed.
+
+The `AddrStream` struct will be completely removed, as it provides no value but
+causes binary bloat.
+
+Similar to `client`, and as describe in the *Design*, the `conn` modules will
+be expanded to support `http1` and `http2` submodules. An `auto` version can be
+explored in `hyper-util`.
+
+### service
+
+A vendored and simplified `Service` trait will be explored.
+
+The error type for `Service`s used for a server will explore having the return
+type changed from any error to one that can become a `hyper::error::Code`.
+
+> **Unresolved**: Both of the above points are not set in stone. We will
+> explore and decide if they are the best outcome during development.
+
+The `MakeService` pieces will be removed.
+
+### Cargo Features
+
+Remove the `stream` feature. The `Stream` trait is not stable, and we cannot
+depend on an unstable API.
+
+Remove the `tcp` and `runtime` features. The automatic executor and timer parts
+are handled by providing implementations of `Executor` and `Timer`. The
+`connect` and `Accept` parts are also moving to `hyper-util`.
+
+### Public Dependencies
+
+- `http`
+- `http-body`
+- `bytes`
+
+Cannot be public while "unstable":
+
+- `tracing`
+
+## `hyper-util`
+
+
+### body
+
+A channel implementation of `Body` that has an API to know when the data has
+been successfully written is provided in `hyper_util::body::channel`.
+
+### client
+
+A `Pool` struct that implements `Service` is provided. It fills a similar role
+as the previous `hyper::Client`.
+
+> **Note**: The `Pool` might be something that goes into the `tower` crate
+> instead. Or it might stay here as a slightly more specialized racing-connect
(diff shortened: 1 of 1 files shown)


docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+focus on the "lower-level" connection types. The `Client` and `Server` helpers
+will be moved to `hyper-util`.
+
+## Public API
+
+### body
+
+The `Body` struct is removed. Its internal "variants" are [separated into
+distinct types](https://github.com/hyperium/hyper/issues/2345), and can start
+in either `hyper-utils` or `http-body-utils`.
```
jplatte, line 189: For consistency with the rest of the doc

```suggestion
in either `hyper-util` or `http-body-utils`.
```

agraven:
Not sure if this is the best place to mention this, but as discussed in rustls/hyper-rustls#167, the `Service` bound on the `Connect` trait only accepting `Uri` as request type is rather limiting, especially when it comes to adjusting behavior for individual requests.

agraven:
> There's something I need to do, that I don't seem to be able to do with the current version. For archival purposes, I would like to fetch an url, and then store not only the bytes of body, but the bytes of the request and the bytes of the returned http headers. It seems like a simple thing, but none of the rust http libraries support it (as far as I can tell). My apologies if this is already available, or somehow impossible.

I've also wanted this for debugging purposes in the past, essentially as an equivalent of `curl -v`

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+how to expose it yet.
+
+
+### Runtime woes
+
+hyper has been able to support different runtimes, but it has sometimes awkward
+default support for Tokio.
+
+- The `runtime` cargo-feature isn't additive
+- Built-in Tokio support can be confusing
```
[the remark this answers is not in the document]
asonix: as someone who [thinks it's fun to build async runtimes](https://git.asonix.dog/safe-async) i selfishly want hyper to keep working with generic runtime-ish traits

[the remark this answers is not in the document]
asonix: I also think it's a good goal to enable hyper to be used on other runtimes so GTK (and other desktop frameworks) can make use of hyper on their one runtime, e.g. https://docs.rs/glib/latest/glib/struct.MainContext.html#method.spawn

liamwarfield:
> > There's something I need to do, that I don't seem to be able to do with the current version. For archival purposes, I would like to fetch an url, and then store not only the bytes of body, but the bytes of the request and the bytes of the returned http headers. It seems like a simple thing, but none of the rust http libraries support it (as far as I can tell). My apologies if this is already available, or somehow impossible.
> 
> I've also wanted this for debugging purposes in the past, essentially as an equivalent of `curl -v`

Two suggestions here:
1. The roadmap mentioned that "for hyper 1.0, the main API will focus on the "lower-level" connection types". Would a `RawConnnection` type that just provides (and takes in) raw bytes be wanted here?
2. I've been working on a flag to preserved header order (#2798) to fix some the libcurl tests. This gets closer to what you want. Currently it feels a little weird to add this overhead to all the connection types (currently its behind certain feature flags, so it's not present in default use), so having something like option 1 might be better.

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+
+It's more common for people to simply use `hyper::server::conn` at this point,
+than to bother with the `hyper::Server`.
+
+While the `hyper::Client` is much easier to use, problems still exist:
+
+- The whole `Connect` design isn't stable.
+  - ALPN and proxies can provide surprising extra configuration of connections.
+  - Some `Connect` implementations may wish to view the path, in addition to    the scheme, host, and port.
```
LucioFranco, line 39: What examples are there of this?

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+distinct types](https://github.com/hyperium/hyper/issues/2345), and can start
+in either `hyper-utils` or `http-body-utils`.
+
+The exported trait `HttpBody` is renamed to `Body`.
+
+A single `Body` implementation in `hyper` is the one provided by receiving
+client responses and server requests. It has the name `Streaming`.
+
+> **Unresolved**: Other names can be considered during implementation. Another
+> option is to not publicly name the implementation, but return `Response<impl
```
LucioFranco, line 197: Why not a `BoxBody`? This would force us to name the Data/Error type though I am not sure if its worthwhile to punt those to later type inference. Mostly thought context here is like tower stacks.

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+### server
+
+The higher-level `hyper::Server` struct, its related `Builder`, and the
+`Accept` trait are all removed.
+
+The `AddrStream` struct will be completely removed, as it provides no value but
+causes binary bloat.
+
+Similar to `client`, and as describe in the *Design*, the `conn` modules will
```
LucioFranco, line 253: Will `conn` modules be promoted to the main `client/server` modules?

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+A vendored and simplified `Service` trait will be explored.
+
+The error type for `Service`s used for a server will explore having the return
+type changed from any error to one that can become a `hyper::error::Code`.
+
+> **Unresolved**: Both of the above points are not set in stone. We will
+> explore and decide if they are the best outcome during development.
+
+The `MakeService` pieces will be removed.
```
LucioFranco, line 267: We should revisit MakeService in tower as well imo

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+A channel implementation of `Body` that has an API to know when the data has
+been successfully written is provided in `hyper_util::body::channel`.
+
+### client
+
+A `Pool` struct that implements `Service` is provided. It fills a similar role
+as the previous `hyper::Client`.
+
+> **Note**: The `Pool` might be something that goes into the `tower` crate
```
LucioFranco, line 299: To also revisit with the remove of balance from `tower` 👍

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+There are some parts of the proposal which are not fully resolved. They are
+mentioned in Design and API sections above, but also collected here for easy
+finding. While they all have _plans_, they are more exploratory parts of the
+API, and thus they have a higher possibility of changing as we implement them.
+
+The goal is to have these questions resolved and removed from the document by
+the time there is a [Release Candidate][timeline].
+
+### Should there be `hyper::io` traits?
```
LucioFranco, line 328: Imo we should follow tokio's 1.0 guidance/solution. I think a tokio 2.0 is fine to mean a hyper 2.0 bump?

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Should returned body types be `impl Body`?
```
LucioFranco, line 330: We should check on the progress of the rustc features to make this easier xD

docs/ROADMAP.md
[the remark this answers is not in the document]
DoumanAsh: again, it is not possible to do without network primitives being portable across runtimes.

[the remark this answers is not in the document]
asonix: maybe not for h3, but h1 at least already works in hyper on 3rd party runtimes.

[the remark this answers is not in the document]
DoumanAsh: hyper uses tokio's primitives for that, tokio is required dependency there.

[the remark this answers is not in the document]
asonix: Ah, I misunderstood "primitive" to mean types like TcpStream, but you were referring to AsyncRead and AsyncWrite. I agree that this is an acceptable compromise.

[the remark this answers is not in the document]
DoumanAsh: Actually I also meant TcpStream and friends as they are tied to runtime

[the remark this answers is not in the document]
asonix: Tokio's TcpStream is not currently required to use hyper. I've implemented an adapter to run hyper clients and servers on my own custom runtime in 0.14 and I did not need tokio's TcpStream for that.

https://git.asonix.dog/safe-async/hyperjive/

seanmonstar:
@avjewe 

> For archival purposes, I would like to fetch an url, and then store not only the bytes of body, but the bytes of the request and the bytes of the returned http headers.

This is possible even now by providing an IO transport that logs the bytes that read and written. For example, reqwest's `ClientBuilder::connection_verbose` ([src](https://github.com/seanmonstar/reqwest/blob/fe0d9688f5b2fd458dbec730c3fbad5434f28e36/src/connect.rs#L1060)) does this.

(Since this is possible, and to tidy up a little, I may hide the comments related to this topic :D)

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+Both the higher-level `Client` and `Server` types have stability concerns.
+
+For the `hyper::Server`:
+
+- The `Accept` trait is complex, and too easy to get wrong. If used with TLS, a
+slow TLS handshake can affect all other new connections waiting for it to
+finish.  - The `MakeService<&IO>` is confusing. The bounds are an assault on
+the eyes.  - The `MakeService` API doesn't allow to easily annotate the HTTP
+connection with `tracing`.  - Graceful shutdown doesn't give enough control.
```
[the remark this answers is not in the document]
seanmonstar: Sigh, I used an auto formatter in vim, to break paragraphs into lines, and it seems it didn't understand lists.

docs/ROADMAP.md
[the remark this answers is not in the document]
seanmonstar: Yes, hyper v0.14 already works with other runtimes, since you can provide your own IO. I've seen it adapted to Google's Fuchsia, and the C API enables curl to adapt its own runtime.

seanmonstar:
@agraven 

> the Service bound on the Connect trait only accepting Uri as request type is rather limiting

Yea, I've heard it both ways: people want a more limited type, or a more expressive type. Either way, I think that proves it's not stable, and should go into `hyper-util`.

docs/ROADMAP.md
seanmonstar: I don't think so. We still want the ability to eventually promote higher-level `Client` or `Server` types.

Maybe having less-nested modules is a good thing, I could be convinced of it.

docs/ROADMAP.md
seanmonstar: I need to fill in a description for this unresolved question... But basically, some is about decoupling from Tokio versions, but another part is to make it less awkward for anything that's using hyper without Tokio. I'm not strongly convinced either way yet.

docs/ROADMAP.md
seanmonstar: One example is Finagle's name resolution (and Dtab): https://twitter.github.io/finagle/guide/Names.html

docs/ROADMAP.md
seanmonstar: Is there a benefit to returning a boxed trait object? If the user wanted one, they could convert to boxed, right?

docs/ROADMAP.md
LucioFranco: I would argue that we don't have that many structs in the conn module if we name things correctly it shouldn't be that hard. In addition, we can always put the `Client`/`Server` at the root of the crate rather than in `client`/`server` modules.

docs/ROADMAP.md
LucioFranco: Yeah, I am just thinking about how it might simplify trait bounds but I think for something like hyper its probably right to not make that choice.

docs/ROADMAP.md
LucioFranco: Does it make sense to add something like dtab? Seems quite complex and I don't 100% understand the scope but it seems quite flexible which we might want to support.

seanmonstar: Oh I have no intention of adding something like that to hyper itself. I was simply saying in this section, "known problems", that _some_ `Connect` implementations have wanted to be able to see the path. That doesn't mean it's what we will do, just that there is more uncertainty around the design.

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+> **Unresolved**: What exactly this looks like will only be known after
+> experimentation.
+
+### client
+
+The high-level `hyper::Client` will be removed, along with the
+`hyper::client::connect` module. They will be explored more in `hyper-util`.
+
+As described in *Design*, the `client::conn` module will gain `http1` and
```
Xuanwo, line 211: Will we handle `Upgrade` automatically?

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+
+### error
+
+The `hyper::Error` struct remains in place.
+
+All errors returned from `Error::source()` are made opaque. They are wrapped an
+internal `Opaque` newtype that still allows printing, but prevents downcasting
+to the internal dependency.
+
+A new `hyper::error::Code` struct is defined. It is an opaque struct, with
```
Xuanwo, line 223: How about following the `std::io::ErrorKind` style? As `Code` may be confused with HTTP Status Code.

lucacasonato:
Wow, this is really great. Thanks for writing it up @seanmonstar!

The most exciting changes for me are the separate h1 and h2 public client APIs. I think this will be a really great change. I have had many uses for this over the last year. It'd be especially great if the `hyper::client::http2::SendRequest` is cloneable like `h2::client::SendRequest`, allowing for multiple simultaneous requests on the same `SendRequest` (unlike `hyper::client::conn::SendRequest`).

The more generic connection pool also seems very useful.

The rest of the changes also seem very reasonable. We'll figure out how they impact our integration in Deno when we have a RC release to integrate :-)

Very exciting! 🎉

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+> Alternative: define a non-exhaustive enum. It's not clear that this is
+> definitely better, though. Keeping it an opaque struct means we can add
+> secondary parts to the code in the future, or add bit flags, or similar
+> extensions.
+
+The purpose of `Code` is to provide an abstraction over the kind of error that
+is encountered. The `Code` could be some behavior noticed inside hyper, such as
+an incomplete HTTP message. Or it can be "translated" from the underlying
+protocol, if it defines protocol level errors. For example, an
+`h2::Reason::CANCEL`.
```
acfoltzer, line 235: My concern with any division between errors we use internally and errors that we expose in the public API is how far the abstraction gap is between the two. I think we agree that there's currently too large of a gap between the error details and the `is_<kind>()` predicates that are the primary way to inspect them. What strategy could we use to prevent `Code` from ending up in a similar position? 

Perhaps `Code` could be part of the internal error definition rather than an abstraction around it, while still leaving us with enough internal fields for implementation flexibility? Like if `hyper::error::Kind` was part of the public API. I suppose this basically ends up being the non-exhaustive enum alternative, though...

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+`http2` sub-modules, providing per-version `SendRequest`, `Connection`, and
+`Builder` structs. An `auto` version can be explored in `hyper-util`.
+
+### error
+
+The `hyper::Error` struct remains in place.
+
+All errors returned from `Error::source()` are made opaque. They are wrapped an
+internal `Opaque` newtype that still allows printing, but prevents downcasting
+to the internal dependency.
```
acfoltzer, line 221: Ruling out downcasting but allowing printing means that our internal dependencies will still leak; applications will end up breaking at runtime due to string matching failures rather than downcasting failures.

The only way we can move breakage from runtime to compile time is to make our internal dependencies part of the public API, therefore subjecting all of hyper to their semver decisions. Not good.

I try to avoid them wherever possible, but perhaps this is a case where feature flags could let users opt in to having dependency error types in the public API at the cost of also having to explicitly opt into non-semver-conforming behavior from hyper with respect to those dependencies.

In other words, if you don't opt in, the best you can do is string matching or downcasting, but you won't have any surprise requirement to update your own `h2` dependency by a major version if hyper bumps its own dependency in a minor release. If you _do_ opt in, then `Error::h2_reason()` becomes `pub` instead of `pub(super)`, but you might have to be more strict about your hyper upgrades.

This is... quite a mess, but I think it's also less messy than depending on stability of `Display` impls for downstream metrics and reporting.

davidpdrsn (approved):
I think it looks good! Have nothing else to add that haven't been brought up already 😊

docs/ROADMAP.md
seanmonstar: This is an interesting topic I'd like to hear more about.

I think keeping the two types separate (internal `Kind` and public `Code`) provides flexibility. The current `Kind` enum uses sub-enums to provide more details. So far we've used new variants when the message should be just a little different, but the general error is still the same. Instead of storing a `&'static str` inside the `Error`, we just use different variants to allow printing the different string. I don't believe programmatically know those minor variant differences is useful to a user.

Personally, I subscribe to the thought that there are a very small class of errors that a user can programmatically do something about, and the rest are only addressable by a human looking at the output message. Thus, I'd want the `Code` to be able to represent those, but otherwise not give more detail than needed since it can make code fragile to internal changes.

docs/ROADMAP.md
seanmonstar: As I mentioned in the other comment, I believe that a small group of errors should be handled programmatically, and the rest should just be printed to a human. I like to believe that most users know that format string matching is very fragile, and hopefully not depended on for much.

I think downcasting an error can be useful, but too much coupling to the specific error types that exist in a stack can make a user's software too fragile. They should be able to inspect errors we purposefully expose, but otherwise we should be allowed to change internal dependencies.

I know @nox was in favor of this change, he might have some more thoughts.

docs/ROADMAP.md
acfoltzer: >Personally, I subscribe to the thought that there are a very small class of errors that a user can programmatically do something about, and the rest are only addressable by a human looking at the output message.

I agree when thinking about "what do we do with this request", but our primary desire for programmatic access is to increase the fidelity of metrics that we report to our customers. Our proxy and even the customer's code they might not decide to do anything differently during the course of one request based on an error. But we do have to report that error back to the customer in a format that doesn't assume it's only for human consumption.

For example, some of our customers depend on being able to distinguish types of errors in downstream automated systems that make decisions about, e.g., how much % of their traffic should go through a particular service at a given time. There are certain classes of errors and variations on those classes that they're particularly sensitive to; just one invalid HTTP header, if it comes from their origin, might be enough to warrant shifting traffic, whereas a certain % of generic connection failures might be acceptable.

It might be that handling these on a case-by-case basis and being eagerly willing to extend `Code` is the best approach, and we can try to start `Code` off with enough variants to cover all of the logic we're currently aware of customers using.

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+    and thus could be managed wholy on top of a read-write transport, HTTP/3 is different. Stream
+    creation is shifted to the QUIC protocol, and HTTP/3 needs to be able to use that directly.
+  - This means the existing `Connection` types for both client and server will not be able to
+    accept a QUIC transport so we can add HTTP/3 support.
+
+### Errors
+
+It's not easy to match for specific errors.
+
+The `Error::source()` can leak an internal dependency. For example, a
```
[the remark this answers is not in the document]
nox: I disagree, I think either `h2` should be a public dependency, or every error coming from h2 should be wrapped as a Hyper type and the `h2` inner bits never exposed through `Error::source`.

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+The `Body` trait will be experimented on to see about making it possible to
+return more frame types beyonds just data and trailers.
+
+> **Unresolved**: What exactly this looks like will only be known after
+> experimentation.
+
+### client
+
+The high-level `hyper::Client` will be removed, along with the
+`hyper::client::connect` module. They will be explored more in `hyper-util`.
```
nox, line 209: +1 on removing `hyper::Client`, that would force us (at work) to implement things differently and then we wouldn't need crutches such as https://github.com/hyperium/hyper/pull/2790

docs/ROADMAP.md
nox: I disagree that we should care about whether we could break things through `Display` and `Debug`, but I do think we should rewrap errors in Hyper so that we keep dependencies truly private.

I don't want `Opaque`, I want a wrapper of `h2::Error`, so that we can still downcast *and* not suffer from h2 bumps *and* not even need to depend on h2 at all.

The only reason some arbitrary crates depend on h2 at work is to be able to do `err.downcast::<h2::Error>()`, if that error was wrapped in Hyper, then I would do `err.downcast::<hyper::H2Error>()` and I wouldn't ever depend on h2 directly.

acfoltzer: > I want a wrapper of `h2::Error`, so that we can still downcast and not suffer from h2 bumps and not even need to depend on h2 at all.

Just to make sure I understand: you want a type in `hyper` that mimics `h2::Error` but does not reëxport it? Would that be substantially different from h2-specific variants on the proposed `Code` type?

seanmonstar: I feel that the `Code` proposal in this document would address that.

docs/ROADMAP.md
seanmonstar: It is meant to sound similar to HTTP/2 and HTTP/3 "Error Codes".

docs/ROADMAP.md
Xuanwo: If we got an `Upgrade` header while using `http1` conn like:

```
HTTP/1.1 101 Switching Protocols
Upgrade: http/2
Connection: Upgrade
```

Will we do this upgrade automatically like what we do in `src/upgrade.rs` `Upgrade` now?

docs/ROADMAP.md
LucioFranco: For my curiosity, what would change at work for yall?

nox: I would design my own connection pool that either gives me a connection (instead of the client doing it itself and then immediately issuing the request), and only then I would issue the request myself, so the two steps wouldn't be combined in the client, and I could handle connection errors from request errors separately more easily than with the current stuff.

docs/ROADMAP.md
nox: I guess it does indeed.

nox:
For the connection types, have we ever considered making HTTP/1 APIs look like HTTP/2 APIs? I've often wished there was something similar to `SendRequest`/`SendBody`/`RecvResponse`/`RecvBody` for HTTP/1.

Also, do we plan to make a h1 crate like there is a h2 crate?

docs/ROADMAP.md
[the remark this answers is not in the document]
erickt: note that Fuchsia currently uses hyper v.0.13, but we're in the process of trying to update it to 0.14 in https://fuchsia-review.googlesource.com/c/fuchsia/+/611683.

nox:
In my previous comment I said:

> For the connection types, have we ever considered making HTTP/1 APIs look like HTTP/2 APIs? I've often wished there was something similar to SendRequest/SendBody/RecvResponse/RecvBody for HTTP/1.

Something that I just realised today: if there was such a API, it would mean I could just use it for all HTTP/1 and HTTP/2 incoming requests, and I could more easily send back a specific HTTP/2 RST_STREAM code whenever I need to on receiving the request, through `SendResponse::send_reset`.

seanmonstar:
I do think we could much of what is in `hyper::proto::h1` into an `http1` crate. That's an eventual goal, but not something that I personally would hope to tackle for hyper 1.0, since we only have so much time and energy, and it wouldn't block hyper 1.0.

docs/ROADMAP.md
```diff
@@ -0,0 +1,382 @@
...
+will start in `hyper-util`.
+
+### Focus on the `Connection` level
+
+As mentioned in the *Known Issues*, the higher-level `Client` and `Server` have
+stability and complexity problems. Therefore, for hyper 1.0, the main API will
+focus on the "lower-level" connection types. The `Client` and `Server` helpers
+will be moved to `hyper-util`.
+
+## Public API
```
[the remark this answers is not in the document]
seanmonstar: Some of those types are public simply to allow propagating the `Send`iness of the user's provide `Service` and `Body` types, so that someone can use a single-threaded executor and use `!Send` types. If it were possible to make that happen without having those types quasi-public, I would do that.

The `MakeServiceRef` will be going away from hyper core.

docs/ROADMAP.md
seanmonstar: hyper doesn't automatically handle HTTP/1 to HTTP/2 upgrades, but it does support generic upgrades by giving back the IO type and letting the user handle the rest. I think this will still be the case.

Issue #2797, opened 2022-03-29
C API for server?


lucifer1004:
It seems that the server-side code has not been exposed by the C API. Is this on the roadmap? Or are there any known obstacles?

I am trying to write a Julia wrapper via the C API and am studying the FFI code.

PR #2807, opened 2022-04-06
[WIP] add server-side C API

lucifer1004:
Related to #2797. The API can be called but does not really work yet.

Since I am not yet familiar with the whole project, I would like to hear more advice from the community, especially those who previously worked on the C API.

### DIFF
--- a/src/ffi/server.rs
+++ b/src/ffi/server.rs
@@ -216,17 +216,12 @@ impl Service<Request<Body>> for hyper_service {
         (self.service_func)(self.userdata, req_ptr, res_ptr);
 
         let hyper_res = non_null! {
-            &mut *res_ptr ?= Box::pin(async { Err(crate::error::Error::new(
+            Box::from_raw(res_ptr) ?= Box::pin(async { Err(crate::error::Error::new(
                 crate::error::Kind::Io
             ))})
         };
 
-        Box::pin(async move {
-            Ok(std::mem::replace(
-                &mut hyper_res.0,
-                Response::new(Body::empty()),
-            ))
-        })
+        Box::pin(async move { Ok((*hyper_res).0) })
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/ffi/server.rs
```diff
@@ -0,0 +1,236 @@
...
+use crate::common::exec::ConnStreamExec;
+use crate::server::conn::{Connection, Http};
+use crate::service::Service;
+
+use super::error::hyper_code;
+use super::http_types::{hyper_request, hyper_response};
+use super::io::hyper_io;
+use super::task::{hyper_executor, hyper_task, WeakExec};
+
+pub struct hyper_http(Http<WeakExec>);
```
seanmonstar, line 20: You couldn't know this, but I'm not a fan of the name that's currently `Http` in hyper. I think this C API type might want to follow the client-side naming, so maybe `hyper_serverconn_options`? And then similarly with the related functions.

src/ffi/server.rs
```diff
@@ -0,0 +1,236 @@
...
+    } ?= ptr::null_mut()
+}
+
+pub struct hyper_service {
+    service_func: hyper_service_callback,
+    userdata: *mut c_void,
+}
+
+type hyper_service_callback =
+    extern "C" fn(*mut c_void, *mut hyper_request, *mut hyper_response) -> ();
```
seanmonstar, line 177: So, it looks like this would require that in C, the callback is synchronous. It must return a response, it can't return a future. I worry that this might make servers in C just do everything in a blocking style, which would be bad for the server.

Would it be hard to have this callback return a `*mut hyper_task`? I haven't thought too hard about it

seanmonstar (changes requested):
Thanks for submitting this, this looks like a nice start! I left some comments inline to continue the discussion. <3

PR #2812, opened 2022-04-11
Refactor(capi): make early returns consistent in C examples

Some-Dood:
Hello there! I saw the blog posts regarding this project's campaign for 1.0. 🥳

I figured I could help out with the C examples by making the code style regarding early returns more consistent. I found that (prior to this PR) the examples tended to mix between coding styles. On a personal note, I much prefer reading code with early returns anyway since there are less indentations/nested braces to worry about. 😅

So with that said, this PR simply refactors instances of nested error-handling logic as early returns. The result is more "linear" code with less indentations. Hooray!

To help make the review easier, I have also separated the commits into the specific decisions I took while refactoring. Hope this helps out a little bit with the documentation for the C API. Thanks!


seanmonstar (approved):
Thanks for making things more consistent 😊

PR #2814, opened 2022-04-17
fix(examples): don't send authority in tower

kiron1:
Sending the authority in the request of this example might be confusing
for new users, since this kind of requests are for proxy not for regular
clients.


seanmonstar:
This example will likely be going away in 1.0, so I'm going to close. Thanks for wanting to make it better!

Issue #2773, opened 2022-03-11
Can't get local address of `AddrStream`


lilyball:
**Version**
hyper v0.14.17

**Description**
`AddrStream` offers a [`remote_addr()`](https://docs.rs/hyper/0.14.17/hyper/server/conn/struct.AddrStream.html#method.remote_addr) method that returns the remote (peer) address of the connection. But there's no equivalent `local_addr()` method. `TcpStream` has [such a method](https://docs.rs/tokio/1.17.0/tokio/net/struct.TcpStream.html#method.local_addr), but the only way to get that from `AddrStream` is [`AddrStream::to_inner()`](https://docs.rs/hyper/0.14.17/hyper/server/conn/struct.AddrStream.html#method.into_inner) that consumes the `AddrStream` entirely (which is to say, there's no `as_inner()` method to get a reference).

This is really frustrating because I want to log the local IP address on incoming connections (I'm listening on `0.0.0.0` and I want to know which interface the connection is coming in on). So I either need an `AddrStream::local_addr()` method or I need an `AddrStream::as_inner()` method[^as_ref] that I can then use to call `TcpStream::local_addr()`.

[^as_ref]: Or maybe an `<AddrStream as AsRef<TcpStream>>` impl.

PR #2816, opened 2022-04-20
fix(server): use local address in AddrStream

ilyatrefilov:
Expose local address of tcp connection in AddrStream for the purpose mentioned in #2773. It is safe to call underlying `getsockname`, because socket is already connected to the peer at the point of `socket.local_addr()?` execution.


seanmonstar (approved):

seanmonstar:
Thanks :)

PR #2817, opened 2022-04-24
Refactor: assorted code readability improvements in `h1/conn.rs`

Some-Dood:
Hello there! This PR features assorted refactoring techniques that hopefully improves readability. No behavior change is expected.

Most of the refactoring involved using the [`matches!` macro](https://doc.rust-lang.org/core/macro.matches.html) instead of explicit Boolean `match` expressions. Other refactoring involved transforming logic to early returns. The overall result is that the code is now a little bit less nested. 🎉

With that said, please feel free to inform me if some of the changes need improvement. Thanks!

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -442,12 +442,16 @@ where
         // determined we couldn't keep reading until we knew how writing
         // would finish.
 
-        if !matches!(self.state.reading, Reading::Init) {
-            return;
-        }
+        match self.state.reading {
+            Reading::Continue(..) | Reading::Body(..) | Reading::KeepAlive | Reading::Closed => {
+                return
+            }
+            Reading::Init => (),
+        };
 
-        if self.can_write_body() {
-            return;
+        match self.state.writing {
+            Writing::Body(..) => return,
+            Writing::Init | Writing::KeepAlive | Writing::Closed => (),
         }
 
         if !self.io.is_read_blocked() {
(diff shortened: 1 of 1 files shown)


src/proto/h1/conn.rs
```diff
@@ -493,22 +485,18 @@ where
...
             _ => false,
         }
     }
 
     pub(crate) fn can_write_body(&self) -> bool {
-        match self.state.writing {
-            Writing::Body(..) => true,
-            Writing::Init | Writing::KeepAlive | Writing::Closed => false,
-        }
+        matches!(self.state.writing, Writing::Body(..))
```
seanmonstar, line 499: In some of these cases, the exhaustive match was on purpose to make sure any changes to the variants were always considered here... I do like that this makes it more succinct. I don't know which is the better choice...

Some-Dood: I totally understand where you're coming from here. State machines are really tricky to get right.

Which specific cases are you referring to aside from the one highlighted here? I'd be glad to revert them just to relieve any doubts for future maintainability. 👍

seanmonstar: Probably just the ones that were already exhaustive (didn't have a `_` branch).

src/proto/h1/conn.rs
```diff
@@ -446,16 +442,12 @@ where
         // determined we couldn't keep reading until we knew how writing
         // would finish.
 
-        match self.state.reading {
-            Reading::Continue(..) | Reading::Body(..) | Reading::KeepAlive | Reading::Closed => {
-                return
-            }
-            Reading::Init => (),
-        };
+        if !matches!(self.state.reading, Reading::Init) {
```
seanmonstar, line 445: I guess and this one? It's hard to remember if that was on purpose of not...

seanmonstar:
Very nice, thanks again!

Issue #2355, opened 2020-12-10
add conn_idle_timeout for server


silence-coding:
Should we add idle timeout for the server? Without idle timeout, the service cannot quickly release idle connections. In this case, the client can set up a large number of idle connections with the hyper server, exhausting the hyper resources.

CodesInChaos:
Even if the server implements idle timeouts, an attacker can simply send the data very slowly [(slowloris attack)](https://en.wikipedia.org/wiki/Slowloris_(computer_security)).

silence-coding:
Although this capability cannot completely solve slow attacks, I think that the server should have this capability to reduce the problems in other scenarios.

PR #2827, opened 2022-05-06
[WAIT] add conn idle timeout

silence-coding:
#2355 
I'd like to ask if it's appropriate to carry out an idle check in these


silence-coding:
This is only a demo to check whether there is a big problem in selecting the time point for connection timeout. I'll re-implement one later on, based on the community's recommendations.

silence-coding:
@paolobarbolini Do you have any good suggestions for me?

deputinizer:
Spelling? Missing `u`
```rust
idle_timeot: Arc<Mutex<Pin<Box<Sleep>>>>
```


Also is changing `Cargo.toml` required?
```toml
log = "0.4.14"
```

And why not use: 
```rust

if let Err(_) = timeout(Duration::from_secs(10), fut).await {
    println!("did not receive value within 10 seconds");
}

```
https://docs.rs/tokio/latest/tokio/time/fn.timeout.html
(im not very familiar with server part of hyper)

silence-coding:
> Spelling? Missing `u`
> 
> ```rust
> idle_timeot: Arc<Mutex<Pin<Box<Sleep>>>>
> ```
> 
> Also is changing `Cargo.toml` required?
> 
> ```toml
> log = "0.4.14"
> ```
> 
> And why not use:
> 
> ```rust
> if let Err(_) = timeout(Duration::from_secs(10), fut).await {
>     println!("did not receive value within 10 seconds");
> }
> ```
> 
> https://docs.rs/tokio/latest/tokio/time/fn.timeout.html (im not very familiar with server part of hyper)

@deputinizer Thank you very much, and I'll take the time to submit a formal pull request later, with the corresponding tests and comments. This submission is my feasibility verification of this function, so I write it casually. My main concern is that there may be problems with this core implementation, especially for h2, resulting in a major rewrite of the submitted code.

silence-coding:
@deputinizer Can you tell me which line of code it is?

```
if let Err(_) = timeout(Duration::from_secs(10), fut).await {
    println!("did not receive value within 10 seconds");
}
```

PR #2830, opened 2022-05-10
Replace transmute with bounded  manual `Send` impl

djkoloski:
Requires https://github.com/hyperium/h2/pull/614 before this will compile.

We can avoid a transmute and some questionable code by using a specially-crafted conditional `Send` impl. `Sendable` and the impl are inside a `const _: () = { ... }` block to prevent the otherwise unsound `Sendable` type from being visible to the rest of the crate.

I believe that both versions are equally sound, but this helps prevent future issues and makes auditing the code more straightforward.

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -5,7 +5,6 @@ use http::HeaderMap;
 use pin_project_lite::pin_project;
 use std::error::Error as StdError;
 use std::io::{self, Cursor, IoSlice};
-use std::mem;
 use std::task::Context;
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 use tracing::{debug, trace, warn};
@@ -409,15 +408,14 @@ fn h2_to_io_error(e: h2::Error) -> io::Error {
     }
 }
 
-struct UpgradedSendStream<B>(SendStream<SendBuf<Neutered<B>>>);
+struct UpgradedSendStream<B>(SendStream<SendBuf<B>>);
 
 impl<B> UpgradedSendStream<B>
 where
     B: Buf,
 {
     unsafe fn new(inner: SendStream<SendBuf<B>>) -> Self {
-        assert_eq!(mem::size_of::<B>(), mem::size_of::<Neutered<B>>());
-        Self(mem::transmute(inner))
+        Self(inner)
     }
 
     fn reserve_capacity(&mut self, cnt: usize) {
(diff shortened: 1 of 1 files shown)


src/proto/h2/mod.rs
```diff
@@ -442,30 +440,24 @@ where
...
-#[repr(transparent)]
-struct Neutered<B> {
-    _inner: B,
-    impossible: Impossible,
-}
-
-enum Impossible {}
-
-unsafe impl<B> Send for Neutered<B> {}
+const _: () = {
```
RalfJung, line 447: Why use an unnamed const rather than a private module?

src/proto/h2/mod.rs
```diff
@@ -409,15 +408,14 @@ fn h2_to_io_error(e: h2::Error) -> io::Error {
     }
 }
 
-struct UpgradedSendStream<B>(SendStream<SendBuf<Neutered<B>>>);
+struct UpgradedSendStream<B>(SendStream<SendBuf<B>>);
```
nox, line 411: I don't like this. The whole point of `Neutered<B>` is that you can't create one, whereas here someone can use the field directly and send a `B` over the channel.

djkoloski: My impression was that `new`/`as_inner_unchecked` were meant to act as guards against sending a `B`. For example, the safety conditions for `new` would be "This must be the only instance of the stream" and `as_inner_unchecked` would be "No messages must be sent through the returned stream".

It's a little unclear to me what the desired API is since this is all private. If the goal is to make the stream impossible to use incorrectly, even in private contexts, then `as_inner_unchecked` shouldn't exist. Even then, someone with private access could access the inner field and do the transmute by hand, so I don't think it's feasible to prevent misuse in private contexts.

nox: > My impression was that `new`/`as_inner_unchecked` were meant to act as guards against sending a `B`. For example, the safety conditions for `new` would be "This must be the only instance of the stream" and `as_inner_unchecked` would be "No messages must be sent through the returned stream".

Yes they are, but as it doesn't live in its own module, the rest of the code in that module is free to do `.0` and access the `SendBuf<B>` without the unsafety of `as_inner_unchecked`.

nox: (Hit cmd-enter too early, sorry)

> If the goal is to make the stream impossible to use incorrectly, even in private contexts, then `as_inner_unchecked` shouldn't exist.

The goal is to make the stream impossible to use incorrectly *in safe code*

djkoloski: Perhaps we should move `UpgradedSendStream` into a separate module and make it public. That way we can use visibility to prevent users from accessing `.0` while providing an API without using `transmute` under the hood.

djkoloski: I attempted this and found that `write` was being called from the `AsyncWrite` impl for `H2Upgraded`. Should these uses additionally have `as_inner_unchecked` for some reason? Or am I misunderstanding why the original implementation `Neutered` the stream?

nox: There is no `H2Upgraded` in this patch, did you mean another type?

The `Neutered<_>` thing is to show that we are not using that variant `SendBuf::Buf` there, only `SendBuf::Cursor`.

seanmonstar:
Sorry for the late update, this has been refactor recently on master: #3967

PR #2831, opened 2022-05-10
Remove unsafe from proto::h2

### LINKED ISSUE #2829: Document how hyper is using UpgradedSendStream safely

On Fuchsia, we are updating our vendored version of Hyper from 0.13 to 0.14 in https://fuchsia-review.googlesource.com/c/fuchsia/+/611683/. As part of this we are reviewing the hyper change. One portion of the codebase we are struggling with is [src/proto/h2](https://github.com/hyperium/hyper/blob/master/src/proto/h2). In #2523, @nox introduced the unsafe `UpgradedSendStream` and `Neutered`, and it's not straightforward what it does, and how we know hyper is using it correctly.

There's a nice [discussion](https://github.com/hyperium/hyper/pull/2523/files#r628043384) though in #2523 that gives some context. It would be helpful if a version of this discussion was added to the code. That would make it easier to review vendored code.

erickt:
Back in #2523, @nox introduced the notion of an UpgradedSendStream, to support the CONNECT method of HTTP/2. This used `unsafe {}` to support `http_body::Body`, where `Body::Data` did not implement `Send`, since the `Data` type wouldn't be sent across the stream once upgraded.

Unfortunately, according to this [thread], I think this may be undefined behavior, because this relies on us requiring the transmute to execute.

This patch fixes the potential UB by adding the unncessary `Send` constraints. It appears that all the internal users of
`UpgradeSendStream` already work with `http_body::Body` types that have `Send`-able `Data` constraints. We can add this constraint without breaking any external APIs, which lets us remove the `unsafe {}` blocks.

I filed https://github.com/rust-lang/unsafe-code-guidelines/issues/337 for formal guidance on this pattern.

Fixes #2829

[thread]: https://users.rust-lang.org/t/is-a-reference-to-impossible-value-considered-ub/31383

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -5,7 +5,6 @@ use http::HeaderMap;
 use pin_project_lite::pin_project;
 use std::error::Error as StdError;
 use std::io::{self, Cursor, IoSlice};
-use std::mem;
 use std::task::Context;
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 use tracing::{debug, trace, warn};
@@ -409,63 +408,32 @@ fn h2_to_io_error(e: h2::Error) -> io::Error {
     }
 }
 
-struct UpgradedSendStream<B>(SendStream<SendBuf<Neutered<B>>>);
+struct UpgradedSendStream<B: Buf>(SendStream<SendBuf<B>>);
 
 impl<B> UpgradedSendStream<B>
 where
     B: Buf,
 {
-    unsafe fn new(inner: SendStream<SendBuf<B>>) -> Self {
-        assert_eq!(mem::size_of::<B>(), mem::size_of::<Neutered<B>>());
-        Self(mem::transmute(inner))
+    fn new(inner: SendStream<SendBuf<B>>) -> Self {
+        Self(inner)
     }
 
     fn reserve_capacity(&mut self, cnt: usize) {
-        unsafe { self.as_inner_unchecked().reserve_capacity(cnt) }
+        self.0.reserve_capacity(cnt)
     }
 
     fn poll_capacity(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<usize, h2::Error>>> {
-        unsafe { self.as_inner_unchecked().poll_capacity(cx) }
+        self.0.poll_capacity(cx)
     }
 
     fn poll_reset(&mut self, cx: &mut Context<'_>) -> Poll<Result<h2::Reason, h2::Error>> {
-        unsafe { self.as_inner_unchecked().poll_reset(cx) }
+        self.0.poll_reset(cx)
     }
 
     fn write(&mut self, buf: &[u8], end_of_stream: bool) -> Result<(), io::Error> {
         let send_buf = SendBuf::Cursor(Cursor::new(buf.into()));
-        unsafe {
-            self.as_inner_unchecked()
-                .send_data(send_buf, end_of_stream)
-                .map_err(h2_to_io_error)
-        }
-    }
-
-    unsafe fn as_inner_unchecked(&mut self) -> &mut SendStream<SendBuf<B>> {
-        &mut *(&mut self.0 as *mut _ as *mut _)
-    }
-}
-
-#[repr(transparent)]
-struct Neutered<B> {
-    _inner: B,
-    impossible: Impossible,
-}
-
-enum Impossible {}
-
-unsafe impl<B> Send for Neutered<B> {}
-
-impl<B> Buf for Neutered<B> {
-    fn remaining(&self) -> usize {
-        match self.impossible {}
-    }
-
-    fn chunk(&self) -> &[u8] {
-        match self.impossible {}
-    }
-
-    fn advance(&mut self, _cnt: usize) {
-        match self.impossible {}
+        self.0
+            .send_data(send_buf, end_of_stream)
+            .map_err(h2_to_io_error)
     }
 }
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -341,7 +341,7 @@ where
                                 let (pending, on_upgrade) = crate::upgrade::pending();
                                 let io = H2Upgraded {
                                     ping,
-                                    send_stream: unsafe { UpgradedSendStream::new(send_stream) },
+                                    send_stream: UpgradedSendStream::new(send_stream),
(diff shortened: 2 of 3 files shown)


src/proto/h2/mod.rs
```diff
@@ -409,63 +408,32 @@ fn h2_to_io_error(e: h2::Error) -> io::Error {
...
-    unsafe fn as_inner_unchecked(&mut self) -> &mut SendStream<SendBuf<B>> {
-        &mut *(&mut self.0 as *mut _ as *mut _)
-    }
-}
-
-#[repr(transparent)]
-struct Neutered<B> {
-    _inner: B,
-    impossible: Impossible,
-}
```
RalfJung, line 453: FWIW, there have even been proposals that would make `Neutered<B>` have size 0, since it cannot ever be inhabited anyway. So this is definitely an incorrect pattern.

erickt:
This is an alternative version of #2830, where this one adds `Send` constraint, but #2830 uses unsafe to not add `Send`.

src/proto/h2/mod.rs
djkoloski: That is checked in `UpgradedSendStream`'s `new`, so I think that should be okay. It's not idea but it shouldn't produce any UB.

nox:
Let's go for #2830 instead, so that we avoid the breaking change.

I'm not sure I understand why you think there is UB though. The transmute doesn't happen on an uninhabited type, it happens on an enum of which one of the variants has an uninhabited field.

erickt:
@nox - we talked about this more on https://github.com/rust-lang/unsafe-code-guidelines/issues/337. I now think this is sound. The UB I thought we had was that directly transmuting a value to something that contains an Impossible type is UB. But instead you are instead ultimately transmuting a pointer to an Impossible type, which is sound. 

The one concern though that’s brought up in the thread is that there’s no guarantee of the layout of a type that includes an impossible type. There’s talk about always treating those types as zero sized. 

> Let's go for https://github.com/hyperium/hyper/pull/2830 instead, so that we avoid the breaking change.

Sure, that’s fine with me. My main interest is to at least document it to save other people’s time when auditing hyper. 

There’s a chance though this isn’t a breaking change (at least for external users). I traced that the proto::h2::client already requires `B::Data: Send`, but I haven’t finished going through all the users of proto::h2::server yet. It’s a little challenging with all the generic type parameters.

src/proto/h2/mod.rs
RalfJung: Ah, you are guarding against the layout optimization there, okay.

This is still most definitely UB though. In fact if there is any uncontroversial UB in Rust, `unreachable_unchecked` would be first on that list, and creating an instance of an uninhabited type is second. [The reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) is also clear about this and never was there any controversy in the UCG that this code should be UB.

RalfJung:
> we talked about this more on https://github.com/rust-lang/unsafe-code-guidelines/issues/337. I now think this is sound. The UB I thought we had was that directly transmuting a value to something that contains an Impossible type is UB. But instead you are instead ultimately transmuting a pointer to an Impossible type, which is sound.

As long as it's a *raw* pointer, that's okay.
For a reference or `Box`, that's not okay.

src/proto/h2/mod.rs
nox: There is never a `Neutered<T>` being created anywhere, that's the thing.

nox:
There is never, ever, a reference to an uninhabited value, or an uninhabited value, being created by my code.

As for `Neutered<B>` having a different size than `B` itself, I think that directly contradicts the meaning of `#[repr(transparent)]`.

nox:
The problem

> There’s a chance though this isn’t a breaking change (at least for external users). I traced that the proto::h2::client already requires `B::Data: Send`, but I haven’t finished going through all the users of proto::h2::server yet. It’s a little challenging with all the generic type parameters.

You may be right. `H2Stream` is marked `pub` (probably because of the `unstable` feature), so to me it's a breaking change as it restricts one of the type's impls. (As a side note I kinda dislike that `unstable` just exposes internal details like that, @seanmonstar do you think we could get rid of it one day?) Maybe we can just add the bound after all.

src/proto/h2/mod.rs
RalfJung: Ah, I see, that makes sense then that Miri didn't complain about hyper. :)

RalfJung:
> There is never, ever, a reference to an uninhabited value, or an uninhabited value, being created by my code.

All right, sorry for the confusion.

> As for Neutered<B> having a different size than B itself, I think that directly contradicts the meaning of #[repr(transparent)].

Yeah, https://github.com/rust-lang/rust/issues/96921 was recently created to discuss this.

erickt:
> You may be right. H2Stream is marked pub (probably because of the unstable feature), so to me it's a breaking change as it restricts one of the type's impls. (As a side note I kinda dislike that unstable just exposes internal details like that, @seanmonstar do you think we could get rid of it one day?) Maybe we can just add the bound after all.

Ah, nuts. So I finished tracing, and `H2Stream` is one of those quasi-public types:

* [hyper::proto::h2::H2Stream](https://github.com/hyperium/hyper/blob/master/src/proto/h2/server.rs#L369) is marked `pub`.
* [hyper::proto::h2](https://github.com/hyperium/hyper/blob/master/src/proto/mod.rs#L17) is `pub(crate)` so it's not directly accessible.
* However, `H2Stream` is used in [hyper::common::exec::ConnStreamExec](https://github.com/hyperium/hyper/blob/master/src/common/exec.rs#L19), which is also `pub`.
* But [hyper::common::exec](https://github.com/hyperium/hyper/blob/master/src/common/mod.rs#L16) is also `pub(crate)`. 
* Finally, `ConnStreamExec` is used as a type constraint in [hyper::server::Server](https://github.com/hyperium/hyper/blob/master/src/server/server.rs#L115).

nox:
> This changes a bounds on a public-ish type, which I believe could be a breaking change. To verify, can the `single_threaded` example in this repo make its `Buf` type `!Send`, and then this change breaks that?

I think it breaks using custom executors with non-`Send` buffers. Pretty niche if you ask me.

erickt:
You are correct, updating `single_threaded` to change `Body` to return a `!Send` impl of `Buf` does val with `*const () cannot be sent between threads safely", so I'll close this. Thanks all!

PR #2836, opened 2022-05-18
fix(http1): don't add content-length to head resp

### LINKED ISSUE #2835: Server incorrectly adds content-length: 0 to Http1 Head responses
**Version**
0.14.18

**Platform**
```
Darwin C02DN4K3MD6R 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64 i386 MacBookPro16,1 Darwin
```

**Description**
Hyper incorrectly sets `content-length: 0` on Head responses without content-length.
According to https://www.ietf.org/archive/id/draft-ietf-httpbis-semantics-19.html#name-content-length: 
"A server MAY send a Content-Length header field in a response to a HEAD request ([Section 9.3.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-semantics-19.html#HEAD)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a response if the same request had used the GET method."

I tried this code:

```
use std::net::SocketAddr;

use hyper::{
    header::HOST,
    service::{make_service_fn, service_fn},
    Body, Client, Request, Server, Version,
};

type Error = Box<dyn std::error::Error + Send + Sync + 'static>;

#[tokio::main]
async fn main() -> Result<(), Error> {

    let on_request = service_fn(|r| {
        let uri = format!("http://{}", r.headers()[HOST].to_str().unwrap());
        let req = Request::builder()
            .method(r.method())
            .uri(uri)
            .version(Version::HTTP_11)
            .body(Body::empty())
            .unwrap();
        let client = Client::new();
        client.request(req)
    });
    let make_svc = make_service_fn(move |_| async move { Ok::<_, Error>(on_request) });

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    let server = Server::bind(&addr).serve(make_svc);

    if let Err(e) = server.await {
        eprintln!("server error: {}", e);
    }

    Ok(())
}
```
and 
```
curl http://localhost:3000 -H "Host: eaufavor.net" -I
curl http://eaufavor.net -I     
```

I expected to see this happen: 
Both curl responses have no `content-length` header.

Instead, this happened:
The response for the request that goes through hyper has `content-length: 0`

jannes:
Head responses should not have content-length implicitly set by hyper.
Fixes #2835

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -1,8 +1,6 @@
 use std::fmt::{self, Write};
 use std::mem::MaybeUninit;
 
-#[cfg(all(feature = "server", feature = "runtime"))]
-use tokio::time::Instant;
 #[cfg(any(test, feature = "server", feature = "ffi"))]
 use bytes::Bytes;
 use bytes::BytesMut;
@@ -10,6 +8,8 @@ use bytes::BytesMut;
 use http::header::ValueIter;
 use http::header::{self, Entry, HeaderName, HeaderValue};
 use http::{HeaderMap, Method, StatusCode, Version};
+#[cfg(all(feature = "server", feature = "runtime"))]
+use tokio::time::Instant;
 use tracing::{debug, error, trace, trace_span, warn};
 
 use crate::body::DecodedLength;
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Looks great, thanks for adding the test!

jannes:
thank you for the quick response :)

PR #2837, opened 2022-05-20
feat(client): add option to HttpConnector to enable or disable

### LINKED ISSUE #2833: Add option to HttpConnector to enable or disable HttpInfo
By default hyper will try to get the `HttpInfo` for each Response even if they are not needed in most cases. The bottom line is that `getpeername` and `getsockname` are actually very slow and they can become a performance bottleneck when making a lot of requests

Just provide an API like client::Builder::set_host

mnpw:
This adds `http_info` configuration option in `http::Config`. It can be used to enable or disable `HttpInfo`. 

`HttpInfo` is set through methods defined on `TcpStream` (`peer_addr()` and `local_addr()`). Configuration options have to be made available to `TcpStream` in order to enforce `http_info`. `HttpConnection` is added to achieve this.
 
Fixes #2833

Ps: Rust and oss newbie so apologies if I missed some obvious issue 😅

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -6,14 +6,16 @@ use std::marker::PhantomData;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::pin::Pin;
 use std::sync::Arc;
-use std::task::{self, Poll};
+use std::task::{self, Poll, Context};
 use std::time::Duration;
+use std::ops::{Deref, DerefMut};
 
 use futures_util::future::Either;
 use http::uri::{Scheme, Uri};
 use pin_project_lite::pin_project;
 use tokio::net::{TcpSocket, TcpStream};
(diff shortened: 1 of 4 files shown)


ho-229:
I think you should open a new PR to submit breaking changes.

src/client/connect/http.rs
```diff
@@ -256,7 +274,7 @@ where
     R: Resolve + Clone + Send + Sync + 'static,
     R::Future: Send,
 {
-    type Response = TcpStream;
+    type Response = HttpConnection;
```
seanmonstar, line 277: This would be a breaking change. Can we provide the configuration without needing a new type?

seanmonstar (changes requested):
Thanks for the PR! I'm in favor of adding the configuration, though I believe we can do so without the breaking change I commented about below. :)

ho-229:
Are you still working on it🤔

Issue #2864, opened 2022-05-21
Hyper currently does not support client body read timeout


qinyushun:
**Version**
0.13.10

**Description**

When the Hyper serves as the server, the client body reading timeout is not implemented. As a result, the connection may be fully occupied by slow attacks from the client and memory may be exhausted.

The following is an example of an attack:

1. Use Hyper to create an HTTP server and perform HTTP1.1 and HTTP2 tests.

   Construct the following test code (for details, see the official example https://github.com/hyperium/hyper/blob/master/examples/params.rs). The sample code uses the `to_bytes` method to obtain the client body packet and print the body content.

   Server code:

   ```rust
   // Respond to all requests and use to_bytes to read the body.
   async fn handle(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
       let bytes = hyper::body::to_bytes(req).await?;
       let body: String = String::from_utf8_lossy(bytes.as_ref()).to_string();
       println!("body : {:?}", body); // Test whether the to_bytes method is suspended.
       Ok(Response::new("Hello, World!\n".into()))
   }
   
   #[tokio::main]
   async fn main() {
       let addr = SocketAddr::from(([0, 0, 0, 0], 1337));
       let make_svc = make_service_fn(|_conn| async {
           Ok::<_, Infallible>(service_fn(handle))
       });
       let server = Server::bind(&addr).serve(make_svc);
       if let Err(e) = server.await {
           eprintln!("server error: {}", e);
       }
   }
   ```

   1）Test the HTTP1 protocol stack. The HTTP1 protocol stack is a hyper project. The code is as follows:

   ![image](https://user-images.githubusercontent.com/84489880/169632886-061b8baf-905c-4ef5-be0e-9fd1b177ec9c.png)

    Key code: conn.rs, io.rs, dispatch.rs, body.rs, and decode.rs.

   The test result shows that no data with the length of Content-Length is received. The io.rs keeps reading the data.

      ```rust
      fn read_mem(&mut self, cx: &mut task::Context<'_>, len: usize) -> Poll<io::Result<Bytes>> {
          if !self.read_buf.is_empty() {
              let n = std::cmp::min(len, self.read_buf.len());
              Poll::Ready(Ok(self.read_buf.split_to(n).freeze()))
          } else {
              println!("291 begin poll_read_from_io {:?}", len);
              let n = ready!(self.poll_read_from_io(cx))?;
              println!("291 end poll_read_from_io {:?}", len);
[...]

PR #2865, opened 2022-05-21
feat(h2): add read body timeout

silence-coding:


silence-coding:
Fix https://github.com/hyperium/hyper/issues/2864

silence-coding:
I'm hesitant to provide uniform configuration in hyper server builder, like http1_header_read_timeout.

seanmonstar:
As discussed in #2864, I don't believe we need to have this in hyper.

lidong14:
@seanmonstar I think this is necessary, because as an HTTP framework, body timeout is a basic capability. like  tomcat , jetty.

silence-coding:
> As discussed in #2864, I don't believe we need to have this in hyper.

When hyper is used as the proxy, req.body_mut().data().await is not applicable.

seanmonstar:
> as an HTTP framework, body timeout is a basic capability

Sure, I agree it is necessary to be able to set some sort of timeout. But since it's possible and easy to do outside of hyper, I don't believe it belongs *in* hyper. A util would be fine.

> When hyper is used as the proxy, req.body_mut().data().await is not applicable.

A body adapter could be written. Some sort of `impl HttpBody for TimeoutPerBuf<impl HttpBody>`.

silence-coding:
-> A body adapter could be written. Some sort of impl HttpBody for TimeoutPerBuf<impl HttpBody>.
This is a good idea. I think it can be out of hyper, but it should be in the ecosystem provided by hyper. Can more security-related features be added to version 1.0, such as idle connection timeout, read/write timeout, and others?

seanmonstar:
> but it should be in the ecosystem provided by hyper

It could be a fair candidate for the eventual `hyper-util` crate!

> Can more security-related features be added to version 1.0

The general answer is yes, but I'd recommend opening an issue to discuss any proposed feature. At the same time, if it's something that can reasonably be implemented *outside* of hyper, then it probably should stay external.

I'm going to close this based on previous discussion.

PR #2878, opened 2022-05-27
feat(client, server): remove the `tcp` cargo feature and related code

### LINKED ISSUE #2856: Remove the `tcp` cargo feature
Remove the `tcp` cargo feature, and related `tokio::net::TcpStream` integration (`TcpStream::connect` and `TcpListener`) (since we are already removing `client::connect` and `server::accept`).

MrGunflame:
Removes the `tcp` cargo feature and related code from server and client. I've kept `client::connect` even thought it doesn't serve much use without tcp-related features.

Closes #2856.


seanmonstar:
Woah, awesome work, thanks!

MrGunflame:
It should be noted that this change currently breaks pretty much every test and example, which all rely on the Server/Client type.

seanmonstar:
Alright, the meta pieces are in place now, so full-speed ahead on the feature work...

Hm, it looks like we'll need some sort of local utility for the unit tests...

MrGunflame:
Yes, I think all tcp feature code is gone now. I've been working on converting the tests  to use conn client/server mod, but I'm not quite happy with it. The examples also pose a similar problem.

seanmonstar:
I'm looking through the tests right now, and it seems like a good time to clean up this mess anyways 😅 . Most of the client `test!` stuff can probably have the inner part of the macro just use `conn` and `TcpStream` instead of `Client`. The `conn` impl tests are probably fine as-is.

... As I scroll more, I'm not sure there's many tests that are checking the higher-level `Client` specifically... If they are, we can comment them out and move them to the https://github.com/hyperium/hyper-util crate as part of the `Client` move.

seanmonstar:
Aww, sorry, it looks like merging #2896 caused conflicts with this 🙈

MrGunflame:
That's fine, I'll have to update the examples anyways before this works.

Robert-Cunningham:
When we merge this PR, will `Exec` struct in `common/exec.rs` have a purpose anymore? Or will we then want to replace `Exec` everywhere `Box<dyn Executor>` (and simultaneously add `TokioExecutor` to `hyper-utils`)?

MrGunflame:
I think you have an argument there to replace `Exec` with an `Arc<dyn Executor>`, but I'll likely just keep change the `Exec` enum into a struct (and remove the `Default` variant).

MrGunflame:
All tests except `get_strip_connection_header` should pass now. For some reason `get_strip_connection_header` gets stuck when using the proxy connection and I haven't been able to find out why that is yet.

seanmonstar:
Thanks for plugging away at this! I'll kick off another CI run.

MrGunflame:
I believe it should finally pass now.

seanmonstar:
Looks like a test or benchmark needs to be updated. I know this has grown big, if you want to just comment out the tests that are failing with a `TODO` to re-enable them, we can merge and file follow-up PRs to get the remaining bits. Whatever you'd find easier and more satisfying <3

seanmonstar:
There's some other items that are somewhat separate, but I think they will cause merge conflicts with this, so getting it merged sooner than later will probably mean less pain. For instance, I was about to start deleting `hyper::Server` and `hyper::server::accept`, but realized that'd cause conflicts with this. But up to you, there's other things to work on too.

MrGunflame:
Alright, I will just comment them out for now I guess, might be able to add them in later in another PR or work on some other changes then.

seanmonstar:
Thanks for taking this so far! I've squashed and fixed up the last couple conflicts and merged in #2929.

PR #2889, opened 2022-06-06
refactor(lib): resolve unused import

Ticsmtc:
running default build command
```rust
cargo build
```
reports
```bash
warning: unused import: `http::header::HeaderName`
 --> src/ext.rs:4:5
  |
4 | use http::header::HeaderName;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -1,6 +1,7 @@
 //! HTTP extensions.
 
 use bytes::Bytes;
+#[cfg(any(feature = "http1", feature = "ffi"))]
 use http::header::HeaderName;
 #[cfg(feature = "http1")]
 use http::header::{IntoHeaderName, ValueIter};


Ticsmtc:
```bash 
cargo build --release 
``` 
reports 
```bash
warning: variant is never constructed: `Internal`
  --> src/error.rs:78:5
   |
78 |     Internal,
   |     ^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

src/error.rs
```diff
@@ -74,7 +74,7 @@ pub(super) enum Parse {
     Header(Header),
     TooLarge,
     Status,
-    #[cfg_attr(debug_assertions, allow(unused))]
+    #[allow(unused)]
```
seanmonstar, line 77: It's always unused? Then we should remove it. But I think it *is* used at times. I'd personally prefer we see the warning when that case is no longer true, then we can just delete the variant...

Ticsmtc: Well , I may put more thought on that. Revert by now .

seanmonstar (approved):
Thanks :D

Issue #2888, opened 2022-06-06
h1_header_read_timeout_running is never set to true


silence-coding:
https://github.com/hyperium/hyper/blob/78de8914eadeab4b9a2c71a82c77b2ce33fe6c74/src/proto/h1/io.rs#L195

silence-coding:
maybe fix 

https://github.com/hyperium/hyper/blob/4545c3ef191ce9b5f5d250ee27c4c96f9b71d2c6/src/proto/h1/role.rs#L83
```
    #[cfg(all(feature = "server", feature = "runtime"))]
    if !*ctx.h1_header_read_timeout_running {
        if let Some(h1_header_read_timeout) = ctx.h1_header_read_timeout {
            let deadline = Instant::now() + h1_header_read_timeout;
            ctx.h1_header_read_timeout_running = true;
            match ctx.h1_header_read_timeout_fut {
                Some(h1_header_read_timeout_fut) => {
                    debug!("resetting h1 header read timeout timer");
                    h1_header_read_timeout_fut.as_mut().reset(deadline);
                }
                None => {
                    debug!("setting h1 header read timeout timer");
                    *ctx.h1_header_read_timeout_fut =
                        Some(Box::pin(tokio::time::sleep_until(deadline)));
                }
            }
        }
    }
```

silence-coding:
@paolobarbolini

paolobarbolini:
I don't currently have time to test it but the fix looks right, if you'd like to send the PR before I get to it

PR #2891, opened 2022-06-09
fix(http1): fix h1_header_read_timeout_running

silence-coding:
Fix: https://github.com/hyperium/hyper/issues/2888


silence-coding:
@seanmonstar

seanmonstar:
@paolobarbolini does it seem like this is all that's needed? Or are there any other places that need to set it too?

paolobarbolini:
> @paolobarbolini does it seem like this is all that's needed? Or are there any other places that need to set it too?

Yeah this is the only place.

seanmonstar (approved):

Issue #2855, opened 2022-05-20
Remove the `stream` cargo feature


seanmonstar:
The `stream` cargo feature should be removed, and any usage of `Stream`. It's not stable, and we cannot depend on it.

- This means `accept::from_stream`, (which `Accept` is already going away anyways).
- `Body::wrap_stream`
- `impl Stream for Body`

PR #2896, opened 2022-06-14
Remove stream cargo feature

oddgrd:
Removes the `stream` cargo feature and any usage of `Stream`, including `Body::wrap_stream`, `impl Stream for Body` and `accept::from_stream`.

A few tests were broken, and some examples. Quite extensive changes were needed to fix some of them, which I wasn't able to do in this PR. As discussed on Discord I left TODO comments on the remaining broken tests, tagged with the PR number (#2896).

This PR partially fixes issue #2855.

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -1,6 +1,5 @@
 #![deny(warnings)]
 
-use futures_util::TryStreamExt;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
 
@@ -16,16 +15,17 @@ async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
         // Simply echo the body back to the client.
         (&Method::POST, "/echo") => Ok(Response::new(req.into_body())),
 
+        // TODO: Fix this, broken in PR #2896
         // Convert to uppercase before sending back to client using a stream.
-        (&Method::POST, "/echo/uppercase") => {
-            let chunk_stream = req.into_body().map_ok(|chunk| {
-                chunk
-                    .iter()
-                    .map(|byte| byte.to_ascii_uppercase())
-                    .collect::<Vec<u8>>()
-            });
-            Ok(Response::new(Body::wrap_stream(chunk_stream)))
-        }
+        // (&Method::POST, "/echo/uppercase") => {
+        // let chunk_stream = req.into_body().map_ok(|chunk| {
+        //     chunk
+        //         .iter()
+        //         .map(|byte| byte.to_ascii_uppercase())
+        //         .collect::<Vec<u8>>()
+        // });
+        // Ok(Response::new(Body::wrap_stream(chunk_stream)))
+        // }
 
         // Reverse the entire body before sending back to the client.
         //
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -1,20 +1,14 @@
 use std::borrow::Cow;
-#[cfg(feature = "stream")]
-use std::error::Error as StdError;
 use std::fmt;
 
 use bytes::Bytes;
 use futures_channel::mpsc;
 use futures_channel::oneshot;
 use futures_core::Stream; // for mpsc::Receiver
-#[cfg(feature = "stream")]
-use futures_util::TryStreamExt;
 use http::HeaderMap;
 use http_body::{Body as HttpBody, SizeHint};
 
 use super::DecodedLength;
-#[cfg(feature = "stream")]
-use crate::common::sync_wrapper::SyncWrapper;
 use crate::common::Future;
 #[cfg(all(feature = "client", any(feature = "http1", feature = "http2")))]
 use crate::common::Never;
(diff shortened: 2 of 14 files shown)


examples/echo.rs
```diff
@@ -16,23 +17,20 @@ async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
...
+        // Convert to uppercase before sending back to client.
         (&Method::POST, "/echo/uppercase") => {
-            let chunk_stream = req.into_body().map_ok(|chunk| {
-                chunk
-                    .iter()
-                    .map(|byte| byte.to_ascii_uppercase())
-                    .collect::<Vec<u8>>()
-            });
-            Ok(Response::new(Body::wrap_stream(chunk_stream)))
+            let body = hyper::body::aggregate(req.into_body()).await?;
```
seanmonstar, line 22: I think we do want to try to keep the body streaming... it may be possible to use [map_data](https://github.com/hyperium/http-body/blob/aa90d1de0b6fb7695037762ffbc3e7bb873eff22/http-body-util/src/lib.rs#L28), and then make it a `BoxBody`.

(A git dependency is fine for now.)

seanmonstar: Oh, actually, if https://github.com/hyperium/http-body/pull/63 merges, it might be easy to just replace most instances of a `StreamBody`.

tests/server.rs
```diff
@@ -1841,117 +1839,117 @@ async fn h2_connect() {
...
-
-    let (socket, _) = listener.accept().await.unwrap();
-    Http::new()
-        .http2_only(true)
-        .serve_connection(socket, svc)
-        .with_upgrades()
-        .await
-        .unwrap();
-}
+// #[tokio::test]
```
seanmonstar, line 1842: I believe this is testing some important behavior for HTTP/2, and that it's using `FuturesUnordered` is fine. Could keep this test mostly as-is.

src/body/body.rs
```diff
@@ -56,12 +50,6 @@ enum Kind {
     },
     #[cfg(feature = "ffi")]
     Ffi(crate::ffi::UserBody),
-    #[cfg(feature = "stream")]
-    Wrapped(
```
seanmonstar, line 60: 🤩

tests/server.rs
```diff
@@ -2192,31 +2190,6 @@ async fn max_buf_size() {
         .expect_err("should TooLarge error");
 }
 
-#[cfg(feature = "stream")]
-#[test]
-fn streaming_body() {
```
seanmonstar, line 2197: I can't decide if this was meant to test `Body::wrap_stream` specifically, or if it was meant to test that the HTTP/1 server dispatcher properly handles a "streaming" (size-not-known) body. If we have other tests that effectively check that, then this one can go. If not, perhaps it could be changed to use `StreamBody`?

tests/server.rs
```diff
@@ -2300,43 +2273,6 @@ async fn http2_service_error_sends_reset_reason() {
     assert_eq!(h2_err.reason(), Some(h2::Reason::INADEQUATE_SECURITY));
 }
 
-#[cfg(feature = "stream")]
-#[test]
-fn http2_body_user_error_sends_reset_reason() {
```
seanmonstar, line 2305: Ah, yea this test is needed... Not because of the `wrap_stream`, but that if any body returns an error with an HTTP/2 reset code, that it results in an HTTP/2 `RST_STREAM` frame.

Could use a `StreamBody`, or could stay commented out and we file an issue to write a small `struct ImmediatelyError` type just for this test...

tests/client.rs
```diff
@@ -477,72 +477,72 @@ test! {
             headers: {
                 "Content-Length" => "5",
             },
-            body: (Body::wrap_stream(Body::from("hello"))),
+            body: (Body::from("hello")),
```
seanmonstar, line 480: This test is making sure that if the `impl Body` is stream-like, the user can specify a `content-length` header explicitly and it won't get overridden to `transfer-encoding: chunked`. So, we'll want the test. Can either comment out with a `TODO` note, or change it to use `StreamBody`.

tests/client.rs
```diff
@@ -2170,162 +2169,162 @@ mod conn {
...
-                conn.await.expect("http conn");
-            });
-
-            let req = Request::builder()
-                .uri("/a")
-                .body(Default::default())
-                .unwrap();
-            let mut res = client.send_request(req).await.expect("send_request");
-            assert_eq!(res.status(), hyper::StatusCode::OK);
-            assert!(res.body_mut().next().await.is_none());
```
seanmonstar, line 2211: Oh, these should be able to use `HttpBody::data()` instead of `StreamExt::next()`.

seanmonstar (changes requested):
Thanks for taking on this issue! (And sorry for not realizing the tests would be annoying to update.) This is pretty close, implementation wise. I left some comments inline about possibly ways to update the tests.

seanmonstar:
Alright, we merged the `StreamBody` PR :D

oddgrd:
Awesome, I'll get to work on solving the remaining issues then. :)

tests/server.rs
oddgrd: I wrote a TODO to consider fixing this, as I was unable to fix it.

tests/server.rs
oddgrd: I wrote a TODO to consider fixing this, as I was unable to fix it.

examples/echo.rs
oddgrd: I wrote a TODO to fix this, as I was unable to fix it.

tests/client.rs
oddgrd: As with the others, I was unable to fix this. I wrote a TODO to fix it.

PR #2900, opened 2022-06-21
feat: wasm32-wasi target support

rjzak:
Work-in-progress `wasm32-wasi` target support for the `server` feature.

Requires: https://github.com/tokio-rs/tokio/pull/4716

Signed-off-by: Richard Zak <richard@profian.com>


src/cfg.rs
```diff
@@ -28,6 +28,7 @@ cfg_proto! {
         ($($item:item)*) => {
             cfg_feature! {
                 #![feature = "client"]
+                #[cfg(not(target_os = "wasi"))] // seems not to work
```
rjzak, line 31: For some reason, the compiler still attempts to build the code behind the `cfg_client!` macro despite including the !Wasi config line.

seanmonstar:
Thanks for hacking away on this! I'm curious, with the split that's currently proposed for hyper 1.0, where hyper no longer directly includes the `TcpListener`/`TcpStream` stuff, but rather pushes that to `hyper-util`, I wonder if that means hyper would need to do anything special. Would that mean the special stuff would only be needed in `hyper-util` (or perhaps some `hyper-wasi` crate)?

rjzak:
I'll take a look, but my guess is that mostly Wasi will just need disabling some components.

rjzak:
@seanmonstar Code updated, would like to revisit this. Could the CI run?

rjzak:
@seanmonstar This should do it, may I have a CI re-run?

rjzak:
It seems the latest CI run skipped 7 of the tests, and the one which failed was `cargo fmt` which had the suggested change of: 
image

rjzak:
All tests pass except Miri, how shall I proceed?

npmccallum:
@seanmonstar 

1. @rjzak Has tested the `master` branch against `wasm32-wasi` and has found that it works with `--features=server,http2`. I've asked him to submit a PR that adds the changes needed to make `http1` work and to look at other features.

2. This smallish PR would enabled hyper to work on `wasm32-wasi` on the `0.14.x` branch. This in turn gives us `axum` support as well as some other notable crates. Given the important position of hyper in the Rust ecosystem, this unlocks a **lot** of value. Please consider this PR.

src/server/tcp.rs
```diff
@@ -439,6 +466,7 @@ mod tests {
         }
     }
 
+<<<<<<< HEAD
```
cschramm, line 469: The latest push contains unmerged changes.

rjzak: Thanks! So weird that Git and `cargo` didn't catch that...

PR #2912, opened 2022-07-09
feat(server): add option to disable mid-message EOF detection

### LINKED ISSUE #2911: Add server option to disable mid-message EOF detection
After reading the request headers and body, and before a response is ready, hyper will read from the socket once more to make sure it has registered interest in case the socket ever hangs up. In some cases, though, users would prefer not to pay the cost of that syscall, and simply notice when the response is written instead.

The action here is:

- Add `http1_mid_message_eof_detection(bool)`, where the default is true.
- Update [`Conn::mid_message_detect_eof`](https://github.com/hyperium/hyper/blob/3660443108e22c063cd98906edffb3cff7e9afb6/src/proto/h1/conn.rs#L413) to check if the detection option is false, and if so, return `Poll::Pending`, just as the `allow_half_close` option does.

oddgrd:
Add an option to disable mid-message EOF detection for users who want to avoid the extra syscall. I tested the change by checking that a mid-message EOF is not detected when the new option is set to false, this seemed like the easiest way. It's copy pasted from the `http1_allow_half_close` test. I am happy to test it differently if you have any suggestions.

Closes #2911

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -46,6 +46,7 @@ where
             io: Buffered::new(io),
             state: State {
                 allow_half_close: false,
+                mid_message_eof_detection: true,
                 cached_headers: None,
                 error: None,
                 keep_alive: KA::Busy,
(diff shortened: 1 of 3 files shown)


oddgrd:
Closed since the change provided insignificant performance gains in benchmarks, and thus is not worth the increased API surface.

PR #2913, opened 2022-07-11
feat(body): add `Body::wrap_body`

davidpdrsn:
This adds `Body::wrap_body` which creates a `Body` from any `impl http_body::Body`. This is required in axum for https://github.com/tokio-rs/axum/pull/1154.

Note that I made the PR against `0.14.x` because the `Body` struct is undergoing some big changes on `master`. If you want me to make this PR against `master` instead and backport this to `0.14.x` then just let me know 😊

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -13,7 +13,6 @@ use http::HeaderMap;
 use http_body::{Body as HttpBody, SizeHint};
 
 use super::DecodedLength;
-#[cfg(feature = "stream")]
 use crate::common::sync_wrapper::SyncWrapper;
 use crate::common::Future;
 #[cfg(all(feature = "client", any(feature = "http1", feature = "http2")))]
@@ -62,6 +61,7 @@ enum Kind {
             Pin<Box<dyn Stream<Item = Result<Bytes, Box<dyn StdError + Send + Sync>>> + Send>>,
         >,
     ),
+    WrappedBody(SyncWrapper<http_body::combinators::UnsyncBoxBody<Bytes, crate::Error>>),
 }
 
 struct Extra {
@@ -137,6 +137,23 @@ impl Body {
     #[inline]
     pub fn channel() -> (Sender, Body) {
         Self::new_channel(DecodedLength::CHUNKED, /*wanter =*/ false)
+    }
+
+    /// Create a `Body` by wrapping another body.
+    #[inline]
+    pub fn wrap_body<B>(body: B) -> Self
+    where
+        B: HttpBody<Data = Bytes> + Send + 'static,
+        B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    {
+        // if `B` is already a `Body` just return that and avoid additional boxing
+        match super::try_downcast(body) {
+            Ok(body) => body,
+            Err(body) => {
+                let body = body.map_err(crate::Error::new_user_body).boxed_unsync();
+                Self::new(Kind::WrappedBody(SyncWrapper::new(body)))
+            }
+        }
     }
 
     pub(crate) fn new_channel(content_length: DecodedLength, wanter: bool) -> (Sender, Body) {
@@ -329,12 +346,12 @@ impl Body {
 
             #[cfg(feature = "ffi")]
             Kind::Ffi(ref mut body) => body.poll_data(cx),
-
             #[cfg(feature = "stream")]
             Kind::Wrapped(ref mut s) => match ready!(s.get_mut().as_mut().poll_next(cx)) {
                 Some(res) => Poll::Ready(Some(res.map_err(crate::Error::new_body))),
                 None => Poll::Ready(None),
             },
+            Kind::WrappedBody(ref mut body) => Pin::new(body.get_mut()).poll_data(cx),
         }
     }
 
@@ -393,6 +410,7 @@ impl HttpBody for Body {
             },
             #[cfg(feature = "ffi")]
             Kind::Ffi(ref mut body) => body.poll_trailers(cx),
+            Kind::WrappedBody(ref mut body) => Pin::new(body.get_mut()).poll_trailers(cx),
             _ => Poll::Ready(Ok(None)),
         }
     }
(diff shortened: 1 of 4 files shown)


sfackler:
Wouldn't [`BoxBody`](https://docs.rs/http-body/latest/http_body/combinators/struct.BoxBody.html) suffice for axum?

davidpdrsn:
> Wouldn't [`BoxBody`](https://docs.rs/http-body/latest/http_body/combinators/struct.BoxBody.html) suffice for axum?

No, I wish 😞

I want to support `impl FromRequest for Request<hyper::Body>` and also fix https://github.com/tokio-rs/axum/issues/1110. That essentially requires being able to convert an `impl http_body::Body` into a `hyper::Body`.

I'll try an elaborate. This is what we have in axum today (everything is generic over the request body):

```rust
#![allow(warnings)]

use http::{Request, Response};
use http_body::combinators::UnsyncBoxBody;
use hyper::{body::Bytes, Body};
use std::{
    collections::HashMap,
    convert::Infallible,
    future::Future,
    pin::Pin,
    task::{Context, Poll},
};
use tower::{util::BoxCloneService, Layer, Service, ServiceExt};

type BoxBody = UnsyncBoxBody<Bytes, hyper::Error>;
type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
type BoxError = Box<dyn std::error::Error + Send + Sync>;

struct Route<B> {
    inner: BoxCloneService<Request<B>, Response<BoxBody>, Infallible>,
}

impl<B> Clone for Route<B> {
    fn clone(&self) -> Self {
        Self {
            inner: self.inner.clone(),
        }
    }
}

impl<B> Service<Request<B>> for Route<B> {
    type Response = Response<BoxBody>;
    type Error = Infallible;
    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        todo!()
    }

    fn call(&mut self, req: Request<B>) -> Self::Future {
        todo!()
    }
}

struct Router<B> {
    routes: HashMap<String, Route<B>>,
}

impl<B> Router<B> {
    fn new() -> Self {
        Router {
            routes: HashMap::new(),
        }
    }

    pub fn layer<L, NewReqBody, NewResBody>(self, _layer: L) -> Router<NewReqBody>
    where
        L: Layer<Route<B>>,
        L::Service:
            Service<Request<NewReqBody>, Response = Response<NewResBody>> + Clone + Send + 'static,
        <L::Service as Service<Request<NewReqBody>>>::Error: Into<Infallible> + 'static,
        <L::Service as Service<Request<NewReqBody>>>::Future: Send + 'static,
        NewResBody: http_body::Body<Data = Bytes> + Send + 'static,
        NewResBody::Error: Into<BoxError>,
    {
        todo!()
    }
}

impl<B> Service<Request<B>> for Router<B> {
    type Response = Response<BoxBody>;
    type Error = Infallible;
    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        Poll::Ready(Ok(()))
    }

    fn call(&mut self, req: Request<B>) -> Self::Future {
        todo!()
    }
}

#[test]
fn a_test() {
    use http_body::Limited;
    use tower_http::limit::RequestBodyLimitLayer;

    let router: Router<Limited<Body>> = Router::new().layer(RequestBodyLimitLayer::new(1024));
}
```

We want `Router` (and thus `Route`) to not be generic to solve https://github.com/tokio-rs/axum/issues/1110 so we need to decide on a request body in

```rust
struct Route<B> {
    inner: BoxCloneService<Request<B>, Response<BoxBody>, Infallible>,
}
```

If we pick `hyper::Body` that forces `Router::layer` to be:

```rust
pub fn layer<L, NewResBody>(self, _layer: L) -> Router
where
    L: Layer<Route>,
    // this says that the service the layer produces must accept `Request<Body>`
    L::Service:
        Service<Request<Body>, Response = Response<NewResBody>> + Clone + Send + 'static,
    <L::Service as Service<Request<Body>>>::Error: Into<Infallible> + 'static,
    <L::Service as Service<Request<Body>>>::Future: Send + 'static,
    NewResBody: http_body::Body<Data = Bytes> + Send + 'static,
    NewResBody::Error: Into<BoxError>,
{
    ...
}
```

and we are forced to implement `Service` as

```rust
impl Service<Request<Body>> for Route { ... }

impl Service<Request<Body>> for Router { ... }
```

However those two (`fn layer` and the `Service` impls) don't work with each other. If we do

```rust
let router = Router::new().layer(RequestBodyLimitLayer::new(1024));
```

Then `RequestBodyLimit` will require the inner service (a `Route`) to accept [`Request<Limited<Body>>`](https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimit.html#impl-Service%3CRequest%3CReqBody%3E%3E) but `Route` only accepts `Request<Body>`. So it doesn't work.

If we try to make the `Service` impl more general we also run into issues:

```rust
impl<B> Service<Request<B>> for Route {
    type Response = Response<BoxBody>;
    type Error = Infallible;
    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        Poll::Ready(Ok(()))
    }

    fn call(&mut self, req: Request<B>) -> Self::Future {
        // but this doesn't work either because `Route` contains a
        // `BoxCloneService<Request<Body>, Response<BoxBody>, Infallible>`
        // which we must call with a `Request<Body>` but we have no way to convert
        // `B` into a `Body`
        todo!()
    }
}
```

So that is why `Body::wrap_body` is needed because it allows me to convert the generic `B` in these `Service` impls into `Body`.

You're right that I could "just" use `BoxBody` for axum and then all this goes away but then I cannot `impl FromRequest for Request<Body>`. Again because that requires converting a `BoxBody` into a `Body`. I'm worried its gonna be a source of confusion if this doesn't work.

davidpdrsn:
axum is moving to its own body type (https://github.com/tokio-rs/axum/pull/1584) so we don't actually need this anymore. Things have also changed for hyper 1.0 so I'll close this for now.

Issue #2914, opened 2022-07-11
Backport fixes to use of `mem::uninitialized` to older versions?


5225225:
Background: In https://github.com/rust-lang/rust/issues/66151 we're adding more strict checks in order to detect invalid use of `mem::uninitialized`/`mem::zeroed`, and older versions of hyper run into this check, so it would be good to have semver compatible versions that people can `cargo update -p hyper` to in order to get their code working again.

0.11 is the oldest version that anyone ran into the issue on, and out of 136 regressions, there were only 4. So just publishing a fix to 0.12 (25 regressions) and 0.13 (52 regressions) would fix most of the issues. The rest of the regressions are on 0.14, but that's got a fixed version published on it, so there's nothing more to do there. (Granted, these numbers might be somewhat off because it's an old crater run, but from checking a few of them at random, there's not many that published an update since).

I'm up for doing the work needed here, I assume I'd just make a pull request to the 0.13.x / 0.12.x branch? It would lead to a perf regression to do the simplest thing of just using mem::zeroed / httparse::EMPTY_HEADER, but that seems acceptable since the crate would still at least work and not panic. Also, 0.12.x might need a `bytes` bump to not run into use of uninit bytes from there, which might be a MSRV bump, but again, I feel that's better than it not working.

PR #2915, opened 2022-07-12
Bump itoa version, don't use mem::uninit (0.13 backport)

5225225:
Marking this as a draft for now while I test some crates, patching their `hyper` version to be this one then seeing if a `cargo update` fixes them with strict init checks. The code is done, though.

See: #2914

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ http-body = "0.3.1"
 httpdate = "0.3"
 httparse = "1.0"
 h2 = "0.2.2"
-itoa = "0.4.1"
+itoa = "1"
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
 pin-project = "1.0"
 tower-service = "0.3"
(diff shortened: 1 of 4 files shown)


5225225:
Hm.... I needed to bump pnet for the tests to pass with strict checks (otherwise you run into uninit memory in there), but that means the MSRV of 1.39 fails tests. And looks like a lot of dependencies depend on the 2021 edition in semver compatible versions with the ones we have in cargo.toml

5225225:
Also, i'm testing according to the checks that are enabled with `-Zstrict-init-checks`, which means "no uninit integers and no zeroed types that can't be zeroed". This is basically the checks that miri does, but is strictly speaking, stricter than what's been ruled to be UB yet (see https://github.com/rust-lang/rust/pull/98919 ).

I figure it's best to fix all the panics that rust can legally insert *now* rather than needing to do another backport if/when we get the chance to make `let x: [u8; 100] = mem::uninit()` panic.

RalfJung:
> This is basically the checks that miri does, but is strictly speaking, stricter than what's been ruled to be UB yet 

Well, the [reference says it is UB](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).

5225225:
Closing this as not really needed anymore, 0.13 is not maintained and there's no risk of anything actually going wrong even in far future LLVM because of the 0x01 filling.

Issue #2885, opened 2022-06-03
Add header line obs-folding option to C API


seanmonstar:
Add a connection option to allow HTTP/1 obsolete line folding of headers in the C API. **This option already exists in hyper, the task is simply to expose it on the C API side.** This will help with curl tests 1274 and 1940.

seanmonstar:
This was done in #2918.

PR #2918, opened 2022-07-16
feat(ffi): add http1_allow_multiline_headers

deantvv:
Close issue #2885

### DIFF
--- a/capi/include/hyper.h
+++ b/capi/include/hyper.h
@@ -403,6 +403,16 @@ enum hyper_code hyper_clientconn_options_headers_raw(struct hyper_clientconn_opt
                                                      int enabled);
 
 /*
+ Set whether HTTP/1 connections will accept obsolete line folding for header values.
+ Newline codepoints (\r and \n) will be transformed to spaces when parsing.
+
+ Pass `0` to disable, `1` to enable.
+
+ */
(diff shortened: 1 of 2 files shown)


seanmonstar (changes requested):
Thanks for the PR! <3

Looks like the function names are out of sync. The `hyper_clientconn_options_http1_allow_multiline_headers` name seems preferable to me, since it is an HTTP/1 specific option. I think just running the `gen_header.sh` script should fix it up for you.

deantvv:
> Thanks for the PR! <3
> 
> Looks like the function names are out of sync. The `hyper_clientconn_options_http1_allow_multiline_headers` name seems preferable to me, since it is an HTTP/1 specific option. I think just running the `gen_header.sh` script should fix it up for you.

Fixed in new commit.

seanmonstar (approved):
Thanks :D

seanmonstar:
Would you want to add the usage in curl? If not, I can a little later, but just checking first.

seanmonstar:
I tried to add it yesterday, and noticed that curl expects the format of multiline headers to be a little different than hyper treats them, so it'll be more work than expected. I'll need to file an issue on the curl repo.

Issue #2853, opened 2022-05-20
Create hyper::Service trait


seanmonstar:
Make [a simplified `Service` trait](https://github.com/hyperium/hyper/blob/master/docs/ROADMAP.md#service) that is only the `call` method. Update the server `Connection` to use it instead of tower-service.

tomkarw:
Hey guys, decided to give this a go next.

Some questions:
- should `Service` be exported from root level, i.e `hyper::Service` or `hyper::service::Service`
- it extended the scope a bit and tried to use the new `Service` trait wherever possible. I had to stop short of removing `tracing-service` dependency completely, as I couldn't refactor `oneshot` without it (it actually uses `poll_ready` heavily)

If you don't like the extended scope, we can merge any subset of commits. I tried to separate it into increments of changes.

tomkarw:
Hey @seanmonstar, I see you being active in other issues that I commented.
Did you miss this one, or do you need bit more time?

seanmonstar:
Sorry, it was still in my inbox waiting for my review, been trying to triage between my deadlines XD

> should Service be exported from root level, i.e hyper::Service or hyper::service::Service

Unknown. Perhaps not root level to start. I dunno... I mean, the reason to have it inside hyper directly is part of the `server` API, I don't think the client stuff would use it. But probably find to stay in a separate module.

> I had to stop short of removing `tracing-service` dependency

I didn't know we even had this dependency? While I do think the community at-large should use `tower::Service`, I just don't think we can justify depending on it publicly in hyper (it's complex and not stable). So, inside hyper, it's gots to go.

tomkarw:
Extending the scope – music to my ears.

I'll get back when `tracing-service` is gone.

tomkarw:
I think I got it. Had to work around the `oneshot::Oneshot` a bit, but it compiles :smile:.  
Not sure if we're not loosing something vital by removing `poll_ready` in context of backpressure.

One test might fail when you trigger the pipeline, `dispatch_impl::drop_response_future_closes_in_progress_connection`. I'll need help finding the root cause, no idea how it is related to the changes.

PR #2920, opened 2022-07-23
feat: create Service trait

tomkarw:
Closes https://github.com/hyperium/hyper/issues/2853.


src/client/client.rs
```diff
@@ -521,7 +522,7 @@ where
     }
 }
 
-impl<C, B> tower_service::Service<Request<B>> for Client<C, B>
+impl<C, B> Service<Request<B>> for Client<C, B>
```
seanmonstar, line 525: Ah, yea, we probably don't need the `Service` impls in the client. Most of these parts are being moved to `hyper-util`, which _can_ integrate with `tower::Service`, since clients are a place I think backpressure makes more sense.

seanmonstar:
Ok, I see. You could leave the `tower-service` dependency in for now, since the work to transfer this client stuff is in a separate PR, and then _this_ PR is just replacing the server side. Then when both PRs are done, we can chop out of the `Cargo.toml`. I think that'll be the least complicated as all this movement happens.

tomkarw:
I might be doing something wrong, but it's impossible to just replace the `server` part. It would require maintaining duplicates of most parts of `src/service` and some parts `src/common`, one with stuff `tower_service::Service` and one with `hyper::service::Service`. The clean sweep throughout the repo was the closest I got to compiling, working code. Trying to do anything local to `src/server` just keeps leaking to other parts of the repo.

@seanmonstar could use some guidance here. I saw you did some work on removing `tcp` feature and parts of `src/client`. Should I maybe wait till `client` part is completely gone? I think clean sweep remove of `tower_service` dependency is the best way to move this MR forward.

tomkarw:
Alright, with the help of cross-cutting removals of code Sean have done to the `client` module, this PR was way easier to do.

Looks good to me now, @seanmonstar I'd ask for CI run and if it's all green (except for `miri`) a review.

tomkarw:
Dumb mistake, I fought the compiler for so long I forgot to run the tests once the project finally compiled.

Locally, all the tests (including examples) pass now.

There is only one reference to `tower` left in the whole project, and it's in doc-string in `client/conn/mod.rs` [here](https://github.com/tomkarw/hyper/blob/feat-create-service-trait/src/client/conn/mod.rs#L16). If we could remove it, it would be a clean sweep of `tower` from the project.

tomkarw:
I can do `fmt + check` or `test`, but I seem to uncapable of remembering to run all 3 of them before pushing :smile:

tomkarw:
Alright, couldn't stop myself from removing the last `tower` reference from `client::conn::SendRequest`. The repo is now all free of the dependency. If you think that's wrong, I can revert the last commit.

tomkarw:
Juuust one more CI trigger.

src/service/service.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// implementations should take care to not call `poll_ready`.
+    ///
+    /// Before dispatching a request, `poll_ready` must be called and return
+    /// `Poll::Ready(Ok(()))`.
+    ///
+    /// # Panics
+    ///
+    /// Implementations are permitted to panic if `call` is invoked without
+    /// obtaining `Poll::Ready(Ok(()))` from `poll_ready`.
+    fn call(&mut self, req: Request) -> Self::Future;
```
sfackler, line 42: Do we want to preserve the `&mut self` here? I know that's how tower's worked, but I think within hyper at least pretty much everything internally would work better with a shared reference so it can be used from multiple tasks without cloning.

seanmonstar: I think it's a valid question, I'd be happy to read more if you're interested in writing up a proposal to change it. I'm tempted to suggest we have the conversation in a separate issue, to not bog down the the work from Tomek.

src/service/service.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// This function is expected to be callable off task. As such,
+    /// implementations should take care to not call `poll_ready`.
+    ///
+    /// Before dispatching a request, `poll_ready` must be called and return
+    /// `Poll::Ready(Ok(()))`.
+    ///
+    /// # Panics
+    ///
+    /// Implementations are permitted to panic if `call` is invoked without
+    /// obtaining `Poll::Ready(Ok(()))` from `poll_ready`.
```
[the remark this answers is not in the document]
tomkarw: Good catch! Simplified the whole doc-string to:
```
/// Process the request and return the response asynchronously.
fn call(&mut self, req: Request) -> Self::Future;
```

src/service/service.rs
tomkarw: It makes sense for immutability to be the default, and if mutable state is required, one needs to utilize an interior mutability pattern.

I'd be happy to implement it and even write the proposal if @sfackler doesn't feel like it, but we should definitely leave it out of scope of this MR.

src/proto/h1/dispatch.rs
```diff
@@ -31,7 +31,6 @@ pub(crate) trait Dispatch {
         cx: &mut task::Context<'_>,
     ) -> Poll<Option<Result<(Self::PollItem, Self::PollBody), Self::PollError>>>;
     fn recv_msg(&mut self, msg: crate::Result<(Self::RecvItem, Body)>) -> crate::Result<()>;
-    fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<Result<(), ()>>;
```
seanmonstar, line 34: We should leave this in, it still have use in detecting that on the client side, the user no longer cares. The server should still say it's not ready if there is already a request `in_flight`.

src/proto/h2/server.rs
```diff
@@ -244,38 +244,6 @@ where
...
-                    }
-                    Poll::Ready(Err(err)) => {
-                        let err = crate::Error::new_user_service(err);
-                        debug!("service closed: {}", err);
-
-                        let reason = err.h2_reason();
-                        if reason == Reason::NO_ERROR {
-                            // NO_ERROR is only used for graceful shutdowns...
-                            trace!("interpreting NO_ERROR user error as graceful_shutdown");
-                            self.conn.graceful_shutdown();
```
seanmonstar, line 268: I had forgotten about this "feature"... Part of me wonders that perhaps it's worth keeping, and we just make `Service::poll_ready` a default method. Then, this kind of stuff still works, if an advanced user wants to, but most people don't have to care about the boilerplate method.

Cargo.toml
```diff
@@ -182,8 +180,8 @@ path = "examples/state.rs"
 required-features = ["full"]
 
 [[example]]
-name = "tower_server"
-path = "examples/tower_server.rs"
+name = "server"
+path = "examples/server.rs"
```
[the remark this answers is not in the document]
tomkarw: There is `example/client.rs`, so I thought a corresponding `server.rs` would still make sense.

src/proto/h2/server.rs
tomkarw: I thought we want to remove `Service::poll_ready` for now, and eventually replace it with a better API for backpressure. Leaving it in the trait for `1.0.0` release would make us stuck with it for a while, even when the replacement API emerges.

src/proto/h1/dispatch.rs
tomkarw: Same as below, this would also mean we have to leave `Service::poll_ready` for `Dispatch` implementation on `Server`.

src/proto/h1/dispatch.rs
```diff
@@ -519,17 +509,6 @@ cfg_server! {
             Ok(())
         }
 
-        fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<Result<(), ()>> {
-            if self.in_flight.is_some() {
-                Poll::Pending
-            } else {
-                self.service.poll_ready(cx).map_err(|_e| {
```
tomkarw, line 526: This would require `poll_ready` on `Server` and thus `Service`.

src/client/conn/mod.rs
```diff
@@ -292,10 +288,6 @@ where
     type Error = crate::Error;
     type Future = ResponseFuture;
 
-    fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<Result<(), Self::Error>> {
```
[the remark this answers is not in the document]
tomkarw: done

src/proto/h2/server.rs
tomkarw: Maybe we could leave it for now, but wait for a replacement before `1.0.0` milestone?

Let me know what you think, I can bring that code back no problem.

Cargo.toml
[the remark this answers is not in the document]
seanmonstar: I've historically suggested the `hello.rs` and `echo.rs` as simple and more complex server examples.

src/proto/h1/dispatch.rs
seanmonstar: Yes, I believe the simplified version is what should be kept.

Cargo.toml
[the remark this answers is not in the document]
tomkarw: Ok, I'll remove it then.

src/proto/h1/dispatch.rs
tomkarw: I'll do that.

tomkarw:
@seanmonstar Could you address https://github.com/hyperium/hyper/pull/2920#discussion_r950670503 and https://github.com/hyperium/hyper/pull/2920#discussion_r950680044, and I'll move this PR forward?

src/proto/h2/server.rs
seanmonstar: While it was "neat", we can probably keep this out for now.

src/proto/h1/dispatch.rs
seanmonstar: No, this could just use the "simple" version for a server, as described in the other comment.

Cargo.toml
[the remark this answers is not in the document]
tomkarw: I see it got removed as part of a different PR.

tomkarw:
@seanmonstar I introduced changes from the discussion threads. 

LGTM, let's wrap and merge this as soon as possible, as the merge conflicts on this branch are killing me ;)

src/proto/h1/dispatch.rs
```diff
@@ -623,17 +611,8 @@ cfg_client! {
...
-                    Poll::Ready(()) => {
-                        trace!("callback receiver has dropped");
-                        Poll::Ready(Err(()))
-                    }
-                    Poll::Pending => Poll::Ready(Ok(())),
-                },
-                None => Poll::Ready(Err(())),
-            }
+        fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> Poll<Result<(), ()>> {
+            Poll::Ready(Ok(()))
```
seanmonstar, line 615: I believe for the client the original stuff should stay. It's how we know that the user gave up on a `client.request(req).await` future, and so _this_ task should shutdown (if the callback receiver is dropped).

tomkarw: Good point, I reverted the code “from memory” and forgot that `Client` also had some additional logic here.

Done.

tomkarw:
CI failed on a doc-test, this should be fixed now.

src/proto/h1/dispatch.rs
```diff
@@ -233,15 +233,6 @@ where
     }
 
     fn poll_read_head(&mut self, cx: &mut task::Context<'_>) -> Poll<crate::Result<()>> {
-        // can dispatch receive, or does it still care about, an incoming message?
-        match ready!(self.dispatch.poll_ready(cx)) {
```
seanmonstar, line 237: Sorry, I didn't notice this needs to be reverted too, otherwise we never actually check the client's `poll_ready` 🙈 .

Everything else looks perfect (I was just about to merge, and notice this.)

tomkarw: Np, good/important catch. TBH, I've been relying on you for soundness ever since CR process started.

Reverted, I also changed the wording of the comment a bit, it didn't sound right to me. It's now:
```
// can dispatch receive, or does it still care about other incoming message?
```

seanmonstar (approved):

seanmonstar:
Thanks so so so much! I know this dragged on a while, but it turns out there were tricky things for us to figure out (and conflicts as all sorts of other things got yanked).

Issue #2922, opened 2022-07-26
Remove Body's `Once` variant


seanmonstar:

seanmonstar:
It was noted that the FFI body type makes use of `Body::empty()`. I think there's 2 ways we can fix that:

1. Add a `pub(crate) fn ffi() -> Body` that just starts the type like this: https://github.com/hyperium/hyper/blob/509672aada0af68a91d963e69828c6e31c44cb7b/src/body/body.rs#L240
2. Or, keep an `Empty` variant. I don't think I like this option as much, but welcome feedback.

seanmonstar:
cc @Xuanwo

Xuanwo:
> 1. Add a `pub(crate) fn ffi() -> Body` that just starts the type like this

Let me give this a try!

PR #2923, opened 2022-07-26
refactor: Remove Body's Once variant

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

Fix https://github.com/hyperium/hyper/issues/2922

I'm not sure whether I correctly did this, PTAL.


src/body/body.rs
```diff
@@ -289,19 +279,16 @@ impl Body {
 
     #[cfg(feature = "http1")]
     pub(super) fn take_full_data(&mut self) -> Option<Bytes> {
-        if let Kind::Once(ref mut chunk) = self.kind {
-            chunk.take()
-        } else {
-            None
-        }
+        None
```
Xuanwo, line 282: Maybe we can remove `take_full_data` entirely?

src/body/body.rs
```diff
@@ -289,19 +279,16 @@ impl Body {
...
 }
 
+#[cfg(feature = "ffi")]
 impl Default for Body {
-    /// Returns `Body::empty()`.
+    /// Returns `Body::ffi()`.
     #[inline]
     fn default() -> Body {
-        Body::empty()
+        Body::ffi()
```
Xuanwo, line 291: Does it make sense to use `ffi()` as default?

src/proto/h1/dispatch.rs
```diff
@@ -248,7 +246,7 @@ where
         match ready!(self.conn.poll_read_head(cx)) {
             Some(Ok((mut head, body_len, wants))) => {
                 let body = match body_len {
-                    DecodedLength::ZERO => Body::empty(),
+                    DecodedLength::ZERO => Body::default(),
```
Xuanwo, line 249: `default` here is `Body::ffi`, LGTY?

src/body/body.rs
seanmonstar: No, I think we'll need to remove the Default impl as well.

Xuanwo: Ok, if so, we need to address the usage like the following:

```rust
DecodedLength::ZERO => Body::empty(),
```

tomkarw: I think it should use this: https://github.com/hyperium/http-body/blob/master/http-body-util/src/empty.rs

seanmonstar: An inline variant is fine, we can't publicly expose those types since they're not stable.

Xuanwo: > I think it should use this: https://github.com/hyperium/http-body/blob/master/http-body-util/src/empty.rs

a trait like `Dispatch` requires using the `Body` struct, so I think returning `http_body_util::Empty` here doesn't work:

```rust
pub(crate) trait Dispatch {
    fn recv_msg(&mut self, msg: crate::Result<(Self::RecvItem, Body)>) -> crate::Result<()>;
}
```

> An inline variant is fine, we can't publicly expose those types since they're not stable.

Can you explain a bit? I don't know how to return an empty body here.

seanmonstar:
Looks like a couple tests that were using the `Default` impl will need to be changed to construct a `Request` manually.

Xuanwo:
```rust
    pub fn get(&self, uri: Uri) -> ResponseFuture
    where
        B: Default,
```

`Client::get` needs `B: Default`, do we need to remove this API? (This will break many code)

seanmonstar:
No, that method doesn't need to be removed (yet). Just change the tests to use `client.request(req)`.

Xuanwo:
I'm feeling challenged to migrate tests like `http2_detect_conn_eof`. Can you give me some help?

In those tests, both request and response depend on `hpyer::Body` (so does `Body::empty()`), I can't replace them with `http_body_util::Empty<bytes::Bytes>` simplely.

Any ideas?

seanmonstar:
> migrate tests like `http2_detect_conn_eof`.

I just took a look, and this should be able to fix it:

```rust
type Empty = http_body_util::Empty<Bytes>;

let service = service_fn(|_:Request<Body>| future::ok::<_, hyper::Error>(Response::new(Empty::new())));

// ...

let req = Request::builder()
            .uri(format!("http://{}/", addr))
            .body(Empty::new())
            .expect("request builder");
```

You don't need to change the `Request<Body>` part of the `service_fn`, since that's the request that hyper is generating.

seanmonstar:
Thanks so much for taking it so far! I took care of merge conflicts and finishing out the examples, and completed it in #2922.

PR #2937, opened 2022-08-09
(feat) reenable benches/end_to_end.rs

Robertoskr:
This benches were disabled becuase of the deprecation of Client and Server, 
this pr is to reenable them with the new conn aproach


benches/end_to_end.rs
```diff
@@ -1,385 +1,400 @@
...
+                let req = make_request();
+                rt.block_on(send_request(req));
+            });
+        } else {
+            for _ in 0..self.parallel_cnt {
+                b.iter(|| {
+                    let req = make_request();
+                    rt.block_on(send_request(req));
+                });
+            }
```
Robertoskr, line 340: as you can see @seanmonstar we do this in sync way... If i try to run this async way, i will have an error because the connection is not ready...
What is the new approach for handling parallel request with the new "conn" things? I tried with poll_ready, to wait until the connection is ready before making the requests... 
What do you think is the best here ? Opening n connections to the server or using an approach like using the poll ready function ?

Robertoskr: also i see on src/client/conn/mod.rs in the example this line: 
`request_sender.ready().await?; `
wich i think is wrong because i can't find that method on the code 💢  (maybe im wrong)

seanmonstar: Yea, we don't want to use `block_on` there. Especially since the parallel option is meant to send them all at the same time.

- For HTTP/1, you'd need to make `parallel_cnt` amount of connections, and then spawn tasks looping on `ready()` and then `send_request()`.
- For HTTP/2, you can make just one connection, and send `parallel_cnt` requests at a time.

Robertoskr: Hey, already did the changes, as you can see in the pr, but when i run the benchs, have the following error, do you know how to solve it ? How is the best way to await the request sender until is ready to send a request ? 
Screenshot 2022-08-13 at 11 29 36

oddgrd: Hey Robertoskr. When I was working on this I called the `ready` method from `tower::ServiceExt` on the request sender: `request_sender.ready().await`. From the comments above it seems like you tried this, but maybe you forgot to import the trait?

Robertoskr: Thanks! Just tried, that and works, but also tried with a tokio::sync::Lock and performs better, don't know if its safe thought, but the code works well.

Robertoskr:
@seanmonstar @oddgrd  Ready for review !

benches/end_to_end.rs
Robertoskr: and for the trait: Yes, did't know that trait existed 😹

oddgrd:
Hey, that's cool that you figured it out! I can have a look at this in the coming week, but I'm a new contributor myself so you'll have to wait for Sean to sign off on this whenever he has time. :smile:

benches/end_to_end.rs
```diff
@@ -1,385 +1,472 @@
...
+                    .http2_initial_connection_window_size(http2_conn_window)
+                    .http2_adaptive_window(http2_adaptive_window)
+                    .serve_connection(tcp_stream, service_fn(handle_request)),
+            );
+        }
+    });
+    local_addr
+}
+
+async fn handle_request(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
```
oddgrd, line 468: In this function you return the request body, but you should be returning the `response_body` from the `Opts` structs.

benches/end_to_end.rs
```diff
@@ -1,385 +1,472 @@
...
+
+        match self.http2 {
+            true => self.bench_http2(b, &rt, &addr),
+            false => self.bench_http1(b, &rt, &addr),
+        };
+    }
+
+    //
+    // Benches http/1 requests
+    //
```
oddgrd, line 303: In Rust we can write [doc comments](https://doc.rust-lang.org/rust-by-example/meta/doc.html) with `///` :slightly_smiling_face:

benches/end_to_end.rs
```diff
@@ -1,385 +1,472 @@
...
+fn spawn_server(rt: &tokio::runtime::Runtime, opts: &Opts) -> SocketAddr {
+    let addr: SocketAddr = "127.0.0.1:0".parse().unwrap();
+    let tcp_listener = rt.block_on(async move { TcpListener::bind(addr).await.unwrap() });
+    let local_addr = tcp_listener.local_addr().unwrap();
+    let mut http = Http::new();
+    //configure http based on opts
+    http.http2_only(opts.http2)
+        .http2_initial_stream_window_size(opts.http2_stream_window)
+        .http2_initial_connection_window_size(opts.http2_conn_window)
+        .http2_adaptive_window(opts.http2_adaptive_window);
```
oddgrd, line 443: You may have forgotten to remove this code, as I don't see this variable being used anywhere. :slightly_smiling_face:

benches/end_to_end.rs
```diff
@@ -1,385 +1,472 @@
...
+
+    let (http2, http2_stream_window, http2_conn_window, http2_adaptive_window) = (
+        opts.http2,
+        opts.http2_stream_window,
+        opts.http2_conn_window,
+        opts.http2_adaptive_window,
+    );
```
oddgrd, line 450: If you `#[derive(Clone)]` for the `Opts` struct you can clone opts here and then use it directly without destructuring. (Except for the `opts.response_body`, which has to be destructured before being passed to the `service_fn` closure.)

```rust
 let opts = opts.clone();
 let body = opts.response_body;
```

oddgrd (commented):
Hey Robertoskr. I had a look at this, and when I tried to run the benchmarks on your branch it seems some of the HTTP2 ones take quite long to finish. On my computer your branch took ~290 seconds to complete, but on the commit before these benchmarks were disabled they completed in ~80s. I left a couple of comments, but unfortunately I wasn't able to solve this myself so I can't be of much help here. 

If you'd like you can have a look at my [attempt](https://github.com/oddgrd/hyper/blob/1af8d6722bb8ba54220337b32581147c55ad9609/benches/end_to_end.rs), maybe it will help you. I was able to get it down to ~100s, but some of the HTTP2 benches are unstable. I might keep looking into this, if I figure it out I'll let you know. Good luck!

Robertoskr:
> Hey Robertoskr. I had a look at this, and when I tried to run the benchmarks on your branch it seems some of the HTTP2 ones take quite long to finish. On my computer your branch took ~290 seconds to complete, but on the commit before these benchmarks were disabled they completed in ~80s. I left a couple of comments, but unfortunately I wasn't able to solve this myself so I can't be of much help here.
> 
> If you'd like you can have a look at my [attempt](https://github.com/oddgrd/hyper/blob/1af8d6722bb8ba54220337b32581147c55ad9609/benches/end_to_end.rs), maybe it will help you. I was able to get it down to ~100s, but some of the HTTP2 benches are unstable. I might keep looking into this, if I figure it out I'll let you know. Good luck!

In my computer it runs the same as the previous version of them in 65 seconds

benches/end_to_end.rs
Robertoskr: nice

benches/end_to_end.rs
Robertoskr: Yep, thanks!

oddgrd:
That's odd, because on my computer the http2 ones take a long time (the http1 ones run great): 
image

seanmonstar:
Thank you for tackling this issue! #3109 was just merged, which re-enables most of these benchmarks, so I'll close this one.

PR #2953, opened 2022-08-19
Fix http_client example

### LINKED ISSUE #2950: http_proxy example no longer works
**Version**
git bisect says the first bad commit is 0c8ee93d7f557afc63ca2a5686d19071813ab2b7

**Platform**
64-bit (Windows)

**Description**
When running the `http_proxy` example, it does not work with powershell's `Invoke-WebRequest -Proxy` command.

Server output:
```
> cargo run --features=full --example http_proxy
Listening on http://127.0.0.1:8100
req: Request { method: CONNECT, uri: example.com:443, version: HTTP/1.1, headers: {"host": "example.com:443", "user-agent": "Mozilla/5.0 (Windows NT 10.0; Microsoft Windows 10.0.19042; en-GB) PowerShell/7.2.5"}, body: Body(Empty) }
upgrade error: upgrade expected but low level API in use
```

Client output:
```
> Invoke-WebRequest https://example.com -Proxy http://127.0.0.1:8100
Invoke-WebRequest: Unable to read data from the transport connection: An established connection was aborted by the software in your host machine..
```
Using the previous commit, I get the expected client output:
```
> iwr https://example.com -Proxy http://127.0.0.1:8100

StatusCode        : 200
StatusDescription : OK
Content           : <!doctype html> ...
```

I will test with curl when I get a chance

andrewhickman:
Fixes #2950

### DIFF
--- a/examples/http_proxy.rs
+++ b/examples/http_proxy.rs
@@ -32,6 +32,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .http1_preserve_header_case(true)
                 .http1_title_case_headers(true)
                 .serve_connection(stream, service_fn(proxy))
+                .with_upgrades()
                 .await
             {
                 println!("Failed to serve connection: {:?}", err);


seanmonstar (approved):

seanmonstar:
Thanks so much :D

Issue #2925, opened 2022-07-26
Set up CI to audit for public dependency types


seanmonstar:
We want strict control over what types from dependencies we expose, and there is a tool that can help us check that in CI: [cargo api-linter](https://github.com/awslabs/smithy-rs/tree/main/tools/api-linter).

We should start by adding a new CI job to `.github/workflows/CI.yml` to use it. Any questions about what goes in the config file can be clarified here or in a PR.

ethanburrell:
Looks like that has been refactored into [cargo-check-external-types](https://github.com/awslabs/smithy-rs/tree/main/tools/cargo-check-external-types). I'll pick up this up.

oliviacrain:
I ran this tool against the master branch (982e6a51385). Any objections to these types being exposed and added to the allowlist, @seanmonstar?

- `bytes::buf::buf_impl::Buf`
- `bytes::bytes::Bytes`
- `http::header`
- `http::header::map::HeaderMap`
- `http::method::Method`
- `http::request::Request`
- `http::response::Response`
- `http::status::StatusCode`
- `http::uri::Uri`
- `http::version::Version`
- `http_body::Body`
- `http_body::frame::Frame`
- `http_body::size_hint::SizeHint`
- `tokio::io::async_read::AsyncRead`
- `tokio::io::async_write::AsyncWrite`


<details>
<summary> Full `cargo-check-external-types` output </summary>
<br>

| Crate | Type | Used In |
| ---   | ---  | ---     |
| bytes | bytes::buf::buf_impl::Buf | src/body/mod.rs:17:16 |
| bytes | bytes::bytes::Bytes | src/body/incoming.rs:155:4 |
| bytes | bytes::bytes::Bytes | src/body/mod.rs:17:21 |
| bytes | bytes::bytes::Bytes | src/client/conn/http1.rs:46:4 |
| bytes | bytes::bytes::Bytes | src/ext/h1_reason_phrase.rs:110:0 |
| bytes | bytes::bytes::Bytes | src/ext/h1_reason_phrase.rs:57:4 |
| bytes | bytes::bytes::Bytes | src/ext/h1_reason_phrase.rs:98:0 |
| bytes | bytes::bytes::Bytes | src/server/conn/http1.rs:65:4 |
| bytes | bytes::bytes::Bytes | src/upgrade.rs:92:4 |
| http | http::header | src/lib.rs:61:22 |
| http | http::header::map::HeaderMap | src/lib.rs:64:0 |
[...]

PR #2957, opened 2022-08-23
feat(CI): add external type check

ethanburrell:
Adding audit for issue #2925.

Uses the nightly build to uses rustdoc to create a diff with the exported types.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -260,3 +260,23 @@ jobs:
         with:
           command: rustdoc
           args: --features full,ffi -- --cfg docsrs --cfg hyper_unstable_ffi -D broken-intra-doc-links
+
+  doc:
+    name: Check exposed types
+    needs: [style, test]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+
+      - name: cargo check-external-types
+        uses: actions-rs/cargo@v1
+        with:
+          command: check-external-types
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ libc = { version = "0.2", optional = true }
 socket2 = { version = "0.4", optional = true }
 
 [dev-dependencies]
+cargo-check-external-types = "0.1.2"
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
 matches = "0.1"
 num_cpus = "1.0"


Cargo.toml
```diff
@@ -45,6 +45,7 @@ libc = { version = "0.2", optional = true }
 socket2 = { version = "0.4", optional = true }
 
 [dev-dependencies]
+cargo-check-external-types = "0.1.2"
```
seanmonstar, line 48: Would it be possible to make the dependency only installed in the CI workflow? So that most local dev doesn't need to download the extra dep?

seanmonstar (commented):
Thanks for taking this on!

ethanburrell:
@seanmonstar Changed pipeline to install the dependency within the pipeline.

seanmonstar:
Looks like the new job failed, can't find:

> Unable to resolve action `actions-rs/install@v1`, unable to find version `v1`

.github/workflows/CI.yml
```diff
@@ -263,3 +263,30 @@ jobs:
...
+  check-external-types:
+    name: Check exposed types
+    needs: [style, test]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
```
oliviacrain, line 276: This action is [unmaintained](https://github.com/actions-rs/toolchain/issues/216). It seems like people are coalescing around [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) as an alternative solution.

.github/workflows/CI.yml
```diff
@@ -263,3 +263,30 @@ jobs:
...
           command: rustdoc
           args: --features full,ffi -- --cfg docsrs --cfg hyper_unstable_ffi -D broken-intra-doc-links
+
+  check-external-types:
+    name: Check exposed types
+    needs: [style, test]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
oliviacrain, line 273: We should use the latest version of this action. Older versions of this action depend on node versions that are deprecated in Github's runners.

```suggestion
        uses: actions/checkout@v3
```

.github/workflows/CI.yml
```diff
@@ -263,3 +263,30 @@ jobs:
...
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
```
oliviacrain, line 279: We should pin to a specific combination of nightly version/cargo-check-external-types that's known to work. The newest nightly version might introduce a new version of rustdoc JSON output that's incompatible with the version expected by `cargo-check-external-type`.

It'd be nicer if we could install the toolchain version listed in their [toolchain file](https://github.com/awslabs/cargo-check-external-types/blob/main/rust-toolchain.toml), but that's probably not necessary to get this PR in.

.github/workflows/CI.yml
```diff
@@ -263,3 +263,30 @@ jobs:
...
+          override: true
+
+      - name: Install cargo-check-external-types
+        uses: actions-rs/install@v1
```
oliviacrain, line 283: Same issue as `actions-rs/toolchain`- I'd consider [`dtolnay/install`](https://github.com/dtolnay/install) instead.

.github/workflows/CI.yml
```diff
@@ -263,3 +263,30 @@ jobs:
...
+        with:
+          crate: cargo-check-external-types
+          version: 0.1.0
+          use-tool-cache: true
+
+      - name: check-external-types
+        uses: actions-rs/cargo@v1
```
oliviacrain, line 290: Same issue as the other `actions-rs/*` actions- I think here we can just invoke `cargo` manually.

seanmonstar:
Thanks again for attempting this. It was merged in #3152, so closing this.

PR #2966, opened 2022-08-25
Rename body to recv temporarily

Michael-J-Ward:
Resolves https://github.com/hyperium/hyper/issues/2963

This currently branches off of https://github.com/hyperium/hyper/pull/2961

PR #2969, opened 2022-08-28
refactor(body): rename HttpBody export to Body

### LINKED ISSUE #2839: Rename the HttpBody export to Body
When the concrete `Body` struct is removed, we can export the `http_body::Body` trait as just `Body`.

RajivTS:
Now that the concrete `Body` type has been temporarily replaced with `Recv` in #2966,
we can rename and export `http_body::Body` as just `Body` instead of `HttpBody`.

Closes #2839


seanmonstar:
Thanks so much! Looks like a run of `cargo fmt` will fixup a couple things.

seanmonstar (approved):

PR #2974, opened 2022-08-31
feat(rt): add Timer trait

### LINKED ISSUE #2846: Add a Timer trait to rt module
Just like the `Executor` trait, add a `Timer` trait to `hyper::rt`.

### LINKED ISSUE #2847: Add Timer generic to Connection builders
Similar to the `E: Executor` generic, add a [`T: Timer` parameter](https://github.com/hyperium/hyper/issues/2846) to the server and client connection builders.

### LINKED ISSUE #2848: Use a generic Timer to implement the timeouts and intervals
After adding a [generic `T: Timer`](https://github.com/hyperium/hyper/issues/2847), update the uses of `tokio::time` to use the generic timer instead.

- [ ] HTTP/2 keep-alive interval
- [ ] HTTP/2 keep-alive timeout
- [ ] HTTP/1 read headers timeout

seanmonstar:
This adds a `hyper::rt::Timer` trait, and it is used in connection
builders to configure a custom timer source for timeouts.

Closes #2846 
Closes #2847 
Closes #2848 

Continues #2904

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -5,8 +5,31 @@
 //! If the `runtime` feature is disabled, the types in this module can be used
 //! to plug in other runtimes.
 
+use std::{
+    future::Future,
+    pin::Pin,
+    time::{Duration, Instant},
+};
+
 /// An executor of futures.
 pub trait Executor<Fut> {
     /// Place the future into the executor to be run.
     fn execute(&self, fut: Fut);
 }
+
+/// A timer which provides timer-like functions.
+pub trait Timer {
+    /// Return a future that resolves in `duration` time.
+    fn sleep(&self, duration: Duration) -> Box<dyn Sleep + Unpin>;
+
+    /// Return a future that resolves at `deadline`.
+    fn sleep_until(&self, deadline: Instant) -> Box<dyn Sleep + Unpin>;
+
+    /// Reset a future to resolve at `new_deadline` instead.
+    fn reset(&self, sleep: &mut Pin<Box<dyn Sleep>>, new_deadline: Instant) {
+        *sleep = crate::common::into_pin(self.sleep_until(new_deadline));
+    }
+}
+
+/// A future returned by a `Timer`.
+pub trait Sleep: Send + Sync + Unpin + Future<Output = ()> {}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,7 +96,6 @@ server = []
 runtime = [
     "tokio/net",
     "tokio/rt",
-    "tokio/time",
 ]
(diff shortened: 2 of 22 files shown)


src/rt.rs
```diff
@@ -5,8 +5,43 @@
...
+    fn sleep_until(&self, deadline: Instant) -> Box<dyn Sleep + Unpin>;
+}
+
+/// The generic version of tokio::time::Sleep, which itself is the output of tokio::time::sleep
+pub trait Sleep: Send + Sync + Unpin + Future<Output = ()> {
+    /// An analogue of tokio::time::Sleep::deadline.
+    fn deadline(&self) -> Instant;
+
+    /// An analogue of tokio::time::Sleep::reset.
+    fn reset(self: Pin<&mut Self>, deadline: Instant);
```
seanmonstar, line 37: [Continuing the conversation](https://github.com/hyperium/hyper/pull/2904#discussion_r957859176), where @dhobsd said:

> Thanks for remembering us and reaching out! (Sorry for delay!)
> 
> Fuchsia's async executor uses deadlines for its time, so that's at least good. Two points to bring up:
> 
> - reset in tokio says "This function can be called both before and after the future has completed." In [fuchsia_async's Timer](https://fuchsia-docs.firebaseapp.com/rust/fuchsia_async/struct.Timer.html), this isn't true -- you can only reset a timer that's already fired. I'm also not sure what it means to reset a timer that hasn't fired because it's hard to guarantee you won't observe the signal. Does hyper reset timers that haven't fired?
> - is_elapsed says that it returns true when the time has elapsed, but is this the same as saying the timer has already fired?
> 
> I guess a third question: will hyper rely on these divergent behaviors? If not, perhaps we can make the documentation stricter in hyper's trait?

seanmonstar: So I looked around how we use these methods, and here's what I've found:

- `reset` is used for the HTTP/1 header read timeout, so that successive reads simply move the timer further out, instead of allocating a new one.
- Doesn't look like we use the other methods, so I've been happy to toss them.

The `reset` method seems to just be to allow for optimizations, but we _could_ simply provide a default implementation of the method somehow... Basically, if it didn't exist, we'd just call `timer.sleep(next_dur)` and replace the old one. Hm, that might mean the method needs to be on `Timer` instead, since we can't make a default implementation that creates a new sleep without the `Timer`. Maybe then a `timer.reset(sleep, next_dur)`.

We even could just leave it off entirely to start with, since adding a method with a default implementation is fine. It would just mean potentially unnecessary allocations when using Tokio.

PR #2976, opened 2022-09-05
Add example for to_bytes and trailers

tiziano88:
It is not obvious how to deal with trailers, especially since in most
cases the body gets consumed by a call to `to_bytes`, even though it is
perfectly fine to call that on a mutable reference to the body, and then
extract the trailers aftewards.


tiziano88:
@mariaschett could you take a look?

mariaschett:
Yes, I agree, we have needed trailers to proxy a gRPC request [here](https://github.com/project-oak/oak/blob/ab93a292f227c7f74a68fdab568f15c3bfb3010d/experimental/trusted_shuffler/server/src/http.rs#L175-L186).

PR #2978, opened 2022-09-11
examples: add an MPTCP using Socket2

Martichou:
Show how one can easily create an MPTCP capable webserver using Socket2.

---

MPTCP is a new protocol added in the Linux Kernel since 5.6. The goal of MPTCP is to support multiple path (cf: ADSL & LTE).
This allow for aggregation of links's bandwidth and for resiliency if any of the link fails.

Even if MPTCP is added in the Linux Kernel since 5.6, you may still need to enable it manually using:
```
sysctl net.mptcp.enabled=1
```

Signed-off-by: Martin Andre <martin.andre@tessares.net>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,6 +103,8 @@ ffi = ["libc"]
 nightly = []
 __internal_happy_eyeballs_tests = []
 
+mptcp = ["socket2"]
+
 [package.metadata.docs.rs]
 features = ["ffi", "full"]
 rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi"]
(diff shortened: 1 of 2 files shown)


Martichou:
_note: this PR will fail until socket2 make a new release with the new Protocol::MPTCP, but the content can still be reviewed_

(see https://github.com/rust-lang/socket2/commit/4c436818cddec880c26c4e79aca7106ba88827ff)

seanmonstar:
Thanks for the PR!

I'm a little hesitant to add a server example using a slightly different IO type. Especially needing to have another dev dependency and check it in CI. If it's already clear you can pass any IO type, is there a benefit to having this example on its own?

Do we just need to make the docs clearer you can pass any async IO?

Martichou:
I understand your totally valid point :)

I just think it would be great to have an explicit example of how to create a MPTCP server using Hyper as it may not be clear to everyone how to do so. I'm not sure about the dependencies, Socket2 was already present and this is why I thaught it wouldn't be too much of an issue (but Socket2 isn't used anywhere so maybe it's just a leftover that need to be cleaned).

It's totally up to you to accept this PR or not.

PR #2979, opened 2022-09-13
cargo clippy

mastrzyz:


src/lib.rs
```diff
@@ -1,11 +1,22 @@
...
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+#![allow(
+    clippy::needless_borrow,
+    clippy::module_inception,
+    clippy::useless_format,
+    clippy::redundant_clone,
+    clippy::manual_non_exhaustive,
+    dead_code,
+    unused_imports
```
mastrzyz, line 17:
```
warning: unused import: `self::body::Sender`
  --> src/body/mod.rs:24:16
   |
24 | pub(crate) use self::body::Sender;
   |                ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

```

.github/workflows/CI.yml
```diff
@@ -45,6 +46,26 @@ jobs:
...
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: clippy
+
+      - name: Clippy Check
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+
```
seanmonstar, line 68: I was thinking we could add an `args: -- -Aclippy::all -Wclippy::specific_lint_name`, and thus not need to change the actual source code (like the fmt job above).

src/lib.rs
dswij: Note that `unused_imports` lint is from `rustc` itself, and not clippy. Also, I think it's best to address both `unused_imports` and `dead_code` lint.

src/lib.rs
```diff
@@ -1,10 +1,18 @@
...
 #![cfg_attr(all(test, feature = "full"), deny(warnings))]
 #![cfg_attr(all(test, feature = "nightly"), feature(test))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![allow(
+    clippy::needless_borrow,
+    clippy::module_inception,
+    clippy::useless_format,
+    clippy::redundant_clone,
+    clippy::manual_non_exhaustive
+)]
```
dswij, line 15: We can avoid adding these lines directly into `src/lib.rs` with an argument when we call clippy in CI, as in [seanmonstar's comment](https://github.com/hyperium/hyper/pull/2979/files#r969625387)

.github/workflows/CI.yml
mastrzyz: I have reverted the change to the main lib file , verified the build is green ! -> https://github.com/mastrzyz/hyperclip/actions/runs/3110424028/jobs/5041594746

mastrzyz (commented):
Done!

.github/workflows/CI.yml
dswij: > My concern is something will fail in CI and not locally .

I agree that changing the top level module would prevent this, but since we're allowing everything except `self_named_module_files`, this should be fine.

Although now that I think about it, changing the top level module might not be a bad idea at all 😄

seanmonstar: Shouldn't we make it `-D`? We want the CI job to fail, not just yell and then get forgotten.

(I would expect this to trigger a complaint about the `ext.rs` module.)

mastrzyz: Sorry I misread -> Updated.


I verified that -D does in fact "Disallow"

image


But for the clippy rule we decided on enabling, there isn't any violation reported in ext.rs :/

PR #2989, opened 2022-09-21
Supports configuration of all TCP Keepalive parameters per  https://en.wikipedia.org/wiki/Keepalive

hansonchar:
*  `Keepalive time` is the duration between two keepalive transmissions in idle condition. TCP keepalive period is required to be configurable and by default is set to no less than 2 hours.
* `Keepalive retry` is the number of retransmissions to be carried out before declaring that remote end is not available

This CR is a proposed enhancement for a backward compatible change on the `0.14.x` branch such that so that all the three TCP Keepalive parameters become configurable.  (In this PR, the next release is assumed to be `0.14.21`, but can easily adjust the proposed deprecation message if that was not the case.)

### DIFF
--- a/src/server/server.rs
+++ b/src/server/server.rs
@@ -2,10 +2,15 @@ use std::error::Error as StdError;
 use std::fmt;
 #[cfg(feature = "tcp")]
 use std::net::{SocketAddr, TcpListener as StdTcpListener};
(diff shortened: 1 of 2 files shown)


hansonchar:
Doesn't seem I can rerun the failed workflow.  Let me close the PR and re-open a new one to see if that would work.

PR #2990, opened 2022-09-22
Supports configuration of all TCP Keepalive parameters per  https://en.wikipedia.org/wiki/Keepalive

hansonchar:
*  `Keepalive time` is the duration between two keepalive transmissions in idle condition. 
* `Keepalive retry` is the number of retransmissions to be carried out before declaring that remote end is not available.

This CR is a proposed enhancement for a backward compatible change on the `0.14.x` branch so that all the three TCP Keepalive parameters would become configurable.  (In this PR, the next release is assumed to be `0.14.21`, but can easily adjust the proposed deprecation message if that was not the case.)

### DIFF
--- a/src/server/server.rs
+++ b/src/server/server.rs
@@ -2,10 +2,15 @@ use std::error::Error as StdError;
 use std::fmt;
 #[cfg(feature = "tcp")]
 use std::net::{SocketAddr, TcpListener as StdTcpListener};
-#[cfg(any(feature = "tcp", feature = "http1"))]
+
+#[cfg(feature = "tcp")]
 use std::time::Duration;
 
 use pin_project_lite::pin_project;
+
+#[cfg(feature = "tcp")]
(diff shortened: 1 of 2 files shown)


hansonchar:
Sorry, need to fix the git commit message.  Will repost CR later.

seanmonstar:
You can just push changes to the same branch, no need to close the PR. As for the commit message, it's just a nudge. I can fix it during merging.

Regarding the change itself, I'm happy to add it to 0.14.x, though I would suggest instead of deprecating the method and exposing the `socket2` type, we instead just add a couple more methods to internally edit the settings. That way we don't tie hyper's version to socket2's.

hansonchar:
> we instead just add a couple more methods to internally edit the settings.

Will look into this.  Thanks!

hansonchar:
Doesn't seem I can reopen this now, so created another PR https://github.com/hyperium/hyper/pull/2991

PR #2991, opened 2022-09-22
Supports configuration of all TCP Keepalive parameters per https://en.wikipedia.org/wiki/Keepalive

hansonchar:
*  `Keepalive time` is the duration between two keepalive transmissions in idle condition. 
* `Keepalive retry` is the number of retransmissions to be carried out before declaring that remote end is not available.

This CR is a proposed enhancement for a backward compatible change on the `0.14.x` branch so that all the three TCP Keepalive parameters would become configurable.

FYI per empirical evidence enabling TCP keepalive with `intervals` and `retries` drastically/significantly improves the stability of persistent TCP connections while reducing the number of file descriptors via closing down dead connections.

### DIFF
--- a/src/server/server.rs
+++ b/src/server/server.rs
@@ -2,10 +2,12 @@ use std::error::Error as StdError;
 use std::fmt;
 #[cfg(feature = "tcp")]
 use std::net::{SocketAddr, TcpListener as StdTcpListener};
-#[cfg(any(feature = "tcp", feature = "http1"))]
+
+#[cfg(feature = "tcp")]
 use std::time::Duration;
 
 use pin_project_lite::pin_project;
+
 use tokio::io::{AsyncRead, AsyncWrite};
 use tracing::trace;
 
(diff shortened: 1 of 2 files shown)


hansonchar:
I believe I've fixed all the stylistic issues reported by the workflow, but it doesn't seem I can trigger a re-run of the workflow.  I also observed some workflow failures unrelated to this change, so they are probably existing failures.

Please let me know if I missed anything.  

If possible, I'd like to have this change pushed/released asap, so I can make further PR to `axum-server` based on this change.

hansonchar:
I see there are platform specific conditional compilations in the `socket2` crate causing some failure.  Looking into them.

seanmonstar:
Could just slap `#[cfg(unix)]` on the extra methods, and not worry about Windows...

hansonchar:
> Could just slap `#[cfg(unix)]` on the extra methods, and not worry about Windows...

Right, but I am trying to make it such that the same API can be provided to the upper layer, so that if a method for a specific parameter that isn't supported on a specific platform calling the method would just be a no-op.  I've just pushed such changes, and added a few unit tests on it.  Hopefully they would pass these tests.

seanmonstar (approved):

seanmonstar:
The master branch has removed both `Server` and `AddrIncoming`. Helpers (such as Tokio integration) will now live in the `hyper-util` crate.

PR #2993, opened 2022-09-22
Use stable rust for building C API

weihanglo:
Rust 1.64.0 stabilizes `cargo rustc --crate-type`, so hyper can use stable instead.


seanmonstar (approved):
Thanks! 🎉

PR #2996, opened 2022-09-25
[WIP] Split server conn types

### LINKED ISSUE #2851: Split the server Connection type into version-specific submodules
The `server::conn` type, `Connection`, should get version-specific types in submodules `http1` and `http2`.

Michael-J-Ward:
Closes #2851

If someone takes a look and provides some guidance, I'll execute the next steps.

# TODO:
- [x] create separate http1 type
- [x] create separate http2 type
- [ ] configuring / constructing?
- [ ] excise old `server::conn::Connection`

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -59,6 +59,7 @@ use crate::upgrade::Upgraded;
 use crate::{common::time::Time, rt::Timer};
 
 pub mod http1;
+pub mod http2;
 
 cfg_feature! {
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for getting started with this! You said you think next steps are configuration? The `Http` type could likely be split into a `Builder` in each module, only including the relevant version methods. Similar to how it is in `client::conn`.

seanmonstar:
Thank you for the big start! I've finished up the split in #3011.

Issue #2995, opened 2022-09-23
Remove the "raw" headers option in C API


seanmonstar:
The C API grew a concept of asking for the "raw" headers, but it turned out to not be useful in practice. We should remove them now while the API is still unstable.

- Remove `hyper_clientconn_options_headers_raw()`.
- Remove `hyper_response_headers_raw()`
- Delete the now unused code in client and proto.
- Don't forget to run `capi/gen_header.sh` to sync up the `hyper.h` file.

PR #3002, opened 2022-10-03
Refactor(ffi): remove the "raw" headers option in C API

dannasman:
Hello,
Solution proposal for issue #2995.


seanmonstar:
Thanks for the PR! I think we could also remove the related parts in proto... I'm surprised it doesn't give a warning about it being unused in CI.

dannasman:
Oh I see. So the whole raw headers option should be removed (like the title of the issue suggests) not just the functions `hyper_clientconn_options_headers_raw()` and `hyper_response_headers_raw()`?

seanmonstar:
Yes, there's no other reason for that internal code to exist.

seanmonstar (approved):
Superb, thank you!

PR #3008, opened 2022-10-11
fix(client): distinguish the H2CheckoutIsClosed error

jiangshibiao:
I have occurred a bug when using hyper and reqwest to send HTTP requests, and I have tried to fix it.

Please first look at the case below the modified code. If the _checkout_ occurs error, it may change to _H2CheckoutIsClosed_ error so that the retry mechanism can be done outside. However, the same logic doesn't appear in the above case, so the retry mechanism can not be activated anymore if the checkout is closed.

I'm not sure which error message (checkout error/connecting error) is more proper to be raised, and it seems not so important.

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -395,7 +395,13 @@ where
             // In both cases, we should just wait for the other future.
             Either::Left((Err(err), connecting)) => {
                 if err.is_canceled() {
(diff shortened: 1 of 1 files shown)


seanmonstar:
I think this looks correct. Have you found that using this code fixes the case you noticed? What did it fix?

PR #3018, opened 2022-10-21
server::conn::http1::Connection always holds a proto

bossmc:
In the recent work to split the `server::conn::Http` into protocol-specific halves, the author of that change left the inner `proto` in an `Option` (and left a comment asking if this was needed any more).  This used to be needed as part of the fallback processing (allowing the fallback to `take()` the HTTP/1 `proto` and break it into `Parts` in order to rebuild the HTTP/2 `proto`).

Now, since we're definitely using HTTP/1 we can just store the `proto` directly and, in the process delete a bunch of `unwrap` calls 💯.  We need to be able to `take` the HTTP/1 `Connection` in a few places (when an `UpgradeableConnection` is upgraded for example) but we can scope down the places where we have to `unwrap` an `Option` that's _almost always_ `Some`.

I also tidied up a few instances of dead code (some are feature-specific) which I ran into when running `cargo test` over this change.

### DIFF
--- a/src/server/conn/http1.rs
+++ b/src/server/conn/http1.rs
@@ -27,7 +27,7 @@ pin_project_lite::pin_project! {
     where
         S: HttpService<Recv>,
     {
-        conn: Option<Http1Dispatcher<T, S::ResBody, S>>,
+        conn: Http1Dispatcher<T, S::ResBody, S>,
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/server/conn/http1.rs
```diff
@@ -116,25 +111,20 @@ where
...
+            _inner: (),
+        }
     }
 
     /// Return the inner IO object, and additional information, if available.
     ///
     ///
-    /// TODO:(mike) does this need to return none for h1 or is it expected to always be present? previously used an "unwrap"
-    /// This method will return a `None` if this connection is using an h2 protocol.
     pub fn try_into_parts(self) -> Option<Parts<I, S>> {
```
seanmonstar, line 126: In fact, this function now seems unnecessary in its entirety, ya?

seanmonstar (commented):
Thanks for cleaning this up! <3

src/server/conn/http1.rs
bossmc: It is correct.  I considered that and then left it in since removing it would be a breaking change... I'll remove 😸

seanmonstar (approved):

PR #3019, opened 2022-10-21
Apply clippy fixes, except in #[test] code

bossmc:
This also adds a CI stage to run `clippy` over the project too, hopefully this will prevent degradation.


bossmc:
Build failure was due to my rustc allowing nightly features by default, I've marked the benches that use `#![feature(test)]` as also needing `#![cfg(feature = nightly)]` so they won't get built without the `nightly` feature enabled.

I've also rebased to pick up the recent `master` changes.

FrankReh:
I will add it's nice if the project would pass clippy cleanly because it makes playing around with the package easier for a new-comer.

Issue #2840, opened 2022-05-20
Explore forwards-compatibile Body trait


seanmonstar:
We want to explore a trait that can be [forwards-compatible in frame types for a body stream](https://github.com/hyperium/hyper/blob/master/docs/ROADMAP.md#how-could-the-body-trait-prepare-for-unknown-frames).

seanmonstar:
# Proposal

## Problem

The `Body` trait provides methods to allow basically 2 frame types to be passed along: data, and trailers. There's a [video][] overview, but I'll summarize: in order to be **Flexible**, we want to allow `Body`s to transmit frame types that hyper doesn't even know about. We can't do that with per-frame-type methods.

### Frames
We want a `Frame` like thing that users can poll (and send). It would be convenient if it behaved like an `enum`, where they could match on the variants, but there's a forwards-compatibility trap. Consider:

```rust
match f {
    Frame::Data(buf) => {},
    Frame::Trailers(map) => {},
    Frame::Unknown(raw) => {},
    // #[non_exhaustive]
    _ => {},
}
```

If we were to start to "understand" a frame type that used to fall into the `Unknown` variant, and created `Frame::Priority(pri)`, then user code that updated would suddenly no longer be handling the frame.

So, we don't want to expose an enum directly. Something like the [dyn match][] concept would be great, but it doesn't exist yet. 

It'd be possible to inspect the frame using a bunch `if`s. But Rust programmers like the feel of `match`, so a macro could do it.

## Proposed Design

```rust
pub trait Body {
    type Data: Buf;
    fn poll_frame(..) -> Result<Option<Frame<Self::Data>>>;
}

pub struct Frame<T>(Kind<T>);

enum Kind<T> {
    // The first two variants are "inlined" since they are undoubtedly
    // the most common. This saves us from having to allocate a
    // boxed trait object for them.
    Data(T),
    Trailers(HeaderMap),
    Unknown(Box<dyn Frameish>),
}

/// The name is just YOLO for now.
pub trait Frameish {
    fn frame_kind(&self) -> u64;
    // Just a byte slice? The DATA type couldn't necessarily fulfill this...
    fn payload(&self) -> &[u8];
}
```

## Usage

### Matching

```rust
while let Some(fr) = body.frame().await? {
    // the macro is optional, but seems helpful
    match_dyn! { fr;
        frame::Data(buf) => {
            
        },
        frame::Trailers(trailers) => {
            
        },
        other => {
[...]

PR #3020, opened 2022-10-24
feat(body): update `Body` trait to use `Frame`s

### LINKED ISSUE #3010: Update to use new http-body trait
With #2840 completed, we need to update hyper to use the new trait.

seanmonstar:
The `Body` trait was adjusted to be forwards compatible with adding new frame types. That resulted in changing from `poll_data` and `poll_trailers` to a single `poll_frame` function. More can be learned from the proposal in https://github.com/hyperium/hyper/issues/2840.

Closes #3010 

BREAKING CHANGE: The polling functions of the `Body` trait have been
  redesigned.

  The free functions `hyper::body::to_bytes` and `aggregate` have been
  removed. Similar functionality is on
  `http_body_util::BodyExt::collect`.


examples/client.rs
```diff
@@ -62,9 +62,11 @@ async fn fetch_url(url: hyper::Uri) -> Result<()> {
 
     // Stream the body, writing each chunk to stdout as we get it
     // (instead of buffering and printing at the end).
-    while let Some(next) = res.data().await {
-        let chunk = next?;
-        io::stdout().write_all(&chunk).await?;
+    while let Some(next) = res.frame().await {
+        let frame = next?;
```
LucioFranco, line 66:
```suggestion
    while let Some(frame) = res.frame().await.transpose()? {
```

I think we can write it like this?

src/lib.rs
```diff
@@ -2,7 +2,7 @@
 #![deny(missing_debug_implementations)]
 #![cfg_attr(test, deny(rust_2018_idioms))]
 #![cfg_attr(all(test, feature = "full"), deny(unreachable_pub))]
-#![cfg_attr(all(test, feature = "full"), deny(warnings))]
+//#![cfg_attr(all(test, feature = "full"), deny(warnings))]
```
LucioFranco, line 5: Was this intentional to commit?

LucioFranco (approved):

src/lib.rs
seanmonstar: Nope 🙃

PR #3034, opened 2022-10-29
chore(body): pub empty

fundon:
Sometimes we want to replace the body.

```rust
BodyExt::collect(replace(req.body_mut(), IncomingBody::empty()))
```

### DIFF
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -104,8 +104,9 @@ impl Incoming {
         Incoming { kind }
     }
 
+    /// Create an empty `Body` stream.
     #[allow(dead_code)]
-    pub(crate) fn empty() -> Incoming {
+    pub fn empty() -> Incoming {
         Incoming::new(Kind::Empty)
     }
 


seanmonstar:
Thanks for the PR!

We actually recently just made it private, since the type is _only_ meant to represent a body streaming from an IO transport. In the example you mentioned, I would usually `map_body` to put the fully collected body type in, so it becomes `Request<Collected>`.

fundon:
Thanks for the reply.

I'll try the `map_body` method first in my case.

fundon:
My case likes below:

https://github.com/seanmonstar/warp/blob/master/src/route.rs#L130-L140

```rust
    pub(crate) fn take_body(&mut self) -> Option<Body> {
        match self.body {
            BodyState::Ready => {
                let body = mem::replace(self.req.body_mut(), Body::empty());
                self.body = BodyState::Taken;
                Some(body)
            }
            BodyState::Taken => None,
        }
    }
```

fundon:
Other case, if we want to limit the body first.

```rust
let body = Limited::new(req.body_mut(), Body::empty()), 1024);
BodyExt::collect(body).await
```

Maybe adds `Default` trait to `Incoming` would be a good solution.
So we can use `std::mem::take` method to take body.

fundon:
Other way to fix that, just define a enum, includes `None`, `Empty`, `Incoming`, `Stream` fileds.

PR #3037, opened 2022-10-29
refactor(rt): Clean up Timer trait

### LINKED ISSUE #3028: `Timer` methods should return `Pin<Box<dyn Sleep>>` and `Sleep` should drop the `Unpin` requirement
Otherwise, you need to double-box `!Unpin` timer futures like Tokio's.

sfackler:
This both avoids an unnecessary extra layer of boxing and removes some dubious unsafe code.

Closes #3028

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -20,16 +20,16 @@ pub trait Executor<Fut> {
 /// A timer which provides timer-like functions.
 pub trait Timer {
     /// Return a future that resolves in `duration` time.
-    fn sleep(&self, duration: Duration) -> Box<dyn Sleep + Unpin>;
+    fn sleep(&self, duration: Duration) -> Pin<Box<dyn Sleep>>;
 
     /// Return a future that resolves at `deadline`.
-    fn sleep_until(&self, deadline: Instant) -> Box<dyn Sleep + Unpin>;
+    fn sleep_until(&self, deadline: Instant) -> Pin<Box<dyn Sleep>>;
 
     /// Reset a future to resolve at `new_deadline` instead.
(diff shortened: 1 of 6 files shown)


src/rt.rs
```diff
@@ -20,16 +20,16 @@ pub trait Executor<Fut> {
 /// A timer which provides timer-like functions.
 pub trait Timer {
```
notgull, line 21: Unrelated: Would it be good to have the resulting futures as an associated type rather than as a boxed future? Both `tokio` and `smol` have well-defined "sleep" types instead of `async` methods, so we may be able to avoid an allocation here by not boxing it.

sfackler: That would require parameterizing the ~entire API over the timer trait which I'm not sure is desired.

notgull: As a workaround, you could just do:

```rust
struct BoxedTimer {
    inner: Box<dyn Timer<Sleep = Pin<Box<dyn Future>>>>>
}
```

But this might not be idiomatic, so I agree with your assessment.

bossmc:
Is there an assumption in hyper that the `Sleep` future(s) cancel on drop?  I could imagine a fairly simple implementation (a `BinaryHeap` of wakers) where sleep-futures continue to fire their wakers after the future has been dropped - would this be a problem?

If this is expected, it would be nice to specify in the trait comment so implementors get this right.

sfackler:
Any `Future` implementation needs to work in the presence of polls that don't make progress, so I can't see how hyper could make that assumption.

bossmc:
Fair enough, deregistering/cancelling the timer on `Drop` of the `Sleep` future is therefore simply an optimization (to reduce unnecessary wakeups) but not needed for correctness, thanks for confirming!

seanmonstar (approved):

Issue #2419, opened 2021-02-05
Http2: Hyper client gets stuck if too many requests are spawned


staticgc:
I have a golang https server to which a rust client sends lot of parallel requests.
After few hundreds of requests, it stops and apparently there is no TCP communication.

The REST API accepts a byte buffer in a body and responds back with its length in a json.

To reproduce, the entire code is at: https://github.com/staticgc/hyper-stuck
It has golang server, Rust Client & Golang Client
The build instructions are very straightforward 

The default max number of futures spawned are [400](https://github.com/staticgc/hyper-stuck/blob/main/test.env)
This apparently is more than the max number of http2 streams at server (which is 250)

If the count is reduced to 200 it works.

Other observations:

1. Golang client works with high goroutine count
2. Increasing connections in Rust client from 1 to say 3 causes more requests to be processed but again gets stuck
3. With 400 futures/requests, when the number of connections = 10 (chosen arbitrarily) it worked ok.
4. When [HTEST_BUF_SIZE](https://github.com/staticgc/hyper-stuck/blob/main/test.env) is reduced to say 1KB then even with 1 connection & high future count it works
5. If the rust client stops, it does so after predictable number of requests

Number of connections here refers to http2 negotiated connection which hyper (I think) creates only 1 per `Client` instance.
So changing [HTEST_CONN_COUNT](https://github.com/staticgc/hyper-stuck/blob/main/test.env) changes the number of `Client` instances that are created.

Also to prevent initial flooding at the server, the rust client makes a single http2 request on each `Client` instance and then issues the parallel requests.

```rust
for i in 0..conn_count {
     send_req_https(client_vec[i].clone(), bufsz, url.as_str()).await?;
}
```

Able to reproduce on: Mac OS & Cent OS Linux

Edit: Apologies for some typos in build instructions in the repo above. I have fixed those. Let me know here if anything remains.

staticgc:
I also changed the `#[tokio::main]` to use single threaded basic schedular and after it got stuck this was the call graph (on mac).

It seems it is waiting for event from OS (kevent) to move ahead.

```
Call graph:
    2865 Thread_8562679   DispatchQueue_1: com.apple.main-thread  (serial)
      2865 start  (in libdyld.dylib) + 1  [0x7fff61883015]
        2865 main  (in hyper-stuck) + 34  [0x10322a072]
[...]

PR #3041, opened 2022-11-02
fix(server): Fix race condition in client dispatcher (#2419)

jfourie1:
There exists a race condition in ClientTask::poll() when the request that is sent via h2::client::send_request() is pending open. A task will be spawned to wait for send capacity on the sendstream. Because this same stream is also stored in the pending member of h2::client::SendRequest the next iteration of the poll() loop can call poll_ready() and call wait_send() on the same stream passed into the spawned task.

Fix this by always calling poll_ready() after send_request(). If this call to poll_ready() returns Pending save the necessary context in ClientTask and only spawn the task that will eventually resolve to the response after poll_ready() returns Ok.

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -7,12 +7,14 @@ use futures_channel::{mpsc, oneshot};
 use futures_util::future::{self, Either, FutureExt as _, TryFutureExt as _};
 use futures_util::stream::StreamExt as _;
 use h2::client::{Builder, SendRequest};
+use h2::SendStream;
 use http::{Method, StatusCode};
 use tokio::io::{AsyncRead, AsyncWrite};
 use tracing::{debug, trace, warn};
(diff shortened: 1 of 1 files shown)


jeromegn:
Worth mentioning this fixes the reproduction:

```
2022-11-03T18:48:08.374569Z  INFO h2_repro: 100 requests on 50 streams
2022-11-03T18:48:08.374591Z  INFO h2_repro: (Set $H2_REQUESTS and $H2_MAX_STREAMS environment variables to adjust)
2022-11-03T18:48:08.407164Z  INFO h2_repro: H1: Completed 100 / 100
2022-11-03T18:48:08.425192Z  INFO h2_repro: H2: Completed 100 / 100
```

seanmonstar (approved):
Phenomenal work debugging this issue!!!

PR #3042, opened 2022-11-03
feat(client) implement Clone for http2::SendRequest

### LINKED ISSUE #3036: Should client::conn::http2::SendRequest be Clone?
Should [`hyper::client::conn::http2::SendRequest`](https://docs.rs/hyper/1.0.0-rc.1/hyper/client/conn/http2/struct.SendRequest.html) be `Clone`, just like [`h2::client::SendRequest`](https://docs.rs/h2/0.3.15/h2/client/struct.SendRequest.html)?

vi:
Resolves #3036

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -24,6 +24,12 @@ pub struct SendRequest<B> {
     dispatch: dispatch::UnboundedSender<Request<B>, Response<IncomingBody>>,
 }
 
+impl<B> Clone for SendRequest<B> {
+    fn clone(&self) -> SendRequest<B> {
+        SendRequest { dispatch: self.dispatch.clone() }
+    }
+}
+
 /// A future that processes all HTTP state for the IO object.
 ///
 /// In most cases, this should just be spawned into an executor, so that it


src/client/conn/http2.rs
```diff
@@ -20,6 +20,7 @@ use crate::proto;
 use crate::rt::{Executor, Timer};
 
 /// The sender side of an established connection.
+#[derive(Clone)]
```
LucioFranco, line 23: I think you need a custom impl of clone here since this will generate a `impl<B: Clone> Clone for SendRequest<B>` which we don't care if B is clone at all. I believe the UnboundedSender type has the same type of bound that doesn't have clone dependent on the T: clone.

vi: Indeed.

Forgot that there is a generic parameter (and probably would have assumed it would add `where dispatch::UnboundedSender<Request<B>, Response<IncomingBody>> : Clone`, not just `where B : Clone` anyway).

Re-uploading (also rebased).

seanmonstar (approved):

PR #3044, opened 2022-11-04
feat(error): preserve `InvalidUri` details

### LINKED ISSUE #3043: URI parse errors should preserve information from the HTTP crate

When an invalid URI is parsed by the `http` crate, it returns  an `InvalidUri` error. When `hyper` encounters one of these errors, it converts it into a `hyper::Error` type with an error kind that indicates that an invalid URI was encountered. The `http::uri::InvalidUri` type contains additional information describing what part of the URI was invalid, which is included in the error's `fmt::Display` output: https://github.com/hyperium/http/blob/c1f309e0c3695f8e14e1b54c738681783d4c7b9e/src/uri/mod.rs#L1063-L1079

Unfrortunately, when `hyper` converts these errors into its `hyper::Error` type, the additional information is discarded: https://github.com/hyperium/hyper/blob/75aac9f47fe0246016e6133cd3cfa35b63c8904e/src/error.rs#L489-L499

The error that's returned to the application will format as just "invalid URI", which is not particularly descriptive. Being able to surface more information about why a URI was invalid would be useful, particularly in applications where URIs are input by the user.

It would be nice if `hyper` preserved the description of what part of the URI was invalid. I think we should probably change the `Parse::Uri` variant in `hyper::error` to contain an `http::uri::InvalidUri`, so that we can format the additional error information in the `fmt::Display` impl for `hyper::Error`.

Since we don't want to expose types from other crates in `hyper`'s public API, for stability reasons, we probably shouldn't allow the user to access the `http::uri::InvalidUri` error in the `hyper::Error` source chain, or allow downcasting the error to it.

- We could include the `http::uri::InvalidUri` as a `source` of `hyper::Error`. This might be conceptually the "right thing", but it would expose a type from an external crate in `hyper`'s public API, which is undesirable as `http` is not yet 1.0.
- We could leave things the way they are now. Unfortunately, that would mean that hyper would completely swallow information describing what part of the URI was invalid, and this would never be exposed to the user.

hawkw:
This branch changes the representation of the `Parse::Uri` error kind to preserve information provided by the `http` crate's `uri::InvalidUri` and `uri::InvalidUriParts` errors. A new `InvalidUri` enum is added that holds one of those two error types, or a custom message (since `hyper` currently returns `Parse::Uri` errors that didn't come from an inner `http` error in some cases). The new enum has a custom `Display` and `Debug` implementation to reduce repetition of the string "invalid URI" in its formatted output.

This is _not_ stored as the error's `cause` currently in order to avoid exposing the `http` crate's error types in the public API. However, when `http` 1.0 is released, we can simplify this code significantly by storing the error as a cause and exposing it in the source chain.

Closes #3043


seanmonstar:
> This is not stored as the error's cause currently in order to avoid exposing the http crate's error types in the public API.

hyper already exposes `http` publicly, and `http` will need to be marked 1.0 before hyper. If it does make things easier, than we can just do that.

hawkw:
@seanmonstar In that case, it might be nicer to expose `http`'s errors as a `source` and `downcast_ref` target!

seanmonstar:
Keeping more details to improve error messages is always our goal. I'm going to close this, but a fixup and re-open is welcome.

PR #3046, opened 2022-11-07
Add test for a large Content-Length header in body::to_bytes

orihjfrog:
This is a quick fix to make sure that the expected body length is less than 10GB before reading all of it. If the expected length is too big, the function will return an empty buffer.

### DIFF
--- a/src/body/to_bytes.rs
+++ b/src/body/to_bytes.rs
@@ -54,8 +54,7 @@ where
 
     let content_length = body.size_hint().lower();
 
-    // If the expected content length is to large, returnan empty body.
-    // It is also possible to return an error instead.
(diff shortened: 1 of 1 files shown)


srmish-jfrog:
This will prevent allocation-related crashes in the default case where there is no limit on `content-length`

orihjfrog:
Possibly consider returning an error instead of an empty body

seanmonstar:
Thanks for the PR! We've discussed this before in email. The best option is likely that people would use `http_body_util::Limited`. (Also sounds like a nice addition to `tower-http` to wrap bodies similar to the body timeout layers). The user can pick the size they want, and it will work with any body collection future, not just this `to_bytes`.

PR #3060, opened 2022-11-18
Update Polish period example

tgross35:
I keep seeing discussion about hyper's [Polish period](https://seanmonstar.com/post/701008919383932928/hyper-polish-period) so figured it might as well be embraced with some Polish.

(feel free to enjoy & merge or close as you see fit)


seanmonstar:
polish soccer fans

(But I'm going to keep the example in English :D)

Issue #3052, opened 2022-11-08
Backport and Deprecations to 0.14.x from 1.0


seanmonstar:
In order to ease the upgrade from 0.14.x to 1.0, a big help would be to backport additions from 1.0, and add deprecation warnings of things that will go away. Ideally, this will allow people using 0.14.x versions to start updating their apps and libraries with guides from the compiler, without having to fully swallow the 1.0 pill yet.

## Backports

- We can't backport _everything_. Some of the changes are breaking in nature. We can only backport **additions**.

## Deprecations

- As things are backported, we can add deprecation warnings to the existing types that are removed in 1.0, guiding them to start using the backported APIs.
- We can also add warnings to types or methods which are just completely gone in 1.0.
- For types that are moving to `hyper-util`, we could add a weaker deprecation warning, especially if the type that is moving is easy to copy.

## Opt-in / Opt-out

- We can create a new cargo feature, such as `1.0-backports` (or some better name), so people only compile the backports if they want to use them.
- I'm undecided about whether deprecations should be on by default, possibly with an opt-out cargo feature, or only opt-in.
  - If they are opt-in, people may not realize they could start preparing early.
  - There is already a built-in way to opt-out of a deprecation warning, `#[allow(deprecation)]`.

kate-shine:
Personally, I'd prefer it to be an opt-out feature.
- opt-in would, in my experience, result in plenty of projects missing it
- enabling it without any attached feature would, as you said, allow disabling by `#[allow(deprecation)]`, but I don't think I like the idea of allowing deprecation warnings for all crates because of one

jcgruenhage:
A way for the backports to be opt-out would still have to be an opt-in feature that's enabled by default. Features are additive, and should only ever enable new functionality. Code that compiles without a feature enabled should never break when a new feature is enabled. Source: https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification

seanmonstar:
Yea, backports can simply be opt-in: we'll make a cargo feature you can enable, and won't put it in `default`.

> I'd prefer it to be an opt-out feature.

I assume you mean being able to opt-out of the deprecation warnings. Users wouldn't have to `#![allow(deprecated)]` and hide all other crates, they can add the `allow` directly on the import/use line.

epage:
clap initially did opt-in but switched to opt-out.  https://github.com/clap-rs/clap/issues/3822 is the issue that led us to making the switch.

[...]

PR #3063, opened 2022-11-25
feat(client): backport split client conn modules

### LINKED ISSUE #3053: Backport the split client conn modules
In 1.0, we've split `hyper::client::conn::Connection` into per-version types. To ease upgrading (see #3052), we can backport the addition of the two modules, `hyper::client::conn::{http1, http2}`. With them in place, we could then add a deprecation to `hyper::client::conn::Connection`.

I don't think this will be actually be `E-medium` in complexity, as it's not adding something that needs many internal changes. But it's not _quite_ easy, since once the files are copied, there might be some confusing errors over internal types that have been renamed.

kxt:
Backport `client::conn` split modules to ease transition to the 1.0 API and deprecate `client::conn::Builder` and `client::conn::handshake`.

Closes #3053

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -62,11 +62,11 @@ jobs:
 
         include:
           - rust: stable
-            features: "--features full"
+            features: "--features full,backports"
           - rust: beta
-            features: "--features full"
+            features: "--features full,backports"
           - rust: nightly
-            features: "--features full,nightly"
+            features: "--features full,nightly,backports"
             benches: true
 
     runs-on: ${{ matrix.os }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,6 +112,9 @@ ffi = ["libc"]
 # 1.0 compatibile client API
 backports = []
 
+# whether or not to display deprecation warnings
+deprecated = []
+
 # internal features used in CI
 nightly = []
 __internal_happy_eyeballs_tests = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,6 +109,9 @@ tcp = [
 # C-API support (currently unstable (no semver))
 ffi = ["libc"]
 
+# 1.0 compatibile client API
+backports = []
+
 # internal features used in CI
 nightly = []
 __internal_happy_eyeballs_tests = []
--- a/examples/tower_client.rs
+++ b/examples/tower_client.rs
@@ -1,20 +1,20 @@
 #![deny(warnings)]
 
-use hyper::client::conn::Builder;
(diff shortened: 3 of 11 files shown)


seanmonstar:
Thanks for getting started on this! I'll try to give CI a kick frequently, looks like it needs a `cargo fmt` run, and perhaps allow the warning for the C API? Looks like compile errors, actually...

seanmonstar:
I'm trying to get Miri fixed, but looks like there is a legit failure reported, that the unit tests are using the deprecated API (as they should, we shouldn't remove tests), and that triggers the `deny(warnings)`. Could fix that either with a `#![allow(deprecated)]` at the top of the file, though that means we won't notice any other crates deprecations, or local `#[allow(deprecated)]` attributes at each usage location.

kxt:
Sorry about the previously failing CI tests, I'm not familiar with GH workflows and it took a while for me to set them up working.
The CI tests ran green bar the Miri tests, I don't know what's up with those.

seanmonstar:
Sorry to delay a little, I wanted to check on something. We've had some useful discussion about deprecations in #3052. It seems worth it to put all deprecation warnings behind a `deprecated` cargo feature, that we will have off by default to start, and eventually add to the `default-features`.

Do you want to add that to this PR? If it's too much to do, I can also do it, just let me know :D

kxt:
I can add it in the next few days. However, the `deprecated` feature sound odd to me, a feature that causes compile errors/warnings without offering a viable alternative. Having the deprecation behind the `1_0-backports` feature would make more sense to me. But I haven't read the mentioned clap issue yet, so there might be a good reason to the `deprecated` feature that I'm missing.

Also, I guess making the deprecated warnings default should be done with a minor version bump (ie. 0.15) per semantic versioning.

howardjohn:
I'd be interesting with helper out here if there is any work needed

kxt:
I also wonder if I can do anything to move this forward.

seanmonstar:
Sorry, there were a couple of new breaking changes that I realized needed to be made to 1.0. We've made them now. It'd be helpful to users to make sure the "backport helpers" match what ends up in 1.0, so we needed to make them first. For instance, in this PR, the builders no longer have their methods prefixed with the HTTP version (so, not `http1_writev()`, etc).

kxt:
Then I guess this PR should be dropped and the issue postponed until it's clear what needs to be backported.

seanmonstar:
Alright, thanks so much @kxt for doing all this. Sorry for the delay, I've taken your commit and amended it with the latest changes in 1.0, and split it into 2, so we could have backports and deprecations be separate. That's in #3155 and #3156.

PR #3065, opened 2022-11-30
docs: add oddgrd to the list of triagers

oddgrd:
Thanks for the invitation, happy to help out however I can!


seanmonstar (approved):
After many excellent PRs through the rc1 dev period, welcome aboard!

Issue #1675, opened 2018-10-18
Clean up Client and Server integration tests


seanmonstar:
Both the client and server tests (`tests/client.rs` and `tests/server.rs`) started with small abstractions, but they weren't flexible enough, and so as more complicated cases were needed, the tests just got more and more complicated, with a lot of repetition. It'd be huge to find a way to refactor them.

JeffLabonte:
I will take this issue and work on it!

tiendq:
@seanmonstar is it still open? :)

seanmonstar:
Yes, this is still open because the integration tests feel very messy, and there's a lot of repeated code. I'm sorry I haven't been able to provide much guidance on how to improve them, if I thought about it a lot, I'd probably fix them myself :)

tiendq:
Well, if it's very messy so we could improve it a little each PR, for example let's remove (as many as possible) repeated code first. If you feel ok with this approach I could try it :) (hopefully it's easy as labelled :D).

ryanyz10:
I'd like to try and tackle this one if it hasn't been done yet! The tests seem quite long though...

ryanyz10:
Ok great, I'll slowly work on this then.

alexs-sh:
Hi @seanmonstar
Could you please take a look at https://github.com/hyperium/hyper/pull/3072

This PR doesn't provide any significant changes but allows to make one step in decreasing code duplication. If it's the wrong way, please, just reject the PR. If it brings something useful, then I can continue with it.

Thanks

PR #3072, opened 2022-12-04
refactor(test):#1675 use fns for creating test server

alexs-sh:
This is a part of https://github.com/hyperium/hyper/issues/1675 issue. It
slightly reduces the number of duplicated code by using one common function for
init a logger and creating a test server.

### DIFF
--- a/tests/client.rs
+++ b/tests/client.rs
@@ -1351,8 +1351,12 @@ mod conn {
 
     use support::{TokioExecutor, TokioTimer};
 
-    async fn setup_test_server() -> (TkTcpListener, SocketAddr) {
+    fn setup_logger() {
         let _ = pretty_env_logger::try_init();
+    }
+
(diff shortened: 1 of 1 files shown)


alexs-sh:
This PR doesn't provide any significant changes but allows to make one step in decreasing code duplication. If it's the wrong way, please, just reject the PR. If it brings something useful, then I can continue with it.

seanmonstar (approved):
This is _exactly_ what I meant in the original issue, that some boilerplate had been copied around way too much. Thank you **so** much for tackling this <3

alexs-sh:
Happy to help :) 

There is yet another possibility to reduce boilerplate code. For example, tests often use the pattern "create server and spawn processing" + "create client and do something". Processing and doing something differ from test to test, but creating a server and connecting a client is often very similar. So I can try to generalize them and show the results.

PR #3075, opened 2022-12-06
feat(examples): re-enable echo uppercase endpoint

oddgrd:
While I was working on the hyper website server echo guide, I noticed this endpoint in the echo example is still disabled (by me :smile:). I was able to re-enable it and it does what it's supposed to, but I'm not sure if this is the idiomatic way to do it now. 

I looked around for another example (or test) for how to do this, but struggled to find something that matched the full flow (get body from request, mutate it without collecting it and return it in a response).


seanmonstar (approved):
Nice!

PR #3076, opened 2022-12-07
Add set_interface connector option

hottea773:
This PR is a duplicate of https://github.com/hyperium/hyper/pull/2484 which I'd like to use, but is closed.

(I've minorly changed it such that we bind to the interface before we bind to the address otherwise it can cause issues with failing to find addresses on VRFs.)

I'd like to use it soon, so I've made it against the `0.14.x` branch. I hope that's ok?

### DIFF
--- a/src/client/connect/http.rs
+++ b/src/client/connect/http.rs
@@ -81,6 +81,7 @@ struct Config {
     reuse_address: bool,
     send_buffer_size: Option<usize>,
     recv_buffer_size: Option<usize>,
+    interface: Option<String>,
 }
 
 // ===== impl HttpConnector =====
@@ -121,6 +122,7 @@ impl<R> HttpConnector<R> {
                 reuse_address: false,
                 send_buffer_size: None,
                 recv_buffer_size: None,
+                interface: None,
             }),
             resolver,
         }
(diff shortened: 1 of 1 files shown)


hottea773:
@seanmonstar sorry for the tag, hope you're the right person. Is there any chance this might get looked at any time soon?

seanmonstar:
So, while I'm supportive of adding the feature in general, I do feel divided about adding new things to 0.14.x. Ideally (from my point of view), this would land in `hyper-util`. Is that fine for your use case? It could mean it's a little harder to start using it right away.

hottea773:
> So, while I'm supportive of adding the feature in general, I do feel divided about adding new things to 0.14.x. Ideally (from my point of view), this would land in `hyper-util`. Is that fine for your use case? It could mean it's a little harder to start using it right away.

Thanks, that makes sense. I think using the `0.14` fork for a few months should be ok for me. AIUI `hyper-util` is where this section of code is going in `1.0.0`, so if I made this change there we'd be able to use it once that is released?

hottea773:
I've opened https://github.com/hyperium/hyper-util/pull/26 which AFAICT essentially re-targets this PR against the `v1` branch. If that is approved, I'll close this PR as I'm happy to wait for v1 to use this feature.

hottea773:
Since https://github.com/hyperium/hyper-util/pull/26 has been merged, so this will be available from v1 onwards, I'm closing this PR

nickjknight:
Since hyper V1 is taking quite a long time. Can we get this useful change applied to 0.14.x as well?

hottea773:
I'll re-open the PR as I'd be happy for this to be merged, but obviously depends on what Sean thinks.

hottea773:
@seanmonstar do you have any views on @nickjknight 's request?

hottea773:
@seanmonstar  do you have any views on @nickjknight 's request?

hottea773:
Hey, thanks. If it is coming next week, then I imagine that should be good for people's use cases. @nickjknight any opinions?

nickjknight:
Yeah should be fine if it's v1 is releasing so soon

PR #3091, opened 2022-12-21
style(client, server): Remove redundant http1, http2 feature gates

oliviacrain:
Remove feature gates on methods where the parent module is already gated on said Cargo feature. These gates seem to be leftovers from the split of connection builders into per-http-version modules.

Since the `client::conn::http1` and `server::conn::http1` modules are gated on the `http1` feature, we can remove `http1` gates within those modules.

Likewise for `client::conn::http2`, `server::conn::http2`, and the `http2` feature.


seanmonstar (approved):

seanmonstar:
Excellent, thank you!

PR #3092, opened 2022-12-21
Revamp `gen_header` using `cargo-expand`

bossmc:
As discussed in #3084 this revamps the header generation logic to use `cargo expand` to create a single input file for `cbindgen`.


capi/cbindgen.toml
```diff
@@ -5,11 +5,8 @@ header = """/*
  * Copyright 2021 Sean McArthur. MIT License.
  * Generated by gen_header.sh. Do not edit directly.
  */"""
-include_guard = "_HYPER_H"
```
seanmonstar, line 8: Should we not have an include guard?

bossmc: This switches to use `#pragma once` instead - there's various pros and cons between the two, but pragma once is well supported and eliminates a few gotchas from include guards (like two headers happening to chose the same include guard variable name).

seanmonstar: After doing some reading about the two, it seems like pragma is "better", but that include guards work on more (older) compilers, and some compilers optimize include guards into pragma once anyways.

So, I figured I'd defer to what curl does, and just follow their example. They use include guards.

bossmc: It's true that `#pragma once` is non-standard and slightly less supported.  You can't optimize include guards to pragmas since the `#define` is visible outside of the guard (whereas some compilers rewrite pragmas to include guards using arbitrary names for the variable that's used for the guard).

I'm not that fussed either way so I'm happy to change back if you prefer?

seanmonstar: Yea, let's keep it as an include guard for now.

bossmc: Done!

PR #3101, opened 2022-12-24
Remove version specific prefixes from builders

### LINKED ISSUE #3085: Remove version-specific prefixes from builder methods
With the split of the builders into per-version kinds (see #2842 and #2851), I feel there's a new question around whether we should remove the prefixes from the method names. They used to exist because the builders were combined over multiple HTTP versions, so the options were scoped to a version. But now:

```rust
mod http1 {
    impl Builder {
        pub fn http1_title_case_headers() {
            // ...
        }
    }
}
```

It seems all those can have the `http1_` prefix removed, and same for the `http2::Builder`s.

Noticed in https://github.com/hyperium/hyper-util/pull/11#discussion_r1035315089.

Michael-J-Ward:
Fixes #3085 

Should I also change thet non-public methods in `proto::h1::conn`?
  - `set_http1_header_read_timeout`
  - `set_h1_parser_config`

### DIFF
--- a/src/server/conn/http1.rs
+++ b/src/server/conn/http1.rs
@@ -224,7 +224,7 @@ impl Builder {
     /// detects an EOF in the middle of a request.
     ///
     /// Default is `false`.
-    pub fn http1_half_close(&mut self, val: bool) -> &mut Self {
+    pub fn half_close(&mut self, val: bool) -> &mut Self {
         self.h1_half_close = val;
         self
     }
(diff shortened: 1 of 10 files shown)


src/server/conn/http1.rs
```diff
@@ -261,7 +261,7 @@ impl Builder {
     /// Note that this setting does not affect HTTP/2.
```
oliviacrain, line 261: There are a few comments with this note- can we remove these as well, since they no longer apply?

Michael-J-Ward: Good catch, done.

seanmonstar (approved):
This looks super, thanks so much! I see there's a `fixup!` commit in the list, did you want to rebase and do the fix locally? The merge button only lets me either rebase merge without interactive mode, or squash it all into a single commit.

Michael-J-Ward:
done ✅

PR #3102, opened 2022-12-27
feat(server): backport the split server conn modules

### LINKED ISSUE #3079: Backport the split server conn modules
In 1.0, we've split hyper::server::conn::Connection into per-version types. To ease upgrading (see https://github.com/hyperium/hyper/issues/3052), we can backport the addition of the two modules, hyper::server::conn::{http1, http2}. With them in place, we could then add a deprecation to hyper::server::conn::Connection.

This is very similar in scope to #3053

oddgrd:
This PR backports the `server::conn::http1` and `server::conn::http2` modules to  `0.14.x`. I've drew a lot on the work already done in https://github.com/hyperium/hyper/pull/3063 (:pray:), but diverged a bit on the feature naming (no strong feelings on this though). I also implemented the `deprecated` feature flag as was suggested in that PR.

I have a couple of questions:
- I cut out the `common::time::Timer` (as is done in split client conn backports), is that OK or should that module also be backported?
- How many tests should be backported? I've started with only one for each protocol, but I'm happy to add more.

I've also only added one deprecation warning so far, but I'll look around and see if more make sense.

Closes #3079

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -773,10 +773,6 @@ where
     /// but it is not desired to actually shutdown the IO object. Instead you
     /// would take it back using `into_parts`.
     pub fn poll_without_shutdown(&mut self, cx: &mut task::Context<'_>) -> Poll<crate::Result<()>>
-    where
-        S: Unpin,
-        S::Future: Unpin,
-        B: Unpin,
     {
         loop {
             match *self.conn.as_mut().unwrap() {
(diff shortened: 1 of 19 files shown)


seanmonstar:
Since I just reviewed #3101, I'm realizing that probably should be part of the backport too, or else we'll need to have deprecated methods _in_ the backport 😅

oddgrd:
Indeed! I'll incorporate their changes tomorrow.

src/server/conn.rs
```diff
@@ -231,6 +236,12 @@ pub struct Parts<T, S> {
 impl Http {
     /// Creates a new instance of the HTTP protocol, ready to spawn a server or
     /// start accepting connections.
+    #[cfg_attr(
```
seanmonstar, line 239: I'd put these attributes on the `struct`s instead. I know it explodes some warnings in here, but you can add `#[cfg_attr(feature = "deprecated", allow(deprecated))]` on the complaining `impl Whatever` blocks. You can see what I did in #3156.

Oh, also, can we make the deprecations a separate PR?

oddgrd: Sure thing!

seanmonstar (approved):

PR #3125, opened 2023-01-23
feat: add downcast on `Sleep` trait

### LINKED ISSUE #3027: Timer::reset can't soundly have an optimized implementation
In 1.0.0-rc.1, the new `Timer` trait has a method to reset an existing `Sleep` to a new deadline. The default implementation simply creates a new `Sleep` and writes it over the old one, but presumably this method exists to it can be overridden with a better implementation when working with a timer like Tokio's that can have its deadline dynamically adjusted.

However, since the `Sleep` is provided as a `dyn Sleep` and `Sleep` has no downcasting APIs, there is no safe way to confirm that the `Sleep` instance is one created by this `Timer` implementation. That effectively means that the current default implementation is the only sound way of implementing the method.

dswij:
closes #3027 

This pr adds a downcast for the `Sleep` trait.

### DIFF
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -6,9 +6,9 @@
 //! to plug in other runtimes.
 
 pub mod bounds;
-pub mod timer;
+mod timer;
 
-pub use timer::*;
+pub use timer::{Sleep, Timer};
 
 /// An executor of futures.
 ///
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -6,12 +6,9 @@
 //! to plug in other runtimes.
 
 pub mod bounds;
+pub mod timer;
 
-use std::{
-    future::Future,
-    pin::Pin,
-    time::{Duration, Instant},
-};
+pub use timer::*;
 
 /// An executor of futures.
 ///
@@ -38,20 +35,3 @@ pub trait Executor<Fut> {
     /// Place the future into the executor to be run.
     fn execute(&self, fut: Fut);
 }
-
-/// A timer which provides timer-like functions.
-pub trait Timer {
-    /// Return a future that resolves in `duration` time.
-    fn sleep(&self, duration: Duration) -> Pin<Box<dyn Sleep>>;
-
-    /// Return a future that resolves at `deadline`.
-    fn sleep_until(&self, deadline: Instant) -> Pin<Box<dyn Sleep>>;
-
-    /// Reset a future to resolve at `new_deadline` instead.
-    fn reset(&self, sleep: &mut Pin<Box<dyn Sleep>>, new_deadline: Instant) {
-        *sleep = self.sleep_until(new_deadline);
-    }
-}
-
-/// A future returned by a `Timer`.
-pub trait Sleep: Send + Sync + Future<Output = ()> {}
--- a/src/rt/timer.rs
+++ b/src/rt/timer.rs
@@ -29,8 +29,8 @@
 //!     }
 //!
 //!     fn reset(&self, sleep: &mut Pin<Box<dyn Sleep>>, new_deadline: Instant) {
-//!         if sleep.downcast_ref::<TokioSleep>().is_some() {
-//!             *sleep = self.sleep_until(new_deadline);
+//!         if let Some(sleep) = sleep.as_mut().downcast_mut_pin::<TokioSleep>() {
+//!             sleep.reset(new_deadline.into())
 //!         }
 //!     }
 //! }
@@ -51,6 +51,12 @@
 //! }
 //!
 //! impl Sleep for TokioSleep {}
+//!
+//! impl TokioSleep {
+//!     pub fn reset(self: Pin<&mut Self>, deadline: Instant) {
+//!         self.project().inner.as_mut().reset(deadline.into());
+//!     }
+//! }
 //! ````
 
 use std::{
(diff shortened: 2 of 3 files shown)


dswij:
I'm not sure I understand #3027 completely, but I'm thinking of something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d2dd45c05a772dbcf7f033425ef4bf82).

Also, I noticed that there is no test for this. Am I missing something, or is this something we can add on?

sfackler:
I think it'd be a bit cleaner to implement downcasting methods directly on Sleep without requiring implementors to deal with it:

```rust
mod private {
    pub struct PrivacyToken;
}

pub trait Sleep: Send + Sync + Future<Output = ()> {
    // PrivacyToken can't be named outside of this crate, so it prevents anyone from overriding this default
    // implementation in another crate. That allows us to trust it to be correct in the downcast methods below.
    #[doc(hidden)]
    fn __private_api_type_id(&self, _: private::PrivacyToken) -> TypeId {
        TypeId::of::<Self>()
    }
}

impl dyn Sleep {
    /// Returns `true` if the sleep's type is `T`.
    pub fn is<T>(&self) -> bool
    where
        T: Sleep,
    {
        self.__private_api_type_id(private::PrivacyToken) == TypeId::of::<T>()
    }

    /// Attempts to downcast the sleep to the type `T` if it has that type.
    pub fn downcast_ref<T>(&self) -> Option<&T>
    where
        T: Sleep,
    {
        if self.is::<T>() {
            unsafe { Some(&*(self as *const dyn Sleep as *const T)) }
        } else {
            None
        }
    }

    /// Attempts to downcast the sleep to the type `T` if it has that type.
    pub fn downcast_arc<T>(self: Arc<Self>) -> Result<Arc<T>, Arc<Self>>
    where
        T: Sleep,
    {
        if self.is::<T>() {
            unsafe { Ok(Arc::from_raw(Arc::into_raw(self).cast::<T>())) }
        } else {
            Err(self)
        }
    }
}
```

dswij:
> I think it'd be a bit cleaner to implement downcasting methods directly on Sleep without requiring implementors to deal with it

That's a good point. We can probably make `Sleep` a derivable sealed trait instead.

Though, `TypeId::of::<T>` requires `T: 'static`, so we might want a workaround similar to [this](https://developers.facebook.com/blog/post/2021/07/20/rust-nibbles-gazebo-any-lifetime/), if we wish to go that direction.

sfackler:
Are there any relevant non-'static Sleep implementations?

dswij:
> Are there any relevant non-'static Sleep implementations?

I don't think so, but we shouldn't be too concerned about it now anyways imo. I have also added some docs to it

src/rt/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Runtime components
+//!
+//! By default, hyper includes the [tokio](https://tokio.rs) runtime.
+//!
+//! If the `runtime` feature is disabled, the types in this module can be used
+//! to plug in other runtimes.
+
+pub mod timer;
+
+pub use timer::*;
```
seanmonstar, line 10: I'd suggest this be an explicit list, just to prevent any accidents now or in the future.

src/rt/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Runtime components
+//!
+//! By default, hyper includes the [tokio](https://tokio.rs) runtime.
+//!
+//! If the `runtime` feature is disabled, the types in this module can be used
+//! to plug in other runtimes.
+
+pub mod timer;
```
seanmonstar, line 8: Perhaps let's start with the module being private, we can always make it public later if it'd be helpful.

seanmonstar (commented):
I think this looks goot to me, @sfackler anything else you'd need?

src/rt/timer.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Reset a future to resolve at `new_deadline` instead.
+    fn reset(&self, sleep: &mut Pin<Box<dyn Sleep>>, new_deadline: Instant) {
+        *sleep = self.sleep_until(new_deadline);
+    }
+}
+
+/// A future returned by a `Timer`.
+pub trait Sleep: Send + Sync + Future<Output = ()> {
+    #[doc(hidden)]
+    /// This method is private and should not be implemented by downstream crate
```
sfackler, line 80: *can* not be implemented, not *should* not be implemented.

src/rt/timer.rs
```diff
@@ -0,0 +1,128 @@
...
+//!     fn sleep_until(&self, deadline: Instant) -> Pin<Box<dyn Sleep>> {
+//!         Box::pin(TokioSleep {
+//!             inner: tokio::time::sleep_until(deadline.into()),
+//!         })
+//!     }
+//!
+//!     fn reset(&self, sleep: &mut Pin<Box<dyn Sleep>>, new_deadline: Instant) {
+//!         if sleep.downcast_ref::<TokioSleep>().is_some() {
+//!             *sleep = self.sleep_until(new_deadline);
+//!         }
```
sfackler, line 34: This doesn't make much sense to me. The whole point of downcasting is to not have to create a new timer.

src/rt/timer.rs
```diff
@@ -0,0 +1,128 @@
...
+    where
+        T: Sleep + 'static,
+    {
+        if self.is::<T>() {
+            unsafe { Some(&mut *(self as *mut dyn Sleep as *mut T)) }
+        } else {
+            None
+        }
+    }
+}
```
sfackler, line 123: I don't think these two downcast methods are going to be too useful, since `reset` is working with a `&mut Pin<Box<dyn Sleep>>`. I think we'd probably need a `pub fn downcast_pin_mut<T>(self: Pin<&mut Self>) -> Option<Pin<&mut T>>` probably?

sfackler:
I think it'd be good to update the test `TokioTimer` to downcast in its reset implementation to test that all of this works in practice: https://github.com/hyperium/hyper/blob/master/benches/support/tokiort.rs#L29-L44

seanmonstar:
Hey @dswij, just wanted to check if you think you'll have time to update this. No worries if you can't, just checking to reduce duplicate work. :)

dswij:
> Hey @dswij, just wanted to check if you think you'll have time to update this. No worries if you can't, just checking to reduce duplicate work. :)

Sorry for the delay, I was on a break a while ago and I haven't caught up to my backlog yet.

I'll try to take a look at this PR on the weekend :slightly_smiling_face:

src/rt/timer.rs
dswij: [8d0e134](https://github.com/hyperium/hyper/pull/3125/commits/8d0e13410e911aaf6d62b178e1fb444003186be5)

src/rt/timer.rs
dswij: Think I understood the issue better now. I updated the example code with an implementation of `tokio::time::Sleep::reset`

src/rt/timer.rs
dswij: [424e396](https://github.com/hyperium/hyper/pull/3125/commits/424e396f427d7ef0bb407a81bc7ed576252cfd96)

src/rt/mod.rs
dswij: [2e9fdb6](https://github.com/hyperium/hyper/pull/3125/commits/2e9fdb69975c15bb8ab70e0a7e70cd275b78155f)

dswij:
@seanmonstar sorry to nag, but is there anything I can do to move this forward?

seanmonstar (approved):

PR #3127, opened 2023-01-24
Make private traits in public api, sealed and public

### LINKED ISSUE #2051: Don't use private traits in public APIs
For example, hyper master currently has this:

```rust
impl<I, B, S, E> Future for Connection<I, S, E>
where
    S: HttpService<Body, ResBody=B>,
    I: AsyncRead + AsyncWrite + Unpin + 'static,
    B: Payload + 'static,
    B::Data: Unpin,
    E: H2Exec<S::Future, B>,
```

`HttpService` and `H2Exec` are both private traits, though. This makes it a huge pain for code working generically with connections to express the proper bounds in their code. You basically have to try to reverse engineer the impl bounds by bouncing off the compiler :(

### LINKED ISSUE #3097: Nameable Future-Proof hyper Executors
## Goal

Make the `Executor` bounds on client and server types publicly nameable, while still preserving the ability to change internal spawning needs.

## Background

hyper's client and server types require a `hyper::rt::Executor` to spawn some related background tasks.

There's two conflicting requirements for the executor:

- The internal types that are spawned must be provided to the `Executor` (via `Executor<InternalFuture>`), so that the `Send`iness can be propagated. This allows users to have `!Send` futures and bodies as long as they provide an `Executor` that can spawn them.
- The internal types expose some of the private architecture, so being able to name them makes it impossible to refactor in the future. Being able to _exhaustively_ name them would prevent us from spawning _new_ background tasks as features are developed.

So far, we have solved this by using privately-public traits, which are blanketly implemented for all generic executors. This allows us to change the types internally, and as long as a user provides a type that is `impl<F: Future> Executor<F> for MyExec`, it just works.

But it has some annoying problems.

- It's not possible to build functions or types on top, since in another crate, you can't write `E: Executor<CantNameMe>`, nor `E: hyper::unnameable::ConnStreamExec` (see #2051; https://github.com/hyperium/hyper-util/pull/11#discussion_r1044221053).
- The bounds are confusing in the documentation (for [example](https://hachyderm.io/@algo_luca/109520185141600746)).

## Proposal

Define **public** trait aliases that are sealed, but can be named externally, and have documentation showing how to provide a `Executor` to match the bounds, and how to express the bounds in your own API.

For example: define `hyper::rt::bounds::ServerConnHttp2Exec`, which is a public trait.

This allows us to continue to not expose the actual internal types, and also to extend the trait to be represent spawning other tasks if need be.

programatik29:
Resolves #2051, resolves #3097 

- Moved, sealed, renamed `ConnStreamExec` to `Http2ConnExec` and made it public.
- Changed `pub(super)` restriction of `HttpService` to `pub`.

### DIFF
--- a/src/common/exec.rs
+++ b/src/common/exec.rs
@@ -3,16 +3,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 
-#[cfg(feature = "server")]
-use crate::body::Body;
-#[cfg(all(feature = "http2", feature = "server"))]
-use crate::proto::h2::server::H2Stream;
 use crate::rt::Executor;
-
-#[cfg(feature = "server")]
-pub trait ConnStreamExec<F, B: Body>: Clone {
-    fn execute_h2stream(&mut self, fut: H2Stream<F, B>);
-}
 
 pub(crate) type BoxSendFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
 
@@ -46,31 +37,6 @@ impl Exec {
 impl fmt::Debug for Exec {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Exec").finish()
-    }
-}
-
-#[cfg(feature = "server")]
-impl<F, B> ConnStreamExec<F, B> for Exec
-where
-    H2Stream<F, B>: Future<Output = ()> + Send + 'static,
-    B: Body,
-{
-    fn execute_h2stream(&mut self, fut: H2Stream<F, B>) {
-        self.execute(fut)
-    }
-}
-
-// ==== impl Executor =====
-
-#[cfg(feature = "server")]
-impl<E, F, B> ConnStreamExec<F, B> for E
-where
-    E: Executor<H2Stream<F, B>> + Clone,
-    H2Stream<F, B>: Future<Output = ()>,
-    B: Body,
-{
-    fn execute_h2stream(&mut self, fut: H2Stream<F, B>) {
-        self.execute(fut)
     }
 }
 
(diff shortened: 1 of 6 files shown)


seanmonstar:
Thanks so much! I do think this is the direction that solves all the problems listed in #3097.

What do others think? @sfackler, @LucioFranco

programatik29:
I think documentation can be better. Where do you think it should be @seanmonstar ?

Some ideas:

 - `serve_connection` method.
 - `hyper::server::conn::http2` module.
 - `Executor` trait.
 - `hyper::rt` module.
 - `hyper::rt::bounds` module.
 - `Http2ConnExec` trait.

programatik29:
In my opinion, there should be a documentation of how should an executor be in `Executor` trait and link to that trait in related places. Also explain in `Http2ConnExec` docs that it can be used with any executor implemented like it is shown in `Executor` trait docs.

seanmonstar:
Good point, more docs would help people. Here's some thoughts of what to put where (anyone is free to suggest differently!):

- `http2::Builder::new` - Can explain briefly what the argument is, linking to the `Http2ConnExec` trait (or `bounds` module, perhaps).
- `serve_connection` - Briefly explain, like above.
- `Http2ConnExec` - Explain that this trait is for bounds, and that any desire to _implement_ the trait should be done by implementing an `Executor`, linking to it.
- `bounds` - Explain that the module contains traits that ease setting bounds, but that traits within that module likely can be implemented automatically by implementing a different trait. I can't say `bounds` will _only_ ever be for `Executor`, so that's what I'm slightly generic here. But there can be a subheading for the `bounds` docs about `Executor`.

seanmonstar (approved):

PR #3135, opened 2023-01-30
feat(client): remove `common::Exec::Default`

### LINKED ISSUE #3128: `http2::handshake()` is useless in 1.0 given no default executor
**Version**

hyper 1.0.0-rc2

**Description**

`hyper::client::conn::http2::handshake(io).await` causes an instant panic. Internally it performs `Builder::new().handshake(io).await` which uses the `Exec::Default` executor. There is no longer a default `tokio` runtime implied since the removal of the `runtime` feature. 

https://github.com/hyperium/hyper/blob/master/src/common/exec.rs#L36-L38

There is a TODO in this file to refactor executors. But I couldn't find a relevant issue, so I wanted to open one to ensure this doesn't get included in the official release

dswij:
Closes #3128 

This PR removes `Exec::Default` so that we avoid panicking by asking users to provide Executor right away. `Exec::Default` serves no purpose after #2975.

Also refactors `hyper::client::conn::Http2::Builder::new` as part of the removal.

### DIFF
--- a/benches/end_to_end.rs
+++ b/benches/end_to_end.rs
@@ -316,11 +316,10 @@ impl Opts {
         let mut client = rt.block_on(async {
             if self.http2 {
                 let io = tokio::net::TcpStream::connect(&addr).await.unwrap();
-                let (tx, conn) = hyper::client::conn::http2::Builder::new()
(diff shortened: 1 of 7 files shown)


seanmonstar (approved):

seanmonstar:
Excellent work, thank you so much! 🙏

PR #3137, opened 2023-02-01
fix(server): prevent sending 100-continue if user drops request body

seanmonstar:


src/proto/h1/conn.rs
```diff
@@ -732,6 +732,12 @@ where
 
     /// If the read side can be cheaply drained, do so. Otherwise, close.
     pub(super) fn poll_drain_or_close_read(&mut self, cx: &mut task::Context<'_>) {
+        if let Reading::Continue(ref decoder) = self.state.reading {
+            // skip sending the 100-continue
+            // just move forward to a read, in case a tiny body was included
+            self.state.reading = Reading::Body(decoder.clone());
```
sfackler, line 738: Will the Reading::Body state properly handle the case where no body is sent and instead headers for the next request are the next thing written?

seanmonstar: If the request included a content-length or transfer-encoding header, the next indicated bytes *must* be the body. If it didn't, then no body can be included.

sfackler: Oh right, duh

sfackler (approved):

Issue #2605, opened 2021-07-26
provide ability to request a fresh connection


rcoh:
**Is your feature request related to a problem? Please describe.**
Certain retry behaviors specify a fresh connection (and sometimes a full DNS re-lookup) to direct traffic away from what might be a problematic host. I need the ability to somehow mark a request to Hyper as _not_ reusing a connection that may be in the pool

**Describe the solution you'd like**
I can imagine a few ways the API might be structured, but in general I'd like some way as marking a request as "needs fresh connection", potentially giving a reference to the old connection / response which was problematic.

**Describe alternatives you've considered**
I could rebuild the entire client & pool when a request fails in a way that necessitates a re-lookup. However, it seems like that may cause unnecessary churn on the client side.

**Additional context**
https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-design-patterns.html#optimizing-performance-timeouts-retries

jdisanti:
I like the idea of having `hyper::client::connect::Fresh`. Usage could look something like:

```rust
use hyper::client::connect::Fresh;

let req = Request::builder()
    .method("GET")
    .uri("https://example.com")
    .extension(Fresh)
    .body(())?;
let resp = client.request(req).await?;
```

This would work for our use-case where we need a new connection on retry since we can just add the `Fresh` extension as part of our Tower retry policy.

I'm not really sure how this would carry over to hyper 1.0 yet though. From reading the code, it looks like the connection pool was moved to `hyper-util`, but I can't find any client code that actually references it.

tjandra-amzn:
This feature would be really helpful to my team too - one thing we've seen is that we can have stale connections that just don't work anymore or the host they're connected to on the backend had some problem. It'd be nice if the ```Fresh``` extension didn't just bypass the lookup in the conn-pool but was a way to force the expiration and replacement of the existing pool in the connection.

jdisanti:
@tjandra-amzn - That seems like a good idea to me.

I talked to @rcoh about this, and he had an interesting idea about having the `PoolClient` place an extension in responses that identifies the `Key` in the connection pool used to produce that response. Then a method could be added to client to evict the connections for a given response.

Thinking about that more, in the case where DNS starts responding with a new IP, I wonder if it would be better to evict only the connection that failed rather than all the connections with the same key, as there could be some newer connections in the pool that are still good.

[...]

PR #3144, opened 2023-02-14
feat(client): add capture_connection

rcoh:
Add `capture_connection` functionality. This allows callers to retrieve the `Connected` struct of the connection that was used internally by Hyper. This is in service of https://github.com/hyperium/hyper/issues/2605.

This pairs with a coming PR that adds a `poison` API to `Connected` — the pair of these APIs makes it possible to avoid reusing connections that the client deems to be bad.

Example usage:
```rust
use hyper::client::connect::{capture_connection, CaptureConnection};
let mut request = http::Request::builder()
   .uri("http://foo.com")
   .body(())
   .unwrap();

let captured_connection = capture_connection(&mut request);

let client = hyper::Client::new();
// success or failure, it will still work fine.
let _ = client.request(request).await;

let connection_info = capture_connection.connection_metadata();
```

Although this uses `http::Extensions` under the hood, the API exposed explicitly hides that detail.

### DIFF
--- a/src/client/connect/mod.rs
+++ b/src/client/connect/mod.rs
@@ -82,9 +82,11 @@
 use std::fmt;
 use std::fmt::{Debug, Formatter};
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::ops::Deref;
 use std::sync::Arc;
 
 use ::http::Extensions;
+use tokio::sync::watch;
 
 cfg_feature! {
     #![feature = "tcp"]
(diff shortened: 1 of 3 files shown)


src/client/connect/mod.rs
```diff
@@ -115,6 +118,114 @@ pub struct Connected {
...
+/// tokio::task::spawn(async move {
+///     let connection_info = captured.wait_for_connection_metadata().await;
+///     println!("we are connected! {:?}", connection_info.as_ref());
+/// });
+///
+/// let client = hyper::Client::new();
+/// client.request(request).await.expect("request failed");
+/// # }
+/// ```
+pub fn capture_connection<B>(request: &mut crate::http::Request<B>) -> CaptureConnection {
```
LucioFranco, line 179: Extension trait?

rcoh: discussed offline, will defer for now but could add later

src/client/connect/mod.rs
```diff
@@ -115,6 +118,114 @@ pub struct Connected {
     pub(super) extra: Option<Extra>,
 }
 
+/// [`CaptureConnection`] allows callers to capture [`Connected`] information
```
LucioFranco, line 121: Does this need to live in this file or can we move it out?

rcoh: can definitely be moved—any preference for final location?

seanmonstar (approved):
Looks real clear to me, love it!

PR #3145, opened 2023-02-15
feat(client): add poison to Connected

rcoh:
Add `poison` method to `Connected`. This allows callers to mark a connection as poisoned which prevents the pool from reusing it on subsequent requests. `is_open` will consider poisoning prior to returning a connection to the pool.


seanmonstar (approved):
LGTM!

PR #3152, opened 2023-02-28
feat(CI): check external types

### LINKED ISSUE #2925: Set up CI to audit for public dependency types

palango:
This adds CI job that runs `cargo-check-external-types` against a whitelist of exported types.

Whitelist in https://github.com/hyperium/hyper/issues/2925#issuecomment-1362473727 didn't change.

For now the nightly for this needs to be pinned to `nightly-2022-11-16`, because `cargo-check-external-types` doesn't work with the latest nightly json doc format.

Resolves #2925
Superseeds #2957

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,6 +24,7 @@ jobs:
       - ffi
       - ffi-header
       - doc
+      - check-external-types
     steps:
       - run: exit 0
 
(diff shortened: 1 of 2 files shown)


palango:
I'm unsure why it cannot find the nightly version. I created an upstream issue: https://github.com/dtolnay/rust-toolchain/issues/76

.github/workflows/external-types.toml
```diff
@@ -0,0 +1,17 @@
...
+    "http::request::Request",
+    "http::response::Response",
+    "http::status::StatusCode",
+    "http::uri::Uri",
+    "http::version::Version",
+    "http_body::Body",
+    "http_body::frame::Frame",
+    "http_body::size_hint::SizeHint",
+    "tokio::io::async_read::AsyncRead",
+    "tokio::io::async_write::AsyncWrite"
```
seanmonstar, line 16: We've since moved away from publicly depending on Tokio, so these two could be removed.

seanmonstar (commented):
Thanks for putting this together! Sorry for the delay, I think we're ready to start having CI check this on master.

.github/workflows/CI.yml
```diff
@@ -210,3 +210,26 @@ jobs:
 
       - name: cargo doc
         run: cargo rustdoc --features full,ffi -- --cfg docsrs --cfg hyper_unstable_ffi -D broken-intra-doc-links
+
+  check-external-types:
```
seanmonstar, line 214: Oh I forgot, to make this required, you'd need to add it to the list under `ci-pass.needs`.

.github/workflows/external-types.toml
palango: Removed.

.github/workflows/CI.yml
palango: Good point, added it there.

seanmonstar (approved):

PR #3153, opened 2023-02-28
Add clippy to CI

### LINKED ISSUE #2977: Add limited clippy to CI
Add a new job to [CI](https://github.com/hyperium/hyper/blob/master/.github/workflows/CI.yml), `clippy`, but that is _extremely_ limited in what lints it checks for. Clippy's defaults are usually too annoying, but there a few specific things that I would like to catch in an automated fashion:

- [`self_named_module_files`](https://rust-lang.github.io/rust-clippy/stable/index.html#self_named_module_files)

Add the job as a `needs` dependency of the `ci-pass` top-level job.

palango:
Adds limited clippy run to CI.

There were two clippy warnings which seemed useful, they're in the second commit. If you don't like them I'm happy to drop the commit.

Resolves #2977
Superseeds https://github.com/hyperium/hyper/pull/2979

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -219,7 +219,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thank you! I'm still surprised this doesn't catch that there is an `ext.rs` and `ext/` folder. When I noticed those two existed, that's why I filed the original issue. Am I misunderstanding what that lint catches?

palango:
Seems like there are some bugs in the lint: https://github.com/rust-lang/rust-clippy/issues/8094

PR #3155, opened 2023-03-02
feat(client): add 1.0 compatible client conn API

seanmonstar:
This patch backports client/conn/http1 and http2 modules from 1.0 to ease transition. It allows code still using 0.14.x to switch to the per-version Connection types available in 1.0.

Updating (part of) #3063

PR #3160, opened 2023-03-06
feat(client): add is_ready to 1.0 compatible client conn API

howardjohn:
This address https://github.com/hyperium/hyper/pull/3155/files#r1124682210, including the is_ready() function that is part of 1.0. As noted in the comment, when_ready is removed from 1.0, so the commented code is also cleaned up.

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -90,7 +90,7 @@ impl<B> SendRequest<B> {
         futures_util::future::poll_fn(|cx| self.poll_ready(cx)).await
     }
 
-    pub(super) fn is_ready(&self) -> bool {
+    pub fn is_ready(&self) -> bool {
         self.dispatch.is_ready()
     }
 
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -90,20 +90,9 @@ impl<B> SendRequest<B> {
         futures_util::future::poll_fn(|cx| self.poll_ready(cx)).await
     }
 
-    /*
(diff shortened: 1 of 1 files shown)


src/client/conn/http2.rs
```diff
@@ -90,20 +90,9 @@ impl<B> SendRequest<B> {
...
-    pub(super) async fn when_ready(self) -> crate::Result<Self> {
-        let mut me = Some(self);
-        future::poll_fn(move |cx| {
-            ready!(me.as_mut().unwrap().poll_ready(cx))?;
-            Poll::Ready(Ok(me.take().unwrap()))
-        })
-        .await
-    }
-
     pub(super) fn is_ready(&self) -> bool {
```
seanmonstar, line 93: A difference is that in 1.0, these are fully public. Well, this one, and the next one.

howardjohn: I think I actually intended to make this `pub` and this was an oversight. Is this (and closed) intentionally not `pub` here but `pub` in 1.0? I can change is_closed or close this PR accordingly

seanmonstar: It just wasn't made public because there hasn't been a need. In that sense, I tend to wait till someone has a request, so as not to expose the API prematurely. 🤷

howardjohn: Fair enough! Our use case is to use this as part of a connection pool; we are working to move to 1.0 to do this but were hoping to implement pooling first while we work on the rest of the migration; exposing this would enable that.

howardjohn:
We moved to 1.0 so this is not needed for us. Happy to re-open if others are interested though

PR #3161, opened 2023-03-07
feat(server): deprecate server conn structs

oddgrd:
This is based on https://github.com/hyperium/hyper/pull/3102 which needs to be merged before this.

I had some difficulty with deprecations inside the `pin-project-lite` macros, it doesn't seem to be possible to set `#[cfg_attr(feature = "deprecated", allow(deprecated))]`on the fields of a struct inside `pin_project!`. That `server::conn::Connection` is inside `pin_project!` makes it difficult to deprecate as well. :thinking: I'll need to come back to this tomorrow. For example, in `hyper/server/server.rs`:

```rust
pin_project! {
    /// A listening HTTP server that accepts connections in both HTTP1 and HTTP2 by default.
    ///
    /// `Server` is a `Future` mapping a bound listener with a set of service
    /// handlers. It is built using the [`Builder`](Builder), and the future
    /// completes when the server has been shutdown. It should be run by an
    /// `Executor`.
    pub struct Server<I, S, E = Exec> {
        #[pin]
        incoming: I,
        make_service: S,
        #[cfg_attr(feature = "deprecated", allow(deprecated))] // <- This doesn't work.
        protocol: Http_<E>,
    }
}
```

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -98,6 +98,12 @@ pub use super::tcp::{AddrIncoming, AddrStream};
 #[derive(Clone, Debug)]
 #[cfg(any(feature = "http1", feature = "http2"))]
 #[cfg_attr(docsrs, doc(cfg(any(feature = "http1", feature = "http2"))))]
+#[cfg_attr(
+    feature = "deprecated",
+    deprecated(
+        note = "This struct will be replaced with `server::conn::http1::Builder` and `server::conn::http2::Builder` in 1.0, enable the \"backports\" feature to use them now."
+    )
+)]
 pub struct Http<E = Exec> {
     pub(crate) exec: E,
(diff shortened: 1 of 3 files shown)


seanmonstar:
What if you just add the `allow` on `struct Server`, instead? Seems fine, since it's only allowing when our own feature is enabled, normal CI will catch any other deprecated stuff we're using from libstd.

oddgrd:
CI passes but it's still showing warnings with the deprecated feature enabled and the allow on the struct itself, unfortunately.

seanmonstar:
That seems weird... (Also, does that mean we need an additional CI job to notice that?)

seanmonstar:
Ok, added in #3162.

seanmonstar (approved):

PR #3163, opened 2023-03-09
feat(client): make `SendRequest::send_request` return a `Send` future

davidpdrsn:
`SendRequest::send_request` currently returns an `impl Future` that isn't `Send`. I was thinking it might nice if it did return a `Send` future. That way you can `spawn` it, or `Box::pin` it.

Changing the type to `impl Future + Send` requires `B: Send` but that probably shouldn't always be required. By making a named future type you can get around that. I considered putting the `B` in `SendRequestFuture` so we could use that in the future in case we needed to store the request. Let me know if you think that would be a good idea.

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -132,23 +132,10 @@ where
     pub fn send_request(
         &mut self,
         req: Request<B>,
-    ) -> impl Future<Output = crate::Result<Response<IncomingBody>>> {
-        let sent = self.dispatch.send(req);
-
-        async move {
-            match sent {
-                Ok(rx) => match rx.await {
-                    Ok(Ok(resp)) => Ok(resp),
-                    Ok(Err(err)) => Err(err),
-                    // this is definite bug if it happens, but it shouldn't happen!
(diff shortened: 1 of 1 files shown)


src/client/conn/http2.rs
```diff
@@ -132,23 +132,10 @@ where
     pub fn send_request(
         &mut self,
         req: Request<B>,
-    ) -> impl Future<Output = crate::Result<Response<IncomingBody>>> {
-        let sent = self.dispatch.send(req);
```
seanmonstar, line 136: What this was just:

```rust
let sent = self.dispatch.send(req).map_err(|_| ());
```

Would that be enough so that it doesn't need a named type?

davidpdrsn:
Turns out we don't actually need this as `impl Future` as "leaks" the `Send`. TIL! So adding `B: Send` to the consuming code makes the future send:

```rust
fn foo<B: Body + Send>(s: SendRequest<B>, req: Request<B>) {
    fn assert_send<T: Send>(_: T) {}
    assert_send(s.send_request(req));
}
```

PR #3171, opened 2023-03-12
Get the MSRV for the build job from the package metadata

nickelc:
That leaves only two places where the MSRV needs to be updated.


seanmonstar (approved):
Makes sense, nice fix, thanks!

Issue #3017, opened 2022-10-20
Make client able to use non-Send executor


seanmonstar:
This likely needs to be done before 1.0, because I suspect it involves adding a generic to a few types. But it's likely too late to fit into RC1.

The client code currently spawns a couple tasks for a connection, but does so by requiring the `Executor<Box<dyn Future + Send>>`, so we can spawn multiple. We can't make the internal type public, but we need to propagate the `Send` bounds of the IO and body types. So, we should change it to have bounds similar to the `server::conn::http2`, essentially an extension trait that means `E: Executor<InternalFutureType<IO, B>>`.

A way we can tell this all works is that we should be able to augment the single-threaded example file with a client as well.

Other relevant issues: #2341

Ruben2424:
I will try to implement this.

PR #3184, opened 2023-03-24
Make client able to use non-Send executor

Ruben2424:
Hi,
I tried to solve #3017. But I got stuck.

I copied and modified existing examples to get one which fails to compile because of the missing send bounds. 

The issue suggests to create a extension trait `Executor<InternalFutureType<IO, B>>`.
But what exactly is the `InternalFutureType<IO, B>`? I had the idea of a `enum InternalFutureType<IO, B>` with variants for every `Future` that gets spawned and implementing `Future` for it. But most the futures that get spawned are Anonymous or have Anonymous Parts. 
https://github.com/hyperium/hyper/blob/47f614fd99bf9e937efc874880e3276caf4c34b1/src/proto/h2/client.rs#L158
https://github.com/hyperium/hyper/blob/47f614fd99bf9e937efc874880e3276caf4c34b1/src/proto/h2/client.rs#L259
https://github.com/hyperium/hyper/blob/47f614fd99bf9e937efc874880e3276caf4c34b1/src/proto/h2/client.rs#L316

Maybe the solution is obvious and I just don't see it? Or am I on the wrong path? 
Any hint would be helpful.


seanmonstar:
So, doing this is a bit unfun. Those anonymous futures would need to become structs, with manual `impl Future` for them. They would need to be pseudo-public. The similar one to see is `proto::h2::server::H2Stream`.

_Then_, we can add in to `rt::bounds` a trait that is implemented for `T: Executor<ConnTask>, T: Executor<Pipe>, T: SendWhen`.

Ruben2424:
I changed the anonymous futures to structs and introduced a trait. And it seams to work mostly. Some cleanup is left.
@seanmonstar is it ok to for `Body` to require `'static + Unpin` ?

Cargo.toml
```diff
@@ -33,6 +33,7 @@ h2 = { version = "0.3.9", optional = true }
 itoa = "1"
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project-lite = "0.2.4"
+pin-project = "1.0.12"
```
seanmonstar, line 36: Is it possible to get away with just using `pin-project-lite`? `pin-project` takes a bit longer to compile...

Cargo.toml
```diff
@@ -112,6 +113,11 @@ name = "client"
 path = "examples/client.rs"
 required-features = ["full"]
 
+[[example]]
+name = "client_single_thread"
```
seanmonstar, line 117: Another option would be to make this part of the existing `single_threaded.rs` example, where it just both starts a server, and also makes use a of a client. That would mean less code that is copied and needs to be updated later on.

src/client/conn/http2.rs
```diff
@@ -35,20 +38,25 @@ impl<B> Clone for SendRequest<B> {
...
 
 /// A builder to configure an HTTP connection.
 ///
 /// After setting options, the builder is used to create a handshake future.
 #[derive(Clone, Debug)]
-pub struct Builder {
-    pub(super) exec: Exec,
+pub struct Builder<Ex>
+where
+    Ex: Executor<BoxSendFuture> + Send + Sync + 'static + Clone,
```
seanmonstar, line 57: This bounds probably don't need to be here on the struct.

src/client/conn/http2.rs
```diff
@@ -60,13 +68,15 @@ pub struct Builder {
...
-) -> crate::Result<(SendRequest<B>, Connection<T, B>)>
+) -> crate::Result<(SendRequest<B>, Connection<T, B, E>)>
 where
     E: Executor<BoxSendFuture> + Send + Sync + 'static,
     T: AsyncRead + AsyncWrite + Unpin + Send + 'static,
     B: Body + 'static,
     B::Data: Send,
     B::Error: Into<Box<dyn StdError + Send + Sync>>,
+    E: ExecutorClient<B, T> + Unpin + Clone,
+    <B as http_body::Body>::Error: std::error::Error + Send + Sync + 'static,
```
seanmonstar, line 79: The line just above, `B::Error: Into<..>` should be enough. Is it not?

src/client/conn/http2.rs
```diff
@@ -240,22 +257,22 @@ where
 
 // ===== impl Builder
 
-impl Builder {
+impl<Ex> Builder<Ex>
+where
+    Ex: Executor<BoxSendFuture> + Send + Sync + 'static + Clone,
```
seanmonstar, line 262: This bounds likely prevents using a local executor. Could probably remove it.

src/client/conn/http2.rs
```diff
@@ -395,12 +406,14 @@ impl Builder {
     pub fn handshake<T, B>(
         &self,
         io: T,
-    ) -> impl Future<Output = crate::Result<(SendRequest<B>, Connection<T, B>)>>
+    ) -> impl Future<Output = crate::Result<(SendRequest<B>, Connection<T, B, Ex>)>> + '_
```
seanmonstar, line 409: The `+ '_` may mess with the lifetime of the returned future...

src/client/conn/http2.rs
Ruben2424: It only requires Clone now

src/client/conn/http2.rs
Ruben2424: I removed it

src/client/conn/http2.rs
Ruben2424: Is it better now?

src/client/conn/http2.rs
Ruben2424: I removed it

Cargo.toml
Ruben2424: I just used `pin-project` for easier development and wanted to switch to `pin-project-lite` once finished.

Cargo.toml
Ruben2424: Done. Do you have an idea why `io::stdout().write_all(&chunk).await?;` didn't work in the combined example? It just wrote nothing to console.

Cargo.toml
Ruben2424: Done

Ruben2424:
I think this is ready for a second review now. 

I am not 100% sure if I really understand everything I did, so please review carefully.

I got some questions:
1. `http1` requires still `Send` for the `io` types. Is this something that needs to be addressed? And if so, in this PR?  
2. One test didn't compile because the new bounds don't allow the error type `Box<dyn std::error::Error + Send + Sync>` is this something that needs to be fixed? And if how can I fix this?
3. I just tested the example and started the integration test. Is there something that needs to be tested manually? Or something that I need to look at when testing? Or is a testcase missing?

I hope I haven't forgotten something important here.

src/client/conn/http2.rs
seanmonstar: Is it required to add something here at all? Because currently, this makes it stricter in that not only is it `Into`, but it must also _be_ `Error`. That means `Box<dyn Error>` no longer qualifies (since `Box<dyn Error>: Error` is not true for confusing reasons).

Cargo.toml
Ruben2424: fixed that in [improve example](https://github.com/hyperium/hyper/pull/3184/commits/3a78a57836417f19ca5b696ac7f3e42a2518a24d). forgot to mention it here.

src/client/conn/http2.rs
```diff
@@ -35,20 +35,24 @@ impl<B> Clone for SendRequest<B> {
...
 }
 
 /// A builder to configure an HTTP connection.
 ///
 /// After setting options, the builder is used to create a handshake future.
 #[derive(Clone, Debug)]
-pub struct Builder {
-    pub(super) exec: Exec,
+pub struct Builder<Ex>
+
```
seanmonstar, line 53: This opening brace should go back on same line. (Surprised rustfmt didn't complain.)

src/client/conn/http2.rs
```diff
@@ -60,13 +64,13 @@ pub struct Builder {
...
-) -> crate::Result<(SendRequest<B>, Connection<T, B>)>
+) -> crate::Result<(SendRequest<B>, Connection<T, B, E>)>
 where
-    E: Executor<BoxSendFuture> + Send + Sync + 'static,
-    T: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    T: AsyncRead + AsyncWrite + Unpin + 'static,
     B: Body + 'static,
     B::Data: Send,
-    B::Error: Into<Box<dyn StdError + Send + Sync>>,
+    B::Error: std::error::Error + Send + Sync + 'static,
```
seanmonstar, line 72: Can this change be undone?

src/client/conn/http2.rs
```diff
@@ -210,22 +215,26 @@ where
...
 }
 
-impl<T, B> fmt::Debug for Connection<T, B>
+impl<T, B, E> fmt::Debug for Connection<T, B, E>
 where
-    T: AsyncRead + AsyncWrite + fmt::Debug + Send + 'static,
+    T: AsyncRead + AsyncWrite + fmt::Debug + 'static + Unpin,
     B: Body + 'static,
+    E: ExecutorClient<B, T> + Unpin,
+    <B as http_body::Body>::Error: std::error::Error + Send + Sync + 'static,
```
seanmonstar, line 223: And this changed to `Into`.

src/common/exec.rs
```diff
@@ -1,50 +1,14 @@
...
-        f.debug_struct("Exec").finish()
-    }
-}
 
 // If http2 is not enable, we just have a stub here, so that the trait bounds
 // that *would* have been needed are still checked. Why?
 //
 // Because enabling `http2` shouldn't suddenly add new trait bounds that cause
 // a compilation error.
-#[cfg(not(feature = "http2"))]
```
seanmonstar, line 43: This cfg is no longer needed?

src/rt/bounds.rs
```diff
@@ -6,43 +6,83 @@
...
     }
 
-    impl<F, B> Http2ConnExec<F, B> for Exec
+    impl<E, B, T> ExecutorClient<B, T> for E
     where
-        H2Stream<F, B>: Future<Output = ()> + Send + 'static,
-        B: Body,
+        E: Executor<H2ClientFuture<B, T>>,
+        B: http_body::Body + 'static,
+        B::Error: std::error::Error + Send + Sync + 'static,
```
seanmonstar, line 42: `Into`?

src/rt/bounds.rs
```diff
@@ -6,43 +6,83 @@
...
     }
 
-    impl<F, B> sealed::Sealed<(F, B)> for Exec
+    impl<E, B, T> sealed_client::Sealed<(B, T)> for E
     where
-        H2Stream<F, B>: Future<Output = ()> + Send + 'static,
-        B: Body,
+        E: Executor<H2ClientFuture<B, T>>,
+        B: http_body::Body + 'static,
+        B::Error: std::error::Error + Send + Sync + 'static,
```
seanmonstar, line 55: `Into`?

seanmonstar (commented):
Very impressive! I think there's just some dangling clean up that I found.

src/common/exec.rs
Ruben2424: I moved it to the module. Is this ok?

src/client/conn/http2.rs
Ruben2424: I changed all `Error` bounds. Now the test works again with `Box<dyn Error ...>`

seanmonstar (approved):
Very well done!

seanmonstar:
Oh, looks like there's merge conflicts. Would you be up to resolving those? Then we can merge. RC4 is nearly here!

PR #3191, opened 2023-04-02
fix(ffi): add repr(transparent) to hyper_context

LegionMammal978:
The default representation does not guarantee the absence of initial padding, which is necessary for the transmute to be sound.

### DIFF
--- a/src/ffi/task.rs
+++ b/src/ffi/task.rs
@@ -66,6 +66,7 @@ struct TaskFuture {
 }
 
 /// An async context for a task that contains the related waker.
+#[repr(transparent)]
 pub struct hyper_context<'a>(Context<'a>);
 
 /// A waker that is saved and used to waken a pending task.
@@ -378,7 +379,7 @@ where
 
 impl hyper_context<'_> {
     pub(crate) fn wrap<'a, 'b>(cx: &'a mut Context<'b>) -> &'a mut hyper_context<'b> {
-        // A struct with only one field has the same layout as that field.
+        // A repr(transparent) struct with only one field has the same layout as that field.
         unsafe { std::mem::transmute::<&mut Context<'_>, &mut hyper_context<'_>>(cx) }
     }
 }


seanmonstar (approved):
Thanks!

seanmonstar:
Sorry, I had to revert this in #3193, since I accidentally merged it before checking CI was happy. At the least, it's changed how the `hyper.h` file is generated. I can't yet tell if the change is important.

LegionMammal978:
Whoops! I hadn't noticed that this project uses cbindgen. Indeed, it doesn't seem to do what we want, referring to a nonexistent `Context` type:

```diff
@@ -158,11 +158,6 @@
 typedef struct hyper_clientconn_options hyper_clientconn_options;
 
 /*
- An async context for a task that contains the related waker.
- */
-typedef struct hyper_context hyper_context;
-
-/*
  A more detailed error object returned by some hyper functions.
  */
 typedef struct hyper_error hyper_error;
@@ -206,15 +201,20 @@
 
 typedef int (*hyper_body_foreach_callback)(void*, const struct hyper_buf*);
 
-typedef int (*hyper_body_data_callback)(void*, struct hyper_context*, struct hyper_buf**);
+/*
+ An async context for a task that contains the related waker.
+ */
+typedef Context hyper_context;
+
+typedef int (*hyper_body_data_callback)(void*, hyper_context*, struct hyper_buf**);
 
 typedef void (*hyper_request_on_informational_callback)(void*, struct hyper_response*);
 
 typedef int (*hyper_headers_foreach_callback)(void*, const uint8_t*, size_t, const uint8_t*, size_t);
 
-typedef size_t (*hyper_io_read_callback)(void*, struct hyper_context*, uint8_t*, size_t);
+typedef size_t (*hyper_io_read_callback)(void*, hyper_context*, uint8_t*, size_t);
 
-typedef size_t (*hyper_io_write_callback)(void*, struct hyper_context*, const uint8_t*, size_t);
+typedef size_t (*hyper_io_write_callback)(void*, hyper_context*, const uint8_t*, size_t);
 
 #ifdef __cplusplus
 extern "C" {
@@ -743,7 +743,7 @@
 /*
  Copies a waker out of the task context.
  */
-struct hyper_waker *hyper_context_waker(struct hyper_context *cx);
+struct hyper_waker *hyper_context_waker(hyper_context *cx);
 
 /*
  Free a waker that hasn't been woken.
```

And if we try replacing it with `#[repr(C)]`, then it writes out the whole struct definition, again including the `Context` type. Looking at cbindgen, the only official way to get it to generate an opaque item is not to place any `repr()` attribute on the structure. Curiously, there's also an unofficial way, which is to do anything that would make cbindgen throw an error on the struct, e.g., prefix it with `/// cbindgen:==`. But this is a sticky issue all around, and it's hard for me to tell which solution would be best.

PR #3194, opened 2023-04-04
Update `socket2` to 0.5.1

nickelc:
The new version switched from `winapi` to `windows-sys`.


nickelc:
I didn't check the MSRV. `socket2` 0.5.1 requires rust 1.63

PR #3201, opened 2023-04-13
h2: add max_pending_accept_reset_streams configuration option

Noah-Kennedy:
This allows users to set the configuration option from https://github.com/hyperium/h2/pull/668.

### DIFF
--- a/src/server/conn.rs
+++ b/src/server/conn.rs
@@ -396,7 +396,8 @@ impl<E> Http<E> {
 
     /// Configures the maximum number of pending reset streams allowed before a GOAWAY will be sent.
     ///
-    /// This will default to whatever the default in h2 is. As of v0.3.17, it is 20.
+    /// This will default to the default value set by the [`h2` crate](https://crates.io/crates/h2).
+    /// As of v0.3.17, it is 20.
     ///
     /// See <https://github.com/hyperium/hyper/issues/2877> for more information.
     #[cfg(feature = "http2")]


src/proto/h2/server.rs
```diff
@@ -35,7 +35,7 @@ const DEFAULT_CONN_WINDOW: u32 = 1024 * 1024; // 1mb
 const DEFAULT_STREAM_WINDOW: u32 = 1024 * 1024; // 1mb
 const DEFAULT_MAX_FRAME_SIZE: u32 = 1024 * 16; // 16kb
 const DEFAULT_MAX_SEND_BUF_SIZE: usize = 1024 * 400; // 400kb
-// 16 MB "sane default" taken from golang http2
+                                                     // 16 MB "sane default" taken from golang http2
```
LPardue, line 38: errm

Noah-Kennedy: huh, fixing

Noah-Kennedy: its rustfmt

Noah-Kennedy: I can do nothing here, it is rustfmt

Cargo.toml
```diff
@@ -13,10 +13,10 @@ categories = ["network-programming", "web-programming::http-client", "web-progra
...
 
 include = [
-  "Cargo.toml",
-  "LICENSE",
-  "src/**/*",
-  #"build.rs",
+    "Cargo.toml",
+    "LICENSE",
+    "src/**/*",
+    #"build.rs",
```
hawkw, line 19: nit: unrelated formatting change?
```suggestion
  "Cargo.toml",
  "LICENSE",
  "src/**/*",
  #"build.rs",
```

src/proto/h2/server.rs
hawkw, line 38:
```suggestion
// 16 MB "sane default" taken from golang http2
```

src/server/conn.rs
```diff
@@ -394,6 +394,17 @@ impl<E> Http<E> {
         self
     }
 
+    /// Configures the maximum number of pending reset streams allowed before a GOAWAY will be sent.
+    ///
+    /// See <https://github.com/hyperium/hyper/issues/2877> for more information.
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
+    pub fn http2_max_pending_accept_reset_streams(&mut self, max: usize) -> &mut Self {
```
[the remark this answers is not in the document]
hawkw: We should definitely mention the default value in the docs. I'm +1 on taking `impl Into<Option<usize>>` here, but :woman_shrugging:

src/proto/h2/server.rs
Noah-Kennedy: Rustfmt was doing this, but ill take this for the same of a small diff

src/server/conn.rs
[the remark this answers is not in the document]
Noah-Kennedy: I have zero preference either way. I'll just go ahead and change it.

seanmonstar:
Does it seem likely people want to configure this on the client? I'm guessing since hyper disables server push, there's nothing to worry about?

Noah-Kennedy:
> Does it seem likely people want to configure this on the client? I'm guessing since hyper disables server push, there's nothing to worry about?

I don't think we need client configuration for now, if someone asks then we can add it IMO.

Noah-Kennedy:
Like you said, we don't support server push anyways for the client.

src/server/conn.rs
```diff
@@ -394,6 +394,22 @@ impl<E> Http<E> {
         self
     }
 
+    /// Configures the maximum number of pending reset streams allowed before a GOAWAY will be sent.
+    ///
+    /// This will default to whatever the default in h2 is. As of v0.3.17, it is 20.
```
hawkw, line 399: nit: is there documentation in `h2` we can link to for this?

hawkw: also, silly docs nit: I might say 
```suggestion
    /// This will default to the default value set by the [`h2` crate](https://crates.io/crates/h2).
    /// As of v0.3.17, it is 20.
```
or something, to make it obvious that when we say `h2` here, we are referring to the library?

hawkw (approved):
some last docs quibbles, but LGTM otherwise!

Noah-Kennedy:
I committed your suggestions

PR #3204, opened 2023-04-16
chore(ci): replace actions-rs with run and use taiki-e/install-action or taiki-e/cache-cargo-install-action

tottoto:
- Replaces actions actions-rs/cargo with run. As it is unmaintained, it warns `Node.js 12 actions are deprecated.`
- Uses taiki-e/install-action, which installs tools without building them and taiki-e/cache-cargo-install-action, which caches the built binaries. These actions saves ci time to install tools.


seanmonstar (approved):
Thanks!

PR #3205, opened 2023-04-18
fix(client): run `connect_to` to completion even if `ResponseFuture` is canceled

### LINKED ISSUE #3199: HTTP2 requests waiting an outstanding connecting task should not be canceled when the connecting task is canceled
**Version**
0.14.25. I guess it applies also to the pool in hyper-util, but haven't really checked. 

**Platform**
Darwin Kernel Version 21.4.0: Mon Feb 21 20:35:58 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T6000 arm64

**Description**

When there are a bunch of HTTP2 requests waiting for a connection from the pool and there is already a connecting task, [if the connecting task is dropped](https://github.com/hyperium/hyper/blob/v0.14.25/src/client/pool.rs#L706-L715) before the waiters receive the new connection [we drop all the waiter senders](https://github.com/hyperium/hyper/blob/v0.14.25/src/client/pool.rs#L397-L400), resulting in the cancelation of the awaiting requests, arguing that the connecting attempt failed. However, this also happens if the connecting attempt is cancelled by the user (ie the `ResponseFuture` is dropped) and not technically failed.

I've written this test that reproduces (what I consider) the bug:

```rust
    #[tokio::test]
    async fn http2_connection_waiters_are_not_canceled_when_outstanding_connecting_task_is_canceled() {
        let _ = pretty_env_logger::try_init();

        let server = TcpListener::bind("127.0.0.1:0").unwrap();
        let addr = server.local_addr().unwrap();

        #[derive(Clone)]
        struct SlowConnector;

        impl hyper::service::Service<Uri> for SlowConnector {
            type Response = TcpStream;
            type Error = hyper::Error;
            type Future = future::Pending<Result<TcpStream, hyper::Error>>;

            fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
                Poll::Ready(Ok(()))
            }

            fn call(&mut self, _dst: Uri) -> Self::Future {
                // An slow connector...
                future::pending()
            }
        }

        let client = Client::builder().http2_only(true).build(SlowConnector);

        // This first request starts the connecting task 
        let req = Request::builder()
            .uri(&*format!("http://{}/a", addr))
            .body(Body::empty())
            .unwrap();
        let mut res1 = client.request(req).map(|_| unreachable!());

        // This second request waits for the conencting task to finish
        // instead of starting a new connecting task
        let req = Request::builder()
            .uri(&*format!("http://{}/a", addr))
            .body(Body::empty())
            .unwrap();
        let mut res2 = client
            .request(req)
            // Test fails here with
            // thread 'dispatch_impl::http2_connection_waiters_are_not_canceled_when_outstanding_connecting_task_is_canceled'
            // panicked at 'internal error: entered unreachable code: res2 should had never resolved, but resulted in
            // Err(hyper::Error(Canceled, "request has been canceled"))', tests/client.rs:1398:22
            .map(|r| unreachable!("res2 should had never resolved, but resulted in {:?}", r));

        // Prime the requests
        assert!(
            future::poll_fn(|ctx| Poll::Ready(Pin::new(&mut res1).poll(ctx).is_pending())).await
        );
        assert!(
            future::poll_fn(|ctx| Poll::Ready(Pin::new(&mut res2).poll(ctx).is_pending())).await
        );

        // The `ResponseFuture` that drives the connecting task is dropped, and the connecting task is canceled
        drop(res1);

        // The request that was waiting for the connecting task fails with Error::Canceled
        assert!(
            future::poll_fn(|ctx| Poll::Ready(Pin::new(&mut res2).poll(ctx).is_pending())).await
        );
    }
```

Thoughts? I'm up to work on a solution if we agree on the assessment.

arnauorriols:
In the previous implementation the `connect_to` future was spawned to let it finish when the checkout future won the race. However, in the event of the user dropping the `ResponseFuture`, the `connect_to` future was canceled as well, and all the http2 requests waiting for the connection would error as a result. This commit prevents that by spawning the `connect_to` future invariably (once it starts) thus covering both cases.

Fixes #3199


arnauorriols:
I've chosen a slightly different approach than that suggested in #3199, and also taken the liberty to remove `common/lazy.rs`, as it is no longer used with this approach. Let me know what you think, I'm open to trying other options.

seanmonstar:
> taken the liberty to remove common/lazy.rs, as it is no longer used with this approach.

I think this still needs to be left in. It's there to prevent _always_ starting a `connect` call, if a checkout would have been just fine. There used to be a unit test for that, but I don't think it survived the transition from futures 0.1 to async/await... (https://github.com/hyperium/hyper/commit/13741f5145eb3dc894d2bc8d8486fc51c29e2e41)

arnauorriols:
I thought the particularity of `Lazy` is to implement `Started`, which we use to check if the connect future has already started when checkout wins. However, as in this PR we are spawning the connect future anyway, we no longer need to check if it has already started. 

However if I'm not mistaken `Lazy` is not lazier than any other future, right? Note that the `connector.connect()` future is spawned only after the future returned by `connect_to` is polled for the first time. The following tests currently cover that requisite (I know because my initial implementation didn't comply with it and these tests fail):
```
    dispatch_impl::alpn_h2
    dispatch_impl::client_keep_alive_0
    dispatch_impl::client_keep_alive_eager_when_chunked
    dispatch_impl::client_keep_alive_when_response_before_request_body_ends
```

arnauorriols:
@seanmonstar could you approve the workflow again please 🙏 ?

seanmonstar:
> However, as in this PR we are spawning the connect future anyway

I'm thinking through this... Is this true _always_? I'm thinking that at least in the HTTP/1 case, after grabbing the `checkout` and `connect_to` futures and racing them, if the `checkout` wins immediately and the `connect_to` was never "started", there's no need to spawn it and trigger a new idle connection, right?

arnauorriols:
> > However, as in this PR we are spawning the connect future anyway
> 
> I'm thinking through this... Is this true _always_? I'm thinking that at least in the HTTP/1 case, after grabbing the `checkout` and `connect_to` futures and racing them, if the `checkout` wins immediately and the `connect_to` was never "started", there's no need to spawn it and trigger a new idle connection, right?

I misspoke in the previous comment. What you say is exactly how it behaves (also for http2), and it is precisely the reason I claim the `Lazy` struct is no longer needed. What I meant is that the current implementation tries to run the `connect_to` future in the same task, and in the event of an error (if the `connect_to` future has started) spawns it in a background task to let it finish. This PR essentially removes this attempt to avoid spawning a new task for the connection attempt, and does it invariably, but instead of spawning the `connect_to` future, it spawns the `connector.connect()` future within (thus not spawning anything if the `connect_to` future is never polled).

arnauorriols:
Let me know how do you want to resolve this. If I haven't convinced you about the `Lazy` I'll revert that change, no worries!

seanmonstar:
Sorry, I was wrapping up a few other things, and wanted to pay this more attention to be sure I understand, because I feel this is pretty subtle. I'll take another look!

src/client/client.rs
```diff
@@ -445,91 +419,102 @@ where
...
+        // connection being established, then this can't take a
+        // second lock. The "connect_to" future is Canceled.
+        let connecting = match pool.connecting(&pool_key, ver) {
+            Some(lock) => lock,
+            None => {
+                let canceled = crate::Error::new_canceled().with("HTTP/2 connection in progress");
+                return Either::Right(future::err(canceled));
+            }
+        };
+        Either::Left(Box::pin(async move {
```
seanmonstar, line 434: Ok, I see, by putting the call to `connector.connect()` inside this sub future, `async move`, it won't call `connect()` until the first poll. Got it.

tests/client.rs
```diff
@@ -1358,6 +1358,70 @@ mod dispatch_impl {
...
+        );
+
+        // The `ResponseFuture` that drives the connecting task is dropped, but
+        // the connecting task should finish in the background
+        drop(res1);
+
+        assert!(
+            // The second response has not been canceled and it is still waiting for the connection
+            // that will never arrive
+            future::poll_fn(|ctx| Poll::Ready(Pin::new(&mut res2).poll(ctx).is_pending())).await
```
seanmonstar, line 1421: But will it actually fulfill with the connection? It'd be good to test that. Otherwise, a _lost_ future here would also make this test pass.

seanmonstar (commented):
OK, I see what you mean about the lazy. It's become a different form of lazy.

One downside with this is it requires spawning another task. Would it be possible to do this without a new task? Or would it require way too much juggling of the pool to get it to work?

tests/client.rs
arnauorriols: Good point. I've implemented the suggestion. Thanks!

arnauorriols:
> One downside with this is it requires spawning another task. Would it be possible to do this without a new task? Or would it require way too much juggling of the pool to get it to work?

I'll give it a try!

PR #3209, opened 2023-04-22
fix RUSTSEC-2023-0034 by updating h2

suryapandian:
Fixes https://rustsec.org/advisories/RUSTSEC-2023-0034 by updating h2


seanmonstar:
This isn't required, you can just run `cargo update` in your own project.

PR #3219, opened 2023-05-05
chore(dependencies): use futures-core via futures-util consistently

tottoto:
`hyper` has already used `futures-core` via `futures-util` at many places. We could reduce some costs to maintain dependencies by doing it consistently.


seanmonstar (approved):
Sure, seems a little cleaner, thanks!

Issue #3040, opened 2022-11-01
Change Service::call to take &self


seanmonstar:
Now that hyper defines its own `Service` trait (#2853), there's a key part we should clarify before hitting 1.0 Final: **Should it take `&self` or `&mut self`?** Whichever the decision, the why should be documented (not necessarily in the API docs, but in code comments or in the `docs/` folder).

The usual arguments are here:

- `&self`
  - It prepares the way for `async fn`, since then the future only borrows `&self`, and thus a `Service` can concurrently handle multiple outstanding requests at once.
  - It's clearer that `Service`s can likely be cloned.
- `&mut self`
  - If a `Service` has state (usually per-connection), it's cheaper to mutate it on each connection. Not being `&mut` would require internal synchronization.

I'm happy to open it up to discussion here. I'd recommend that comments are arguments be well-formed and reasoned, and try to provide something new, so that we keep the signal high. [Chat](https://discord.gg/kkwpueZ) is a fine place to bounce an idea around, and then come back and write it up so we can keep the history (decision history is extremely valuable).

davidpdrsn:
My experience is that most services need to be `Clone` for two reasons:

1. To get an owned reference that can be moved into the response future.
2. When using `tower::make::Shared` (or [similar](https://docs.rs/axum/latest/axum/routing/struct.IntoMakeService.html))

Therefore to share state across clones you generally need `Arc<Mutex<_>>`. That means you're not really using the `&mut self` and could do with a `&self`. So while I haven't done much experimentation with writing `&self` services my gut feeling is that `&self` is better.

LucioFranco:
I will write up more once I have explored what the async fn in traits feature provides us. But from my experience tonic has used `async fn handler(&self, Req) -> Res` setup and had no issues. In fact, I think it expresses very well what the api does and allows users to add a mutex or atomic ops or static read only types etc. That said, we need this api to be future proof for what tower might be and does that mean we need to wait for GATs?

asonix:
This may have been brought up elsewhere, but actix-web's [`Service`](https://docs.rs/actix-service/latest/actix_service/trait.Service.html) trait takes `&self`, but is otherwise very similar to Tower's trait

fundon:
This is one of my practices [`Handler`](https://docs.rs/viz-core/0.4.3/src/viz_core/handler/mod.rs.html#39-66) in Viz. `&self` should be sufficient for most scenarios.

rob2244:
[...]

PR #3223, opened 2023-05-12
refactor(service): change Service::call to take &self

rob2244:
change Service::call to take &self instead of &mut self. Because of this change, the trait bound in the service::util::service_fn and the trait bound in the impl for the ServiceFn struct were changed from FnMut to Fn. This change was decided on for the following reasons:
- It prepares the way for async fn, since then the future only borrows &self, and thus a Service can concurrently handle multiple outstanding requests at once.
- It's clearer that Services can likely be cloned
- To share state across clones you generally need Arc<Mutex<_>> that means you're not really using the &mut self and could do with a &self

This commit closses issue: #3040
BREAKING CHANGE: The Service::call function no longer takes a mutable reference to self. The FnMut trait bound on the service::util::service_fn function and the trait bound on the impl for the ServiceFn struct were changed from FnMut to Fn

### DIFF
--- a/examples/service_struct_impl.rs
+++ b/examples/service_struct_impl.rs
@@ -8,6 +8,7 @@ use tokio::net::TcpListener;
 use std::future::Future;
 use std::net::SocketAddr;
 use std::pin::Pin;
+use std::sync::Mutex;
 
 type Counter = i32;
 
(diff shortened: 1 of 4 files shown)


seanmonstar (approved):
Awesome, thanks!

seanmonstar:
@programatik29 what's an adapater in hyper-util going to need to look like, do you think?

rob2244:
> @programatik29 what's an adapater in hyper-util going to need to look like, do you think?

When you say adapter, are there structs in hyper-util that implement this trait? Couldn't we just change them there as well? Or am I missing something?

programatik29:
@seanmonstar With current tower services, I think our best bet is something like this: 

https://github.com/davidpdrsn/tower-hyper-http-body-compat/blob/cbaa7df5246587082b770d91283daf646303c34e/src/service.rs#L31-L46

seanmonstar:
> are there structs in hyper-util that implement this trait?

Not yet, but we could add them. Or there could be a separate crate, as was just linked. :shrug:

PR #3230, opened 2023-05-17
feat(rt): replace IO traits with hyper::rt ones

### LINKED ISSUE #3110: hyper should use its own IO traits
This was actually brought up as [an unresolved question in the ROADMAP](https://github.com/hyperium/hyper/blob/master/docs/ROADMAP.md#should-there-be-hyperio-traits). I think I mentally dismissed it at some point as just being annoying, but in the past week as I've thought about how we could make use of io-uring in hyper, I've been thinking about this again. If nothing else, this _should_ be a public record of the decision, whether for or against. This will help others, including Future Us.

All other integration with a runtime (Tokio) has been removed, and helpers exist in `hyper-util`. hyper 1.0 (as of of rc.2) still depends on `tokio` with all features turned off, to ask that IO transports implement `tokio::io::{AsyncRead, AsyncWrite}`. By doing so, it makes it _easier_ for users to simply supply a `tokio::net::TcpStream`. But, let me at least bring up some downsides.

## Reasons for `hyper::io::{AsyncRead, AsyncWrite}`

- For people integrating with _other_ runtimes (curl, fuchsia, etc), it can feel odd to need to pull in `tokio` just to implement the IO traits hyper wants. People have expressed concern that they don't want to compile multiple runtimes. While we can explain it's just the traits (when the features are turned off), owning the traits could make people less confused.
- If we own the traits, we can decide to try different things that perhaps Tokio itself wouldn't want to do. This includes differences in opinion of how vectored reads and writes should be supported, but also means we could try to provide specialization paths for completion-based IO (such as io-uring).

We can provide `hyper_util::io::Tokio(T)` that implements the traits for Tokio types, to reduce friction.

seanmonstar:
This replaces the usage of `tokio::io::{AsyncRead, AsyncWrite}` in hyper's public API with new traits in the `hyper::rt` module.

Closes #3110 

## Before merging

- [x] Determine naming (`AsyncRead` vs `Read`; `ReadBuf` and `ReadBufCursor`)
- [x] Determine minimal public methods for `ReadBuf` and `Cursor`
- [x] Document the types and methods
- [ ] Prove that it is forwards-compatible with completion-based IO

### DIFF
--- a/benches/end_to_end.rs
+++ b/benches/end_to_end.rs
@@ -4,8 +4,7 @@
 extern crate test;
 mod support;
 
-// TODO: Reimplement Opts::bench using hyper::server::conn and hyper::client::conn
-// (instead of Server and HttpClient).
+// TODO: Reimplement parallel for HTTP/1
 
 use std::convert::Infallible;
 use std::net::SocketAddr;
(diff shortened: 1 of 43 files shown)


seanmonstar:
cc @Noah-Kennedy, regarding making sure it's forwards-compatible.

Issue #2730, opened 2022-01-01
Connection::graceful_shutdown always waits for the first request.


sfackler:
**Version**
hyper 0.14.16

**Platform**
Linux DESKTOP-DHO88R7 4.19.104-microsoft-standard #1 SMP Wed Feb 19 06:37:35 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

**Description**
If you gracefully shut down a server connection future before the first request, Hyper will not actually shut the connection down until a request is processed:

```rust
use hyper::server::conn::Http;
use hyper::Response;
use tokio::io::AsyncReadExt;
use tokio::net::{TcpListener, TcpStream};
use tokio::pin;

#[tokio::main]
async fn main() {
    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
    let addr = listener.local_addr().unwrap();

    tokio::spawn(server(listener));

    let mut stream = TcpStream::connect(addr).await.unwrap();
    println!("connected");
    let mut buf = vec![];
    stream.read_to_end(&mut buf).await.unwrap();
}

async fn server(listener: TcpListener) {
    let socket = listener.accept().await.unwrap().0;

    let service = hyper::service::service_fn(|_: hyper::Request<hyper::Body>| async {
        Err::<Response<hyper::Body>, _>("no")
    });

    let future = Http::new()
        .http1_only(true)
        .serve_connection(socket, service);
    pin!(future);
    future.as_mut().graceful_shutdown();

    future.await.unwrap();
}
```

I would expect this program to exit almost instantly since there is no request being processed when the graceful_shutdown is invoked. However, it instead blocks forever waiting on the client to send headers.

The behavior actually appears to be that the shutdown is processed immediately after the first request is fully handled.

seanmonstar:
Yea, I think this is behavior is currently on purpose. Whether it _should_ be is a fair question. I think at the time, I assumed that a _new_ connection would usually have a request incoming ASAP, whereas an idle connection might not, and so it was better to allow that request to come in.

We could alter that behavior, if we document a good reason for _why_ we're making that change. Also, I imagine using the `http1_header_read_timeout` option could help here.

sfackler:
[...]

PR #3233, opened 2023-05-23
fix(http1): http1 server graceful shutdown fix

rob2244:
fix issue in the graceful shutdown logic
which causes the connection future to hang
when graceful shutdown is called prior to any
requests being  made. This fix checks to see
if the connection is still in its initial state
when disable_keep_alive is called, and starts the
shutdown process if it is.

This addresses issue #2730

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -175,6 +175,11 @@ where
         }
     }
 
+    #[cfg(feature = "server")]
+    pub(crate) fn has_initial_read_write_state(&self) -> bool {
+        matches!(self.state.reading, Reading::Init) && matches!(self.state.writing, Writing::Init)
+    }
+
     fn should_error_on_eof(&self) -> bool {
         // If we're idle, it's probably just the connection closing gracefully.
         T::should_error_on_parse_eof() && !self.state.is_idle()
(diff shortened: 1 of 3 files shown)


rob2244:
I'm not sure if this is the best solution or not, it feels like it's catering a little too much to this one edge case if you know what I mean. But it does pass all the tests. Let me know if there's a better way to do this.

rob2244:
Actually, it seems like the tests failed on mac-OS. I think this solution may introduce a race condition.

seanmonstar:
Ok, after taking a look: the problem is that the connection is closing on all platforms, but macOS is triggering an error trying to write the second half of the request.

@sfackler in your mind, what behavior did you expect? If the request has already _started_, as in we've received some of the bytes, but not the full headers, should that be allowed to continue? Or should that be closed, and we only allow requests to continue if they are streaming their body?

I think the goal was to prevent pre-emptive idle connections from keeping the server alive, not to interrupt new connections that had their headers split between packets, right?

seanmonstar:
Fixed this up in #3261.

sfackler:
Yeah, I would expect that the idle timer would start counting as soon as the connection is created, and stop as soon as the first byte of the headers are received. The header timeout should handle the second case I think.

PR #3242, opened 2023-06-09
fix(client): Do not strip `path` and `scheme` components from URIs for HTTP/2 Extended CONNECT requests

bdbai:
As per [RFC 8441](https://datatracker.ietf.org/doc/html/rfc8441#section-4):

> On requests that contain the :protocol pseudo-header field, the
      :scheme and :path pseudo-header fields of the target URI (see
      [Section 5](https://datatracker.ietf.org/doc/html/rfc8441#section-5)) MUST also be included.

I was trying to implement WebSocket handshake based on HTTP/2 Extended CONNECT. The current implementation crafted a request which was considered invalid by the server (HAProxy 2.8):

image

Compared to the [example](https://datatracker.ietf.org/doc/html/rfc8441#section-5.1) illustrated in RFC 8441, the pseudo headers `:scheme` and `:path` were missing.

After the fix, the client could successfully establish a WebSocket tunnel to the server.

image

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -277,8 +277,6 @@ where
             } else {
                 origin_form(req.uri_mut());
             }
-        } else if req.method() == Method::CONNECT {
-            authority_form(req.uri_mut());
         }
 
         let mut res = match pooled.send_request_retryable(req).await {
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -16,6 +16,7 @@ use crate::common::{
     exec::BoxSendFuture, lazy as hyper_lazy, sync_wrapper::SyncWrapper, task, Future, Lazy, Pin,
     Poll,
 };
+#[cfg(feature = "http2")]
 use crate::ext::Protocol;
 use crate::rt::Executor;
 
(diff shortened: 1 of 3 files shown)


bdbai:
Hi @nox , can I nudge a little bit for reviewing this PR?👀

nox (commented):
Would be nice to have a test but this looks good to me.

bdbai:
@nox I have added two integrations tests for client `CONNECT` requests.

@seanmonstar Since the Extended CONNECT requests only apply to h2, I've made a few changes to the helper macro in the integration test so that certain tests will not run in http/1 mode. Please take a look and see if you are okay with this approach.

seanmonstar:
I'm happy to get this in with the test, thanks so much! Looks like CI noticed a combination of features that fail to compile, so probably a `cfg(feature = ...)` flag is needed in a new place.

bdbai:
@seanmonstar I've updated it with the feature flags. Hopefully it will pass this time.

seanmonstar (approved):

seanmonstar:
Thank you for being patient with us!

seanmonstar:
Yes, a similar thing could be made a PR to https://github.com/hyperium/hyper-util

PR #3250, opened 2023-06-14
Bump msrv to 1.60

Ruben2424:
CI fails in #3184 because the log crate now requires rustc version 1.60.


seanmonstar:
Do we need to bump it? Is log a dependency? Or just a dev-dependency?

Ruben2424:
i just ran `cargo tree` it is under the normal dependency tree under `want` and `mio`.

seanmonstar:
If we upgrade to want 0.3.1, it no longer depends on `log`. And by removing the `default-features` from mio, we'll be free!

Ruben2424:
It looks like want`0.3.1` fixes the issue. No change needed.
Can you re-run CI in #3184

PR #3255, opened 2023-06-16
Cargo fmt workaround

Ruben2424:
`cargo fmt` does not work for all files. This adds the same workaround as in [tokio](https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L498)


seanmonstar (approved):
Thanks!

PR #3261, opened 2023-07-06
fix(http1): http1 server graceful shutdown fix

### LINKED ISSUE #2730: Connection::graceful_shutdown always waits for the first request.

seanmonstar:
This is from #3233, I've fixed up the failing test.

---

fix issue in the graceful shutdown logic
which causes the connection future to hang
when graceful shutdown is called prior to any
requests being  made. This fix checks to see
if the connection is still in its initial state
when disable_keep_alive is called, and starts the
shutdown process if it is.

Closes #2730

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -175,6 +175,13 @@ where
         }
     }
 
+    #[cfg(feature = "server")]
+    pub(crate) fn has_initial_read_write_state(&self) -> bool {
+        matches!(self.state.reading, Reading::Init)
+            && matches!(self.state.writing, Writing::Init)
+            && self.io.read_buf().is_empty()
+    }
+
     fn should_error_on_eof(&self) -> bool {
         // If we're idle, it's probably just the connection closing gracefully.
         T::should_error_on_parse_eof() && !self.state.is_idle()
--- a/src/proto/h1/dispatch.rs
+++ b/src/proto/h1/dispatch.rs
@@ -82,7 +82,11 @@ where
     #[cfg(feature = "server")]
     pub(crate) fn disable_keep_alive(&mut self) {
         self.conn.disable_keep_alive();
-        if self.conn.is_write_closed() {
+
+        // If keep alive has been disabled and no read or write has been seen on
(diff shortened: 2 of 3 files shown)


seanmonstar:
This was included in RC4.

seanmonstar:
If you mean for 0.14.x, it seemed like a behavior change and was marked as a breaking change. I've forgotten the argument for why. I don't feel strongly, unless it would be *break* expected behavior.

djc:
From a brief look at the changes and the PR description, it's not obvious to me why this should be a breaking change.

Issue #3192, opened 2023-04-03
Document Drop behavior for Connection types


DoumanAsh:
This is not per se feature, but rather documentatation issue.
Current documentation on `Connection` types is very limited to understand its behavior

https://docs.rs/hyper/1.0.0-rc.3/hyper/client/conn/http1/struct.Connection.html
https://docs.rs/hyper/1.0.0-rc.3/hyper/client/conn/http2/struct.Connection.html

It asks to poll it, but it doesn't say what will happen if you drop it (due to panic or other reasons)
It would be good to add more details on its behavior so that user could be confident in how to handle it after request had been handled

fercevik729:
I don't see documentation for Drop behavior on Connection types. Could I try my hand at this issue?

seanmonstar:
Sure, always welcome!

fercevik729:
#3264

DoumanAsh:
@fercevik729 you should add documentation on `Connection` types directly rather than poll methods
This is where user would expect to see all details on how to use `Connection`

fercevik729:
Should I add it as a note under the existing documentation of the structs?

fercevik729:
Sure thing!

ugochukwu-850:
Has this issue been handled

PR #3264, opened 2023-07-13
Resolves issue #3192

fercevik729:
Added notes about potential issues with calling poll() on a Connection type after it has been dropped, resolves issue #3192

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -46,6 +46,11 @@ pub struct Parts<T> {
 ///
 /// In most cases, this should just be spawned into an executor, so that it
 /// can process incoming and outgoing messages, notice hangups, and the like.
+///
+/// # Note
+///
+/// Attempting to poll a Connection after it has been dropped may lead
+/// to undefined behavior and is not recommended.
 #[must_use = "futures do nothing unless polled"]
 pub struct Connection<T, B>
 where
@@ -255,10 +260,6 @@ where
 {
     type Output = crate::Result<()>;
 
-    /// # Note
-    ///
-    /// Attempting to poll a Connection after it has been dropped may lead
-    /// to undefined behavior and is not recommended
     fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
         match ready!(Pin::new(self.inner.as_mut().unwrap()).poll(cx))? {
             proto::Dispatched::Shutdown => Poll::Ready(Ok(())),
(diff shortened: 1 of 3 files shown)


src/client/conn/http1.rs
```diff
@@ -255,6 +255,10 @@ where
 {
     type Output = crate::Result<()>;
 
+    /// # Note
+    ///
+    /// Attempting to poll a Connection after it has been dropped may lead
```
DoumanAsh, line 260: does it though?

you should not use term undefined behavior lightly

seanmonstar:
Thanks for the PR!

In this case, it's not possible to poll a connection after it's been dropped. That's because once dropped, you know longer have the `Connection` to be able to call `conn.poll()` on it.

I think the original issue was asking what happens if you drop the `Connection` before `Connection::poll()` has returned `Poll::Ready()`. The general answer is that since there isn't a concept of async drop, it doesn't do anything extra with the socket. The socket just gets closed abruptly, in whatever state is was (well, technically that's dependent on what `T::drop()` does).

src/client/conn/http1.rs
fercevik729: https://doc.rust-lang.org/std/future/trait.Future.html. It says that once a Future is resolved, calling poll might result in panic, indefinite blocking, or other problems. Should I revise the documentation to say that?

DoumanAsh:
> I think the original issue was asking what happens if you drop the Connection before Connection::poll() has returned Poll::Ready()

Yes, but was also clarifying how to handle `Connection` in context of handshake:
https://docs.rs/hyper/latest/hyper/client/conn/fn.handshake.html

Current state of hyper documentation is not very helpful for users
The example only shows that you should spawn it on task but that seems to be quite oversimplified and lazy.
Normally you would just concurrently poll both.
But there is also problem of connection pooling as ideally you want to preserve `IO` so you need to destroy `Connection`

When I was checking source code I think `SendRequest` future would yield error once you drop `Connection`?
I don't remember now since it is been a while since I used hyper v1

In any case I created issue because I think it is oversight that there is not much details on behavior of `SendRequest` and `Connection`

src/client/conn/http1.rs
DoumanAsh: Well, technically you cannot poll `Connection` in safe manner after it has been dropped :)
You should consider `Connection` in context of `SendRequest` as you use them together

fercevik729: In that case should I document the `handshake` method as well?

DoumanAsh: I think `conn` module documentation would be good place:
https://github.com/hyperium/hyper/blob/master/src/client/conn/mod.rs#L1

fercevik729: I have pushed some changes to the conn, http1, and http2 modules. I appreciate any feedback :)

PR #3265, opened 2023-07-14
Add `--generate-link-to-definition` option when building on docs.rs

GuillaumeGomez:
This option generates links in source code pages, allowing to jump to definition and to jump back to doc. It makes browsing the source code pages much nicer. You can see it in action [here](https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_middle/lib.rs.html#90).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ nightly = []
 
 [package.metadata.docs.rs]
 features = ["ffi", "full"]
-rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi"]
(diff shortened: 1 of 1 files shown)


seanmonstar:
Is there a reason this isn't enabled by default?

GuillaumeGomez:
Some members of the rustdoc team are opposed to this feature. The RFC is ready but not open (and very likely won't be before long): https://github.com/GuillaumeGomez/rfcs/pull/1

seanmonstar:
I think I'd lean towards leaving it to the defaults. If the rustdoc team eventually feels it should be enabled, we'll get it automatically.

GuillaumeGomez:
As you prefer. You know you have the possibility to enable it now so don't hesitate to use it, it's quite a convenient feature. Anyway, closing in the meantime.

PR #3275, opened 2023-07-23
fix(client): early server response shouldn't propagate NO_ERROR

### LINKED ISSUE #2872: Client: handle `RST_STREAM` with `NO_ERROR` set for the reason
**Version**
```
hyper = "0.14.18"
h2 = "0.3.13"
```

**Platform**
```
> uname -a
Linux <REDACTED> 5.17.5-76051705-generic #202204271406~1651504840~22.04~63e51bd SMP PREEMPT Mon May 2 15: x86_64 x86_64 x86_64 GNU/Linux
```

**Description**
I've found that Google Cloud Storage's API can respond with HTTP/2 `RST_STREAM` frame with `NO_ERROR` set for the reason, which appears to mean "stop sending the request body and read my response" according to https://datatracker.ietf.org/doc/html/rfc7540#section-8.1

> A server can send a complete response prior to the client sending an entire
   request if the response does not depend on any portion of the request
   that has not been sent and received.  When this is true, a server MAY
   request that the client abort transmission of a request without error
   by sending a RST_STREAM with an error code of NO_ERROR after sending
   a complete response (i.e., a frame with the END_STREAM flag).
   Clients MUST NOT discard responses as a result of receiving such a
   RST_STREAM, though clients can always discard responses at their
   discretion for other reasons.

I believe this is happening in response to a `PutObject` request when the bucket is being rate limited for writes. The server is trying to tell the client to stop sending the request body because it won't be processed, and instead it should immediately read the response to discover the `429 Too Many Requests` error code.

However, Hyper's client implementation appears to just return the `RST_STREAM` message as an error and discards the response instead of handling it, which gives a hilariously confusing error message of:
```
error reading a body from connection: stream error received: not a result of an error
```

To be compliant with the spec, the implementation should stop sending the body and immediately read the response and return it.

For context, I'm using the Gcloud Storage API via https://crates.io/crates/aws-sdk-s3 (because the Gcloud Rust SDK doesn't support streaming bodies, but thankfully Gcloud Storage exposes an S3-compatible API), which uses Hyper internally. `aws-sdk-s3` appears to be returning the error from Hyper verbatim, however.

DDtKey:
Closes #2872

Added test is failing against the current version (before this fix)

Backport for `0.14.x`: https://github.com/hyperium/hyper/pull/3274

### DIFF
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -201,7 +201,16 @@ impl Body for Incoming {
                             ping.record_data(bytes.len());
                             return Poll::Ready(Some(Ok(Frame::data(bytes))));
                         }
-                        Some(Err(e)) => return Poll::Ready(Some(Err(crate::Error::new_body(e)))),
+                        Some(Err(e)) => {
+                            return match e.reason() {
+                                // These reasons should cause stop of body reading, but don't fail it.
+                                // The same logic as for `Read for H2Upgraded` is applied here.
+                                Some(h2::Reason::NO_ERROR) | Some(h2::Reason::CANCEL) => {
+                                    Poll::Ready(None)
+                                }
+                                _ => Poll::Ready(Some(Err(crate::Error::new_body(e)))),
+                            }
+                        }
                         None => {
                             *data_done = true;
                             // fall through to trailers
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -203,13 +203,13 @@ impl Body for Incoming {
                         }
                         Some(Err(e)) => {
                             return match e.reason() {
-                                // These reasons should cause stop of body reading, but don't fail it.
+                                // These reasons should cause the body reading to stop, but not fail it.
                                 // The same logic as for `Read for H2Upgraded` is applied here.
                                 Some(h2::Reason::NO_ERROR) | Some(h2::Reason::CANCEL) => {
                                     Poll::Ready(None)
                                 }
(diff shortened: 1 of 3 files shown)


src/body/incoming.rs
```diff
@@ -201,7 +201,16 @@ impl Body for Incoming {
                             ping.record_data(bytes.len());
                             return Poll::Ready(Some(Ok(Frame::data(bytes))));
                         }
-                        Some(Err(e)) => return Poll::Ready(Some(Err(crate::Error::new_body(e)))),
+                        Some(Err(e)) => {
+                            return match e.reason() {
+                                // These reasons should cause stop of body reading, but don't fail it.
+                                // The same logic as for `Read for H2Upgraded` is applied here.
```
DDtKey, line 207: I'm referring to this line here: https://github.com/hyperium/hyper/blob/a45d5d5a04369f93334fc893875d8d1a49054e04/src/proto/h2/mod.rs#L296

Both of them call the same operation (`self.recv_stream.poll_data(cx)`) and seems that logic should be the same. But we can make it more specific.

DDtKey: Btw, an additional example from another lang & lib where it was also solved by avoiding error if response has been read: https://github.com/square/okhttp/pull/6295

nox (approved):
Seems good to me given we already do something similar for streams upgraded from HTTP/2.

seanmonstar (approved):

PR #3281, opened 2023-08-02
Replace tokio with futures.

afranchuk:
This replaces some `tokio` sync things with equivalent `futures-channel` types, and makes `tokio` an optional dependency. It is still needed for `http2` since `h2` needs it (though FWIW it looks like `h2` could use `futures::io` stuff and also make `tokio` optional).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,9 @@ httpdate = "1.0"
 httparse = "1.8"
 h2 = { version = "0.3.9", optional = true }
 itoa = "1"
+tokio = { version = "1", optional = true }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project-lite = "0.2.4"
-tokio = { version = "1", features = ["sync"] }
 want = "0.3"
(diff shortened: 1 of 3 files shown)


Noah-Kennedy:
The tokio sync things do not depend on the tokio runtime, and have a number of advantages over the ones from the futures crate. This is probably not a beneficial change.

seanmonstar:
Yea, the tokio channel is used here because in benchmarks, it's faster. I know the `futures-channel` one is used in another place, and it'd be nice to not have two implementations, but again, it's because that made the benchmarks faster.

Noah-Kennedy:
> Yea, the tokio channel is used here because in benchmarks, it's faster. I know the `futures-channel` one is used in another place, and it'd be nice to not have two implementations, but again, it's because that made the benchmarks faster.

The tokio channel implementations also interact (when running on a tokio runtime) with tokio's budgeting system, which is very useful for preventing starvation, and, in rare cases, deadlocks.

Issue #3289, opened 2023-08-18
h2/client(1.0.0-rc4): panic on a double poll on Select


lucab:
**Version**

hyper 1.0.0-rc4

**Platform**

Linux 5.15.109+ SMP x86_64 GNU/Linux

**Description**

Deno is currently using the RC versions of hyper 1.0 and it recently upgraded from `-rc3` to `-rc4`: https://github.com/denoland/deno/pull/19987.

After that we started seeing occasional but frequent panics due to a double poll on a Select future:
```
thread 'main' panicked at 'cannot poll Select twice', ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.28/src/future/select.rs:112:42
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at ./rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
   2: core::panicking::panic_display
             at ./rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:150:5
   3: core::panicking::panic_str
             at ./rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:134:5
   4: core::option::expect_failed
             at ./rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/option.rs:1932:5
   5: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
   6: <hyper::proto::h2::client::H2ClientFuture<B,T> as core::future::future::Future>::poll
   7: tokio::runtime::task::harness::Harness<T,S>::poll
   8: tokio::runtime::scheduler::current_thread::Context::run_task
   9: tokio::macros::scoped_tls::ScopedKey<T>::set
  10: tokio::runtime::scheduler::current_thread::CurrentThread::block_on
  11: tokio::runtime::runtime::Runtime::block_on
  12: <omitted>::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I don't have a reproducer or a test for this, which I think is the result of a race.

lucab:
[...]

PR #3290, opened 2023-08-18
fix(http2/client): avoid double-polling a Select future

lucab:
This reworks http2 client connection task in order to avoid storing a `Select` future. Under some scheduling cases the future was polled multiple times, resulting in runtime panics:

```
thread 'main' panicked at 'cannot poll Select twice', /index.crates.io/futures-util-0.3.28/src/future/select.rs:112:42
stack backtrace:
[...]
   5: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
   6: <hyper::proto::h2::client::H2ClientFuture<B,T> as core::future::future::Future>::poll
[...]
```

Closes: https://github.com/hyperium/hyper/issues/3289

### DIFF
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -6,7 +6,7 @@ use crate::rt::{Read, Write};
 use bytes::Bytes;
 use futures_channel::mpsc::{Receiver, Sender};
 use futures_channel::{mpsc, oneshot};
-use futures_util::future::{self, Either, FutureExt as _, Select};
+use futures_util::future::{Either, Fuse, FutureExt as _};
 use futures_util::stream::{StreamExt as _, StreamFuture};
 use h2::client::{Builder, Connection, SendRequest};
 use h2::SendStream;
@@ -245,10 +245,11 @@ pin_project! {
         T: Unpin,
     {
         #[pin]
-        select: Select<ConnMapErr<T, B>, StreamFuture<Receiver<Never>>>,
+        drop_rx: Fuse<StreamFuture<Receiver<Never>>>,
         #[pin]
         cancel_tx: Option<oneshot::Sender<Never>>,
-        conn: Option<ConnMapErr<T, B>>,
+        #[pin]
+        conn: Fuse<ConnMapErr<T, B>>,
     }
 }
 
@@ -263,9 +264,9 @@ where
         cancel_tx: oneshot::Sender<Never>,
     ) -> Self {
         Self {
-            select: future::select(conn, drop_rx),
+            drop_rx: drop_rx.fuse(),
             cancel_tx: Some(cancel_tx),
-            conn: None,
+            conn: conn.fuse(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


lucab:
@seanmonstar unfortunately I've only observed these panics on a remote service, but I didn't manage to easily reproduce it manually or through a test.
We did run this patch on the staging environment and didn't observe panics, but maybe it was just a lucky run where we didn't hit the same race condition.

I put together this patch based on the stacktrace and some educated guesses, but I'm not really familiar with this codebase.

The code was introduced in https://github.com/hyperium/hyper/pull/3184 by @Ruben2424, you may want to double-check this.

src/proto/h2/client.rs
```diff
@@ -280,25 +281,20 @@ where
...
-                    // the connection some more should start shutdown
-                    // and then close
-                    trace!("send_request dropped, starting conn shutdown");
-                    drop(this.cancel_tx.take().expect("Future polled twice"));
-                    this.conn = &mut Some(b);
-                    return Poll::Pending;
-                }
-            }
-        }
+        if let Poll::Ready(_) = this.conn.poll_unpin(cx) {
```
piscisaureus, line 284: As far as I'm aware, fusing a future doesn't allow it to be polled multiple times. You have to explicitly check whether the future is already terminated using the [`is_terminated()`](https://docs.rs/futures/latest/futures/future/trait.FusedFuture.html#tymethod.is_terminated) method.

src/proto/h2/client.rs
```diff
@@ -280,25 +281,20 @@ where
...
+            // ok or err, the `conn` has finished.
+            return Poll::Ready(());
+        };
+
+        if let Poll::Ready(_) = this.drop_rx.poll_unpin(cx) {
```
piscisaureus, line 289: Same comment applies here.

src/proto/h2/client.rs
lucab: Thanks for the feedback. My understanding on this is different, but maybe I read it wrong.

The outer future here (and below too) is a `Fuse` (see https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.fuse) which internally already takes care of shunting subsequent polls to `Pending`:
https://github.com/rust-lang/futures-rs/blob/89c06f5e89b8499102a198736ccc859949e049a3/futures-util/src/future/future/fuse.rs#L82-L90
Did I misunderstand something here?

Or maybe are you suggesting to avoid introducing two `Fuse` struct wrappers here and directly try to implement the `FusedFuture` trait on the two existing underlying futures (and then check `is_terminated()`)?

Ruben2424: > Thanks for the feedback. My understanding on this is different, but maybe I read it wrong.
The outer future here (and below too) is a Fuse (see https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.fuse) which internally already takes care of shunting subsequent polls to Pending:
https://github.com/rust-lang/futures-rs/blob/89c06f5e89b8499102a198736ccc859949e049a3/futures-util/src/future/future/fuse.rs#L82-L90
Did I misunderstand something here?
Or maybe are you suggesting to avoid introducing two Fuse struct wrappers here and directly try to implement the FusedFuture trait on the two existing underlying futures (and then check is_terminated())?


I agree. As far as i understand the `Future` implementation of `Fuse`, it is already checked in the `poll` method.

src/proto/h2/client.rs
```diff
@@ -245,10 +245,11 @@ pin_project! {
         T: Unpin,
     {
         #[pin]
-        select: Select<ConnMapErr<T, B>, StreamFuture<Receiver<Never>>>,
```
Ruben2424, line 248: What speaks against `select: Select<Fuse<ConnMapErr<T, B>>, Fuse<StreamFuture<Receiver<Never>>>>,

Ruben2424 (commented):
Thanks for the PR!
All in all, I'm not sure if this truly addresses the underlying issue or merely treats the symptoms.
As of my understanding the future should never be polled twice.

src/proto/h2/client.rs
lucab: I pushed a commit on top which gets rid of additional `Fuse` structs and directly uses the `FusedFuture` trait (also found out that `StreamFuture` already implements that).

The two versions should be functionally equivalent, so I'm happy to keep whatever form is preferred by the maintainers.

src/proto/h2/client.rs
lucab: Uhm, not sure without further context, but I think this won't be enough and the actual fix has to be done in the `poll()` code.

Right now, if the `ConnTask->Select->ConnMap` side returns `Ready` first, it is theoretically possible for a subsequent/spurious poll to go through `ConnTask->Select` again, which would then immediately panic.

This PR rewrites the `ConnTask::poll()` logic (to avoid such execution flow) and also drops the stored `Select` (to avoid part of the issue, but also because it doesn't really seems useful in this context).

lucab:
Alternatively I think we could also keep the `Select` (which already implements `FusedFuture`) and perform the `is_terminated()` check on it before polling.
But honestly I think that getting rid of it makes the `ConnTask::poll()` logic simpler to follow.

src/proto/h2/client.rs
Ruben2424: Oh yes, you are right, the select is panicing not the futures in the select.

seanmonstar (approved):
Makes sense, thanks for catching and providing the fix! <3

PR #3294, opened 2023-08-22
fix(server): Respect Expect header only when http proto > 1.0

owarai:
1. fix server Expect handling(from the perspective of server correctness)
    Refer to: https://www.rfc-editor.org/rfc/rfc9110.html#name-expect
    > A server that receives a 100-continue expectation in an HTTP/1.0 request MUST ignore that expectation
3. adds HTTP/1.0 Expect 100-continue test


seanmonstar (approved):
Thanks!

PR #3295, opened 2023-08-24
ReasonPhrase Polish

seanmonstar:
This removes the `from_bytes_unchecked()` method, since it's utility seems dubious (we can always add back if it's badly needed). It also makes `from_static` a const fn.


acfoltzer (approved):
Looks fine to me; we are not currently using this method and, like you say, it can be easily added back if we're missing an anticipated use case

PR #3303, opened 2023-09-03
Add graceful_shutdown example.

aaronriekenberg:
I wanted to have graceful shutdown after a timeout in my server application.  There is not an existing example that does exactly this.

This is mostly the existing `hello.rs` example with graceful shutdown added.

Thanks!


seanmonstar (approved):
Wow, this is exceptionally well explained, thank you so much!

It'd probably be great if the graceful shutdown guide on https://hyper.rs/guides pointed to this as a fuller example.

PR #3305, opened 2023-09-05
Start http1 header read timeout even when 0 bytes are read

jeromegn:
As mentioned in https://github.com/hyperium/hyper/pull/2675#issuecomment-1666309018, the timeout didn't apply if no bytes were written yet.

This moves the start of the timeout's polling before the check for an empty buffer.


seanmonstar:
FWIW, this sounds very similar to this other PR, which I paused because of the follow comment: https://github.com/hyperium/hyper/pull/3185#issuecomment-1489170787.

seanmonstar (approved):
Yea, this is much wiser. If browsers cannot handle this with preconnects, they should stop looking like bad actors.

PR #3307, opened 2023-09-08
chore(dependencies): Remove unused num_cpus and tokio-util dependencies

tottoto:
These dependencies seem not to be used.


seanmonstar (approved):
Good catch, thanks!

PR #3312, opened 2023-09-11
feat(error): change `Display for Error` to only print top error

### LINKED ISSUE #2844: Figure out hyper::Error Display including source chain
The current `Display` output of `Error` [doesn't match what many people think it should](https://github.com/hyperium/hyper/blob/master/docs/ROADMAP.md#errors), namely that it prints the error chain. We need to either:

- Change `hyper::Error` to not print the source as well.
- Influence the errors-wg to allow printing the chain in `Display`, like proposed in https://github.com/seanmonstar/errors/issues/1.

seanmonstar:
hyper's `Error` used to print the error source automatically, preferring to provide a better default for users who do not know about `Report`. But, to fit better with the wider ecosystem, this changes the format to only print the hyper `Error` itself, and not its source.

Closes #2844

BREAKING CHANGE: The format no longer prints the error chain. Be sure to
  check if you are logging errors directly.

  The `Error::message()` method is removed, it is no longer needed.

  The `Error::into_cause()` method is removed.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,6 +8,18 @@ pub type Result<T> = std::result::Result<T, Error>;
 type Cause = Box<dyn StdError + Send + Sync>;
 
 /// Represents errors that can occur handling HTTP streams.
+///
+/// # Formatting
+///
+/// The `Display` implementation of this type will only print the details of
+/// this level of error, even though it may have been caused by another error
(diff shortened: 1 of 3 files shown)


faern:
Finally, amazing! Error printing where hyper is now part of the error stack will make so much more sense and blend in so much better from now on. Good decision :+1:

Issue #3067, opened 2022-12-01
Polish the API docs


seanmonstar:
This is issue about what appears at https://docs.rs/hyper. As a general rule, we should incrementally improve them with the following in mind:

- Types should have a description of what they are used for, not just a repeat of the type name.
- Methods should almost always include a code example.
- Modules should explain concepts of what its types are meant to be used for, and how to combine those types with others.
- If a question comes up more than once from a user, it should be somewhere in the API docs.
	- If it is there, and people still ask, it's probably in the wrong place, or not written clearly.

A248:
Hi, maybe it would be useful to mention the repository https://github.com/hyperium/hyperium.github.io, for the sake of future contributors who come to this issue looking to contribute to the documentation.

seanmonstar:
@A248 Ah, what I meant by "API docs" is those that are within the source code here, that get rendered at docs.rs/hyper. The website repository is the subject of #3068.

seanmonstar:
I'm going through the currently rendered docs (of rc4) and writing down some notes in the top comment about what needs to be improved. By making them separate items in a task list, it should be easy to convert any single one into an issue. If you'd like to fix any, start a pull request mentioning it, and I'll click the button to convert the task into an issue.

PR #3325, opened 2023-09-24
docs(server): polish server connection docs

oddgrd:
This PR aims to address some of the tasks in https://github.com/hyperium/hyper/issues/3067, specifically:

> conn module: mention hyper-util providing an AutoConnection and builder

>  struct conn::http1::Connection: Explain that this is returned from Builder::serve_connection, and that it is essential an impl Future that must be polled (or awaited) for anything to happen.

The change in service is to make it an automatic link, which was suggested by cargo doc.

The changes have been tested by running `cargo doc --features "server,http1,client,http2" --open` and inspecting the docs.

Feedback on phrasing or if anything should be expanded on is most welcome!

### DIFF
--- a/src/server/conn/mod.rs
+++ b/src/server/conn/mod.rs
@@ -5,8 +5,14 @@
 //! are not handled at this level. This module provides the building blocks to
 //! customize those things externally.
 //!
-//! This module is split by HTTP version. Both work similarly, but do have
-//! specific options on each builder.
+//! This module is split by HTTP version, providing a connection builder for
+//! each. They work similarly, but they each have specific options.
+//!
+//! If your server needs to support both versions, an auto connection builder is
+//! provided in the [`hyper-util`](https://github.com/hyperium/hyper-util/tree/master)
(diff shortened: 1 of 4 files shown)


src/server/conn/mod.rs
```diff
@@ -5,8 +5,14 @@
...
-//! This module is split by HTTP version. Both work similarly, but do have
-//! specific options on each builder.
+//! This module is split by HTTP version, providing a connection builder for
+//! each. They work similarly, but they each have specific options.
+//!
+//! If your server needs to support both versions, an auto-connection builder is
+//! provided in the [`hyper-util`](https://github.com/hyperium/hyper-util/tree/master)
+//! crate. This builder wraps the HTTP/1 and HTTP/2 connection builders from this
+//! module, allowing you to set configuration for both. The builder will then check
+//! the version of the incoming connection and serve it accordingly.
```
oddgrd, line 15: This should link to the hyper-util docs when those are generated. Perhaps we should open an issue for creating an example for auto-connection in hyper-util as well?

seanmonstar (approved):

Issue #2874, opened 2022-05-26
Meta: Audit public dependencies


seanmonstar:
As part of reaching hyper 1.0, we need to audit the "public" dependencies to make sure that:

- We actually mean for that dependency to be exposed.
- The dependency is stable.

This list might be incomplete, but currently has:

- [x] `http` must reach 1.0
- [x] `http-body` must reach 1.0
- [x] #3319
- [x] `bytes` must be 1.0
- [x] Tokio 1.0 or remove for `hyper::io` traits

hawkw:
I think the `Tokio` 1.0 checkmark should be checked off, given that Tokio's `io` traits are stable :)

PR #3326, opened 2023-09-27
feat(upgrade): introduce tracing as an optional unstable feature

### LINKED ISSUE #3319: Make `tracing` an unstable feature

RamziA961:
This change allow users to opt out of tracing via the `tracing` crate by adding tracing as an optional feature. This change is part of the effort, outlined in #2874, to reach hyper 1.0.

Closes #3319
BREAKING CHANGES: tracing is disabled by default and requires users to opt in to revert to previous behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,8 +91,8 @@ tracing = ["dep:tracing"]
 nightly = []
 
 [package.metadata.docs.rs]
-features = ["ffi", "full"]
-rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi"]
+features = ["ffi", "full", "tracing"]
+rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi", "--cfg", "hyper_unstable_tracing"]
 
 [package.metadata.playground]
 features = ["full"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ httpdate = "1.0"
 httparse = "1.8"
 h2 = { version = "0.3.9", optional = true }
 itoa = "1"
-tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project-lite = "0.2.4"
 tokio = { version = "1", features = ["sync"] }
 want = "0.3"
@@ -38,6 +37,7 @@ want = "0.3"
 # Optional
 
 libc = { version = "0.2", optional = true }
+tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
 
 [dev-dependencies]
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
@@ -83,6 +83,9 @@ server = []
 
 # C-API support (currently unstable (no semver))
 ffi = ["dep:libc", "dep:http-body-util"]
+
+# Utilize tracing (currently unstable)
+tracing = ["dep:tracing"]
 
 # internal features used in CI
 nightly = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,6 +68,8 @@ pub use crate::error::{Error, Result};
 #[macro_use]
 mod cfg;
 #[macro_use]
+mod trace;
+#[macro_use]
 mod common;
 pub mod body;
 mod error;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,7 +69,6 @@ pub use crate::error::{Error, Result};
 mod cfg;
 
 #[macro_use]
-#[cfg_attr(docsrs, doc(all(feature = "tracing"), hyper_unstable_tracing))]
 mod trace;
 
 #[macro_use]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,7 +51,20 @@
 //! - `server`: Enables the HTTP `server`.
 //!
 //! [feature flags]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section
-
+//!
+//! # Unstable Features
+//! hyper includes a set of unstable optional features that can be enabled through the use of a
+//! feature flag and a [configuration flag].
+//!
+//! The following is a list of feature flags and their corresponding `RUSTFLAG`:
+//! - `ffi`: Enables C API for hyper `hyper_unstable_ffi`.
+//! - `tracing`: Enables debug logging with `hyper_unstable_tracing`.
+//!
+//! Enabling an unstable feature is possible with the following `cargo` command, as of version `1.64.0`:
+//! ```notrust
+//! RUSTFLAGS="--cfg hyper_unstable_tracing" cargo rustc --features client,http1,http2,tracing --crate-type cdylib
+//!```
+//! [configuration flag]: https://doc.rust-lang.org/reference/conditional-compilation.html
 #[doc(hidden)]
 pub use http;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,8 +67,11 @@ pub use crate::error::{Error, Result};
 
 #[macro_use]
 mod cfg;
+
 #[macro_use]
+#[cfg_attr(docsrs, doc(all(feature = "tracing"), hyper_unstable_tracing))]
 mod trace;
+
 #[macro_use]
 mod common;
 pub mod body;
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -243,7 +243,7 @@ where
             *this.is_terminated = true;
         }
         polled.map_err(|_e| {
-            debug!("connection error: {}", _e)
+            debug!("connection error: {}", _e);
         })
     }
 }
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -12,8 +12,6 @@ use h2::client::{Builder, Connection, SendRequest};
 use h2::SendStream;
 use http::{Method, StatusCode};
 use pin_project_lite::pin_project;
-#[cfg(feature = "tracing")]
-use tracing::{debug, trace, warn};
 
 use super::ping::{Ponger, Recorder};
 use super::{ping, H2Upgraded, PipeToSendStream, SendBuf};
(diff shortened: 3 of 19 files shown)


seanmonstar:
Thanks for taking this on!

We need to make it unstable similar to the FFI feature, with a check in `lib.rs` that a `cfg(hyper_unstable_tracing)` was also passed, or trigger a compiler error. And then document it in the unstable features section of the same file.

Separately, one possible way I think this could be done is making an internal module with trace and debug macros, and inside those out the cfg checks. It's not a big deal, it just might be easier.

RamziA961:
Thanks for the feedback. An internal module would definitely be a cleaner approach, and would help with unused variable warnings. I'll get started on that.

RamziA961:
I think a quick overview of the changes I made might be helpful in review the PR, so here goes:

 - I introduced a module called `trace`, which essentially exposes macros that act as wrappers around `tracing`'s macros but handle conditional compilation internally. The caveat is `tracing` has `span` macros aren't so easily replaceable. Their uses throughout the codebase are sparse and all uses comprised creating  a `Span` and the calling the method `enter()`. Both the `Span` and `Entered` instances need to be bound to a variable so that they are not dropped. The only approach that I could think of the preserves the conciseness of the original syntax is the following but it is greatly restrictive:

```rust
macro_rules! debug_span {
    ($($arg:tt)*) => {
        #[cfg(feature = "tracing")]
        {
            #[cfg(not(hyper_unstable_tracing))]
            compile_error!(
                "\
                The `tracing` feature is unstable, and requires the \
                `RUSTFLAGS='--cfg hyper_unstable_tracing'` environment variable to be set.\
            "
            );
            let span = tracing::debug_span!($($arg)+);
            let _ = span.enter();
        }
    }
}
```

- I replaced uses of `tracing`'s macros with the new internal `trace` module's macro. In most cases, it simply required deleting `use tracing::{x, y, z}`, except when the macros were used as part of an expression, e.g., `something.map_err(|e| tracing::debug!("{e:?}"))`.

- Some variables had to be prefix with an underscore since unused variables warnings are emitted when compiling with `tracing` disabled.

- To ensure that the codes compiles when the `tracing` feature is off, the `trace` module must always be declared in `lib` so that the compiler can expand invocations of its macros into nothing.

- I introduced a new feature `tracing` and a new cfg flag `--hyper_unstable_tracing`. The feature must be enabled and the flag must be passed to `rustc` for tracing to be enabled, otherwise a compiler error is thrown in similar fashion to `ffi`. This is the cause of the failing CI test.

- Documentation for the module, feature, and cfg flag are provided.

I think that covers it.

src/lib.rs
```diff
@@ -67,6 +67,11 @@ pub use crate::error::{Error, Result};
 
 #[macro_use]
 mod cfg;
+
+#[macro_use]
+#[cfg_attr(docsrs, doc(all(feature = "tracing"), hyper_unstable_tracing))]
```
seanmonstar, line 72: Is this attribute required? Since it's not public, the module won't be rendered on docs.rs.

src/trace.rs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            #[cfg(not(hyper_unstable_tracing))]
+            compile_error!(
+                "\
+                The `tracing` feature is unstable, and requires the \
+                `RUSTFLAGS='--cfg hyper_unstable_tracing'` environment variable to be set.\
+            "
+            );
+            let span = tracing::debug_span!($($arg)+);
+            let _ = span.enter();
```
seanmonstar, line 60: I believe this will drop the enter guard immediately, since the block ends on the next line. You could change this to return `span.entered()`, and then the usage would be `let _s = debug_span!(blah);`.

src/trace.rs
```diff
@@ -0,0 +1,212 @@
...
+//! The [`trace`][crate::trace] module is an internal module that contains wrapper macros encapsulating
+//! [`tracing`][tracing]'s macros. These macros allow for conditional expansion of
+//! [`tracing`][tracing] macros during compilation when the `tracing` feature is enabled, or trimming
+//! them when the feature is disabled, all in a concise manner.
+//!
+//! The macros from the [`trace`][crate::trace] module are declared by default and can be used throughout the
+//! crate. However, as the contents of these macros are conditionally compiled, they are effectively trimmed
+//! during inline expansion when the `tracing` feature is disabled.
+//!
+//! # Unstable
```
seanmonstar, line 16: This makes me realize that in `lib.rs`, there isn't a section about unstable features. We should move the essence of this there (without describing what the internals of this module does).

seanmonstar (commented):
Thanks so much, the work is very thorough, and your write up is spectacular.

Cargo.toml
```diff
@@ -29,14 +29,14 @@ http-body-util = { version = "=0.1.0-rc.3", optional = true }
...
 itoa = "1"
-tracing = { version = "0.1", default-features = false, features = ["std"] }
 pin-project-lite = "0.2.4"
 tokio = { version = "1", features = ["sync"] }
 
 # Optional
 
 httpdate = { version = "1.0", optional = true }
 libc = { version = "0.2", optional = true }
+tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
```
hawkw, line 39: is `tracing`'s "std" feature actually required here? i think we should be fine without enabling it (although it doesn't matter, since hyper doesn't support no-std...)

src/trace.rs
hawkw: yeah, that's correct. we should either have the macro return an un-entered span that's entered at the callsite, or change this to 
```suggestion
            let _enter = span.enter();
```
or similar.

src/trace.rs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            #[cfg(not(hyper_unstable_tracing))]
+            compile_error!(
+                "\
+                The `tracing` feature is unstable, and requires the \
+                `RUSTFLAGS='--cfg hyper_unstable_tracing'` environment variable to be set.\
+            "
+            );
+            let span = tracing::error_span!($($arg)+);
+            let _ = span.enter();
```
hawkw, line 93: this also drops the span immediately and will need to be changed as discussed above.

src/trace.rs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            #[cfg(not(hyper_unstable_tracing))]
+            compile_error!(
+                "\
+                The `tracing` feature is unstable, and requires the \
+                `RUSTFLAGS='--cfg hyper_unstable_tracing'` environment variable to be set.\
+            "
+            );
+            let span = tracing::info_span!($($arg)+);
+            let _ = span.enter();
```
hawkw, line 126: same issue as above.

src/trace.rs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            #[cfg(not(hyper_unstable_tracing))]
+            compile_error!(
+                "\
+                The `tracing` feature is unstable, and requires the \
+                `RUSTFLAGS='--cfg hyper_unstable_tracing'` environment variable to be set.\
+            "
+            );
+            let span = tracing::trace_span!($($arg)+);
+            let _ = span.enter();
```
hawkw, line 159: and here

src/proto/h2/client.rs
```diff
@@ -243,7 +242,9 @@ where
         if polled.is_ready() {
             *this.is_terminated = true;
         }
-        polled.map_err(|e| debug!("connection error: {}", e))
+        polled.map_err(|_e| {
+            debug!("connection error: {}", _e);
```
hawkw, line 246: while we're here, we may want to change these events to record the error as a structured field, like
```suggestion
            debug!(error = %_e, "connection error");
```
this isn't a big deal, but it could be worth doing, and we're changing all the existing tracing macros anyway...

src/trace.rs
```diff
@@ -0,0 +1,212 @@
...
+    ($($arg:tt)+) => {
+        #[cfg(feature = "tracing")]
+        {
+            #[cfg(not(hyper_unstable_tracing))]
+            compile_error!(
+                "\
+                The `tracing` feature is unstable, and requires the \
+                `RUSTFLAGS='--cfg hyper_unstable_tracing'` environment variable to be set.\
+            "
+            );
```
hawkw, line 42: this will result in the same compile error being emitted for *every* use of a tracing macro, if the unstable cfg is unset. this means there will be a giant pile of the same error in the rustc output. i wonder if it would be nicer to instead just stick the compile error in this module, so that we generate a single error if the feature flag is enabled without the unstable cfg?

src/upgrade.rs
```diff
@@ -233,6 +231,7 @@ impl fmt::Debug for OnUpgrade {
 #[cfg(any(feature = "http1", feature = "http2"))]
 impl Pending {
     pub(super) fn fulfill(self, upgraded: Upgraded) {
+        #[cfg(any(feature = "http1", feature = "http2"))]
```
hawkw, line 234: why was it necessary to add these additional cfg attributes? the entire impl block is only enabled with `#[cfg(any(feature = "http1", feature = "http2"))]`, so AFAICT, this shouldn't be necessary?

src/trace.rs
RamziA961: The issue with `let _s = debug_span!(blah);` is that the compiler does not support removing an expression. When tracing is off, this effectively becomes `let _s = ;`. 

Similarly, `let _enter  = span.enter()` is also dropped at the end of the block.

The following seems to work:
```rust
use tracing_subscriber::prelude::*;

macro_rules! test {
    ($($arg:tt)+) => {
        #[cfg(debug_assertions)]
        { 
            let _sp = tracing::error_span!($($arg)+);
            _sp.entered()
        }
    };
}

fn main() {
    let stdout = tracing_subscriber::fmt::layer().pretty();

    tracing_subscriber::registry()
        .with(
            stdout
                .with_span_events(tracing_subscriber::fmt::format::FmtSpan::EXIT)
        )
        .init();
     
    let _e = { test!("asd", value = 32 ) };
    
    tracing::error!("test one");
    tracing::error!("test two");
}
```

Should I move forward with this approach?

src/upgrade.rs
RamziA961: Good catch. Thank you for pointing this out.

src/trace.rs
RamziA961: I agree, I will make the change.

seanmonstar:
All the things look good to me, thanks! Good job! You could include the change to the CI features check in this PR, so that it merges with the relevant code. Also, looks like there's merge conflicts.

seanmonstar (approved):

PR #3327, opened 2023-09-28
chore(dependencies): reduce default dependencies

tottoto:
`httpdate` and `want` are used in some combinations of features. Although Cargo does not provide such a combined features configuration, the scopes could be reduced.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,18 +26,18 @@ futures-util = { version = "0.3", default-features = false }
 http = "0.2"
 http-body = "=1.0.0-rc.2"
 http-body-util = { version = "=0.1.0-rc.3", optional = true }
-httpdate = "1.0"
 httparse = "1.8"
 h2 = { version = "0.3.9", optional = true }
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks!

(Now I wonder, is there a CI job we could setup to catch these?)

tottoto:
[`cargo-udeps`](https://github.com/est31/cargo-udeps) can handle feature aware checking. For `hyper`'s default and full feature, it can be checked with `cargo-udeps` (like [`tonic` project](https://github.com/hyperium/tonic/blob/b3fca19104bf001d3a3dac74221b7c9bede13cf1/.github/workflows/CI.yml#L63), in this ci, tonic checks each feature with `cargo-hack`). Regarding the another features, as `hyper` uses some crates in the combination of multiple feature, it seem not to be able to achieve minimal dependencies for any possible features combination.

tottoto:
Opend #3334 to check unused dependencies.

PR #3334, opened 2023-10-03
chore(ci): check unused dependencies

tottoto:
Adds a CI job to check unused dependencies.


tottoto:
In my local test, `itoa` seems not to be detected although `httpdate` and `want` are detected.

seanmonstar (approved):

PR #3335, opened 2023-10-04
chore(common): replace Never with Infallible

tottoto:
`hyper` might be able to use `std::convert::Infallible` to describe the error which never happens, instead of `hyper::common::never::Never`.


seanmonstar (approved):
Nice clean up, thanks!

PR #3344, opened 2023-10-12
refactor: resolve unused warning

tottoto:
Adds `cfg` feature attributes to resolve unused warnings, which occur in each features and the combinations, and ci config to check `unused_imports` and `dead_code`.


seanmonstar (approved):
Amazing!

PR #3346, opened 2023-10-13
refactor(lib): remove common re-export

tottoto:
There have been mix of the way of using these types, both re-exported and direct using. This pull request proposes to use them directly consistently as using re-exported ones consistently seems to be difficult and using them directly seems simple.


seanmonstar (approved):
Impressive!

PR #3348, opened 2023-10-15
refactor(lib): resolve FIXME messages

aliu:
Remove outdated FIXME comments, and resolve FIXME regarding usage of `MaybeUninit`.

Since hyper's MSRV is higher now, we can use some new stabilized apis. I also removed some `unsafe` blocks related to `MaybeUninit` usage.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -135,18 +135,13 @@ impl Http1Transaction for Server {
         let len;
         let headers_len;
 
-        // Unsafe: both headers_indices and headers are using uninitialized memory,
+        // Both headers_indices and headers are using uninitialized memory,
         // but we *never* read any of it until after httparse has assigned
         // values into it. By not zeroing out the stack memory, this saves
         // a good ~5% on pipeline benchmarks.
-        let mut headers_indices: [MaybeUninit<HeaderIndices>; MAX_HEADERS] = unsafe {
-            // SAFETY: We can go safely from MaybeUninit array to array of MaybeUninit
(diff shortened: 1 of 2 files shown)


src/proto/h1/role.rs
```diff
@@ -135,18 +135,13 @@ impl Http1Transaction for Server {
...
-        let mut headers_indices: [MaybeUninit<HeaderIndices>; MAX_HEADERS] = unsafe {
-            // SAFETY: We can go safely from MaybeUninit array to array of MaybeUninit
-            MaybeUninit::uninit().assume_init()
-        };
+        let mut headers_indices = [MaybeUninit::<HeaderIndices>::uninit(); MAX_HEADERS];
         {
-            /* SAFETY: it is safe to go from MaybeUninit array to array of MaybeUninit */
-            let mut headers: [MaybeUninit<httparse::Header<'_>>; MAX_HEADERS] =
-                unsafe { MaybeUninit::uninit().assume_init() };
+            let mut headers = [MaybeUninit::<httparse::Header<'_>>::uninit(); MAX_HEADERS];
```
seanmonstar, line 144: Does this result in the same generated code? Or at least same performance? I mean, logically it makes sense. But it doesn't initialize an array to then declare the stuff inside is uninitialized, right?

aliu: The generated code in debug-mode is slightly different, but in release mode it's identical. Here is a link to the assembly: https://godbolt.org/z/6rvbz8qWd

seanmonstar (approved):

Issue #3195, opened 2023-04-06
no method named `send_replace` found for struct `Arc<Sender<Option<Connected>>>`


jauderho:
**Version**
hyper v0.14.25

**Platform**
Linux 6.2.0-19-generic #19-Ubuntu SMP PREEMPT_DYNAMIC Sat Mar 25 10:22:33 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Description**
While trying to build rustybgp using the following command:

```
RUSTFLAGS="-C link-arg=-fuse-ld=mold" cargo build --release
```

Build stopped with the error message:

```
   Compiling hyper v0.14.25
error[E0599]: no method named `send_replace` found for struct `Arc<Sender<Option<Connected>>>` in the current scope
   --> /home/jauderho/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.25/src/client/connect/mod.rs:225:17
    |
225 |         self.tx.send_replace(Some(connected.clone()));
    |                 ^^^^^^^^^^^^ method not found in `Arc<Sender<Option<Connected>>>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `hyper` due to previous error

```

Looks like it came from this commit: https://github.com/hyperium/hyper/pull/3144/files

seanmonstar:
Running `cargo update` should fix it for you. Likely that method is in a newer version of Tokio than you have in your directory.

palfrey:
Upgrading minimum tokio to 1.13 as per https://github.com/hyperium/hyper/pull/3345 fixes this one

PR #3350, opened 2023-10-17
chore(dependencies): relax minimum tokio version to 1

tottoto:
This pull request reverts #3345. #3195 is about 0.14.25 and `tokio::sync::watch::Sender::send_replace` seems not to be used in this branch.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,6 +26,7 @@ jobs:
       - doc
       - check-external-types
       - udeps
+      - minimal-versions
     steps:
       - run: exit 0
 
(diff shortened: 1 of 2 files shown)


hawkw:
AFAICT, [hyper's CI](https://github.com/hyperium/hyper/blob/8ebe25cab5953da2cb5163e58c1561b1a2db0b88/.github/workflows/CI.yml) does not currently check that Hyper compiles with [`-Z minimal-versions`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#minimal-versions). If we're going to use very permissive dependencies such as this one, we should be checking that we can build with minimal version selection --- it will ensure that we are not accidentally using newer APIs that don't exist in the minimum allowed version of the dependency.

tottoto:
Rebased. And dropped the crates' version fix which is not needed now.

seanmonstar (approved):

PR #3357, opened 2023-10-19
chore(ci): update to cargo-check-external-types 0.1.9

tottoto:
Updates to `cargo-check-external-types` 0.1.9, and moves its config to Cargo metadata introduced at this version.


Cargo.toml
```diff
@@ -97,6 +97,23 @@ rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi", "--cfg", "hype
 [package.metadata.playground]
 features = ["full"]
 
+[package.metadata.cargo_check_external_types]
```
seanmonstar, line 100: Is this change required? Did they drop support for a separate file? I kinda like that this stuff is over in the CI folder, and not bothering the files that get published...

tottoto: It is not required. `cargo-check-external-types` supports a separate config file.

tottoto: Fixed.

seanmonstar (approved):

PR #3366, opened 2023-10-20
docs(client,server): added note about stability of Builder struct default options

newmanifold:
Completes #3336

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -97,6 +97,9 @@ where
 /// A builder to configure an HTTP connection.
 ///
 /// After setting options, the builder is used to create a handshake future.
+/// 
+/// **Note**: The default values of options are *not considered stable*. They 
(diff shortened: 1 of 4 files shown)


src/client/conn/http1.rs
```diff
@@ -97,6 +97,9 @@ where
 /// A builder to configure an HTTP connection.
 ///
 /// After setting options, the builder is used to create a handshake future.
+///
+/// **Note**: The default values of options are *not considered stable*. They
+/// are subject to change at any time
```
seanmonstar, line 102: Could you add a period to the end of the sentences?

seanmonstar (commented):
Thanks so much!

src/client/conn/http1.rs
newmanifold: Done.

seanmonstar (approved):

PR #3370, opened 2023-10-23
FFI: fix deadlock in hyper_executor::poll_next

### LINKED ISSUE #3369: C API deadlock in hyper_executor::poll_next
Version: current master, dd638b5b34225d2c5ad0bd01de0ecf738f9a0e12
Platform: Linux 6.2.0-34-generic #34-Ubuntu SMP PREEMPT_DYNAMIC Mon Sep  4 13:06:55 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

In https://github.com/curl/curl/issues/11203#issuecomment-1733313614, @marlonbaeten reported that removing `hyper_clientconn_free(client);` from curl's c-hyper code caused curl to never send the request.

I did a little poking around and I believe what's happening is a deadlock. I'm not sure yet why freeing the client triggers the deadlock, but I can see in the Hyper code where it happens.

In hyper_executor::poll_next, `self.driver` is locked and then `drain_queue` is called :

https://github.com/hyperium/hyper/blob/dd638b5b34225d2c5ad0bd01de0ecf738f9a0e12/src/ffi/task.rs#L129-L134

`drain_queue` then tries to lock `self.driver` a second time:

https://github.com/hyperium/hyper/blob/dd638b5b34225d2c5ad0bd01de0ecf738f9a0e12/src/ffi/task.rs#L151-L156

The docs for [Mutex::lock](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock) say:

> The exact behavior on locking a mutex in the thread which already holds the lock is left unspecified. However, this function will not return on the second call (it might panic or deadlock, for example).

Steps to reproduce:

Build Hyper's FFI library using this patch to add some println debugging lines:

```patch
diff --git a/src/ffi/task.rs b/src/ffi/task.rs
index e75dfc1a..e868f39d 100644
--- a/src/ffi/task.rs
+++ b/src/ffi/task.rs
@@ -131,6 +131,7 @@ impl hyper_executor {
                 Poll::Pending => {
                     // Check if any of the pending tasks tried to spawn
                     // some new tasks. If so, drain into the driver and loop.
+                    eprintln!("draining queue??");
                     if self.drain_queue() {
                         continue;
                     }
@@ -149,11 +150,14 @@ impl hyper_executor {
 
     fn drain_queue(&self) -> bool {
         let mut queue = self.spawn_queue.lock().unwrap();
+        eprintln!("locked spawn queue :D");
         if queue.is_empty() {
             return false;
         }
 
+        eprintln!("trying to lock driver. this might deadlock :/");
         let driver = self.driver.lock().unwrap();
+        eprintln!("locked driver! :D :D :D");
 
         for task in queue.drain(..) {
             driver.push(task);
```

Build curl against that version of Hyper's FFI library, using this patch against curl:

```patch
diff --git a/lib/c-hyper.c b/lib/c-hyper.c
index 5726ff1cc..7a5e764c8 100644
--- a/lib/c-hyper.c
+++ b/lib/c-hyper.c
@@ -1214,9 +1214,6 @@ CURLcode Curl_http(struct Curl_easy *data, bool *done)
   }
   sendtask = NULL; /* ownership passed on */
 
-  hyper_clientconn_free(client);
-  client = NULL;
-
   if((httpreq == HTTPREQ_GET) || (httpreq == HTTPREQ_HEAD)) {
     /* HTTP GET/HEAD download */
     Curl_pgrsSetUploadSize(data, 0); /* nothing */
```

Observe this output, followed by a hang:

```
$ ~/curl/src/curl https://example.com/
locked spawn queue :D
trying to lock driver. this might deadlock :/
locked driver! :D :D :D
locked spawn queue :D
trying to lock driver. this might deadlock :/
locked driver! :D :D :D
draining queue??
locked spawn queue :D
trying to lock driver. this might deadlock :/
```

jsha:
poll_next locks the driver, and also calls drain_queue while holding that lock. Since drain_queue locks the driver too, that results in a deadlock.

To fix, unlock the driver before calling drain_queue.

Fixes #3369.

### DIFF
--- a/src/ffi/task.rs
+++ b/src/ffi/task.rs
@@ -126,27 +126,35 @@ impl hyper_executor {
         let mut cx = Context::from_waker(&waker);
 
         loop {
-            match Pin::new(&mut *self.driver.lock().unwrap()).poll_next(&mut cx) {
-                Poll::Ready(val) => return val,
-                Poll::Pending => {
-                    // Check if any of the pending tasks tried to spawn
-                    // some new tasks. If so, drain into the driver and loop.
-                    if self.drain_queue() {
-                        continue;
-                    }
-
-                    // If the driver called `wake` while we were polling,
-                    // we should poll again immediately!
-                    if self.is_woken.0.swap(false, Ordering::SeqCst) {
-                        continue;
-                    }
-
-                    return None;
-                }
+            {
+                // Scope the lock on the driver to ensure it is dropped before
+                // calling drain_queue below.
+                let mut driver = self.driver.lock().unwrap();
+                match Pin::new(&mut *driver).poll_next(&mut cx) {
+                    Poll::Ready(val) => return val,
+                    Poll::Pending => {}
+                };
             }
+
+            // poll_next returned Pending.
+            // Check if any of the pending tasks tried to spawn
+            // some new tasks. If so, drain into the driver and loop.
+            if self.drain_queue() {
+                continue;
+            }
+
+            // If the driver called `wake` while we were polling,
+            // we should poll again immediately!
+            if self.is_woken.0.swap(false, Ordering::SeqCst) {
+                continue;
+            }
+
+            return None;
         }
     }
 
+    /// drain_queue locks both self.spawn_queue and self.driver, so it requires
+    /// that neither of them be locked already.
     fn drain_queue(&self) -> bool {
         let mut queue = self.spawn_queue.lock().unwrap();
         if queue.is_empty() {


seanmonstar:
The fix looks great, thanks! Seems CI is grumpy about a missing comma, of course.

src/ffi/task.rs
```diff
@@ -126,27 +126,35 @@ impl hyper_executor {
...
-                    return None;
-                }
+            {
+                // Scope the lock on the driver to ensure it is dropped before
+                // calling drain_queue below.
+                let mut driver = self.driver.lock().unwrap();
+                match Pin::new(&mut *driver).poll_next(&mut cx) {
+                    Poll::Ready(val) => return val,
+                    Poll::Pending => {}
+                };
```
hawkw, line 136: nit, take it or leave it: this could be 
```suggestion
                if let Poll::Ready(val) =  Pin::new(&mut *driver).poll_next(&mut cx) {
                    return val;
                }
```

hawkw (approved):

src/ffi/task.rs
jsha: Thanks! Yeah, I thought about doing it this way but I like making it really explicit that the only other possibility for the match is "pending", since we have following code that assumes the result was pending.

PR #3371, opened 2023-10-23
feat(client): allow `!Send` IO with HTTP/1 client

### LINKED ISSUE #3363: Allow client IO to be !Send
Just like `server::conn::http1::Connection` can work with an IO type that isn't `Send`, because it has a `with_upgrades()` variant, we should make it so the client `http1::Connection` type doesn't _require_ `Send`. This would be a breaking change, and so should happen before tagging 1.0.

I believe doing this is probably somewhat involved, let me lay out what I think needs to be done. If any step is unclear, or another way seems better, please comment :)

- [ ] Remove `Send` from the `T: Read + Write` bounds for `conn::http1::Connection`
- [ ] Provide a `UpgradeableConnection` in similar style to what is in `server`.
- [ ] Update the `single_threaded.rs` example to use a `!Send` IO type for the client, to prove it works.

seanmonstar:
This removes the requirement of the IO type from being `Send` for the HTTP/1 client connection. To do so, the ability to perform `hyper::upgrade`s had to be moved to a separate type which does require the `Send` bound. This mirrors how the server types do it.

The `Connection` type now has a `with_upgrades()` method to convert.

Closes #3363

BREAKING CHANGE: If you use client HTTP/1 upgrades, you must call
  `Connection::with_upgrades()` to still work the same.

### DIFF
--- a/examples/single_threaded.rs
+++ b/examples/single_threaded.rs
@@ -138,7 +138,7 @@ async fn http1_server() -> Result<(), Box<dyn std::error::Error>> {
     loop {
         let (stream, _) = listener.accept().await?;
 
-        let io = TokioIo::new(stream);
+        let io = IOTypeNotSend::new(TokioIo::new(stream));
 
         let cnt = counter.clone();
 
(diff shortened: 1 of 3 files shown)


mstyura:
Further relaxation of Send on IO during upgrade is not possible due to IO type put into http request Extension, right? Is it theoretically possible to refactor code to avoid usage of request extension so the IO can be !Send and used with with request upgrades? Current bounds a bit confusing because at the same time h2 upgrades does not require IO to be Send.

PR #3377, opened 2023-10-29
chore(dependencies): remove libc crate from dependencies

tottoto:
`hyper` uses `libc` in `ffi` feature for the C language's type aliases. As `hyper`'s `ffi` feature is an unstable feature, `hyper` could use these types via `std::ffi` which require Rust 1.64, except for `c_size_t` which is a nightly only API. As `libc::size_t` and `core::ffi::c_size_t` are both defined as `usize`, this type might be able to be defined as type alias in `hyper` until it stabilizes.


kkocdko (approved):
Happy to see the deps strip.

seanmonstar (approved):

PR #3381, opened 2023-11-01
ping: eliminates unnecessary ping when the connection is idle.

### LINKED ISSUE #3380: eliminates unnecessary ping when the connection is idle
**Version**
master

**Platform**
Linux 5.4.0-154-generic #171-Ubuntu SMP Fri Jun 16 16:29:04 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Description**
Setting PermitWithoutStream to false on the client side eliminates the need to send pings when the connection is idle.

otherwise output `too many ping`

can check code here https://github.com/grpc/grpc-go/blob/70f1a4045da95b93f73b6dbdd7049f3f053c0680/internal/transport/http2_server.go#L891-L894

grpc-client will check https://github.com/grpc/grpc-go/blob/70f1a4045da95b93f73b6dbdd7049f3f053c0680/internal/transport/http2_client.go#L1685-L1693

I tried this code:

maybe need to check idle here
```
fn maybe_ping(&mut self, cx: &mut task::Context<'_>, is_idle: bool, shared: &mut Shared) {
        match self.state {
            KeepAliveState::Scheduled(at) => {
                ...
                if !self.while_idle && is_idle {
                    trace!("keep-alive while idle disabled");
                    return;
                }
               ...
            }
            KeepAliveState::Init | KeepAliveState::PingSent => (),
        }
    }
```

I expected to see this happen: [explanation]
do not output `too many ping`

Instead, this happened: [explanation]
unnecessary ping from client side

HuSharp:
close #3380

rely on https://pkg.go.dev/google.golang.org/grpc/keepalive
server-side will output `too many pings`

### DIFF
--- a/src/proto/h2/ping.rs
+++ b/src/proto/h2/ping.rs
@@ -264,7 +264,7 @@ impl Ponger {
 
         if let Some(ref mut ka) = self.keep_alive {
             ka.maybe_schedule(is_idle, &locked);
-            ka.maybe_ping(cx, &mut locked);
+            ka.maybe_ping(cx, is_idle, &mut locked);
         }
 
         if !locked.is_ping_sent() {
@@ -284,7 +284,7 @@ impl Ponger {
                 if let Some(ref mut ka) = self.keep_alive {
                     locked.update_last_read_at();
                     ka.maybe_schedule(is_idle, &locked);
-                    ka.maybe_ping(cx, &mut locked);
+                    ka.maybe_ping(cx, is_idle, &mut locked);
                 }
 
                 if let Some(ref mut bdp) = self.bdp {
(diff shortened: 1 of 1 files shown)


seanmonstar:
I think I understand the bug. The problem is that a ping was _scheduled_ while the connection was busy (not idle), but then streams finished, and the timer fires and wants to send a ping, but _then_ it shouldn't send the ping. Is that right?

Have you tried out this patch in your workflow?

HuSharp:
> I think I understand the bug. The problem is that a ping was _scheduled_ while the connection was busy (not idle), but then streams finished, and the timer fires and wants to send a ping, but _then_ it shouldn't send the ping. Is that right?

Yes, I think this is the root cause.

> Have you tried out this patch in your workflow?

Yes, I have tried this patch, and it did not meet the error `too many pings` while can be handled by this pr check

seanmonstar (approved):

PR #3406, opened 2023-11-15
feat(lib): update to `http` 1.0

seanmonstar:
Updates dependencies `http` and `http-body` to 1.0.

To do so, implements `Clone` for `OnUpgrade`.

BREAKING CHANGE: Upgrade to `http` 1.0.

### DIFF
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -46,6 +46,7 @@ use std::future::Future;
 use std::io;
 use std::marker::Unpin;
 use std::pin::Pin;
+use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
 
 use crate::rt::{Read, ReadBufCursor, Write};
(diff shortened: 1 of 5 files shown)


davidpdrsn (approved):
LGTM

src/upgrade.rs
```diff
@@ -69,8 +70,9 @@ pub struct Upgraded {
 /// A future for a possible HTTP upgrade.
 ///
 /// If no upgrade was available, or it doesn't succeed, yields an `Error`.
+#[derive(Clone)]
 pub struct OnUpgrade {
-    rx: Option<oneshot::Receiver<crate::Result<Upgraded>>>,
+    rx: Option<Arc<Mutex<oneshot::Receiver<crate::Result<Upgraded>>>>>,
```
seanmonstar, line 75: I'd love this to be optimized a bit, to not need the extra arc/mutex. But it works for now.

PR #3413, opened 2023-11-15
Implement Read/Write for Pin<P>

### LINKED ISSUE #3412: Missing impl Read/Write for Pin<P>
Original: https://github.com/hyperium/hyper/issues/3110#issuecomment-1813249595

Without `{Read,Write} for Pin<P>` implementations, something like `Pin<Box<T: Read>>` doesn't actually impl `Read`. It seems like a pretty obvious blanket impl to have (e.g. Future has `impl<P> Future for Pin<P> where P: DerefMut<Target: Future>`, tokio has  `impl<P> Async{Read,Write} for Pin<P> where P: DerefMut<Target: Async{Read,Write}> + Unpin`), and because Pin is `#[fundamental]` it would actually be a breaking change to add one in the future.

coolreader18:
Closes #3412

### DIFF
--- a/src/rt/io.rs
+++ b/src/rt/io.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 use std::mem::MaybeUninit;
+use std::ops::DerefMut;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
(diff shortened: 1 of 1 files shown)


src/rt/io.rs
```diff
@@ -294,6 +295,20 @@ impl<T: ?Sized + Read + Unpin> Read for &mut T {
     deref_async_read!();
 }
 
+impl<P> Read for Pin<P>
+where
+    P: DerefMut,
```
seanmonstar, line 300: I notice in Tokio, it's `P: DerefMut + Unpin`. Do you know why this should be different, or if it should be the same as Tokio?

coolreader18: My guess is that they wanted to avoid having to use unsafe due to `Pin::as_deref_mut` being unstable (or not existing at all, it seems like that impl dates verbatim back to 2019), or maybe the semantics of Pin weren't as fleshed out at the time(?)

coolreader18: In any case, `Future` has [this same bound](https://doc.rust-lang.org/nightly/src/core/future/future.rs.html#118-127) for its impl for Pin.

seanmonstar (approved):

coolreader18:
Oh, woah, wild, I was looking in the Rust changelog for something totally unrelated to this and found rust-lang/rust#81363, which shows it was sorta an under-specification of Pin at the time.

PR #3419, opened 2023-11-16
v1.0.1

seanmonstar:


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
...
+
+
+#### Features
+
+* **rt:** Make ReadBuf::new public ([7161f562](https://github.com/hyperium/hyper/commit/7161f56274a30bfbe4a718bbe21d35beaf86b00b))
+
+
+#### Breaking Changes
+
+* Pin is #[fundamental], so this theoretically conflicts
```
coolreader18, line 17: Oops, I didn't realize how those were supposed to be formatted; this doesn't give much context. Probably could stick the commit subject in front of here.

CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
...
+  with existing user Read/Write for Pin<...> impls. However, those impls
+  probably don't exist yet.
```
coolreader18, line 19:
```suggestion
* Pin is #[fundamental], so providing a Read/Write impl for it theoretically conflicts
  with existing user Read/Write for Pin<...> impls. However, those impls
  probably don't exist yet.
```
Whoops, I didn't realize how these were supposed to be formatted.

PR #3422, opened 2023-11-17
Update README.md: added hyper-util to the dependencies section

sheroz:
There is a need to include the hyper-util crate to compile examples, otherwise, it cannot solve the dependency for TokioIo.  It looks like the examples also need to be updated as they refer to an internal "../benches/support/mod.rs" for TokioIo.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -22,6 +22,7 @@ serde_json = "1.0"
 form_urlencoded = "1"
 http = "1"
 futures-util = { version = "0.3", default-features = false }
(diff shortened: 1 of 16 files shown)


seanmonstar:
I don't think the examples can actually depend on `hyper-util`, since the examples are _part of_ of the `hyper` crate, and `hyper-util` depends on `hyper`. So it becomes circular. The only way we could is if the examples were made separate, with their own `Cargo.toml` file(s), then they could depend on both `hyper` and `hyper-util`.

PR #3430, opened 2023-11-20
fix(client): add missing without_shutdown method

kiron1:
in `client::conn::http1::Connection` the method `poll_without_shutdown` references `without_shutdown` in its documentation, but the method `without_shutdown` is not present in the code. This PR copies `without_shutdown` from hyper `0.14.x`: https://github.com/hyperium/hyper/blob/0.14.x/src/client/conn.rs#L520-L528

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -91,12 +91,13 @@ where
 
     /// Prevent shutdown of the underlying IO object at the end of service the request,
     /// instead run `into_parts`. This is a convenience wrapper over `poll_without_shutdown`.
-    pub fn without_shutdown(self) -> impl Future<Output = crate::Result<Parts<T>>> {
+    pub async fn without_shutdown(self) -> crate::Result<Parts<T>> {
         let mut conn = Some(self);
         futures_util::future::poll_fn(move |cx| -> Poll<crate::Result<Parts<T>>> {
             ready!(conn.as_mut().unwrap().poll_without_shutdown(cx))?;
             Poll::Ready(Ok(conn.take().unwrap().into_parts()))
         })
+        .await
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/client/conn/http1.rs
```diff
@@ -88,6 +88,16 @@ where
     pub fn poll_without_shutdown(&mut self, cx: &mut Context<'_>) -> Poll<crate::Result<()>> {
         self.inner.poll_without_shutdown(cx)
     }
+
+    /// Prevent shutdown of the underlying IO object at the end of service the request,
+    /// instead run `into_parts`. This is a convenience wrapper over `poll_without_shutdown`.
+    pub fn without_shutdown(self) -> impl Future<Output = crate::Result<Parts<T>>> {
```
seanmonstar, line 94: I'm thinking that perhaps for 1.0, we make this an `async fn`. It might be a little clearer in the docs, and we can always change from `async fn` to `-> impl Future`, but the other direction I believe is a breaking change.

seanmonstar (approved):

src/client/conn/http1.rs
aliu: The body of this function, and the one from `server/conn/http1.rs`, is really outdated. I'm not quite sure why it was written like this to begin with (maybe async blocks didn't exist at the time?). Anyways, we should update them:

```rust
pub async fn without_shutdown(mut self) -> crate::Result<Parts<T>> {
    futures_util::future::poll_fn(|cx| self.conn.poll_without_shutdown(cx)).await?;
    Ok(self.into_parts())
}
```
or
```rust
pub fn without_shutdown(mut self) -> impl Future<Output = crate::Result<Parts<T>>> {
    async move {
        futures_util::future::poll_fn(|cx| self.conn.poll_without_shutdown(cx)).await?;
        Ok(self.into_parts())
    }
}
```

PR #3431, opened 2023-11-20
fix(client): add missing without_shutdown method

kiron1:
in `client::conn::http1::Connection` the method `poll_without_shutdown` references `without_shutdown` in its documentation, but the method `without_shutdown` is not present in the code. This PR copies `without_shutdown` from hyper `0.14.x`: https://github.com/hyperium/hyper/blob/0.14.x/src/client/conn.rs#L520-L528


seanmonstar (approved):
Thanks!

PR #3440, opened 2023-11-23
add a unix socket example

danielecr:
I think this example would save time for new users, like me

### DIFF
--- a/examples/hello_unix_socket.rs
+++ b/examples/hello_unix_socket.rs
@@ -1,0 +1,66 @@
+#![deny(warnings)]
+
+use std::convert::Infallible;
+use std::net::SocketAddr;
+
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR!

My feeling here is that it's very similar to existing example, and so I don't think it teaches enough to overcome the time/cost of compiling it when checking tests. It would be a fair blog post, or similar, if you'd like to publish it :)

danielecr:
I understand (I did cut&paste), but really I spent a morning for just 3 lines to replace.
Also hyperlocal is a project that exists just for that, and now it is obsoleted.

Does it makes sense to add comments in `hello.rs` for the unix socket version?

danielecr:
#3443 like this, maybe?

PR #3443, opened 2023-11-23
add comments for unix socket usage

danielecr:
in the comment there is code to be used as placeholder for using unix socket as transport


seanmonstar:
I appreciate your desire to improve the documentation, thank you! However, I personally think adding commented out code to the example can be distracting to people who don't need it.

If anything, I think adding one comment above the line for `TcpListener::bind()` could be helpful, saying `// You can use any kind of listener or source of new connections.`.

danielecr:
I completely disagree on that, anyway ok, nevermind

Issue #2565, opened 2021-06-01
Client support for receiving 1xx informational responses


seanmonstar:
This is to discuss adding support for accessing 1xx informational responses on the client side (see #2426 for server side). Currently, hyper parses all 1xx responses and ignores them, as RFC 7230 says it must at least be able to receive and ignore them.

The Client (and `client::conn`) interfaces return an `impl Future<Output = Result<Response>>`, which only allows awaiting it for a single response. We can't really add new methods to allow also awaiting for informational responses, unless we stuck to always return a specialized type that is both a `Future` and has extra methods. However, _using_ those together wouldn't be that clear to a user. You can't be sure if or how many informational responses would be received before the full response, so you would want to await "both" types.

Instead, we can use a callback format. We could define a type (`hyper::client::ext::Informational`? `On1xx`? names...), and the user would store that in the request extensions. We'd then check for that in the h1 proto code and if it exists, call the callback with the 1xx headers.

nox:
I much prefer the idea of adding methods on `ResponseFuture`, hooks make bookkeeping harder for the caller, and I don't think it is more confusing to have both those methods and the `Future` impl than say, using `poll_data` and `poll_trailers` on `HttpBody`.

bagder:
In the curl project, there's right now 105 test cases disabled when built with hyper. 46 of them use 100-continue in one way or another and we really need to know when we get 1xx responses (or not) to make curl+hyper behave like the built-in HTTP code does. It could allow us to make significant progress in my task to make curl+hyper match the built-in HTTP code.

seanmonstar:
Right-o, I got distracted with other work. I'll do some thinking and work on this in the next few days.

seanmonstar:
I've spent some time researching options (prior art of APIs), and considering the existing code, and I think I've settled on staged approach. I agree with you, @nox, we _can_ add a method to `ResponseFuture` on the Rust side, and make it work right. However, it'd be less easy to access in the exposed C API, and will take more time to get right. So:

1. First, add support for a callback extension, which will be exposed in the C API. This is the least complicated to do on hyper's side, so we can do it quickly. Importantly, though, most of it can be reused or only slightly altered to support the next steps.
[...]

PR #3472, opened 2023-12-11
feat(client): support Expect: 100-continue

iasoon:
This patch modifies hyper clients to respect the `Expect: 100-Continue` header.
When the header is present on a request, the client will not send the request body
until the `100 Continue` informational response is received from the server.
Related issue: #2565

The implementation follows the existing `on_informational` callback mechanism,
threading the connection state through to the reponse header parsing.

This seems like the most straightforward modification to get `Expect: 100-Continue`
support, but happy to discuss alternatives.

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -75,6 +75,8 @@ where
                 // We assume a modern world where the remote speaks HTTP/1.1.
                 // If they tell us otherwise, we'll downgrade in `read_head`.
                 version: Version::HTTP_11,
+                #[cfg(feature = "client")]
+                awaiting_100_continue: false,
             },
             _marker: PhantomData,
         }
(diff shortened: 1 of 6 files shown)


seanmonstar:
Thanks for the PR and taking this on!

One thing I was wondering about is how would someone choose to wait only a certain amount of time for the server to respond? We _could_ add another timeout into the connection state, but if it's not super hard to do outside, I think that's preferable.

A different way of doing this would be to make the `Body` be responsible for pausing, and thus it could hold onto a timeout there. Perhaps an extension callback would allow a user to signal a oneshot channel that the 100 response was received.

I'm not sure which is best, at the moment.

rfairfax:
FWIW - I hit this bug recently as well and was just about to file something.  Thanks @iasoon for proposing a fix.  I work on some storage services where this feature is key to avoiding sending large amounts of data that would otherwise be rejected anyway so this is super high value to my users and my service's scalability.  I'm glad to test patches if it helps validate things.

> A different way of doing this would be to make the `Body` be responsible for pausing

I actually started going down this route as it seemed like a way to avoid requiring patching hyper but ended up having to give up as there was no good way to signal the continue came through.  However, after thinking about what this would look like my 2c is that the connection is the right place to do this as it's part of the protocol and not a property of the body.  If you wire up the continue logic in the body's poll method, for example, then failure to choose the right body implementation that can pause results in not speaking the protocol correctly which seems error prone.

PR #3478, opened 2023-12-12
Fix compile error: borrow of moved value: `header`

chrislearn:


seanmonstar:
What do you mean compile error? It seemed to compile without this change.

chrislearn:
Use command ` cargo build --features full --features tracing` or ` cargo build --all-features full`.

I tested on windows11 stable-x86_64-pc-windows-msvc  and wsl2 ubuntu stable-x86_64-unknown-linux-gnu.

seanmonstar (approved):
Ah, I see what you mean. I guess our CI doesn't check building with unstable tracing. Should fix that too. But I can verify this fixes things, thanks!

Issue #3453, opened 2023-12-01
Documentation link in hyper::client::conn for "Higher-level Client API" should probably go to hyper-util now


taladar:
**Version**
1.0.1

https://docs.rs/hyper/1.0.1/hyper/client/conn/index.html

The link just goes to the module hyper::client above which only contains conn now. Presumably it refers to the code now moved to hyper-util.

ketzu:
After looking out what happend to `hyper::Client` after stumbling over it in some examples (for sharing a pooled connection in axum), I ended up here. (New to rust, so sorry if I break some conventions or seem rude😅 )

My adventure through this issue, thanks for that :) But the resulting question kind of plays into this:

 * hyper-util, as a new target for the link, already labels the `Client` as legacy, should anyone be encouraged to use legacy for new code?
 * reqwest (the link provided in the commit 5236ac5) here, depends on [hyper 0.14](https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L103C25-L103C25), using `hyper::Client` [internally](https://github.com/seanmonstar/reqwest/blob/master/src/async_impl/client.rs#L1654), as far as I can tell, implying the recommendation of hyper to get a pooled connection is to use an older version of hyper.

As I am new to this, I might misunderstand some things around this :) I'll just leave it as food for thought though (and because I secretly hope I'll end up with a recommendation for a higher level client api)

PR #3501, opened 2023-12-22
replace client link in docs

GlenDC:
fixes issue #3453

### DIFF
--- a/src/client/conn/mod.rs
+++ b/src/client/conn/mod.rs
@@ -6,7 +6,9 @@
 //! customize those things externally.
 //!
 //! If you are looking for a convenient HTTP client,
-//! then you may wish to consider [reqwest](https://github.com/seanmonstar/reqwest).
+//! then you may wish to consider [reqwest](https://github.com/seanmonstar/reqwest)
+//! for a high level client or [`hyper-util`'s client](https://docs.rs/hyper-util/latest/hyper_util/client/index.html)
+//! if you want to keep it more low level / basic.
 //!
 //! ## Example
 //!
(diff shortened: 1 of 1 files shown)


src/client/conn/mod.rs
```diff
@@ -5,8 +5,8 @@
 //! are not handled at this level. This module provides the building blocks to
 //! customize those things externally.
 //!
-//! If don't have need to manage connections yourself, consider using the
-//! higher-level [Client](super) API.
+//! If you are looking for a convenient HTTP client,
+//! then you may wish to consider [reqwest](https://github.com/seanmonstar/reqwest).
```
dswij, line 9: I think we might want to point to `hyper-util` instead of `reqwest`

GlenDC: That’s not a convenient / high level http client though?

dswij: It's higher level than this, I'd say we can just point to both of them here.

GlenDC:
```suggestion
//! then you may wish to consider [reqwest](https://github.com/seanmonstar/reqwest)
//! for a high level client or [`hyper-util`'s client](https://docs.rs/hyper-util/latest/hyper_util/client/index.html)
//! if you want to keep it more low level / basic.
```

Like this @dswij ?

dswij (approved):
Thank you! @GlenDC 

Can you help to rebase? then I think it's good to merge afterwards.

GlenDC:
Done @dswij .

dswij:
Thank you.

PR #3503, opened 2023-12-25
use share data in service

getong:
fix two issues here

- [x] all connections share the same data here, not one state per connection .
- [x] when uri not found, the couter does not add one.


seanmonstar (approved):
Thanks for noticing and sending a fix!

PR #3515, opened 2024-01-05
Add a safe way to fill a ReadBuf

### LINKED ISSUE #3509: Add the ability to safely read into ReadBuf
It is impossible to write an implementation of `hyper::rt::Read` because `ReadBufCursor` cannot be written to safely.

Have a method that initializes the read buffer by zeroing it and then returning an `&mut [u8]` that can be filled safely. E.g. [this](https://docs.rs/tokio/latest/tokio/io/struct.ReadBuf.html#method.initialize_unfilled).

N/A

notgull:
Closes #3509

### DIFF
--- a/src/rt/io.rs
+++ b/src/rt/io.rs
@@ -244,6 +244,41 @@ impl<'data> ReadBufCursor<'data> {
         self.buf.init = self.buf.filled.max(self.buf.init);
     }
 
+    /// Read into a byte slice safely.
+    ///
+    /// This function provides a safe alternative to the [`as_mut`] and
+    /// [`advance`] APIS. It initializes the uninitialized part of the
+    /// buffer (if one exists) and passed it as a slice to the provided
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR!

My initial thoughts were that we could provide a couple methods like in [`BorrowedCursor`](https://doc.rust-lang.org/std/io/struct.BorrowedCursor.html). Looking through them, it occurs to me that the only "basic" safe way to write data is to like `append(&[u8])`. Otherwise, even if you can safely access all `initialized_mut()`, it would still be unsafe to state an arbitrary number of bytes have been written.

notgull:
> My initial thoughts were that we could provide a couple methods like in [`BorrowedCursor`](https://doc.rust-lang.org/std/io/struct.BorrowedCursor.html).

Any specific methods?

Issue #3283, opened 2023-08-03
What's the maximum header size limit?


bagder:
**Version**
current from master

**Platform**
Linux

**Description**
I'm using hyper in curl, and when I receive a HTTP/1 response with **many** response headers, hyper cuts off the response and returns an error due to `message head is too large`. (With a `HYPERE_INVALID_PEER_MESSAGE` error)

I cannot find what this exact limit is. I can't even find it in the hyper source code.

seanmonstar:
To be more specific, are you asking about the list of headers, or the length of any single header? If receiving a _lot_ of headers, the current limit of name/value pairs in hyper is 100. That was picked somewhat arbitrarily, and could be increased if more than that were common, but so far no one has ever needed it 🤷 .

bagder:
Thanks for this.

I can't tell you what's right but it seems Firefox, Chrome (and curl) all accept name/value pairs all the way until they reach the maximum response header size. Which seems to be limited to somewhere between 256KB and 1MB.

ajgh1:
I'm not sure about client-side, but Apache httpd 2.4's default `LimitRequestFields` (100) and `LimitRequestFieldSize` (8190) settings imply if a request's headers go much over 800KB, default httpd configurations will reject it.  These limits can be changed of course, but the manual cautions that should not be necessary under normal circumstances.

seanmonstar:
No, that limit is for the HTTP/1 decoder. HTTP/2 has a setting that you can send, `SETTINGS_MAX_HEADER_LIST_SIZE `, to tell peer how much you're willing to accept, and hyper will enforce it for you.

f3lixding:
> the current limit of name/value pairs in hyper is 100. That was picked somewhat arbitrarily, and could be increased if more than that were common, but so far no one has ever needed it

I ran into the same problem (message head is too large) for reqwest and I found my way here to this thread. 
Just wanted to +1 on this. It would be fantastic if we can have the MAX_HEADERS be configurable.

Millione:
need to be increased or can be configurable.

PureWhiteWu:
Hi, we've encountered same problem here. Our HTTP request has 100+ headers and hyper failed with error.

Is there any plan to make it configurable?

grodin:
I think this was resolved in #3532 but the issue seems to have been left open

seanmonstar:
Good call, thanks for conneting the issue!

PR #3523, opened 2024-01-12
feat(http1): support configurable `max_headers`

wfly1998:
Recently we encountered a problem when using `hyper`.

After the client sent the request, it received response `431 Request Header Fields Too Large`. After investigation, we found that `hyper` sets an upper limit on the number of `headers`, and this upper limit is a fixed and non-configurable value.

#3283 also encountered similar problems.

This PR sets `max_headers` to a configurable value. Its default value remains the same as the previous code, and we can set this value when needed.

Additionally, unit tests have been added and the current code passes.

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -115,6 +115,7 @@ pub struct Builder {
     h1_writev: Option<bool>,
     h1_title_case_headers: bool,
     h1_preserve_header_case: bool,
+    h1_max_headers: Option<usize>,
     #[cfg(feature = "ffi")]
     h1_preserve_header_order: bool,
     h1_read_buf_exact_size: Option<usize>,
@@ -309,6 +310,7 @@ impl Builder {
             h1_parser_config: Default::default(),
             h1_title_case_headers: false,
             h1_preserve_header_case: false,
+            h1_max_headers: None,
             #[cfg(feature = "ffi")]
             h1_preserve_header_order: false,
             h1_max_buf_size: None,
@@ -439,6 +441,24 @@ impl Builder {
         self
     }
 
+    /// Set the maximum number of headers.
+    ///
+    /// When a response is received, the parser will reserve a buffer to store headers for optimal
+    /// performance.
+    ///
+    /// If client receives more headers than the buffer size, the error "message header too large"
+    /// is returned.
+    ///
+    /// Note that headers is allocated on the stack by default, which has higher performance. After
+    /// setting this value, headers will be allocated in heap memory, that is, heap memory
+    /// allocation will occur for each response, and there will be a performance drop of about 5%.
+    ///
+    /// Default is 100.
+    pub fn max_headers(&mut self, val: usize) -> &mut Self {
+        self.h1_max_headers = Some(val);
+        self
+    }
+
     /// Set whether to support preserving original header order.
     ///
     /// Currently, this will record the order in which headers are received, and store this
@@ -518,6 +538,9 @@ impl Builder {
             }
             if opts.h1_preserve_header_case {
                 conn.set_preserve_header_case();
+            }
+            if let Some(max_headers) = opts.h1_max_headers {
+                conn.set_http1_max_headers(max_headers);
             }
             #[cfg(feature = "ffi")]
             if opts.h1_preserve_header_order {
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -13,6 +13,7 @@ use http::header::Entry;
 use http::header::ValueIter;
 use http::header::{self, HeaderMap, HeaderName, HeaderValue};
 use http::{Method, StatusCode, Version};
+use smallvec::{smallvec, smallvec_inline, SmallVec};
 
 use crate::body::DecodedLength;
 #[cfg(feature = "server")]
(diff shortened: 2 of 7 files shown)


src/proto/h1/role.rs
```diff
@@ -139,9 +139,10 @@ impl Http1Transaction for Server {
         // but we *never* read any of it until after httparse has assigned
         // values into it. By not zeroing out the stack memory, this saves
         // a good ~5% on pipeline benchmarks.
-        let mut headers_indices = [MaybeUninit::<HeaderIndices>::uninit(); MAX_HEADERS];
+        let mut headers_indices = vec![MaybeUninit::<HeaderIndices>::uninit(); ctx.h1_max_headers];
```
seanmonstar, line 142: This will result in a heap allocation for every request (well, two, the one a couple lines below too). This line is carefully crafted to be fast, which is why it uses uninitialized memory. You can compare the `cargo bench --bench pipeline --features full` of master and this patch to see if there's much change.

I suspect this will require a more involved solution. Perhaps an `enum` that, if the max hasn't been changed, can still use the uninitialized stack array, and if it has, then it uses a `Vec`. (Maybe the `Vec` could even be stored and reused?)

seanmonstar (changes requested):
Thanks for the thorough PR! I welcome adding the option, but I want to make sure we continue to stay as fast as possible.

src/proto/h1/role.rs
aviramha: https://docs.rs/smallvec/latest/smallvec/index.html I think that's exactly what you want @seanmonstar ?

aviramha:
We encountered an issue that requires this as well, thanks for sending PR! would love to help if needed :)

src/proto/h1/role.rs
aviramha: Thinking about it, I'd consider either dropping the configuration, or changing it to contain the pre-allocated number while using smallvec trusting other part of the code (axum for example) to make sure it doesn't exceed header size in kbs and not in count.

wfly1998: > https://docs.rs/smallvec/latest/smallvec/index.html I think that's exactly what you want @seanmonstar ?

It seems that there is too little data on the stack and does not meet the requirements 😢

wfly1998: > Thinking about it, I'd consider either dropping the configuration, or changing it to contain the pre-allocated number while using smallvec trusting other part of the code to make sure it doesn't exceed header size in kbs and not in count.

I'm also thinking about solutions, maybe const generic has the best performance, but the implementation is ugly.

Using `enum` is also a solution. If you use the default implementation to keep the data on the stack, you can ensure no performance degradation. But if the data is stored on the heap, on the one hand, as @seanmonstar said, it will introduce the overhead of memory application; on the other hand, it will also waste a lot of stack memory, because this huge `enum` uses only a small part on stack.

aviramha: another option - https://github.com/Lokathor/tinyvec

Noah-Kennedy: > Thinking about it, I'd consider either dropping the configuration, or changing it to contain the pre-allocated number while using smallvec trusting other part of the code to make sure it doesn't exceed header size in kbs and not in count.

both total header size and header count can be useful to limit for different reasons

wfly1998: @aviramha Hey I tried `smallvec` and it works great! The code has also been updated!

When using the default value of "max headers", headers are allocated on the stack with no performance penalty. When "max headers" is set, headers will be allocated on the heap and performance will be reduced by approximately 5%.

Perhaps this is the best solution for configurable capacity, cc @seanmonstar

aviramha: > > Thinking about it, I'd consider either dropping the configuration, or changing it to contain the pre-allocated number while using smallvec trusting other part of the code to make sure it doesn't exceed header size in kbs and not in count.
> 
> both total header size and header count can be useful to limit for different reasons

Yeah but I was under the assumption there's other part of the code that validates that, but I might be wrong.

src/client/conn/http1.rs
```diff
@@ -439,6 +441,24 @@ impl Builder {
...
 
+    /// Set the maximum number of headers.
+    ///
+    /// When a response is received, the parser will reserve a buffer to store headers for optimal
+    /// performance.
+    ///
+    /// If client receives more headers than the buffer size, the error "message header too large"
+    /// is returned.
+    ///
+    /// Note that headers is allocated on the stack by default, which has higher performance. After
```
seanmonstar, line 452: Is there a need to include this paragraph? While it is interesting, perhaps we don't need really need it...

(Same with server).

PureWhiteWu: I feel that as a user, this paragraph can clearly tell me the behavior of this API and help me make decisions.

Otherwise, I might just set a very large value (whether it is used or not).

wfly1998: If this option has any impact on performance, I think it needs to be explained.

seanmonstar: That's fair! I suppose my concern was about it "promising" how internals worked. If we later updated this to be able to allocate less. But thinking this through as I respond, I don't think anyone would mind if  a later update made things faster. It's not like they would be depending on it allocating on the heap or stack, just that it's fast or why it's less fast.

Seems good!

seanmonstar (approved):
Phenomenal work!

PR #3524, opened 2024-01-13
feat(http2): add `initial_max_send_streams` method to HTTP/2 client builder

magurotuna:
This adds `initial_max_send_streams` method to the HTTP/2 client builder.

The value set via this method will then be forwarded to [`h2::client::Builder::initial_max_send_streams`]. If not set, the default value defined in `h2` crate will be used in case some users might rely on this behavior. However, in the following version we plan to explicitly set `100` as a reasonable default value (see https://github.com/hyperium/h2/issues/731 for rationale on why this can be considered a reasonable value).

[`h2::client::Builder::initial_max_send_streams`]: https://docs.rs/h2/0.4.1/h2/client/struct.Builder.html#method.initial_max_send_streams

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -310,15 +310,11 @@ where
     /// This value will be overwritten by the value included in the initial
     /// SETTINGS frame received from the peer as part of a [connection preface].
     ///
-    /// Passing `None` will do nothing.
-    ///
-    /// If not set, hyper will use a default.
+    /// The default value is determined by the `h2` crate, but may change.
     ///
     /// [connection preface]: https://httpwg.org/specs/rfc9113.html#preface
     pub fn initial_max_send_streams(&mut self, initial: impl Into<Option<usize>>) -> &mut Self {
-        if let Some(initial) = initial.into() {
-            self.h2_builder.initial_max_send_streams = initial;
-        }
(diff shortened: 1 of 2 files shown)


seanmonstar:
I had this thought, and I'd love to hear other opinions. I can be convinced either way. But:

Should the feature be exposed in 1.2, and the default changed in 1.3? I wonder if there's value in giving people some time to specify their preferred option, before perhaps limiting them more than expected.

At the same time, getting the default to people faster may be more helpful to more people.

magurotuna:
Given that hyper is forming the foundation of HTTP in the Rust ecosystem, being conservative makes more sense to me. Your first idea, exposing the builder method in the next minor version and then setting the default value in one more next minor version, sounds a better one. I'd be happy to make a change to this PR to remove the default value part.

From our (developers of [Deno](https://github.com/denoland/deno)) perspective, our eventual goal is to fix the `fetch` implementation that's powered by `reqwest` create. So it would be ideal if we could ship hyper's next version including this new builder method, and incorporate the new version into `reqwest` finally. (Then we can configure the initial_max_send_streams value on our side)

seanmonstar:
Alright, yea, sounds good to me too. If you want to remove the smaller default here, we can include this in 1.2 with a note that the default will change.

seanmonstar (approved):

PR #3528, opened 2024-01-18
feat(http2): add config for max_local_error_reset_streams in server

Noah-Kennedy:
This change exposes a tunable for the max_local_error_reset_streams parameter in h2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ http = "0.2"
 http-body = "0.4"
 httpdate = "1.0"
 httparse = "1.8"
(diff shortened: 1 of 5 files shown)


Noah-Kennedy:
The warnings change was needed because the nightly tests are failing right now due to warnings.

seanmonstar:
> The warnings change was needed because the nightly tests are failing right now due to warnings.

As in 0.14.x will have warnings from now on?

Noah-Kennedy:
@seanmonstar only on nightly

seanmonstar (approved):

PR #3529, opened 2024-01-18
Bumb h2 version to 0.3.24

QnJ1c2kNCg:
This is to address the following advisory:
https://rustsec.org/advisories/RUSTSEC-2024-0003.html


seanmonstar:
This is not required, you can run `cargo update -p h2` in your own project to get the latest version.

PR #3532, opened 2024-01-19
test(body): re-enable miri on a few channel tests

seanmonstar:
After prodding from https://github.com/hyperium/hyper/issues/3015#issuecomment-1829333578, checking to see if this works now.

### DIFF
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -554,7 +554,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(miri))] // TODO issue #3015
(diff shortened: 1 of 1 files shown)


RalfJung:
The title says "a few" -- what about the others?

seanmonstar:
I started with just the tests that were changed in the previous issue. Looking at the rest, I think they don't like `#[tokio::test]` because it uses epoll, right?

RalfJung:
Ah yes, that is https://github.com/rust-lang/miri/issues/602.

PR #3536, opened 2024-01-21
fix(http2): typo in trace logging

erebe:
fix simple typo in trace logging for http2 connection handshake.

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -426,7 +426,7 @@ where
         let opts = self.clone();
 
         async move {
-            trace!("client handshake HTTP2");
+            trace!("client handshake HTTP/2");
 
             let (tx, rx) = dispatch::channel();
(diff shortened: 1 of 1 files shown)


src/client/conn/http2.rs
```diff
@@ -426,7 +426,7 @@ where
         let opts = self.clone();
 
         async move {
-            trace!("client handshake HTTP/1");
+            trace!("client handshake HTTP2");
```
hawkw, line 429: nit: how about
```suggestion
            trace!("client handshake HTTP/2");
```

erebe: I don't mind personally.
I picked `HTTP2` because in the doc on the same file, this HTTP version is most often referred to as `HTTP2`.
But there is also 1 mention of `HTTP/2`

seanmonstar (approved):
Thanks for catching this!

PR #3573, opened 2024-02-19
chore: use `Trait` rather than `Into<Box<dyn Trait>>`

wfly1998:
All types with `Error + Send + Sync` can be boxed into `Box<dyn Error + Send + Sync>`, but using `Error ...` directly is more flexible and less restrictive for the user.

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -61,7 +61,7 @@ impl<T, B> Connection<T, B>
 where
     T: Read + Write + Unpin + 'static,
     B: Body + 'static,
-    B::Error: Into<Box<dyn StdError + Send + Sync>>,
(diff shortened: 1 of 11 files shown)


seanmonstar:
The way Rust's standard library currently is, the former (`Into<...>`) is more flexible. That's because `Box<dyn Error>` does **not** implement `Error`, for "reasons".

So, this change makes it so `Error = Box<dyn Error>` no longer works.

wfly1998:
> The way Rust's standard library currently is, the former (`Into<...>`) is more flexible. That's because `Box<dyn Error>` does **not** implement `Error`, for "reasons".
> 
> So, this change makes it so `Error = Box<dyn Error>` no longer works.

Got it, thanks for your answer.

Issue #3590, opened 2024-03-01
Expose Service and HttpService without features


tottoto:
I would suppose that `hyper-util` would be good to allow users to enable `http1` or/and `http2` feature separately when using `server-auto` feature (although it is a breaking change). As `hyper-util` uses `Service` and `HttpService` to implement `server-auto` feature, these traits are required without `http{1,2}` features. Since these traits do not actually require optional dependencies, the drawback to expose them seems small comparing the benefit.

tottoto:
This allows `hyper-util` to enable its service util without `http1` and `http2` features as well.

refs: https://github.com/hyperium/hyper-util/pull/79

tottoto:
https://github.com/hyperium/hyper-util/pull/105 is a motivation for this proposal.

PR #3591, opened 2024-03-02
feat(service): expose Service and HttpService trait unconditionally

tottoto:
Exposes `Service` and `HttpService` trait unconditionally.

refs:  #3590

### DIFF
--- a/src/service/mod.rs
+++ b/src/service/mod.rs
@@ -25,12 +25,6 @@ mod http;
 mod service;
 mod util;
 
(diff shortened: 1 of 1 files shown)


dswij:
What's the benefit of having this without http1/http2?

tottoto:
I think it would be a benefit in the cases that crates which collaborate with `hyper` need to use `service` and do not themselves use the others in http1 or http2 features.

seanmonstar (approved):

PR #3595, opened 2024-03-05
feat: relax connection IO trait bounds

ehaydenr:
I can't tell if this was intentional. hyper 0.14.x didn't have this requirement and as far as I can tell it's not actually needed.


dswij (approved):

dswij:
For context, I believe that this is here for https://github.com/hyperium/hyper/blob/4d0c184f935adf93ed05b86ec3a16e432aa81706/src/server/conn/http1.rs#L180.

But the `Connection` impl itself, and `Dispatch` does not require this bound, so it's safe to remove this from `impl Future`.

dswij:
Thank you! @ehaydenr

PR #3596, opened 2024-03-07
perf: optimize error type

Noah-Kennedy:
The allocations and frees for these errors can get expensive in some cases. Any reduction in allocator usage here helps some use cases a lot it seems.

### DIFF
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -253,7 +253,7 @@ impl Body for Incoming {
                                 Some(h2::Reason::NO_ERROR) | Some(h2::Reason::CANCEL) => {
                                     Poll::Ready(None)
                                 }
-                                _ => Poll::Ready(Some(Err(crate::Error::new_body(e)))),
+                                _ => Poll::Ready(Some(Err(crate::Error::new_body_h2(e)))),
                             };
                         }
                         None => {
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,11 +1,28 @@
 //! Error and Result module.
 use std::error::Error as StdError;
 use std::fmt;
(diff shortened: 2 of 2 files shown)


Noah-Kennedy:
I'm leaving this as a draft while I look into ways to maybe make `Cause` faster as well.

Noah-Kennedy:
Marked as ready and squashing to make review easier. Gonna leave as two commits and squash again on merge.

tottoto:
Looks good to me. However, as it increases a little complexity, I would think that there would needs to be a benchmark to show in which cases it is actually efficient or not, and how the efficiency is.

seanmonstar:
For context, the reason it was boxed before was:

- To reduce return sizes of `Result<T, Error>`.
- Because errors should be less common, it seemed better to make the error case more expensive if it allowed the good case to be any faster.

Noah-Kennedy:
> Looks good to me. However, as it increases a little complexity, I would think that there would needs to be a benchmark to show in which cases it is actually efficient or not, and how the efficiency is.

I'll try and construct something, although I suspect that for most users this won't matter. That said, the effect of the double allocation is surprisingly significant.

Noah-Kennedy:
> For context, the reason it was boxed before was:
> 
>     * To reduce return sizes of `Result<T, Error>`.
> 
>     * Because errors should be less common, it seemed better to make the error case more expensive if it allowed the good case to be any faster.

Hmm, the non-error case is a good point. I'll see if I can shrink this down.

I'm investigating more right now the actual production issue I saw with this, because there's a chance that this might not be a valid use of HTTP/2 semantics I'm looking at which is causing this. I might potentially close this if I decide that any case where this matters is an issue of "you're holding it wrong".

PR #3598, opened 2024-03-11
Remove reference to Client in SendRequest::send_request

TomFryersMidsummer:
`Client` no longer exists. This is similar to #3501.


dswij (approved):
Thank you.

PR #3600, opened 2024-03-12
refactor(upgrade): make parts of `upgrade`` public

apersomany:
Making `Pending`, `pending`, and `Pending::fulfill` public allows external libraries to provide the `OnUpgrade` extension to HTTP server frameworks (such as Axum or Warp) that solely depend on Hyper's upgrade mechanism.


seanmonstar:
Thanks for the PR!

I'm hesitant to make the other parts of this public. It sorta reveals that the internals are a channel, and we'd thus need to maintain that forever.

apersomany:
I see. Let me close the PR for now.

PR #3607, opened 2024-03-21
forward implement Service for reference types

lperlaki:
simplifies the usage of `dyn Service`s and making `Service` types `Clone` by wrapping them in a `Rc`or `Arc`

PR #3608, opened 2024-03-22
docs(body): cleanup some references to `Incoming`

dswij:
`Incoming::channel` is only used internally for tests. I'm planning to get back to those later to clean those up.


dswij:
Actually just saw #3607

PR #3616, opened 2024-03-30
fix: avoid panic when calling `graceful_shutdown` if H1 connection was upgraded

### LINKED ISSUE #3615: Panic on graceful shutdown for http/1
**Version**
Encountered with `1.x`

**Platform**
Doesn't matter

**Description**

Attempt to call `graceful_shutdown` for H1 connection which is upgraded (=> `Poll::Ready`) leads to panic:
```
panic was raised: panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-1.2.0/src/server/conn/http1.rs:483:38:
called `Option::unwrap()` on a `None` value
```

The reason is this line:

```rs
Pin::new(self.inner.as_mut().unwrap()).graceful_shutdown()
```

When connection is Upgraded (& future  is `Ready`), it's `None` https://github.com/hyperium/hyper/blob/bc9a86f58f8bd5c35b2bfd7e632ec132280d79ba/src/server/conn/http1.rs#L502-L506 

So we should avoid unwrapping

DDtKey:
Closes #3615

Added test which fails without these changes

### DIFF
--- a/src/server/conn/http1.rs
+++ b/src/server/conn/http1.rs
@@ -482,7 +482,11 @@ where
     /// This `Connection` should continue to be polled until shutdown
     /// can finish.
     pub fn graceful_shutdown(mut self: Pin<&mut Self>) {
-        Pin::new(self.inner.as_mut().unwrap()).graceful_shutdown()
+        // Connection (`inner`) is `None` if it was upgraded (and `poll` is `Ready`).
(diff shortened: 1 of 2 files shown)


dswij (approved):
thank you!

PR #3618, opened 2024-04-02
refactor(h1): use UninitSlice::as_uninit_slice_mut() instead of cast

seanmonstar:


src/proto/h1/io.rs
```diff
@@ -246,8 +245,9 @@ where
             self.read_buf.reserve(next);
         }
 
-        let dst = self.read_buf.chunk_mut();
-        let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
+        // SAFETY: ReadBuf and poll_read promise not to set any uninitialized
+        // bytes onto `dst`.
+        let dst = unsafe { self.read_buf.chunk_mut().as_uninit_slice_mut() };
```
tottoto, line 250: As `UninitSlice::as_uninit_slice_mut` is introduced at `bytes` 1.2, the minimum version seems to be needed to be  bumped to the version 1.2 from 1.

https://github.com/tokio-rs/bytes/blob/v1.2.0/CHANGELOG.md

tottoto (approved):

dswij (approved):

PR #3621, opened 2024-04-05
fix: Do not include user information in Host header

tindzk:
According to RFC 9110, section 7.2, the Host header should only comprise the URI host and an optional port.

Currently, the examples set the Host header to the URI's authority which may also contain user information (see RFC 3986, section 3.2).

Update the examples to construct the Host header manually to avoid sensitive information from showing up in server logs and to ensure that the server's routing logic works correctly when a username and password are supplied.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -53,12 +53,10 @@ async fn fetch_url(url: hyper::Uri) -> Result<()> {
         }
     });
 
-    let authority = url.authority().unwrap().clone();
-
     let path = url.path();
     let req = Request::builder()
         .uri(path)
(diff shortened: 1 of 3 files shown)


tindzk:
I agree with the sentiment of keeping the examples as minimal as possible. I have shortened the fix to always include the port number. This should be compatible with the majority of servers out there.

The same issue is also present on the website: https://hyper.rs/guides/1/client/basic/

sfackler:
@seanmonstar I think this was closed by accident?

seanmonstar:
While technically true, I'm so far inclined to keep the examples simple. There's many other things that a user would need to do to make a good client that are also not included.

PR #3622, opened 2024-04-05
Default HTTP/2 max_header_list_size to 16kb

seanmonstar:
The RFC doesn't specify a default. `h2` has an "emergency" default that's quite high. It makes sense for hyper to pick a safer default to protect users who don't think about it. When checking other libraries, many set a default of either 8kb or 16kb.


Noah-Kennedy (approved):

nox:
FYI when logged into Google, console.cloud.google.com sends headers bigger than this default size, thus that seems too low to me.

seanmonstar:
Do you have a suggested default instead?

Part of the choice here was noting that _most_ other libraries either picked 8kb or 16kb, and rely on users to opt-in to more. I suppose one possibility we could consider is different defaults for server vs client.

nox:
The previous value was just fine in 2024 IMO.

PR #3628, opened 2024-04-10
multi layer proxy example.

ohaddahan:
In this configuration we have a `public` `master` server, which accepts
  **outgoing** connections from `endpoint` servers.
  The reason for using outgoing connections, is to avoid the need to open firewall ports.
  The `master` will receive requests and forward them one of the servers connected to it.

(since it's in `examples` I merged it into one file/binary, in the real world the two servers are obviously not in the same process)

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,16 +50,7 @@ spmc = "0.3"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tokio = { version = "1", features = [
-    "fs",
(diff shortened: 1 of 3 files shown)


ohaddahan:
@hatoo helped a lot in building this.

seanmonstar:
I appreciate the interesting example! However, I'm trying to now limit how many examples are in the repository, since they increase compile/test time and maintenance. It would certainly be an interesting blog post, though!

PR #3629, opened 2024-04-11
fix warnings / lint errors and format

OneOfOne:
- cargo.toml: format and upgrade bytes
- fix warnings / lint errors


seanmonstar (changes requested):
Thanks for the PR!

I'd rather not make most of these changes. The only ones I would merge are the removal of the static lifetime.

PR #3630, opened 2024-04-14
fix(client): send content-length even with no body

seanmonstar:
Most request methods define a payload. If hyper detects that no body has been included, it will now include a `content-length: 0` header automatically.

It will not do this for methods that don't have defined payloads (GET, HEAD, and CONNECT).

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -1288,6 +1288,13 @@ impl Client {
             body
         } else {
             head.headers.remove(header::TRANSFER_ENCODING);
+            // If we know there's body coming, set a content-length.
+            // But only if the method normally has a body.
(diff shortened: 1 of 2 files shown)


dswij (approved):

src/proto/h1/role.rs
```diff
@@ -1468,6 +1474,11 @@ impl Client {
     }
 }
 
+#[cfg(feature = "client")]
+fn is_method_assumed_empty(method: &Method) -> bool {
+    matches!(method, &Method::GET | &Method::HEAD | &Method::CONNECT)
```
ckcr4lyf, line 1479: I'd think `OPTIONS` and `DELETE` should be included here? Probably also `TRACE`.

PR #3632, opened 2024-04-16
fix: auto content-length failing on curl

dswij:
This PR now adds `content-length` to header automatically only when `content-type` is present.

This is also the case according to https://github.com/seanmonstar/reqwest/issues/2240#issuecomment-2044260263

See https://github.com/hyperium/hyper/commit/172fdfaf0e0d9222917f271a83339238082e2657#commitcomment-141005865


seanmonstar:
I'm just going to revert it entirely: https://github.com/hyperium/hyper/pull/3633

PR #3637, opened 2024-04-17
feat(http1): add support for receiving trailer fields

### LINKED ISSUE #2703: Receiving HTTP/1.1 trailers

hjr3:
Closes #2703

In https://github.com/hyperium/hyper/pull/3375 we were strict about sending trailers per RFC 7230. In this PR, I have been much more accepting of trailers. Is this the behavior we want?

Some questions:

- If the client does not send `TE: trailers` but the server sends trailer headers, should trailers be parsed and passed along?
- If the server does not include a `trailers: ...` header indicating which trailer headers are being returned, should trailers be parsed and passed along? I think this one is almost certainly "yes".
- If the server includes trailers with invalid header names, such as `Content-Length: 5`, should that still be parsed and passed along?

### DIFF
--- a/src/proto/h1/decode.rs
+++ b/src/proto/h1/decode.rs
@@ -4,8 +4,10 @@ use std::io;
 use std::task::{Context, Poll};
 use std::usize;
 
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use futures_util::ready;
+use http::{HeaderMap, HeaderName, HeaderValue};
+use http_body::Frame;
 
 use super::io::MemRead;
 use super::DecodedLength;
@@ -26,7 +28,8 @@ pub(crate) struct Decoder {
     kind: Kind,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq)]
+// FIXME: can we keep Copy trait?
+#[derive(Debug, Clone, PartialEq)]
 enum Kind {
     /// A Reader used when a Content-Length header is passed with a positive integer.
     Length(u64),
@@ -35,6 +38,8 @@ enum Kind {
         state: ChunkedState,
         chunk_len: u64,
         extensions_cnt: u64,
+        trailers_buf: Option<BytesMut>,
+        trailers_cnt: u64,
     },
     /// A Reader used for responses that don't indicate a length or chunked.
     ///
@@ -87,6 +92,8 @@ impl Decoder {
                 state: ChunkedState::new(),
                 chunk_len: 0,
                 extensions_cnt: 0,
+                trailers_buf: None,
+                trailers_cnt: 0,
             },
         }
     }
@@ -123,12 +130,12 @@ impl Decoder {
         &mut self,
         cx: &mut Context<'_>,
         body: &mut R,
-    ) -> Poll<Result<Bytes, io::Error>> {
+    ) -> Poll<Result<Frame<Bytes>, io::Error>> {
         trace!("decode; state={:?}", self.kind);
         match self.kind {
             Length(ref mut remaining) => {
                 if *remaining == 0 {
-                    Poll::Ready(Ok(Bytes::new()))
+                    Poll::Ready(Ok(Frame::data(Bytes::new())))
                 } else {
                     let to_read = *remaining as usize;
                     let buf = ready!(body.read_mem(cx, to_read))?;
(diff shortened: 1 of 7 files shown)


tests/server.rs
```diff
@@ -2743,6 +2743,31 @@ fn http1_trailer_fields_not_allowed() {
...
+        0\r\n\
+        chunky-trailer: header data\r\n\
+        \r\n\
+    ",
+    )
+    .expect("writing");
+
+    assert_eq!(server.body(), b"hello");
+
+    // FIXME: add support for checking trailers that server received
```
hjr3, line 2768: I will add this tomorrow.

hjr3: added in [1d5eaf4](https://github.com/hyperium/hyper/pull/3637/commits/1d5eaf4979628d51a274c560bbab88e21864b3ed)

src/proto/h1/decode.rs
```diff
@@ -26,7 +28,8 @@ pub(crate) struct Decoder {
     kind: Kind,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq)]
+// FIXME: can we keep Copy trait?
```
hjr3, line 31: is it fine to drop `Copy` here? this type is internal and nothing seemed to break.

src/proto/h1/decode.rs
```diff
@@ -143,45 +150,72 @@ impl Decoder {
...
+                    ))?;
                     if *state == ChunkedState::End {
                         trace!("end of chunked");
-                        return Poll::Ready(Ok(Bytes::new()));
+
+                        if trailers_buf.is_some() {
+                            trace!("found possible trailers");
+                            match decode_trailers(
+                                &mut trailers_buf.take().expect("Trailer is None"),
+                                // decoder enforces that trailers count will not exceed usize::MAX
```
dswij, line 182: where are we enforcing that?

src/proto/h1/decode.rs
dswij: It's fine, I think. If the compiler doesn't complain then we're not really using `Copy` anywhere internally.

src/proto/h1/decode.rs
```diff
@@ -442,19 +478,46 @@ impl ChunkedState {
...
+        trailers_buf: &mut Option<BytesMut>,
+        trailers_cnt: &mut u64,
     ) -> Poll<Result<ChunkedState, io::Error>> {
-        match byte!(rdr, cx) {
-            b'\n' => Poll::Ready(Ok(ChunkedState::EndCr)),
+        let byte = byte!(rdr, cx);
+        match byte {
+            b'\n' => {
+                if *trailers_cnt == usize::MAX as u64 {
+                    return Poll::Ready(Err(io::Error::new(
```
dswij, line 507: Can we write some tests covering this as well?

dswij (commented):
Thanks for the PR! It looks great overall.

> If the server does not include a trailers: ... header indicating which trailer headers are being returned, should trailers be parsed and passed along? I think this one is almost certainly "yes".

Yes. I think hyper should pass this along and let users decide what to do with the extra trailers.

> If the server includes trailers with invalid header names, such as Content-Length: 5, should that still be parsed and passed along?

I'm not too sure about this one. I see two possible options:

1. Pass it along but mark it as unexpected
2. Err it

I'm interested to see what the others think about this

src/proto/h1/decode.rs
hjr3: here: https://github.com/hyperium/hyper/pull/3637/files#diff-2ef0d3180cacbd6f072b1728bd8c8a9d0c132df381876f264b67e18bb4a87d5fR506

hjr3: i also added a check in [545f8c4](https://github.com/hyperium/hyper/pull/3637/commits/545f8c48d3da589cf7a4154aa6433db1bbe07076)

src/proto/h1/decode.rs
hjr3: added in [545f8c4](https://github.com/hyperium/hyper/pull/3637/commits/545f8c48d3da589cf7a4154aa6433db1bbe07076)

i also lowered the trailer limit to 1024 as i cannot imagine a use case for sending a million trailers.

hjr3: I added a size limit in [f194be1](https://github.com/hyperium/hyper/pull/3637/commits/f194be13cd6fd15823101a8c775252269136c07c)

src/proto/h1/decode.rs
```diff
@@ -19,8 +19,11 @@ use self::Kind::{Chunked, Eof, Length};
 /// This limit is currentlty applied for the entire body, not per chunk.
 const CHUNKED_EXTENSIONS_LIMIT: u64 = 1024 * 16;
 
-/// Maximum number of trailers allowed in a chunked body.
-const TRAILERS_LIMIT: u64 = 1024;
+/// Maximum number of trailer fields allowed in a chunked body.
+const TRAILERS_FIELD_LIMIT: u64 = 1024;
+
+/// Maximum number of bytes allowed for all trailer fields.
+const TRAILER_LIMIT: u64 = TRAILERS_FIELD_LIMIT * 64;
```
dswij, line 26: I wonder if we want to expose an option to the user for this.

hjr3: sure. when i think about the use cases for this:

- the client sends `TE: trailers` and wants to limit the amount of trailers to read in the server's response
- the client sends a chunked body with trailer fields and the server wants to limit the amount of trailers to read in the client's request

Looking at prior art:

- We can include `h1_max_trailer_size` and `h1_max_trailer_fields` in `client::conn::http1::Builder`
- We can include `h1_max_trailer_size` and `h1_max_trailer_fields` in `server::conn::http1::Builder`

dswij: Let's add this in this PR too, I think it's pretty simple

dswij: Let's add this in this PR too, I think it's simple enough to include this

seanmonstar: I've been thinking of adding a `max_header_size()` for the HTTP/1 builder, to mirror the HTTP/2 side. Does it make sense for that to be used by both headers and trailers? Or do people need to allow different sizes?

hjr3: Most uses of trailer fields, that I have observed, add 1-2 headers that are relatively short. I don't think people would want more capacity than what was for `max_header_size()`. People may want to set a lower capacity for trailers than what `max_header_size()` allows, but I am unsure how common this would be.

I think we should use `max_header_size()` to enforce both for now. If there is a real use case that people bring up, we can always add `max_trailers()` and `max_trailers_size()` in the future.

If we want to go this route, then I can update the trailer parsing logic to check against `max_headers()`. In a future PR, I can:

- Add `max_header_size()` to the client and server builders.
- Update the trailer parsing logic to check against `max_headers_size()`
- Update the header parsing logic `src/proto/h1/role.rs` to enforce `max_headers_size()`

hjr3: i added 6c01c879e9f62a543e8865e536dd7d1561619a1f to resolve this

dswij (approved):
LGTM, thanks for the PR!

Anything you want to add? @seanmonstar

seanmonstar (approved):
Alright, just got to review it all. Looks really good, thank you!

The one thing I think that could help is to add some docs about the support, including what is and isn't enforced. But then, as I looked around so as to suggest where, I realized that there's likely several behavioral things for HTTP/1 or HTTP/2 that aren't really documented. So, in that case, it doesn't seem fair to block this on wanting the docs. Rather, I'll open an issue to decide on where to put those, and what things are missing.

hjr3:
@seanmonstar Please feel free to assign me any relevant issues. I already some blog posts on sending trailers:

- https://hermanradtke.com/streaming-body-trailing-headers-axum-0-6/
- https://hermanradtke.com/streaming-body-trailing-headers-hyper-1-0-and-axum-0-7/

I had planned on writing one for receiving trailers. I had a note to ask you if you wanted this sort of content in examples, the hyper.rs guides, etc.

PR #3643, opened 2024-04-22
feat(http1): add option allowng leading whitespaces before first header

### LINKED ISSUE #3285: Leading whitespace in first HTTP/1 response header?
**Version**
current git master

**Platform**
Linux

**Description**
hyper rejects HTTP responses where the first header has leading whitespace. This is spec compliant.

The popular browsers accept such headers and therefore we are now converting curl to (again) also allow them (in https://github.com/curl/curl/pull/11607).

hyper's current behavior prevents it from getting content other HTTP widely used clients can fetch. Do you think that is perhaps reason to reconsider this stance and allow this violation? It would make life in curl easier as right now curl+hyper builds have different behavior than curl with its native HTTP implementation.

dswij:
closes #3285


hjr3:
This appears to be blocked on:

- a new version of httparse that includes https://github.com/seanmonstar/httparse/pull/147
- running `sh capi/gen_header.sh`

dswij:
Yes. This PR is waiting for a new release of httparse to include the httparse changes you mentioned.

seanmonstar:
There was an httpparse v1.9 release that you can refer to, now.

nox:
Do browsers actually parse the header, or do they ignore that header line just like they do for other kinds of invalid header lines? There was already an option to ignore invalid header lines in httparse, did it not make httparse silently ignore those initial headers with leading whitespace?

seanmonstar:
The feature here is welcome, but I'm going to close this due to inactivity. If someone wants to update and re-submit, you're welcome to!

Issue #912, opened 2016-09-09
why `send` method automatic set `Date` header?


fengcen:
I found the code in `write_head()`:

    if !self.headers.has::<header::Date>() {
        self.headers.set(header::Date(header::HttpDate(now_utc())));
    }

I don't want to set `Date` header. What should I do?

seanmonstar:
The spec says it's required. See the last sentence of the second paragraph in [RFC7231#7.1.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.1.2):

> An origin server MUST send a Date header field[...].

smithsps:
RFC7231#7.1.1.2 - ` An origin server MUST NOT send a Date header field if it does not
   have a clock capable of providing a reasonable approximation of the
   current instance in Coordinated Universal Time.`

It'd be uncommon, but it's technically possible that hyper could be used on such a system.

fengcen:
Ok. Thank you for the information.

seanmonstar:
@smithsps it could be, so I wouldn't object to some sort of config, I suppose, but I'm left wondering if it realistically ever happens.

smithsps:
@seanmonstar agreed, the only thing I could think of could be clockless embedded device that for some reason didn't have a NTP sync.

PR #3644, opened 2024-04-22
Customization of whether server date headers are sent

edwardwc:
Initially starting with issue #912 then followed up with pull request #2751, this pull request addresses the need for customizing whether the server sends the `date` header for http/2 and http/1.1.

According to RFC7231#7.1.1.2, it is not always preferable to send the `date` header if the system may not have access to time or may not be accurate. Additionally, disabling the `date` header will result in slight performance gains. 

This pull request aims to allow users to customize when building the http1 or http2 service whether the date headers should be sent in responses, with a similar API to other boolean options. It also integrates tests for both http1 and 2 to prevent regression.

If you have suggestions, questions, comments, please comment on this pull request! I will promptly make changes.

Have an incredible rest of your day.


tusharmath (approved):

src/server/conn/http1.rs
```diff
@@ -359,6 +361,16 @@ impl Builder {
         self
     }
 
+    /// Set whether the `date` header should be included in HTTP responses.
+    ///
+    /// Note that including the `date` header is recommended by RFC 7231.
+    ///
+    /// Default is true.
+    pub fn add_date_header(&mut self, enabled: bool) -> &mut Self {
```
seanmonstar, line 369: I think I'd lean towards making the name `auto_date_header(bool)`, which matches some of the other "auto" options.

edwardwc:
@seanmonstar auto_date_header does make more sense!

seanmonstar (approved):

PR #3648, opened 2024-04-25
fix(http2): "Trailer" header should be allowed in HTTP/2 responses as per RFC 9110

lqs:
This pull request addresses the incorrect handling of the `Trailer` header in HTTP/2 responses within the hyper project. According to RFC 9110 "HTTP Semantics", the `Trailer` header is not listed as a "hop-by-hop" header in section 7.6.1. [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html) specifies that a sender intending to generate one or more trailer fields in a message should use the `Trailer` header in the header section to indicate potential trailer fields, as mentioned in section 6.6.2.

The current implementation in hyper may result in unnecessary warnings for gRPC calls, as gRPC implementations, such as grpc-go, routinely use the `Trailer` header to indicate the presence of trailers after the response body. (See this behavior in [grpc-go's implementation](https://github.com/grpc/grpc-go/blob/d32e66ce27447a0a217464a36fdd3935801c0453/internal/transport/handler_server.go#L306).

This fix updates hyper to correctly allow the `Trailer` header in HTTP/2 responses, aligning it with RFC 9110 and preventing misleading warnings in environments using gRPC, such as linkerd2-proxy, which prints a warning `Connection header illegal in HTTP/2: trailer` on every gRPC call.

Note: [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection) mistakenly lists `Trailer` as a hop-by-hop header, which is incorrect.

### DIFF
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -8,7 +8,7 @@ use std::task::{Context, Poll};
 use bytes::{Buf, Bytes};
 use futures_util::ready;
 use h2::{Reason, RecvStream, SendStream};
-use http::header::{HeaderName, CONNECTION, TE, TRAILER, TRANSFER_ENCODING, UPGRADE};
+use http::header::{HeaderName, CONNECTION, TE, TRANSFER_ENCODING, UPGRADE};
 use http::HeaderMap;
 use pin_project_lite::pin_project;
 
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Nice catch! When paying attention, it seems obvious. We just blindly trusted the MDN page without thinking, woops!

Issue #3652, opened 2024-04-29
Issues with client termination of H2 CONNECT streams


howardjohn:
**Version**
Hyper 1.3

**Platform**
```
6.8.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 17 Apr 2024 15:20:28 +0000 x86_64 GNU/Linux
```

**Description**

We are seeing a few issues around using H2 CONNECT. Our application is using Tokio, Rustls, and Hyper to communicate between a client and server both using the same stack. We have multiple streams on one TCP connection sometimes (though the issue occurs regardless of multiplexing).

Our first issue that popped up was leaking connections. This was debugging and an (attempted) fix was opened in https://github.com/hyperium/hyper/pull/3647. More details there. This was like not detected before because:
* The CONNECT tests in the repo all close from the server side
* When using hyper's pool this isn't seen since the pool will hold onto the connection after all streams are closed (until the idle timeout). We only started seeing this when we stopped using hyper's pool and started closing the connection immediately after the last stream closed.

With that fix, everything was working fine. However, a later refactor in our broke things again; we started dropping the `SendRequest` before we were complete with IO operations on `Upgraded` (before, we dropped the `SendRequest` _after_). This causes the stream/connection to be closed unexpectedly. This can be reproduced easily by moving the `drop(client)` in https://github.com/hyperium/hyper/pull/3647 to before the write/read operations. This is easy to workaround, but its not documented and @seanmonstar suggested it was not expected on Discord.

seanmonstar:
That is surprising. The `Connection` is design to stay alive until all shared references to it are gone. Those are the `SendRequest`, and any `ResponseFuture`, `SendStream`, and `RecvStream`. Perhaps something about the way the types are stored inside the `Upgraded` makes it drop the reference and no longer be counted for liveness?

cc @nox, who worked on this previously.

howardjohn:
Ok I think I figured it out and put up two PRs:
* Test in hyper https://github.com/hyperium/hyper/pull/3655
* Fix in h2 https://github.com/hyperium/h2/pull/772

PR #3655, opened 2024-05-01
Add test for HTTP2 CONNECT termination

howardjohn:
For https://github.com/hyperium/hyper/issues/3652.

In https://github.com/hyperium/hyper/pull/3647, I thought I tracked down the root cause of https://github.com/hyperium/hyper/issues/3652. However, with the help of @seanmonstar  in https://github.com/hyperium/hyper/issues/3652#issuecomment-2085359895 I realized that this was too forcefully terminating the connection even if things other than SendRequest were still alive. I also realized the original test case I wrote was just wrong and failed to properly reproduce the issue.

This adds a test that properly reproduces the issue. On my machine, it fails about 5% of the time:
```
1876 runs so far, 100 failures (94.94% pass rate). 95.197349ms avg, 1.097347435s max, 5.398457ms min
```

With further investigation, I believe this bug actually originates in `h2` itself. https://github.com/hyperium/h2/pull/772. With that PR, this test is 100% reliable
```
64010 runs so far, 0 failures (100.00% pass rate). 44.484057ms avg, 121.454709ms max, 1.872657ms min
```

### DIFF
--- a/tests/client.rs
+++ b/tests/client.rs
@@ -2009,6 +2009,82 @@ mod conn {
             .expect_err("client should be closed");
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn http2_connect_detect_close() {
+        // Regression test for failure to fully close connections when using HTTP2 CONNECT
+        // We send 2 requests and then drop them. We should see the connection gracefully close.
+        use futures_util::future;
+        let (listener, addr) = setup_tk_test_server().await;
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):

seanmonstar:
Thank you! I was waiting to release `h2` with the fix, so we don't add a flaky test to hyper. The h2 v0.4.5 is out now :)

howardjohn:
Ah makes sense. Thanks!

PR #3656, opened 2024-05-03
Remove incorrect comment in example

Tacklebox:
There was a comment in the service_struct_impl.rs example indicating that the default match arm would somehow return a 404 and not increment the counter. This wasn't the case. That arm also had a redundant return.


seanmonstar (approved):
Thanks :)

PR #3667, opened 2024-05-13
remove 'static lifetime bound on http1/2 client IO

jgraef:
The method [`hyper::client::conn::http1::Builder::handshake`](https://docs.rs/hyper/latest/hyper/client/conn/http1/struct.Builder.html#method.handshake) has a `'static` lifetime bound on the type parameter `T` (the underlying connection). This lifetime bound is not necessary and prohibits users of the library to borrowing connections to a hyper client.

There are similar bounds on the `Connection` struct, `handshake` method, plus all those for http2.

This PR removes these `'static` lifetime bounds.

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -51,7 +51,7 @@ pub struct Parts<T> {
 #[must_use = "futures do nothing unless polled"]
 pub struct Connection<T, B>
 where
-    T: Read + Write + 'static,
+    T: Read + Write,
     B: Body + 'static,
 {
     inner: Dispatcher<T, B>,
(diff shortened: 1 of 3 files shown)


jgraef:
There's at least one `'static` that I missed[1]. I will update this PR soon.

There are also some `Unpin` trait bounds that are not necessary. Should I open a new PR for them?

[1]: https://github.com/jgraef/hyper/blob/7b78fc6b36b1972fb084fde9b3246643f87eabfc/src/client/conn/http1.rs#L62

seanmonstar:
Rebase or new commits, whichever you find easier to push. We can always squash when merging.

seanmonstar:
Seems like rustfmt isn't happy. Could you run `rustfmt --check --edition 2021 $(git ls-files '*.rs')`?

seanmonstar (approved):

PR #3668, opened 2024-05-18
refactor(lib): apply small refactoring

tottoto:
Applies small refactoring.

### DIFF
--- a/src/ext/h1_reason_phrase.rs
+++ b/src/ext/h1_reason_phrase.rs
@@ -147,7 +147,7 @@ const fn is_valid_byte(b: u8) -> bool {
     //
     // The 0xFF comparison is technically redundant, but it matches the text of the spec more
     // clearly and will be optimized away.
-    #[allow(unused_comparisons)]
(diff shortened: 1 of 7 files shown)


src/ext/h1_reason_phrase.rs
```diff
@@ -147,7 +147,7 @@ const fn is_valid_byte(b: u8) -> bool {
     //
     // The 0xFF comparison is technically redundant, but it matches the text of the spec more
     // clearly and will be optimized away.
-    #[allow(unused_comparisons)]
+    #[allow(unused_comparisons, clippy::absurd_extreme_comparisons)]
```
seanmonstar, line 150: Why does clippy think this is absurd? They look like valid comparisons to me...

seanmonstar (approved):

src/ext/h1_reason_phrase.rs
tottoto: clippy seems to treat comparisons with values beyond what the type can represent as absurd.

https://rust-lang.github.io/rust-clippy/rust-1.78.0/index.html#/absurd_extreme_comparisons

PR #3678, opened 2024-05-27
Add unexpected cfgs to `[lints]` table

Urgau:
With the release of rust-lang/cargo#13913 (in nightly-2024-05-19), there is no longer any need for the kind of workarounds employed in #3660. Cargo has now gain the ability to declare `--check-cfg` args directly inside the `[lints]` table with [`[lints.rust.unexpected_cfgs.check-cfg]`](https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html#check-cfg-in-lintsrust-table)[^1]:

`Cargo.toml`:
```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(hyper_unstable_tracing,hyper_unstable_ffi)'] }
```

This PR adds those unexpected cfgs inside the `[lints]`, reverts #3660 and remove the manual `--cfg docsrs` (since it's automaticly added by docs.rs).

[^1]: take effect on Rust 1.80 (current nightly), is ignored on Rust 1.79 (current beta), and produce an unused warning below

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,7 +102,7 @@ check-cfg = [
 
 [package.metadata.docs.rs]
 features = ["ffi", "full", "tracing"]
-rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi", "--cfg", "hyper_unstable_tracing"]
+rustdoc-args = ["--cfg", "hyper_unstable_ffi", "--cfg", "hyper_unstable_tracing"]
 
 [package.metadata.playground]
 features = ["full"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,13 @@ tracing = ["dep:tracing"]
 # internal features used in CI
 nightly = []
 
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = [
+    'cfg(hyper_unstable_tracing)',
+    'cfg(hyper_unstable_ffi)'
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -93,9 +93,16 @@ tracing = ["dep:tracing"]
...
+level = "warn"
+check-cfg = [
+    'cfg(hyper_unstable_tracing)',
+    'cfg(hyper_unstable_ffi)'
+]
+
 [package.metadata.docs.rs]
 features = ["ffi", "full", "tracing"]
-rustdoc-args = ["--cfg", "docsrs", "--cfg", "hyper_unstable_ffi", "--cfg", "hyper_unstable_tracing"]
+rustdoc-args = ["--cfg", "hyper_unstable_ffi", "--cfg", "hyper_unstable_tracing"]
```
tottoto, line 105: Is this a related change?

Urgau: Yes, I'm trying to make sure that all custom cfgs are declared, and `docsrs` is no longer a custom cfg since https://github.com/rust-lang/docs.rs/pull/2390. So, for consistency with above I removed it.

tottoto: Ok. I understand what you are addressing, but I would think it should be a separated pull request.

seanmonstar (approved):
Thank you!

PR #3682, opened 2024-06-05
Make Future implementation on Connection unconditional on executor being Send + Sync.

mstyura:
This PR is created to raise the question if `Future` implementation on `Connection` is mandatory to request Executor to be `Send` + `Sync`.
Question initially raised here https://discord.com/channels/500028886025895936/670880858630258689/1247563996068970557.

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -230,7 +230,7 @@ where
     B::Data: Send,
     E: Unpin,
     B::Error: Into<Box<dyn Error + Send + Sync>>,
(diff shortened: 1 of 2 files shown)


mstyura:
Interesting, CI does `rustfmt  --edition 2021 $(git ls-files '*.rs')` and it formats code, while `cargo fmt` does not.

seanmonstar:
I think this looks good, any reason it was marked a draft?

mstyura:
It was initially supposed to raise a question, that's why I've marked it as draft. But as long as you are ok with changes I will unmask as draft and fix issues if any  according to review

seanmonstar (approved):

PR #3693, opened 2024-06-27
improve uri docs for send_request of http1 send request

GlenDC:
This behaviour surprised me as discussed over Discord.
As such here is a small patch in an attempt to be more clear about the fact
that hyper has a hands-off approach regarding the Uri field in the subject line of Http1 requests.


seanmonstar (approved):

seanmonstar:
There was a couple terms that needed to be adjusted to be "right", so I did them locally and committed, but kept most the rest of it. Thanks!

PR #3700, opened 2024-07-10
Expose ReadBufCursor::remaining and ReadBufCursor::put_slice

nox:
There is currently no way to write to ReadBufCursor without unsafe code, even though writing a slice to it like one would do to a Buf shouldn't require unsafe code.

### DIFF
--- a/src/rt/io.rs
+++ b/src/rt/io.rs
@@ -244,19 +244,29 @@ impl<'data> ReadBufCursor<'data> {
         self.buf.init = self.buf.filled.max(self.buf.init);
     }
 
+    /// Returns the number of bytes that can be written from the current
+    /// position until the end of the buffer is reached.
+    ///
+    /// This value is equal to the length of the slice returned by `as_mut()``.
     #[inline]
-    pub(crate) fn remaining(&self) -> usize {
+    pub fn remaining(&self) -> usize {
         self.buf.remaining()
     }
 
+    /// Transfer bytes into `self`` from `src` and advance the cursor
+    /// by the number of bytes written.
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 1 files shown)


nox:
There is also #3515 that is similar, but more involved.

src/rt/io.rs
```diff
@@ -244,19 +244,27 @@ impl<'data> ReadBufCursor<'data> {
         self.buf.init = self.buf.filled.max(self.buf.init);
     }
 
+    /// Returns the number of bytes that can be written from the current
+    /// position until the end of the buffer is reached.
+    ///
+    /// This value is equal tho the length of the slice returned by `as_mut()``.
```
seanmonstar, line 250:
```suggestion
    /// This value is equal to the length of the slice returned by `as_mut()``.
```

src/rt/io.rs
```diff
@@ -244,19 +244,27 @@ impl<'data> ReadBufCursor<'data> {
...
     #[inline]
-    pub(crate) fn remaining(&self) -> usize {
+    pub fn remaining(&self) -> usize {
         self.buf.remaining()
     }
 
+    /// Transfer bytes into `self`` from `src` and advance the cursor
+    /// by the number of bytes written.
+    ///
+    /// `self` must have enough remaining capacity to contain all of `src`.
```
seanmonstar, line 259: How about adding a `# Panics` heading in between these two lines?

seanmonstar (changes requested):
Seems fair to add these :)

nox:
@seanmonstar I addressed both comments.

nox:
Can we release 1.5 to get this please?

PR #3705, opened 2024-07-16
Allow building an HTTP/1 connection with prebuffered bytes

### LINKED ISSUE #3704: `Upgraded::downcast_ref` is unusable with `hyper_util::server::conn::auto`
`hyper_util::server::conn::auto::Connection<'a, I, S, E>` internally wraps `I` in `hyper_util::common::rewind::Rewind<_>`, which is a private type, so we cannot ever make use of `Upgraded::downcast_ref` anymore.

While I could expose `Rewind<_>` in `hyper_util` and improve docs in `hyper_util::server::conn::auto` to signal that one should be downcasting to `Rewind<I>`, I am filing this in `hyper` because I thought of a potentially better resolution:

1. Introduce `hyper::server::conn::http1::Builder::serve_buffered_connection`:<br><br>
    ```rust
    impl Builder {
        pub fn serve_buffered_connection<I, S>(&self, buffered: Bytes, io: I, service: S) -> Connection<I, S>
        where
            S: HttpService<IncomingBody>,
            S::ResBody: 'static,
            <S::ResBody as Body>::Error: Into<Box<dyn StdError + Send + Sync>>,
            I: Read + Write + Unpin,
        {
            let mut conn = proto::Conn::new_buffered(buffered, io);
            …
        }
    }
    ```
2. Use that in `hyper_util::server::conn::auto::Builder::serve_connection` instead of wrapping `I` into `Rewind<_>`.

nox:
Fixes #3704.

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -47,9 +47,14 @@ where
     B: Buf,
     T: Http1Transaction,
 {
+    #[cfg(any(all(feature = "client", feature = "http1"), test))]
     pub(crate) fn new(io: I) -> Conn<I, B, T> {
+        Self::new_buffered(Bytes::new(), io)
+    }
+
+    pub(crate) fn new_buffered(buffered: Bytes, io: I) -> Conn<I, B, T> {
         Conn {
-            io: Buffered::new(io),
+            io: Buffered::new_buffered(buffered, io),
             state: State {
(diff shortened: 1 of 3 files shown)


seanmonstar:
The issue was solved in a different way, by providing `hyper_util::server::conn::auto::upgrade::downcast()`.

PR #3718, opened 2024-07-30
fix(http1): make `date_header` effective

yhx-12243:
`date_header` option is not work in http1 because the argument in `Builder` is forgotten to pass into the further connection object.


seanmonstar (approved):
Thank you!

PR #3723, opened 2024-08-02
tests: demonstrate max_header_list_size hang

ajwerner:
This test demonstrates a bug in hyper where large response headers can result in a hang.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -61,6 +61,30 @@ t! {
             ;
 }
 
+// Test that clients get responses when the server sends a response with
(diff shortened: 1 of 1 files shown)


ajwerner:
This test hangs until https://github.com/hyperium/h2/pull/791 gets merged. After that h2 is released, I'll revive this in order to close https://github.com/hyperium/hyper/issues/3724. Until then @tottoto could you cancel the actions associated with this PR so they don't keep burning CPU cycles?

tottoto:
Canceled the action.

PR #3725, opened 2024-08-02
fix(http1): send 'connection: close' during graceful shutdown

### LINKED ISSUE #3720: No `Connection: close` on HTTP1 Connection Drain
When HTTP1 connection draining is activated, the Connection: close header is not attached to responses sent to active connections. This prevents active clients from realizing that the server is requesting that the connection be closed and prevents graceful draining of HTTP1 connections.

seanmonstar:
Closes #3720


seanmonstar:
This is probably a good idea? Is there any reason this would we bad?

dayvejones:
Thanks! I'll be waiting for a new release with this fix:)

PR #3731, opened 2024-08-10
feat(client): Add more Http2 client settings options

0x676e67:
Hmm, I noticed that hyper is missing some configuration options for http2 that apply to my scenario. If that makes sense, merged to the mainline branch? I thought there might be someone else who might need it

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -363,22 +363,6 @@ where
         self
     }
 
-    /// Enables or disables server push promises.
-    ///
-    /// This value is included in the initial SETTINGS handshake.
-    /// Setting this value to value to
-    /// false in the initial SETTINGS handshake guarantees that the remote server
-    /// will never send a push promise.
-    ///
-    /// This setting can be changed during the life of a single HTTP/2
-    /// connection by sending another settings frame updating the value.
-    ///
-    /// If not set, hyper will use a default.
-    pub fn enable_push(&mut self, enabled: impl Into<Option<bool>>) -> &mut Self {
-        self.h2_builder.enable_push = enabled.into();
-        self
-    }
-
     /// Sets the maximum number of concurrent streams.
     ///
     /// The maximum concurrent streams setting only controls the maximum number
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Thank you!

src/client/conn/http2.rs
```diff
@@ -355,6 +353,62 @@ where
...
+    /// This value is included in the initial SETTINGS handshake.
+    /// Setting this value to value to
+    /// false in the initial SETTINGS handshake guarantees that the remote server
+    /// will never send a push promise.
+    ///
+    /// This setting can be changed during the life of a single HTTP/2
+    /// connection by sending another settings frame updating the value.
+    ///
+    /// Default value of crate `h2`: `true`.
+    pub fn enable_push(&mut self, enabled: bool) -> &mut Self {
```
seanmonstar, line 379: Actually, on second thought: since we don't have a way for users to handle push promises, it probably doesn't make as much sense to enable the option.

0x676e67: Well, we will change it back to the default disabled option value `false`

src/client/conn/http2.rs
```diff
@@ -355,6 +351,62 @@ where
...
+    /// This value is included in the initial SETTINGS handshake.
+    /// Setting this value to value to
+    /// false in the initial SETTINGS handshake guarantees that the remote server
+    /// will never send a push promise.
+    ///
+    /// This setting can be changed during the life of a single HTTP/2
+    /// connection by sending another settings frame updating the value.
+    ///
+    /// If not set, hyper will use a default.
+    pub fn enable_push(&mut self, enabled: impl Into<Option<bool>>) -> &mut Self {
```
seanmonstar, line 377: I don't think it needs an option bool. The method just shouldn't exist yet, until accessing the push promises is possible.

seanmonstar (changes requested):

src/client/conn/http2.rs
0x676e67: Okay, let's revert enable push back to its original state. We'll add a method to set it when accessing push promises becomes possible.

seanmonstar (approved):

PR #3742, opened 2024-08-23
docs: link from Connection to handshake

WhyNotHugo:
When reading the documentation for `Connection` it is not entirely obvious how instances of this type are obtained. Add a helpful link, mostly for those less familiar.

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -48,6 +48,8 @@ pub struct Parts<T> {
 ///
 /// In most cases, this should just be spawned into an executor, so that it
 /// can process incoming and outgoing messages, notice hangups, and the like.
+///
+/// Instances of this type are typically created via the [`handshake`] function
 #[must_use = "futures do nothing unless polled"]
 pub struct Connection<T, B>
 where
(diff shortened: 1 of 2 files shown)


tottoto:
Although I think these links are helpful, having independent sections sounds a bit excessive. How about adding it as inline explanations, similar to the server's documentation?

- https://docs.rs/hyper/1.4.1/hyper/server/conn/http1/struct.Connection.html
- https://docs.rs/hyper/1.4.1/hyper/server/conn/http2/struct.Connection.html

tottoto:
Although I think these links are helpful, having independent sections sounds a bit excessive. How about adding them as inline explanations, similar to the server's documentation?

- https://docs.rs/hyper/1.4.1/hyper/server/conn/http1/struct.Connection.html
- https://docs.rs/hyper/1.4.1/hyper/server/conn/http2/struct.Connection.html

WhyNotHugo:
Yeah, the new section is pointless. Updated.

WhyNotHugo:
The explicit link was also unnecessary; the `handshake` function is in the same module.

tottoto (approved):

tottoto:
Thank you!

PR #3743, opened 2024-08-24
feat(server): add h1 `idle_timeout`

finnbear:
## Motivation
Older browsers sometimes open 6 or more H1 connections and keep them alive and idle after resources are loaded, costing server resources and distracting from DoS attackers. In my specific case, DoS mitigation required limiting TCP connections per IP address, but limits had to be high to account for multiple legitimate H1 browsers on the same IP address.

Related: #1628
Related: #2355

## Changes
- [x] Added `idle_timeout` setting, optimized to avoid creating an additional timer per request
- [ ] ~~Automatically send `keep-alive: timeout=#` header~~ (deferred due to: https://github.com/hyperium/http/issues/142)

## Testing

- [x] Fixed multiple pre-existing and broken, but relevant, tests
- [x] Added a test with idle timeout
- [x] Added a test without idle timeout (improves coverage)
- [x] Tested in my app

```toml
# Cargo.toml

[patch.crates-io]
hyper = { git = "https://github.com/finnbear/hyper", branch = "idle_timeout" }
hyper-util = { git = "https://github.com/finnbear/hyper-util", branch = "idle_timeout" }
```

## Future work
- [x] Support in `hyper-util` (https://github.com/hyperium/hyper-util/pull/146)
- [ ] Consider other timeouts (reading body, writing response)

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -61,9 +61,15 @@ where
                 #[cfg(feature = "server")]
                 h1_header_read_timeout: None,
                 #[cfg(feature = "server")]
-                h1_header_read_timeout_fut: None,
-                #[cfg(feature = "server")]
                 h1_header_read_timeout_running: false,
+                #[cfg(feature = "server")]
+                h1_idle_timeout: None,
+                #[cfg(feature = "server")]
+                h1_idle_timeout_running: false,
+                #[cfg(feature = "server")]
+                h1_timeout_fut: None,
+                #[cfg(feature = "server")]
+                first_head: true,
                 #[cfg(feature = "server")]
                 date_header: true,
                 #[cfg(feature = "server")]
@@ -148,6 +154,11 @@ where
     }
 
     #[cfg(feature = "server")]
+    pub(crate) fn set_http1_idle_timeout(&mut self, val: Duration) {
+        self.state.h1_idle_timeout = Some(val);
+    }
+
+    #[cfg(feature = "server")]
     pub(crate) fn set_allow_half_close(&mut self) {
         self.state.allow_half_close = true;
     }
@@ -216,25 +227,8 @@ where
         debug_assert!(self.can_read_head());
         trace!("Conn::read_head");
 
-        #[cfg(feature = "server")]
-        if !self.state.h1_header_read_timeout_running {
-            if let Some(h1_header_read_timeout) = self.state.h1_header_read_timeout {
-                let deadline = Instant::now() + h1_header_read_timeout;
-                self.state.h1_header_read_timeout_running = true;
-                match self.state.h1_header_read_timeout_fut {
-                    Some(ref mut h1_header_read_timeout_fut) => {
-                        trace!("resetting h1 header read timeout timer");
-                        self.state.timer.reset(h1_header_read_timeout_fut, deadline);
-                    }
-                    None => {
-                        trace!("setting h1 header read timeout timer");
-                        self.state.h1_header_read_timeout_fut =
-                            Some(self.state.timer.sleep_until(deadline));
-                    }
-                }
-            }
-        }
-
+        #[cfg_attr(not(feature = "server"), allow(unused))]
+        let mut progress = false;
         let msg = match self.io.parse::<T>(
             cx,
             ParseContext {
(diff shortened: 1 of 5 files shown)


src/proto/h1/conn.rs
```diff
@@ -61,10 +62,16 @@ where
...
                 #[cfg(feature = "server")]
-                h1_header_read_timeout_fut: None,
-                #[cfg(feature = "server")]
                 h1_header_read_timeout_running: false,
                 #[cfg(feature = "server")]
+                h1_idle_timeout: None,
+                #[cfg(feature = "server")]
+                h1_idle_timeout_running: false,
+                #[cfg(feature = "server")]
+                h1_timeout_fut: None,
```
finnbear, line 71: Used one timer for both header read and idle timeouts. In the case of idle -> header read, it is going to be reset instead of recreated.

src/proto/h1/conn.rs
```diff
@@ -61,10 +62,16 @@ where
...
+                #[cfg(feature = "server")]
+                first_request: true,
```
finnbear, line 73: Side note: It would be \*trivial\* to apply the idle timeout, instead of the header read timeout, to the first request (simply by removing this boolean and treating all requests like `!first_request`).

tests/server.rs
```diff
@@ -1590,31 +1586,40 @@ async fn header_read_timeout_slow_writes_multiple_requests() {
...
+            Is: 0\r\n\
+        ",
+        )
+        .expect("write 6");
+        thread::sleep(Duration::from_secs(3));
+        // header read timed out 1s ago
+        tcp.write_all(
+            b"\
+            Wrong: 0\r\n\
         ",
```
finnbear, line 1611: The header read timeout test was broken (failed due to invalid header after finished request, not timeout). This may explain how the broken read timeout went unnoticed. I have fixed it, and now it doubles as a test in which the idle timeout would trigger if not for the header read timeout.

tests/server.rs
```diff
@@ -1593,28 +1593,39 @@ async fn header_read_timeout_slow_writes_multiple_requests() {
...
         )
-        .expect_err("write 6");
+        .and(tcp.flush())
+        .expect_err("write 7");
     });
```
finnbear, line 1621: The header read timeout test was broken (failed due to invalid header after finished request, not timeout). This may explain how the broken read timeout went unnoticed. I have fixed it, and now it doubles as a test in which the idle timeout would trigger if not for the header read timeout.

src/proto/h1/conn.rs
```diff
@@ -217,24 +226,10 @@ where
...
-                        self.state.timer.reset(h1_header_read_timeout_fut, deadline);
-                    }
-                    None => {
-                        trace!("setting h1 header read timeout timer");
-                        self.state.h1_header_read_timeout_fut =
-                            Some(self.state.timer.sleep_until(deadline));
-                    }
-                }
-            }
-        }
```
finnbear, line 236: I was forced to move the code that starts the timeout after the parsing is polled, because which timeout to use depends on whether the connection is idle (no bytes read on all poll attempts) or not (some bytes read on at least one poll attempt).

My earlier attempts to predict whether bytes would be available based on the read buffer failed (the read buffer looks empty until the first poll of the io in parsing).

I also considered starting the idle timeout elsewhere, but the code would be more fragile as a single rouge poll of read_head would start the header read timer when it shouldn't be started.

src/proto/h1/conn.rs
```diff
@@ -1106,6 +1155,12 @@ impl State {
         self.reading = Reading::Init;
         self.writing = Writing::Init;
 
+        #[cfg(feature = "server")]
+        if self.h1_idle_timeout.is_some() {
+            // Next read will start and poll the idle timeout.
+            self.notify_read = true;
+        }
+
```
finnbear, line 1163: Based on the code immediately after, this function could unconditionally set `self.notify_read = true` with a slight performance penalty if the user disables the idle timeout (the only case in which that extra polling isn't necessary).

However, the benefit would be increased maintainability, as the polling would be easier to predict (the non-polling case, which probably lacks test coverage, would be eliminated). Let me know if you want me to make this change.

src/proto/h1/conn.rs
```diff
@@ -217,24 +226,10 @@ where
...
-                    }
-                    None => {
-                        trace!("setting h1 header read timeout timer");
-                        self.state.h1_header_read_timeout_fut =
-                            Some(self.state.timer.sleep_until(deadline));
-                    }
-                }
-            }
-        }
+        let first_head = !self.io.is_read_blocked();
```
finnbear, line 229: Anecdotally, the read is only unblocked for the first poll of the first head. If you prefer, I could add a `bool` to the connection to track whether the first request has been made.

MarkusSintonen:
@sfackler / @seanmonstar is there a possibility on getting this PR reviewed? I am also interested in getting H1 idle timeouts available :)

finnbear:
> I am also interested in getting H1 idle timeouts available :)
If you want, you can use the TOML snippet in my PR description to avoid needing to wait for review :)

finnbear:
Today I investigated further and reached the conclusion that this PR is unnecessary for my use-case because, contrary to my [prior belief](https://discord.com/channels/500028886025895936/670880858630258689/1276568656276619406), `header_read_timeout` is applied to all requests, not just the first one. I'm not sure why I didn't realize this before.

In other words, unmodified `hyper` works like this:
1. accept connection
2. start header read timeout
3. read header
4. stop header read timeout
5. read remainder of request and write response
6. skip to step 2

This PR would only be relevant for users who want to set _different_ values for header read timeout and idle timeout. Any such user is welcome to:
- explain their use case.
- resurrect this PR, possibly with modifications.
- if idle_timeout < header_read_timeout, make a [custom stream type](https://github.com/getsentry/relay/blob/730317cbfc42c0f0f7ac84f7f3d3a78abd8e9073/relay-server/src/services/server/io.rs#L11-L128) and use it with `header_read_timeout`.
- if idle_timeout > header_read_timeout, make a similar custom stream type, disable `header_read_timeout`, and reproduce the behavior of `header_read_timeout` by communicating with a middleware.

Considering the above, I will close this PR. I intend to open a new PR that only includes my fixes to the existing test cases.

finnbear:
It seems I confused myself by forgetting the details of my changes. Unmodified `hyper` does not "skip to step 2." That requires a simple modification, which was part of this PR:
image

PR #3753, opened 2024-09-12
chore(rt): delete the unnecessary symbol in the comments

hongmengning:
Delete the unnecessary symbol in the comments


hongmengning:
```
CI error: error: package `tokio-util v0.7.12` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.63.0
Error: Process completed with exit code 101.
```

It should have nothing to do with this change

seanmonstar (approved):

Issue #3752, opened 2024-09-11
Relax body requirements by removing `'static` bound


nanoqsh:
For now to send any requests you need request's body to be `'static`: [handshake](https://docs.rs/hyper/latest/hyper/client/conn/http1/fn.handshake.html)

I think this restriction is unnecessary for many async runtimes that allow borrowing external environment. My proposal is to remove this restriction, allowing types such as [`Full<&'a [u8]>`](https://docs.rs/http-body-util/0.1.2/http_body_util/struct.Full.html) to be used as a body. Besides, it may simplify the public API a little bit.

This can be done for http1. But the implementation of the http2 requires a [downcasting](https://doc.rust-lang.org/std/any/struct.TypeId.html#method.of) of body's [`Data`](https://docs.rs/hyper/latest/hyper/body/trait.Body.html#associatedtype.Data), so it must be `'static`. Anyway, even `<B as Body>::Data: 'static` is less restrictive than `B: Body + 'static`, it allows to use some borrowed body that produces static chunks of data. I'm not sure if it's possible to relax this restriction too. Maybe something can be invented in the future

raftario:
Would very much like the bounds to be relaxed too

PR #3755, opened 2024-09-12
fix: remove `'static` bounds for client's body

nanoqsh:
impl #3752


nanoqsh:
I closed the problem. I made workarounds for myself. If someone needs it you can open it again.

Issue #2845, opened 2022-05-20
Create a hyper::error::Code type


seanmonstar:
Create a new opaque struct `Code` in `hyper::error`, that [contains constants of errors](https://github.com/hyperium/hyper/blob/master/docs/ROADMAP.md#error) that could happen in hyper.

```rust
pub struct Code(Code_);

enum Code_ {
    Canceled,
    // etc
}

impl Code {
    pub const CANCELED: Code = Code(Code_::Canceled);
}
```

Should it be an opaque struct, or a non-exhaustive enum?

dswij:
How will this differ with `hyper::error::Error`?

dswij:
Since this is related to https://github.com/hyperium/hyper-util/pull/3, figured I'll take a look at this.

One thing I'm not too sure of is: how exactly does this `Code` differ from `hyper::error::Kind`?

seanmonstar:
It's not required to make progress on https://github.com/hyperium/hyper-util/pull/3, we can figure out the relationship afterwards. But to the question: the major difference is that `Code` is a public thing. The internal `Kind` is meant to be a cheaper way to assign specific error information, vs allocating a string that might not ever be printed. I expect `Kind` to probably hold more specific information than `Code` might expose.

GlenDC:
I think I do right by responding to this issue instead of making an ew one, but if not sorry for hijacking and I'll make a new one none the less, @seanmonstar .

I think this is the kind of thing I need and wouldn't mind contributing one of these weeks, just would need a bit of guidance.

Currently the error reporting of Hyper is pretty limited as exposed by the current Error struct <https://docs.rs/hyper/1.0.0-rc.4/hyper/struct.Error.html> (seems unchanged from 0.14).

In general it is fine, but if you want to be able to handle specific kinds of errors you're often now forced to downcast_ref, perhaps even a chain of such checks, which becomes very painful very quickly. E.g. you might be looking for a specific kind of IO Error (which can happen in different ways), or some specific h2 issue (e.g. invalid frame size), etc etc.. Right now checking these kind of things so you can handle failures in a proper manner (so instead of just logging and accept failure, that you can actually correct your mistake on the fly, or retry, or w/e else.

For this I would argue against another opaque type.. I mean the type can be opaque in its implementation (so if you prefer a struct over an Enum, I'm fine with that), but the other approach is fine as well. Most important however is that any kind of error can just be accessed just as easily, while now some seem more exposed then others.

[...]

PR #3760, opened 2024-10-04
add is_parse_method to hyper::Error

mzr:
Similarly to `is_parse_status` I'd like to have a mechanism I can depend on to have custom logic for HTTP method parse errors.

The `Kind`s of errors are private outside of the crate's code, so seems like helper functions are they way. There are some already, I'm just adding another one.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -189,6 +189,11 @@ impl Error {
         matches!(self.inner.kind, Kind::Parse(Parse::Status))
     }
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! I appreciate the goal, to make more of the error types actionable. However, I'd really much rather make progress on #2845, and come up with a nicer way to expose details about the error.

mzr:
Sure thing!

PR #3772, opened 2024-10-22
docs(lib): do not inline http rexports

seanmonstar:
I think this will make the docs clearer. Before this change, `HeaderMap` is separate as a re-export, and the other types all look native to hyper. This will show that most of the top level types in hyper are re-exports from `http`.


tottoto (approved):
I agree that this makes the docs clearer.

PR #3781, opened 2024-11-10
fix(http1): is_timeout works on HeaderTimeout

### LINKED ISSUE #3780: `hyper::Error::is_timeout` returns `false` on a `HeaderTimeout`.
**Version**
```
hyper = {version="1.5.0", features=["http1", "server"]}

# other dependencies
tokio = {version="1.41.1", features=["full"]}
hyper-util = {version = "0.1.10", features = ["full"]}
```

**Platform**
Output from `uname --all`:
```
Linux computer 6.7.11-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 27 16:50:39 UTC 2024 x86_64 GNU/Linux
```

**Description**
`hyper::Error::is_timeout` returns `false` on a `HeaderTimeout`.

On an HTTP/1.1 server, when I set a timeout with [`header_read_timeout`], a slow client will cause a `HeaderTimeout` error.
Calling [`is_timeout`] on the error should return `true`, but it returns `false`.

[`header_read_timeout`]: https://docs.rs/hyper/1.5.0/hyper/server/conn/http1/struct.Builder.html#method.header_read_timeout
[`is_timeout`]: https://docs.rs/hyper/1.5.0/hyper/struct.Error.html#method.is_timeout

To reproduce:
```rust
use std::convert::Infallible;
use std::net::SocketAddr;

use hyper::server::conn::http1;
use hyper::service::service_fn;
use hyper::Response;
use hyper_util::rt::{TokioIo, TokioTimer};
use tokio::time::Duration;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    let addr = SocketAddr::from(([127, 0, 0, 1], 1234));
    let listener = tokio::net::TcpListener::bind(addr).await?;
    loop {
        let (stream, _) = listener.accept().await?;
        let io = TokioIo::new(stream);
        let r: Result<Response<String>, Infallible> = Ok(Response::new("hello".to_string()));

        tokio::task::spawn(async move {
            if let Err(e) = http1::Builder::new()
                .timer(TokioTimer::new())
                .header_read_timeout(Duration::from_secs(1))
                .serve_connection(io, service_fn(|_| async { r.clone() }))
                .await
            {
                println!("Is {e:?} a timeout? {}", e.is_timeout());
            }
        });
    }
}
```
Open a TCP socket (`nc localhost 1234`) and hold it open for a second. This triggers the timeout, and the program prints `Is hyper::Error(HeaderTimeout) a timeout? false`.

T-aian:
Changed `hyper::Error::is_timeout()` to also check if the kind of the error is a `HeaderTimeout`. I also added a test at the end of `server.rs` to check for this kind of timeout error.

Fixes #3780

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -240,16 +240,11 @@ impl Error {
 
     /// Returns true if the error was caused by a timeout.
     pub fn is_timeout(&self) -> bool {
-        #[cfg(any(
-            all(feature = "http1", feature = "server"),
-            feature = "ffi"
-        ))]
-        return matches!(self.inner.kind, Kind::HeaderTimeout) || self.find_source::<TimedOut>().is_some();
+        #[cfg(any(all(feature = "http1", feature = "server"), feature = "ffi"))]
+        return matches!(self.inner.kind, Kind::HeaderTimeout)
+            || self.find_source::<TimedOut>().is_some();
 
-        #[cfg(not(any(
-            all(feature = "http1", feature = "server"),
-            feature = "ffi"
-        )))]
+        #[cfg(not(any(all(feature = "http1", feature = "server"), feature = "ffi")))]
         self.find_source::<TimedOut>().is_some()
     }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -240,6 +240,16 @@ impl Error {
 
     /// Returns true if the error was caused by a timeout.
     pub fn is_timeout(&self) -> bool {
+        #[cfg(any(
+            all(feature = "http1", feature = "server"),
+            feature = "ffi"
+        ))]
+        return matches!(self.inner.kind, Kind::HeaderTimeout) || self.find_source::<TimedOut>().is_some();
+
+        #[cfg(not(any(
+            all(feature = "http1", feature = "server"),
+            feature = "ffi"
+        )))]
         self.find_source::<TimedOut>().is_some()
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the PR! Looks like rustfmt is grumpy. `rustfmt --edition 2021 $(git ls-files '*.rs')` should fix it up.

src/error.rs
```diff
@@ -240,6 +240,11 @@ impl Error {
 
     /// Returns true if the error was caused by a timeout.
     pub fn is_timeout(&self) -> bool {
+        #[cfg(any(all(feature = "http1", feature = "server"), feature = "ffi"))]
+        return matches!(self.inner.kind, Kind::HeaderTimeout)
+            || self.find_source::<TimedOut>().is_some();
+
+        #[cfg(not(any(all(feature = "http1", feature = "server"), feature = "ffi")))]
```
seanmonstar, line 247: Looks like the feature flag combinations are tricky. Here's what I'd suggest:

```rust
#[cfg(all(feature = "http1", feature = "server"))]
if matches!(self.inner.kind, Kind::HeaderTimeout) {
    return true;
}

self.find_source::<TimedOut>().is_some()
```

PR #3795, opened 2024-11-26
fix: content length assertion for HEAD

### LINKED ISSUE #3794: HEAD request with empty body but set content-length triggers panic
**Version**
hyper 1.5.0

**Platform**
MacOS

**Description**

With HEAD requests it seems it is often the case that content-length is set to relevant GET body response, but payload is empty.
That's what for example tower_http [ServeDir](https://docs.rs/tower-http/latest/tower_http/services/struct.ServeDir.html) does. In debug mode this triggers panic:
https://github.com/hyperium/hyper/blob/master/src/proto/h1/role.rs#L704C36-L704C89

I guess this assertion should skip HEAD method completely or maybe it's more of a tower_http bug and body should define different sizehints here?

Release version works as expected, as assertion is skipped.

23doors:
fixes #3794

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -701,15 +701,15 @@ impl Server {
                             // same to help developers find bugs.
                             #[cfg(debug_assertions)]
                             {
-                                if let Some(len) = headers::content_length_parse(&value)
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks!

PR #3799, opened 2024-11-28
fix(server): fix failing tests for server feature

### LINKED ISSUE #3790: Make client tests work without server feature
Running `cargo test --features http1,http2,client` fails to compile because it uses some types only available when the `server` feature is also enabled. Fixing this would just require adding or adjusting `#[cfg(feature = ...)]` attributes, paying attention to the compiler error messages.

For CI, the `ffi` job can be adjusted to no longer enable `server` when running the tests.

ionionascu:
Update tests that require the server feature to only be compiled when the feature is enabled. After updating these tests, the server feature is no longer required for the ffi job when running tests.

Closes #3790

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -195,7 +195,7 @@ jobs:
       - name: Run FFI unit tests
         env:
           RUSTFLAGS: --cfg hyper_unstable_ffi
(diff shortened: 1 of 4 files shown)


seanmonstar (approved):
Thanks so much!

Issue #3708, opened 2024-07-17
The script that generates the header file is not available


spongehah:
When I delete the header file under `capi/include`, run the command: 
```SH
(base) spongehah@MacBook capi % ./gen_header.sh
cp: /Users/spongehah/Document/src/hyper/capi/include/hyper.h: No such file or directory
rm: /Users/spongehah/Document/src/hyper/capi/include/hyper.h.backup: No such file or directory
```

seanmonstar:
Why are you deleting the file first?

---

As I look through the script again, I see it creates a backup to be able to `diff` them, if you passed `--verify` (which is what we used in CI). I suppose the script could be modified to only create a backup if `--verify` was passed.

spongehah:
> Why are you deleting the file first?
> 
> As I look through the script again, I see it creates a backup to be able to `diff` them, if you passed `--verify` (which is what we used in CI). I suppose the script could be modified to only create a backup if `--verify` was passed.

I just want to verify that the script works

PR #3800, opened 2024-11-29
chore: Only backup when passing --verify

linyihai:
## What this PR want to do
Fixes https://github.com/hyperium/hyper/issues/3708

## How to test 
Testing with `--verify` and without `--verify`.


seanmonstar (approved):
Thanks!

Issue #3802, opened 2024-11-30
Reducing even eliminating all `clippy` warning 


linyihai:
**Is your feature request related to a problem? Please describe.**

Running `clippy` cmd, current workspace generate serval warnings 

`cargo clippy --workspace --all-targets --no-deps -- -D warnings`

output is

```
...

error: non-binding `let` on a future
   --> src/client/dispatch.rs:455:9
    |, 
455 |         let _ = tx.try_send(Custom(3)).unwrap();
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider awaiting the future or dropping explicitly with `std::mem::drop`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_future

error: could not compile `hyper` (lib test) due to 49 previous errors
```

**Describe the solution you'd like**

- Some of the more obvious fixes and can be automatically adopted, can be repaired
- Something that requires a little bit of thinking about how to improve, like defining types for complex expressions, like `Poll<Option<crate::Result<(MessageHead<T::Incoming>, DecodedLength, Wants)>>>`, 
- Add `clippy` CI workflow

**Describe alternatives you've considered**

**Additional context**

linyihai:
I would love to reduce clippy error prompts :)

seanmonstar:
Some of the things clippy finds are worth fixing! But just enough things I disagree with that I therefore don't automatically run clippy, nor do I keep it in a CI job.

So, if you see some things that seem like obvious wins, those are welcome in a PR! I might comment on some of them that I disagree with what clippy recommends, just so you know ahead of time :)

linyihai:
> Some of the things clippy finds are worth fixing! But just enough things I disagree with that I therefore don't automatically run clippy, nor do I keep it in a CI job.
> 
> So, if you see some things that seem like obvious wins, those are welcome in a PR! I might comment on some of them that I disagree with what clippy recommends, just so you know ahead of time :)

Thanks for you clarification.

PR #3805, opened 2024-12-03
refactor(clippy): Reduce some Clippy warnings

linyihai:
## What this PR want to do 

Tried to complete https://github.com/hyperium/hyper/issues/3802 as well as possible.

This PR wants to reduce some Clippy warnings that worth fixing.  Almost all of the changes are based on clippy prompts. And passed the MSRV 1.63 test.

Some of the warnings have not changed, mainly based on Clippy's false positives, and some clippy rules are newer than the current MSRV 1.63.

## How to test and review

Needs downgrade some crate versions to run on MSRV 1.63.
```
cargo update tokio --precise 1.38.1 
cargo update tokio-util --precise 0.7.11
cargo update hashbrown --precise 0.15.0
cargo update regex --precise 1.9.6
cargo update async-stream --precise 0.3.5
cargo update tokio-stream --precise 0.1.15
```

Run `cargo +1.63-x86_64-unknown-linux-gnu  clippy --workspace --no-deps --all-targets --features="full"  -- -D warnings`
The warnings/error reduces to 7.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -100,10 +100,8 @@ fn is_complete_fast(bytes: &[u8], prev_len: usize) -> bool {
             if bytes[i + 1..].chunks(3).next() == Some(&b"\n\r\n"[..]) {
                 return true;
             }
-        } else if b == b'\n' {
-            if bytes.get(i + 1) == Some(&b'\n') {
-                return true;
-            }
+        } else if b == b'\n' && bytes.get(i + 1) == Some(&b'\n') {
+            return true;
         }
     }
 
@@ -1622,7 +1620,7 @@ fn write_headers_original_case(
 struct FastWrite<'a>(&'a mut Vec<u8>);
 
 #[cfg(feature = "client")]
-impl<'a> fmt::Write for FastWrite<'a> {
+impl fmt::Write for FastWrite<'_> {
     #[inline]
     fn write_str(&mut self, s: &str) -> fmt::Result {
         extend(self.0, s.as_bytes());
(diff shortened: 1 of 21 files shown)


src/client/conn/http1.rs
```diff
@@ -302,6 +302,12 @@ where
 
 // ===== impl Builder
 
+impl Default for Builder {
```
seanmonstar, line 305: Since this _adds_ something new, let's leave it as a feature PR to be separate.

src/proto/h1/role.rs
```diff
@@ -77,7 +77,7 @@ where
         return Ok(None);
     }
 
-    let _entered = trace_span!("parse_headers");
+    trace_span!("parse_headers");
```
seanmonstar, line 80: I think this changes the functionality.

src/server/conn/http1.rs
```diff
@@ -227,6 +227,12 @@ where
 
 // ===== impl Builder =====
 
+impl Default for Builder {
```
seanmonstar, line 230: Leave this for a feature PR, too.

tests/client.rs
```diff
@@ -1827,7 +1827,7 @@ mod conn {
             sock.set_write_timeout(Some(Duration::from_secs(5)))
                 .unwrap();
             let mut buf = [0; 4096];
-            sock.read(&mut buf).expect("read 1");
+            let _ = sock.read(&mut buf).expect("read 1");
```
seanmonstar, line 1830: Why does clippy want this change? Seems like more noise, why can't we just ignore the integer?

seanmonstar (commented):
Thanks, several of these are fine changes. I comment inline on the ones I think clippy is wrong about.

src/client/conn/http1.rs
linyihai: OK.

src/proto/h1/role.rs
linyihai: Yes, This is a Clippy false positive warnings.

tests/client.rs
linyihai: My Vscode shows that it needs handle the return amount. 

Well, this isn't a very valuable change by comparison, and 'let _' is another way to ignore the return value

PR #3815, opened 2024-12-18
prototype support for 1xx informational responses

dicej:
This is a prototype intended to spur discussion about what support for 1xx informational responses should look like in a Hyper server.  The good news is that it works great (for HTTP/1 only, so far).  The bad news is it's kind of ugly.  Here's what I did:

- Add `ext::InformationalSender`, a type which wraps a `futures_channel::mspc::Sender<Response<()>>`.  This may be added as an extension to an inbound `Request` by the Hyper server, and the application and/or middleware may use it to send one or more informational responses before sending the real one.

- Add code to `proto::h1::dispatch` and friends to add such an extension to each inbound request and then poll the `Receiver` end along with the future representing the final response.  If the app never sends any informational responses, then everything proceeds as normal.  Otherwise, we send those responses as they become available until the final response is ready.

TODO items:
- [ ] Also support informational responses in the HTTP/2 server.
- [ ] Determine best way to handle when the app sends an informational response with a non-1xx status code.  Currently we just silently ignore it.
- [ ] Come up with a less hacky API?
- [ ] Add test coverage.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -143,10 +143,6 @@ pub(super) enum User {
     #[cfg(any(feature = "http1", feature = "http2"))]
     #[cfg(feature = "server")]
     UnexpectedHeader,
-    /// User tried to respond with a 1xx (not 101) response code.
-    #[cfg(feature = "http1")]
-    #[cfg(feature = "server")]
-    UnsupportedStatusCode,
 
     /// User tried polling for an upgrade that doesn't exist.
     NoUpgrade,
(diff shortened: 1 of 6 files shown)


seanmonstar:
Cool start, thanks!

I'd suggest trying to figure out what the API looks like for users as the next step. That's the part that we need to get right, and can't really change once it's stable. The implementation details inside we can optimize and refactor whenever.

dicej:
> I'd suggest trying to figure out what the API looks like for users as the next step.

@vikanezrimaya's [proposal](https://gist.github.com/vikanezrimaya/037101dc7b28de37ef03a47569213236#api-based-on-the-old-server-push-proposal) looks fine to me, and could easily be implemented using this PR as a foundation.  Any concerns with that approach?

PR #3820, opened 2025-01-01
fix(server): `max_local_error_reset_streams` function `mut self` -> `&mut self`

finnbear:
This PR fixes an inconsistency with the HTTP/2 builder, which I assume was unintentional. This is a breaking change. Hopefully it counts as acceptable breakage.

I searched and only found one public, non-fork user of `max_local_error_reset_streams` on GitHub (https://github.com/IGrok2/monorepo/blob/955a9b9bbed159b99637cf7576482af831acac7d/src/opnieuw/src/main.rs#L380-L388) and I think this PR wouldn't even break it.


seanmonstar:
Well that's a major 🤯🤡

dswij (approved):

PR #3821, opened 2025-01-03
docs(examples): add an example of handler based routing

randomairborne:
Adds an example showing more complex routing functionality than that in web_api. Builds off web_api with a more complex, reusable router framework.
This is a pretty complex example, it has more lines of code than any other example except single_thread. It may need to be pared down- do not hesitate to let me know if there's anything that's overcomplicated or needs to be changed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -193,12 +193,15 @@ name = "upgrades"
 path = "examples/upgrades.rs"
 required-features = ["full"]
 
-
 [[example]]
 name = "web_api"
 path = "examples/web_api.rs"
 required-features = ["full"]
 
+[[example]]
+name = "routing"
+path = "examples/routing.rs"
+required-features = ["full"]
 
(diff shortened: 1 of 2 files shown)


randomairborne:
I am curious- I noticed that the other examples tend to contain fairly few comments. Is this intentional?

dswij:
> I am curious- I noticed that the other examples tend to contain fairly few comments. Is this intentional?

I won't say it's intentional. More likely, the author thought the parts are easy to understand so a comment is unnecessary. 

If the comment can explain what's happening, it's good to have. But if it's more of a guide, then we can put it in the [website](https://hyper.rs/guides/1/) instead.

seanmonstar:
I appreciate the interesting example! However, I'm trying to now limit how many examples are in the repository, since they increase compile/test time and maintenance. It would certainly be an interesting blog post, though!

PR #3823, opened 2025-01-06
misc: mark `Body::channel()` as deprecated in 0.14

cratelyn:
this commit marks `Body::channel()` as deprecated in the 0.14 release.

#2970 (_see #2962_) made this interface private, and hyper 1.0 no longer includes this function. instead, users should wrap an `mpsc::Receiver<T>` in a `StreamBody`. this should help users migrating from hyper 0.14 to 1.0, by indicating that this interface is affected by the breaking changes in hyper 1.0, and indicating a suitable upgrade path.

### DIFF
--- a/src/body/body.rs
+++ b/src/body/body.rs
@@ -137,6 +137,13 @@ impl Body {
     ///
     /// Useful when wanting to stream chunks from another thread.
     #[inline]
+    #[cfg_attr(
+        feature = "deprecated",
+        deprecated(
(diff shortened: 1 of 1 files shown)


cratelyn:
with https://github.com/hyperium/http-body/pull/140 landed, there's now a much cleaner path for migration. calls to `hyper::Body::channel()` can be replaced with `http_body_util::channel::Channel::new(..)`. this mostly aligns with other types that have moved out to libraries like `hyper_util` or `http_body_util`.

now that there is a more direct 1:1 equivalent for those migrating to use, i think this can be closed.

PR #3824, opened 2025-01-09
http1 expose ignore_invalid_headers_in_requests opt

GlenDC:
Fixes https://github.com/hyperium/hyper/issues/3825

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ futures-channel = { version = "0.3", optional = true }
 futures-util = { version = "0.3", default-features = false, optional = true }
 h2 = { version = "0.4.2", optional = true }
 http-body-util = { version = "0.1", optional = true }
-httparse = { version = "1.8", optional = true }
+httparse = { version = "1.9", optional = true }
 httpdate = { version = "1.0", optional = true }
 itoa = { version = "1", optional = true }
 pin-project-lite = { version = "0.2.4", optional = true }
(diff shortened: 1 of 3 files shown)


seanmonstar:
This probably requires a newer version of `httparse` in the Cargo.toml, to the version that method was introduced.

GlenDC:
> This probably requires a newer version of `httparse` in the Cargo.toml, to the version that method was introduced.

Seems so: <https://github.com/seanmonstar/httparse/releases/tag/v1.9.0>

src/client/conn/http1.rs
```diff
@@ -402,6 +402,19 @@ impl Builder {
...
     }
 
+    /// Set whether HTTP/1 connections will silently ignored malformed header lines.
+    ///
+    /// If this is enabled and a header line does not start with a valid header
+    /// name, or does not include a colon at all, the line will be silently ignored
+    /// and no error will be reported.
+    ///
+    /// Default is false.
+    pub fn ignore_invalid_headers_in_requests(&mut self, enabled: bool) -> &mut Builder {
```
seanmonstar, line 412: This should be in the `server` module. We likely can also drop off the `_in_requests` suffix, since it's the only message that would make sense for a server (the responses cannot be constructed with invalid headers at all, before passing to the connection).

seanmonstar (changes requested):

src/client/conn/http1.rs
GlenDC: would be same for `ignore_invalid_headers_in_responses` though, but I guess you cannot drop that for backwards compatibility

seanmonstar (approved):
Perfect, thanks!

PR #3836, opened 2025-01-22
feat(http1): support request body limit

myyrakle:
I propose the ability to limit the size of the request body.

This could be implemented at the application level as well, but I think it would be more efficient to handle it at the hyper implementation level.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -70,6 +70,12 @@ pub(super) enum Kind {
         any(feature = "http1", feature = "http2")
     ))]
     Body,
+    /// Error when body is too large
+    #[cfg(all(
+        any(feature = "client", feature = "server"),
+        any(feature = "http1", feature = "http2")
+    ))]
+    BodyTooLarge,
     /// Error while writing a body to connection.
     #[cfg(all(
         any(feature = "client", feature = "server"),
@@ -333,6 +339,14 @@ impl Error {
         any(feature = "client", feature = "server"),
         any(feature = "http1", feature = "http2")
     ))]
+    pub(super) fn new_body_too_large() -> Error {
+        Error::new(Kind::BodyTooLarge)
+    }
+
+    #[cfg(all(
+        any(feature = "client", feature = "server"),
+        any(feature = "http1", feature = "http2")
+    ))]
     pub(super) fn new_body_write<E: Into<Cause>>(cause: E) -> Error {
         Error::new(Kind::BodyWrite).with(cause)
     }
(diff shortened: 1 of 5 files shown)


sfackler:
How would it be more efficient to handle it in Hyper?

myyrakle:
> How would it be more efficient to handle it in Hyper?

From what I understand, the current hyper implementation receives and processes the request body by dividing it into chunks when the request body is large.

If I process TooLarge at the application level, I can reject it only after receiving all the chunks, but if I process it in hyper, I think I can reduce resources by stopping before all the chunks are received.

seanmonstar:
You can make use of this wrapper: https://docs.rs/http-body-util/latest/http_body_util/struct.Limited.html

myyrakle:
> You can make use of this wrapper: https://docs.rs/http-body-util/latest/http_body_util/struct.Limited.html

I'm a little confused because I don't know the overall structure of hyper. 
If I'm misunderstanding anything, please point out. 

As you advised, I used Limited to set a limit. 
```rust 
...
async fn hello(
    request: Request<hyper::body::Incoming>,
) -> Result<Response<Full<Bytes>>, Infallible> {
    let limited = Limited::new(request, 100);

    match limited.collect().await {
        Ok(body) => {
            let bytes = body.to_bytes();

            println!(
                "Received request: {:?}",
                String::from(std::str::from_utf8(&bytes).unwrap())
            );

            Ok(Response::new(Full::new(Bytes::from("Hello, World!"))))
        }
        Err(err) => {
            eprintln!("Error: {}", err);
            Ok(Response::new(Full::new(Bytes::from(
                "Request body too large",
            ))))
        }
    }
}
...
```
And it worked fine.

However, even if the limit was set to `Limited`, it did not seem to stop the frame read loop of `poll_read_body` midway. 
image
image

Is this a sufficiently efficient structure? Or did I use it wrong?

seanmonstar:
hyper streams the request body only as you ask for it. It does so in efficient chunks. So, what happens when you use the `Limited` helper is it will keep polling for another chunk (since you called `collect`), until it notices that the length has exceed the limit you set. It's possible that total of the chunks is a little bigger than the limit, but don't worry about that.

The tools we mentioned are sufficiently efficient.

myyrakle:
> hyper streams the request body only as you ask for it. It does so in efficient chunks. So, what happens when you use the `Limited` helper is it will keep polling for another chunk (since you called `collect`), until it notices that the length has exceed the limit you set. It's possible that total of the chunks is a little bigger than the limit, but don't worry about that.
> 
> The tools we mentioned are sufficiently efficient.

I understand, thank you for your kind reply.

I'll close this pr.

PR #3843, opened 2025-01-29
export Incoming::empty

myyrakle:
Change the existing `Incoming::empty` function from `pub(crate)` to `pub. 
Previously, it was almost impossible to create an Incoming object outside of hyper. 
 
If you can use this function, it may be of some help when writing common code. 
(At least in my case. It's a bit of an exception.)

This is a bit minor, but I think it has to do with API design direction. 

What do you think about this?

### DIFF
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -139,8 +139,8 @@ impl Incoming {
         Incoming { kind }
     }
 
-    #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


seanmonstar:
It purposefully has no public constructors, as mentioned in the docs of the type.

Users are not supposed to create this, but use a type more specific to their needs, like from `http-body-util`.

PR #3417, opened 2023-11-16
Introduce hyper::ext::Http1RawMessage

nox:

PR #3846, opened 2025-02-17
Allow recording the raw HTTP/1 headers sent and received

dis-archive:
This allows feeding them into a WARC file (https://en.wikipedia.org/wiki/WARC_(file_format)). The rest of the request and response is already available as it's either the explicitly set body or the received response body.

This includes the final `\r\n\r\n` between the headers and the response (so that it can be distinguished from bare `\n\n`).

Needed for a similar request in reqwest: https://github.com/seanmonstar/reqwest/issues/1229.

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -112,6 +112,8 @@ pub struct Builder {
     h1_parser_config: ParserConfig,
     h1_writev: Option<bool>,
     h1_title_case_headers: bool,
+    h1_record_raw_request_headers: bool,
+    h1_record_raw_response_headers: bool,
     h1_preserve_header_case: bool,
     h1_max_headers: Option<usize>,
     #[cfg(feature = "ffi")]
@@ -312,6 +314,8 @@ impl Builder {
             h1_read_buf_exact_size: None,
             h1_parser_config: Default::default(),
             h1_title_case_headers: false,
+            h1_record_raw_request_headers: false,
+            h1_record_raw_response_headers: false,
             h1_preserve_header_case: false,
             h1_max_headers: None,
             #[cfg(feature = "ffi")]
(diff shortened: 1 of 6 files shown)


seanmonstar:
We had something kinda like this for a little while for the C API, but then removed it. There's https://github.com/hyperium/hyper/pull/3417 that re-introduces it, though it needed to target master.

It looks like this does a little more: you distinguish between request and response. Why not use use a single type? I think it's because you're trying to capture what is encoded too. FWIW, that could only be done for the client, the server wouldn't have a way to give that to the user.

Depending on the purpose, a logging connection type could also do this without any help from hyper, such as reqwest's `connection_verbose()` option.

dis-archive:
> Why not use use a single type? I think it's because you're trying to capture what is encoded too.

Yes, that's why. I want to capture both what headers are sent and what headers are received, in a byte-perfect fashion.

> Depending on the purpose, a logging connection type could also do this without any help from hyper, such as reqwest's `connection_verbose()` option.

The purpose is recording requests I make into [WARC files](https://en.wikipedia.org/wiki/WARC_(file_format)), i.e. permanently storing requests I make and responses I get through reqwest. I could try parsing that out of `connection_verbose()` logs, but that would be more brittle than I'd like.

dis-archive:
I fixed the rustfmt, can I get another CI run?

dis-archive:
I fixed the errors I saw in CI, previously I hadn't tested the `nightly` feature flag locally. Hopefully, everything works now.

I'd like to request another CI run.

It won't help for this PR, because the setting is not applied retroactively, but there's a setting to allow CI runs also for new contributors, it's in repository settings → "Actions" → "Approval for running fork pull request workflows from contributors " → "Require approval for first-time contributors who are new to GitHub (Only users who are both new on GitHub and who have never had a commit or pull request merged into this repository will require approval to run workflows.)".

dis-archive:
Hi. Can I get a decision on this PR? A "no" would also help me to find out that I'll have to permanently fork this crate or look for something else if I don't want to keep my fork forever.

I have followup PRs to expose this in hyper-util or reqwest.

PR #3855, opened 2025-03-16
chore: remove redundant backticks

jimmycathy:
remove redundant backticks


seanmonstar (approved):
Ah thanks for catching this!

PR #3863, opened 2025-03-23
feat: add Error::is_shutdown()

joshka:
Benchmarking tools often cause connections to be ungracefully shutdown.
This makes it possible to filter these errors out.

Replaces #2745

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -238,6 +238,15 @@ impl Error {
         matches!(self.inner.kind, Kind::User(User::BodyWriteAborted))
     }
 
+    /// Returns true if the error was caused while calling `AsyncWrite::shutdown()`.
+    pub fn is_shutdown(&self) -> bool {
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -238,6 +238,14 @@ impl Error {
         matches!(self.inner.kind, Kind::User(User::BodyWriteAborted))
     }
 
+    /// Returns true if the error was caused while calling `AsyncWrite::shutdown()`.
+    pub fn is_shutdown(&self) -> bool {
+        #[cfg(all(feature = "http1", any(feature = "client", feature = "server")))]
+        matches!(self.inner.kind, Kind::Shutdown);
```
dswij, line 244: You'll need to remove the semicolon

```suggestion
        matches!(self.inner.kind, Kind::Shutdown)
```

joshka: oops - yep

joshka:
ping

seanmonstar (approved):

PR #3864, opened 2025-03-24
Gate tests with features they test

WhyNotHugo:
Running `cargo test` fails to compile because tests rely on types which are behind a #[cfg(…)] which is disabled by default.

Add a #[cfg(…)] directive to tests which rely on types which are also behind a #[cfg(…)] directive, so that these tests run only if the types on which they depend exist.


seanmonstar (approved):
Thanks for the clean up!

PR #3865, opened 2025-03-25
refactor: use cfg-aliases for feature flags

joshka:
Introduces aliases for each feature flag as well as for the main
combinations of http version and client/server roles. This makes it
significantly easier to understand which feature flags are required on
any given line as the aliases require less parsing of nested `cfg`
expressions.

- Drops 122 LoC.
- Each cfg attribute is about half the width
- No more complex any() within all() conditions

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -135,7 +135,7 @@ pub(super) enum User {
     ManualUpgrade,
 
     /// The dispatch task is gone.
-    #[cfg(any(http1_client, http1_server))]
+    #[cfg(any(http_client))]
     DispatchGone,
 
     /// User aborted in an FFI callback.


seanmonstar:
It does look cleaner, true! I'd rather not take a build-dependency, though.

I tried to make things easier with a few cfg_blah macros, but they didn't work out well in practice.

joshka:
> It does look cleaner, true! I'd rather not take a build-dependency, though.

My personal calculus would be to rely on a well used crate like this (69M downloads on 6 versions, vs hyper's 278M on 244 versions), but I can see why you might want to avoid that in hyper. I will expand the macros into their effect. It's mostly just a feature gated println.

> I tried to make things easier with a few cfg_blah macros, but they didn't work out well in practice.

I think there's some useful ideas there in grouping code and adding doc comments, but using [doc_auto_cfg](https://doc.rust-lang.org/rustdoc/unstable-features.html#doc_auto_cfg-automatically-generate-doccfg) is often a bit simpler unless there's specific situations where that doesn't work.

src/mock.rs
```diff
@@ -1,37 +1,37 @@
 // FIXME: re-implement tests with `async/await`
 /*
-#[cfg(feature = "runtime")]
+#[cfg(runtime)]
```
joshka, line 3: Doing some more digging, this feature doesn't exist in the cargo.toml. Should it be removed from the source or added to the features list?

seanmonstar:
I should have said more! Not just a build-dependency, but the build script. There's a light positive pressure generally to avoid them unless _building_ something, because they slow down compiler parallelism.

joshka:
> I should have said more! Not just a build-dependency, but the build script. There's a light positive pressure generally to avoid them unless _building_ something, because they slow down compiler parallelism.

That sounds like a good general rule to follow, but also one which leads to negative outcomes if followed without thinking (let's not fall into a "cargo" cult mentality here). Can we measure the impact of this change on build times perhaps? I would intuitively expect that it's so negligible as to be meaningless as a blocker, but I'm often surprised when such assumptions are false. How would you confirm that that? Do you perhaps have a very low CPU system that you can point at that would show the difference like a Raspberry Pi 1 or similar?

hanna-kruppe:
That's the wrong benchmark: it's measuring (1) incremental builds (2) with a no-op change to the library itself but none to the build script, but (3) unnecessarily including the cost of cargo asking `git` for changes that may require rebuilding the build script (because there's no `cargo:rerun-if-changed=...` output). For the many thousands projects that depend on hyper, what matters is the scenario where hyper and its build script are both built from scratch. Besides the impact that has on critical path length, the time for building and linking even trivial build scripts can be significant as previously discussed e.g. in https://github.com/rust-lang/log/issues/489

joshka:
Can you provide a different number  that shows the impact of this?

hanna-kruppe:
Your diff doesn't replace hyper in the dependency tree of tokio-console, it adds the local version *on top* (you can verify this with `cargo tree -i hyper`). That version also has no features enabled, and I believe path dependencies are treated somewhat differently by Cargo compared to git and registry dependencies. The right way to test this is the `[patch]` table (which also requires updating the existing hyper dependency to match the version in git), like this:

```diff
diff --git a/Cargo.lock b/Cargo.lock
index bc83a65..8daee0f 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -821,9 +821,8 @@ dependencies = [

 [[package]]
 name = "hyper"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bbbff0a806a4728c99295b254c8838933b5b082d75e3cb70c8dab21fdfbcfa9a"
+version = "1.6.0"
+source = "git+https://github.com/joshka/hyper.git?rev=83488c120d10dfd5a92aa5f6ff7f92355a9ebba1#83488c120d10dfd5a92aa5f6ff7f92355a9ebba1"
 dependencies = [
  "bytes",
  "futures-channel",
diff --git a/Cargo.toml b/Cargo.toml
index 1eb3e8a..4b9ff78 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,3 +32,6 @@ install-updater = false
 [profile.dist]
 inherits = "release"
 lto = "thin"
+
+[patch.crates-io]
+hyper = { git = "https://github.com/joshka/hyper.git", rev = "83488c120d10dfd5a92aa5f6ff7f92355a9ebba1" }
```

With this change, I get these numbers (your branch first, hyper 1.5.0 from crates.io second):

```
$ hyperfine --prepare 'cargo clean' 'cargo build'
Benchmark 1: cargo build
  Time (mean ± σ):     38.198 s ±  0.501 s    [User: 229.797 s, System: 19.679 s]
  Range (min … max):   37.632 s … 39.110 s    10 runs
$ git stash -- Cargo.toml Cargo.lock
Saved working directory and index state WIP on main: 2bd1afd update(api,subscriber)!: upgrade tonic to 0.13 (#615)
$ hyperfine --prepare 'cargo clean' 'cargo build'
Benchmark 1: cargo build
  Time (mean ± σ):     38.053 s ±  0.230 s    [User: 229.485 s, System: 19.600 s]
  Range (min … max):   37.769 s … 38.387 s    10 runs
```

Is this significant? It's probably hard for a human to notice because 38 seconds is way outside "I sit there watching the progress bar fill up" territory. On the other hand, if you multiply the CPU time by how many times hyper is compiled every month, it doesn't seem negligible any more. (The Rust ecosystem in general scores pretty terribly on this metric, but that's no reason to ignore it entirely.) Note that this computer has a CPU from 2017 (Intel i7-6700K) but it was high end back then and still works perfectly fine for most things -- definitely better than the newer laptop I sometimes use for coding while traveling.

That aside, tokio-console has a relatively large dependency graph (245 "units" in `cargo --timings`) and isn't exactly a small and lightweight thing to compile (far from the worst in the Rust world, but 38 seconds is nothing to sneeze at). I'm personally much more interested in the lower end of the scale, basically what you need for a minimal HTTP1 server with hyper + tokio and little else. (Or HTTP1 client, but for that there's also ureq as lighter alternative). I would have liked to test the effect your branch has for something a binary with only these dependencies (with main.rs adapted from hyper's hello world example):

```toml
[dependencies]
bytes = "1.8.0"
http-body-util = "0.1.2"
hyper = { version = "1.5.0", features = ["http1", "server"] }
hyper-util = { version = "0.1.10", features = ["tokio"] }
tokio = { version = "1.41.1", features = ["net", "rt-multi-thread"] }
```

... but unfortunately your branch doesn't compile with that feature set:

```
   Compiling hyper v1.6.0 (https://github.com/joshka/hyper.git?rev=83488c120d10dfd5a92aa5f6ff7f92355a9ebba1#83488c12)
error[E0004]: non-exhaustive patterns: `error::Kind::User(User::DispatchGone)` not covered
   --> /home/hanna/.cargo/git/checkouts/hyper-8d72082f643922ed/83488c1/src/error.rs:369:15
    |
369 |         match self.inner.kind {
    |               ^^^^^^^^^^^^^^^ pattern `error::Kind::User(User::DispatchGone)` not covered
    |
```

Regardless, if the *absolute* impact were to be similar to the effect I saw in tokio-console, I'd consider that noteworthy on a project that otherwise builds in 4 seconds. In this example it probably won't have the same wall time impact because there's a lot of time where everything waits on tokio and there's spare CPU cores to use for a small build script. But I have other projects (not public) that build in 10 seconds or so where I might use hyper in the future which have more dependencies and more than enough work for all my CPU cores during most of the build.

src/error.rs
```diff
@@ -42,125 +42,104 @@ pub(super) enum Kind {
...
     NoUpgrade,
 
     /// User polled for an upgrade, but low-level API is not using upgrades.
-    #[cfg(all(any(feature = "client", feature = "server"), feature = "http1"))]
+    #[cfg(any(http1_client, http1_server))]
     ManualUpgrade,
 
     /// The dispatch task is gone.
-    #[cfg(all(feature = "client", any(feature = "http1", feature = "http2")))]
+    #[cfg(any(http1_client, http1_server))]
```
joshka, line 138: This is wrong, it should be `http_client` here.

joshka:
> ... but unfortunately your branch doesn't compile with that feature set:

Looks like I failed on a boolean expansion somewhere (which is a pain because if I've missed one then all the expansions are now suspect). Mostly I did these by regex replacement, but it's entirely possible that I missed a bracket somewhere and broke things. If this was to be considered for merge, I'd do a comprehensive audit of the changes to ensure that this is correct.

joshka:
Ugh on not noticing that it was using 1.5 there. Apologies.

> I would have liked to test the effect your branch has for something a binary with only these dependencies (with main.rs adapted from hyper's hello world example):

I fixed just the buggy enum that you mentioned (noting that I'm not sure if this problem is more widespread - I'd have to check it out fully and redo the conversions to be sure)

Using the hello-world example from axum source code (@ latest commit as of now af6a595f)

```diff
diff --git i/Cargo.toml w/Cargo.toml
index cfbfb615..de75834b 100644
--- i/Cargo.toml
+++ w/Cargo.toml
@@ -53,3 +53,6 @@ verbose_file_reads = "warn"
 # These have been fixed in the past, but are still present in the changelog.
 DefaultOnFailedUpdgrade = "DefaultOnFailedUpdgrade"
 OnFailedUpdgrade = "OnFailedUpdgrade"
+
+[patch.crates-io]
+hyper = { version = "1.6.0", path = "../hyper" }
```

```
❯ cargo tree -p example-hello-world
\example-hello-world v0.1.0 (/Users/joshka/local/axum/examples/hello-world)
├── axum v0.8.3 (/Users/joshka/local/axum/axum)
│   ├── axum-core v0.5.2 (/Users/joshka/local/axum/axum-core)
│   │   ├── bytes v1.10.1
│   │   ├── futures-core v0.3.31
│   │   ├── http v1.3.1
│   │   │   ├── bytes v1.10.1
│   │   │   ├── fnv v1.0.7
│   │   │   └── itoa v1.0.15
│   │   ├── http-body v1.0.1
│   │   │   ├── bytes v1.10.1
│   │   │   └── http v1.3.1 (*)
│   │   ├── http-body-util v0.1.3
│   │   │   ├── bytes v1.10.1
│   │   │   ├── futures-core v0.3.31
│   │   │   ├── http v1.3.1 (*)
│   │   │   ├── http-body v1.0.1 (*)
│   │   │   └── pin-project-lite v0.2.16
│   │   ├── mime v0.3.17
│   │   ├── pin-project-lite v0.2.16
│   │   ├── rustversion v1.0.20 (proc-macro)
│   │   ├── sync_wrapper v1.0.2
│   │   ├── tower-layer v0.3.3
│   │   ├── tower-service v0.3.3
│   │   └── tracing v0.1.41
│   │       ├── log v0.4.27
│   │       ├── pin-project-lite v0.2.16
│   │       └── tracing-core v0.1.33
│   │           └── once_cell v1.21.3
│   ├── bytes v1.10.1
│   ├── form_urlencoded v1.2.1
│   │   └── percent-encoding v2.3.1
│   ├── futures-util v0.3.31
│   │   ├── futures-core v0.3.31
│   │   ├── futures-task v0.3.31
│   │   ├── pin-project-lite v0.2.16
│   │   └── pin-utils v0.1.0
│   ├── http v1.3.1 (*)
│   ├── http-body v1.0.1 (*)
│   ├── http-body-util v0.1.3 (*)
│   ├── hyper v1.6.0 (/Users/joshka/local/hyper)
│   │   ├── bytes v1.10.1
│   │   ├── futures-channel v0.3.31
│   │   │   └── futures-core v0.3.31
│   │   ├── futures-util v0.3.31 (*)
│   │   ├── http v1.3.1 (*)
│   │   ├── http-body v1.0.1 (*)
│   │   ├── httparse v1.10.1
│   │   ├── httpdate v1.0.3
│   │   ├── itoa v1.0.15
│   │   ├── pin-project-lite v0.2.16
│   │   ├── smallvec v1.14.0
│   │   └── tokio v1.44.1
│   │       ├── bytes v1.10.1
│   │       ├── libc v0.2.171
│   │       ├── mio v1.0.3
│   │       │   └── libc v0.2.171
│   │       ├── parking_lot v0.12.3
│   │       │   ├── lock_api v0.4.12
│   │       │   │   └── scopeguard v1.2.0
│   │       │   │   [build-dependencies]
│   │       │   │   └── autocfg v1.4.0
│   │       │   └── parking_lot_core v0.9.10
│   │       │       ├── cfg-if v1.0.0
│   │       │       ├── libc v0.2.171
│   │       │       └── smallvec v1.14.0
│   │       ├── pin-project-lite v0.2.16
│   │       ├── signal-hook-registry v1.4.2
│   │       │   └── libc v0.2.171
│   │       ├── socket2 v0.5.9
│   │       │   └── libc v0.2.171
│   │       └── tokio-macros v2.5.0 (proc-macro)
│   │           ├── proc-macro2 v1.0.94
│   │           │   └── unicode-ident v1.0.18
│   │           ├── quote v1.0.40
│   │           │   └── proc-macro2 v1.0.94 (*)
│   │           └── syn v2.0.100
│   │               ├── proc-macro2 v1.0.94 (*)
│   │               ├── quote v1.0.40 (*)
│   │               └── unicode-ident v1.0.18
│   ├── hyper-util v0.1.11
│   │   ├── bytes v1.10.1
│   │   ├── futures-util v0.3.31 (*)
│   │   ├── http v1.3.1 (*)
│   │   ├── http-body v1.0.1 (*)
│   │   ├── hyper v1.6.0 (/Users/joshka/local/hyper) (*)
│   │   ├── pin-project-lite v0.2.16
│   │   ├── tokio v1.44.1 (*)
│   │   └── tower-service v0.3.3
│   ├── itoa v1.0.15
│   ├── matchit v0.8.4
│   ├── memchr v2.7.4
│   ├── mime v0.3.17
│   ├── percent-encoding v2.3.1
│   ├── pin-project-lite v0.2.16
│   ├── rustversion v1.0.20 (proc-macro)
│   ├── serde v1.0.219
│   ├── serde_json v1.0.140
│   │   ├── itoa v1.0.15
│   │   ├── memchr v2.7.4
│   │   ├── ryu v1.0.20
│   │   └── serde v1.0.219
│   ├── serde_path_to_error v0.1.17
│   │   ├── itoa v1.0.15
│   │   └── serde v1.0.219
│   ├── serde_urlencoded v0.7.1
│   │   ├── form_urlencoded v1.2.1 (*)
│   │   ├── itoa v1.0.15
│   │   ├── ryu v1.0.20
│   │   └── serde v1.0.219
│   ├── sync_wrapper v1.0.2
│   ├── tokio v1.44.1 (*)
│   ├── tower v0.5.2
│   │   ├── futures-core v0.3.31
│   │   ├── futures-util v0.3.31 (*)
│   │   ├── pin-project-lite v0.2.16
│   │   ├── sync_wrapper v1.0.2
│   │   ├── tokio v1.44.1 (*)
│   │   ├── tower-layer v0.3.3
│   │   ├── tower-service v0.3.3
│   │   └── tracing v0.1.41 (*)
│   ├── tower-layer v0.3.3
│   ├── tower-service v0.3.3
│   └── tracing v0.1.41 (*)
└── tokio v1.44.1 (*)
```

```
❯ hyperfine --parameter-list branch master,jm/cfg-aliases --prepare "(cd ../hyper; git switch {branch}); cargo clean" "cargo build -p example-hello-world"
Benchmark 1: cargo build -p example-hello-world (branch = master)
  Time (mean ± σ):      7.422 s ±  0.123 s    [User: 20.397 s, System: 3.141 s]
  Range (min … max):    7.314 s …  7.708 s    10 runs

Benchmark 2: cargo build -p example-hello-world (branch = jm/cfg-aliases)
  Time (mean ± σ):      7.892 s ±  0.125 s    [User: 21.126 s, System: 3.259 s]
  Range (min … max):    7.748 s …  8.123 s    10 runs

Summary
  cargo build -p example-hello-world (branch = master) ran
    1.06 ± 0.02 times faster than cargo build -p example-hello-world (branch = jm/cfg-aliases)
```

> Is this significant? It's probably hard for a human to notice because 38 seconds is way outside "I sit there watching the progress bar fill up" territory. On the other hand, if you multiply the CPU time by how many times hyper is compiled every month, it doesn't seem negligible any more. (The Rust ecosystem in general scores pretty terribly on this metric, but that's no reason to ignore it entirely.) Note that this computer has a CPU from 2017 (Intel i7-6700K) but it was high end back then and still works perfectly fine for most things -- definitely better than the newer laptop I sometimes use for coding while traveling.

I think it would be worth setting an actual target of what's good enough performance for build rather than this being some ambiguous idea. Without this everyone's tradeoffs are invisible constraints that are impossible to meet. I'm running a max  spec-ed at the time, but now 2 years old laptop here as my daily, which I acknowledge probably still puts it still well into the upper end of the performance bracket. For me personally the time vs complexity tradeoff is an obvious win on the side of simplicity here. I'd say that anything sub second is immaterial here.

hanna-kruppe:
I'm only a user of hyper, so I can't say anything worthwhile about how to balance improved maintainability (which only matters *directly* for maintainers) with build time regressions. If I was a maintainer, I could look at the value this provides to me and try to come up with a number for how much build time I'm willing to trade for it. I trust that @seanmonstar and others will do this in the best way for their own health and the health of the project.

But as user of hyper, there is no measurable regression I'd unconditionally accept as immaterial for me. My projects depend on many crates, not just hyper, and overall build time is influenced by all of them -- both by how long each crate takes to build in isolation and by interactions of the overall dependency graph (available parallelism, critical paths, feature unification, etc.). If every one of my 200 dependencies would accept a 5% or 0.1s build regression because none of them are noticeable for humans, build times for my project get very noticeably slower.

Naturally, this doesn't apply equally to all dependencies: it's more effective to spend effort on keeping build times low for crates that are compiled often and have non-trivial impact on downstream build times. Hyper scores relatively high on both metrics because of its popularity and because of its position in the dependency graph (downstream of tokio, upstream of most of the HTTP-related Rust ecosystem). That's why I thought it was worth my time to engage here in the first place. But again, that's just my priorities as a user of hyper: weighing that against maintainability of the code base is the maintainers' job.

PR #3869, opened 2025-04-02
docs(service): add `HttpService` documentation

cratelyn:
this commit introduces some additional documentation to the `HttpService` trait, and the `Service` trait.

notably, this commit introduces some intradoc links, so that rustdoc will render links to types like `http::Request` and `http::Response`, or to the `Body` trait.

additionally, mention of `hyper-util` is added to the `Service` trait, to direct users to the glue implementations that they will likely need to interact with e.g. `tower`.


seanmonstar (approved):
💪

PR #3887, opened 2025-05-08
fix(date): improve caching accuracy

Patryk27:
`CachedDate` gets updated every second, but the current logic doesn't take into account milliseconds and nanoseconds - if the application starts at `12:00:00.600`, hyper will report `date: ...T120000` up until `12:00:01.600`, which is overzealous.

We can sidestep this by subtracting the nanoseconds part (which includes milliseconds) from the time passed to `CachedDate::update()`, essentially rounding `self.next_update` to the beginning of the next closest second.

### DIFF
--- a/src/common/date.rs
+++ b/src/common/date.rs
@@ -1,7 +1,7 @@
 use std::cell::RefCell;
 use std::fmt::{self, Write};
 use std::str;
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Cool, thanks!

PR #3889, opened 2025-05-16
chore: further simplify BufList::copy_to_bytes

tisonkun:
@seanmonstar


tisonkun:
Ah .. My local fork specialize `T` as `Bytes` but hyper use a `T: Buf` generic param. So it does be different. Sorry for the invalid patch.

PR #3890, opened 2025-05-17
refactor(lib): drop futures-util except in ffi

hanna-kruppe:
Make hyper usable for h1/h2 and client/server without this heavyweight dependency. It's about 17k lines of code and takes up to 1.7 seconds to compile on my machine, but hyper is only using a tiny fraction of it. Larger applications probably still pull in futures-util by other means, but it's no longer as unavoidable as in the early days of the ecosystem.

To remove futures-util without raising MSRV, I took these steps:

* When futures-util just re-exports something from its dependencies, use it directly from the source.
* Inline trivial helpers like `poll_unpin` that "only" communicate intent a little better but don't save any significant amount of code.
* Refactor the h2 client code to avoid `StreamFuture` for the "Client has been dropped" detection -- just poll the mpsc channel directly.
* Implement a couple of small helpers from scratch when they're straightforward and fit on one screen each. The majority of this is polyfills for standard library APIs that would require a higher MSRV.
* Use `AtomicWaker` from the `atomic-waker` crate, a separately published copy of the futures-util type of the same name. While the two crates are owned by different organizations (smol-rs vs. rust-lang), it's mostly the same people maintaining both copies.

The uses of future-util in hyper's tests/benches/examples and in the `ffi` module seem much harder to remove entirely, so I did not touch those modules at all.

### DIFF
--- a/src/common/future.rs
+++ b/src/common/future.rs
@@ -1,0 +1,30 @@
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+// TODO: replace with `std::future::poll_fn` once MSRV >= 1.64
+pub(crate) fn poll_fn<T, F>(f: F) -> PollFn<F>
+where
+    F: FnMut(&mut Context<'_>) -> Poll<T>,
+{
+    PollFn { f }
+}
+
+pub(crate) struct PollFn<F> {
+    f: F,
+}
+
+impl<F> Unpin for PollFn<F> {}
+
+impl<T, F> Future for PollFn<F>
+where
+    F: FnMut(&mut Context<'_>) -> Poll<T>,
+{
+    type Output = T;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        (self.as_mut().f)(cx)
+    }
+}
--- a/src/proto/h2/client.rs
+++ b/src/proto/h2/client.rs
@@ -11,9 +11,7 @@ use crate::rt::{Read, Write};
 use bytes::Bytes;
 use futures_channel::mpsc::{Receiver, Sender};
 use futures_channel::{mpsc, oneshot};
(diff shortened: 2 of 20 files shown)


src/common/future.rs
hanna-kruppe, line 1: I'm not a fan of all these polyfills but I included them for now to show it's possible (and what it costs) to do without MSRV bump. I'd also be happy to bump the MSRV if that's acceptable.

src/proto/h2/client.rs
```diff
@@ -305,7 +302,7 @@ pin_project! {
         T: Unpin,
     {
         #[pin]
-        drop_rx: StreamFuture<Receiver<Infallible>>,
+        drop_rx: Receiver<Infallible>,
```
hanna-kruppe, line 305: I kept the changes here to a minimum to minimize the risk of regressions, but I wonder if this couldn't be a plain old `AtomicBool`?

src/proto/h2/client.rs
```diff
@@ -164,21 +163,19 @@ where
     // 'Client' has been dropped. This is to get around a bug
     // in h2 where dropping all SendRequests won't notify a
     // parked Connection.
```
hanna-kruppe, line 165: I definitely don't want to touch this in this PR, but I do wonder if that bug still exists. https://github.com/hyperium/h2/issues/502 seems relevant.

seanmonstar: Possibly! Would be happy to try having it removed in a separate PR.

src/proto/h2/client.rs
seanmonstar: I guess it depends if the bug mention above is handled. I think this was a oneshot channel so that the task would be woken up again.

seanmonstar (approved):
I think all this looks great, thanks so much!

Any other thoughts? I'd be happy to merge this as-is.

hanna-kruppe:
If you're happy with it then I'm happy as well :) Thanks for the quick review!

PR #3898, opened 2025-06-03
Backport `initial_max_send_streams` support to 0.14.x

Ralith:
This backports the relevant changes from https://github.com/hyperium/h2/issues/731. Happy to drop the changed default if minimizing disruption to existing users is preferred, but I suspect the current behavior is detrimental to most users (see e.g. https://github.com/facebook/buck2/issues/316) and the tweaked default will make things better.

### DIFF
--- a/src/client/client.rs
+++ b/src/client/client.rs
@@ -1228,6 +1228,26 @@ impl Builder {
         self
     }
 
+    /// Sets the initial maximum of locally initiated (send) streams.Add commentMore actions
+    ///
+    /// This value will be overwritten by the value included in the initial
+    /// SETTINGS frame received from the peer as part of a [connection preface].
+    ///
+    /// Passing `None` will do nothing.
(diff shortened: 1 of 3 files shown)


seanmonstar:
Just as a general note, v1 came out 1.5 years ago, and I'm reducing the amount of maintenance done to 0.14. I know there's less initial work if someone else backports, but if a new patch disrupts someone unexpectedly, I need to put more maintenance in a dying branch.

That's not to say we can't do this...

Ralith:
Yeah. Luckily it's a clean backport, so risk should be pretty low. Just need a decision on the changed default.

seanmonstar:
I'd rather not roll out a change to defaults in 0.14.x. Support from active contributors has ended.

Ralith:
It's been long enough that the downstream I care about has since updated to 1.x anyway.

PR #3908, opened 2025-06-27
refactor(proto/h1): simplify match version

dianqk:
This should mitigate https://github.com/rust-lang/rust/issues/140686 and https://github.com/seanmonstar/reqwest/issues/2687.

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -405,15 +405,18 @@ impl Http1Transaction for Server {
         {
             extend(dst, b"HTTP/1.1 200 OK\r\n");
         } else {
-            match msg.head.version {
-                Version::HTTP_10 => extend(dst, b"HTTP/1.0 "),
-                Version::HTTP_11 => extend(dst, b"HTTP/1.1 "),
-                Version::HTTP_2 => {
-                    debug!("response with HTTP2 version coerced to HTTP/1.1");
-                    extend(dst, b"HTTP/1.1 ");
-                }
-                other => panic!("unexpected response version: {:?}", other),
-            }
+            extend(
+                dst,
+                match msg.head.version {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! I agree with this statement in the upstream issue:

> The underlying issue causes miscompilation of so many things besides hyper that I would recommend not using such workarounds. I. recommend using either use an earlier toolchain (1.85?) or a later one (upcoming 1.89?) with the fix.

That the miscompilation causes the output of bad HTTP message in hyper is almost a good thing: it's very noticeable. So many other miscompilations likely exist that are much more invisible. It'd be better for anyone experiencing this to downgrade rustc for a while.

dianqk:
That's fine with me. BTW, upstream LLVM, which is upstream of rustc, can only add a workaround for this too. 🫠

roblabla:
> That's fine with me. BTW, upstream LLVM, which is upstream of rustc, can only add a workaround for this too. 🫠

I don't understand this. The problem also occurs with LLD. In this case, the whole toolchain is open source and controlled by either rustc or llvm: Rustc generates the LLVM IR, llvm turns that into machine code, and llvm's LLD turns it into a mach-o. So how is it that only a "workaround" can be added?

dianqk:
> > That's fine with me. BTW, upstream LLVM, which is upstream of rustc, can only add a workaround for this too. 🫠
> 
> I don't understand this. The problem also occurs with LLD. In this case, the whole toolchain is open source and controlled by either rustc or llvm: Rustc generates the LLVM IR, llvm turns that into machine code, and llvm's LLD turns it into a mach-o. So how is it that only a "workaround" can be added?

First, ld64 is a commonly used linker for Mach-O, which is the main issue to be addressed. We cannot ask users to switch to LLD.
Second, Yes, LLD is open source, but we can't ask someone to fix it immediately.

Issue #3833, opened 2025-01-17
Update Hyper Client to support handling `expect: 100-continue` header


heiwais25:
**Is your feature request related to a problem? Please describe.**

When I use Hyper Client based on the example (https://github.com/hyperium/hyper/blob/3817a79b213f840302d7e27fac8508de9caada0f/examples/client.rs#L49-L54), it doesn't correctly handle the `expect: 100-continue` header in the request. 

**Describe the solution you'd like**

Hyper Client needs to wait for the `100 continue` response from the S3WSServer after sending the request header parts for certain time and then send the request body if it gets the response from the server or after certain time threshold.

This is to have the same functionality as https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100

**Describe alternatives you've considered**

Expose the client parts where it sends the request header and body, but it will be inefficient for the developers to implement the same functionality.

**Additional context**

seanmonstar:
I'm not certain I want to add explicit support inside hyper, because a timeout is needed, and I'd prefer as much as possible that those be done outside of hyper. That doesn't mean we couldn't eventually someday add explicit support, of course. But, in order to support it, it needs two things:

1. Support for knowing when a 100 response is received (see #2565)
2. A body which combines waiting for that response or a timeout, whichever comes first.

I think we can provide both, the first one in hyper, and a helper body type in `hyper-util` which registers a callback and then provides a body wrapper that make use of it and a timmer.

- [x] #3818 
- [ ] Provide an `expect_continue` helper in `hyper-util`

heiwais25:
Thank you for sharing the earlier discussions and the follow up.

It would be great if the client supports this behavior somehow even though Hyper doesn't have explicit support.

I understand the point of `on_information()`, but have some questions about the second one.

It seems `hyper-util` uses `SendRequest` https://github.com/hyperium/hyper/blob/3817a79b213f840302d7e27fac8508de9caada0f/src/client/conn/http1.rs#L221 to send the request eventually which is in Hyper crate.

Could you share the high level idea about the second one without updating Hyper? I think we ended up updating Hyper.

heiwais25:
Also, to add few thoughts, I think it makes sense to add it to the Hyper explicitly since it is listed in RFC https://www.rfc-editor.org/rfc/rfc9110#name-expect.

heiwais25:
Hi, I was able to implement 100 Continue behavior by using `set_on_informational_hook` that you exposed.

[...]

PR #3910, opened 2025-07-07
feat(example): add client 100 Continue with fixed body POST example

heiwais25:
## Notes

This commit adds a new client example demonstrating HTTP 100 Continue functionality. 

## Background

Hyper Client originally didn't have support for the HTTP 100 Continue functionality.

As per [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.1.1), the client is expected to wait for the 100 Continue from the server after sending only the header parts.

- Other Doc: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/100

Following the discussion in https://github.com/hyperium/hyper/issues/3833, the `on_informational` hook has been officially exposed, allowing the Hyper client to support 100 Continue functionality based on its usage.

This example aims to help new developers easily implement 100 Continue functionality with the Hyper client. The included server is intended only to provide a complete, working scenario.

## Tests

Since it is not easy to test 100 Continue with public APIs, this was tested in the following scenario using the included mock server.

Run mock TCP server:
```bash
cargo run --example client_100_continue --features="full" -- --server 8080
```

Run client:
```bash
cargo run --example client_100_continue --features="full" -- http://127.0.0.1:8080
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,6 +137,11 @@ path = "examples/client_json.rs"
 required-features = ["full"]
 
 [[example]]
+name = "client_100_continue"
+path = "examples/client_100_continue.rs"
+required-features = ["full"]
+
+[[example]]
 name = "echo"
 path = "examples/echo.rs"
 required-features = ["full"]
--- a/examples/README.md
+++ b/examples/README.md
@@ -32,6 +32,8 @@ futures-util = { version = "0.3", default-features = false }
 
 * [`client_json`](client_json.rs) - A simple program that GETs some json, reads the body asynchronously, parses it with serde and outputs the result.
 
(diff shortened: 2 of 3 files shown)


seanmonstar:
Thanks for the PR! I'm glad you were able to get it working 😄 

Two things here:

1. I suspect we could an easier utility in hyper-util, providing a body implementation and calling `on_informational` for you.
2. I'd like to limit how many more examples we add to the repository here. I do think it'd be a great candidate to put on your blog.

heiwais25:
Hi, @seanmonstar Thank you for response.

1. I think we can put to hyper-util package as you said instead of having an example to help others to easily use.
2. I'm okay to put this into my blog, but prefer to contribute to the hyper or hyper-util package. One thing I found lacking from Hyper project was the documentation and explanations. Because of this, I think it’s better to consolidate the information in one place rather than scattering it across blog posts.

PR #3913, opened 2025-07-09
refactor(upgrade): explicitly specify type parameter in downcast method

1911860538:
Make the type parameter passing more explicit by changing
`io.__hyper_downcast()` to `io.__hyper_downcast::<T>()` in the
downcast method. This improves code clarity and avoids potential
type inference ambiguities in multi-generic scenarios.

### DIFF
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -150,7 +150,7 @@ impl Upgraded {
     /// `Upgraded` back.
     pub fn downcast<T: Read + Write + Unpin + 'static>(self) -> Result<Parts<T>, Self> {
         let (io, buf) = self.io.into_inner();
-        match io.__hyper_downcast() {
+        match io.__hyper_downcast::<T>() {
             Ok(t) => Ok(Parts {
                 io: *t,
                 read_buf: buf,


seanmonstar:
Thanks for the PR! What motivated this change? I think the original code is pretty clear and impossible for inference to mess up...

1911860538:
> Thanks for the PR! What motivated this change? I think the original code is pretty clear and impossible for inference to mess up...

The original code rarely encounters type inference issues, and my modification primarily aims to enhance readability. Improved code clarity is achieved as explicit type parameters clearly indicate the target downcast type, making the code more readable and self-documenting.

1911860538:
> > Thanks for the PR! What motivated this change? I think the original code is pretty clear and impossible for inference to mess up...
> 
> Yes, because this method has only one generic parameter, the original code is impossible to encounter type inference issues. My modification primarily aims to enhance readability. Improved code clarity is achieved as explicit type parameters clearly indicate the target downcast type, making the code more readable and self-documenting.

seanmonstar:
Hm, ok. I lean towards leaving it alone, but if another reviewer feels differently, I could be swayed. @hyperium/triage

1911860538:
> Hm, ok. I lean towards leaving it alone, but if another reviewer feels differently, I could be swayed. @hyperium/triage

😂

src/upgrade.rs
```diff
@@ -150,7 +150,7 @@ impl Upgraded {
     /// `Upgraded` back.
     pub fn downcast<T: Read + Write + Unpin + 'static>(self) -> Result<Parts<T>, Self> {
         let (io, buf) = self.io.into_inner();
-        match io.__hyper_downcast() {
+        match io.__hyper_downcast::<T>() {
```
cratelyn, line 153: usually, the convention i am familiar with is to rely on "turbofish" syntax like this when the type parameter is ambiguous, and a specific concrete type must be provided. for example, `iter.collect::<Vec<_>>()`.

my disposition is also to leave this call as is.

PR #3920, opened 2025-07-25
docs(rt): improve rt module overview

seanmonstar:


jlizen:
Nice improvement, thanks for this!

PR #3922, opened 2025-07-29
feat(client/dispatch): `TrySendError<T>: std::error::Error`

cratelyn:
this commit introduces a `std::error::Error` implementation for `hyper::client::conn::TrySendError`.

this allows callers of
`hyper::client::conn::http2::SendRequest::try_send_request()` or `hyper::client::conn::http1::SendRequest::try_send_request()` to box a `TrySendError<T>` without discarding a potentially recovered message.

a `std::fmt::Display` implementation is added in this commit, because it is a prerequisite for implementations of `std::error::Error`.

for some previous discussion on this topic, see "_other options_" here: https://github.com/hyperium/hyper/pull/3883#issuecomment-2860212062.

see also:
* https://github.com/hyperium/hyper/pull/3883
* https://github.com/hyperium/hyper/pull/3892

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -1,3 +1,5 @@
+use std::error::Error as StdError;
+use std::fmt;
 use std::task::{Context, Poll};
 #[cfg(feature = "http2")]
 use std::{future::Future, pin::Pin};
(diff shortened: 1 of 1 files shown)


Samielakkad (commented):
The impl itself looks straightforward. I?d add a tiny unit test for the new public behavior so this does not regress silently:

- `format!("{}", err)` delegates to the inner hyper error
- `std::error::Error::source(&err)` returns that inner error
- `err.into_message()` still recovers the request/message after formatting/source checks on a borrowed error

The last point matters because the main value of `TrySendError<T>` is preserving `T`; a test can make sure the new `Error` impl does not encourage an API shape that accidentally makes recovery awkward later.

PR #3924, opened 2025-08-04
Introduce QUIC traits

seanmonstar:
This adds initial support for QUIC to hyper, a prerequisite for HTTP/3 support.

This adds a new module, `hyper::rt::quic`, and new traits which largely match what was [originally proposed for the h3 crate](https://github.com/hyperium/h3/blob/master/docs/PROPOSAL.md). The idea is similar to the other transport traits in `hyper::rt`, which is that a user can describe to hyper how their QUIC implementation works, and then hyper will be able to use that for HTTP/3.

While the `h3` crate has drifted some from the original proposal, I wasn't sure the additional complexity that had been added was necessary to expose at the hyper level. If during the implementation process it proves we do need it, we can make the change.

These traits are put behind a `cfg(hyper_unstable_quic)` flag, until they are stabilized.

This should likely stay as a PR proposal until a separate PR is ready to merge that integrates these traits with HTTP/3 glue in `hyper::proto::h3`.

**Please keep comments to implementation details.**

cc #1818

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,7 +102,8 @@ nightly = []
 level = "warn"
 check-cfg = [
     'cfg(hyper_unstable_tracing)',
-    'cfg(hyper_unstable_ffi)'
+    'cfg(hyper_unstable_ffi)',
+    'cfg(hyper_unstable_quic)',
 ]
 
 [lints.clippy]
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -17,6 +17,10 @@ pub mod bounds;
 mod io;
 mod timer;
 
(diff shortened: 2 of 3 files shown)


Ruben2424 (commented):
Left some thought about things learned in h3 to the traits below.

If I understand the current proposal/implementation correctly it requires quic implementations to implement both traits if they want to be used with hyper and h3 alone right?

Using h3s traits publicly in hyper is not possible because of the stability guaranties right?

Would it be possible to `impl rt::quic::Connection for F where F: h3::quic::Connection`
or would i limit the possible changes in h3 to much?

PR #3938, opened 2025-08-27
Update documented default values for `http1::Builder`

Will-Low:
1. Document the default value for `allow_multiple_spaces_in_request_line_delimiters`.
2. Update the formatting of the default boolean values to be monospaced. This is used by `half_close` and I personally found it easier to read, since the default values pop out.


seanmonstar (approved):
Nice improvement, thanks!

PR #3946, opened 2025-09-06
fix(client): `http2_connect_detect_close` uses in-memory i/o

cratelyn:
see https://github.com/hyperium/hyper/issues/3896.

this commit rewrites this test to use `tokio::io::duplex(..)`.

this should help avoid spurious failures, by replacing `setup_tk_test_server()` with an in-memory i/o type.


cratelyn:
https://github.com/hyperium/hyper/pull/3947 supersedes this. closing!

PR #3951, opened 2025-09-10
Drop dependency on futures-channel

WhyNotHugo:
Tokio already provides mpsc and oneshot, and it is an unconditional dependency. Use those implementations and drop a superfluous dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ tokio = { version = "1", features = ["sync"] }
 # Optional
 
 atomic-waker = { version = "1.1.2", optional = true }
-futures-channel = { version = "0.3", optional = true }
 futures-core = { version = "0.3.31", optional = true }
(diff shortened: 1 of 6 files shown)


WhyNotHugo:
v2: applied cargo fmt

WhyNotHugo:
Odd, there are some failures which didn't come up locally. I'll leave this as draft until I've resolve these.

0x676e67:
The API of futures-channel is somewhat different from Tokio’s channel. Doesn’t it need to be changed?

WhyNotHugo:
We're relying on Receiver implementing Stream, which tokio's doesn't. I don't think this change is feasible.

Sorry for the noise.

WhyNotHugo:
(See b12f119e289161a2a989bd729e0929da911e775d)

PR #3959, opened 2025-09-25
feat: Allow MAX_URI_LEN configuration via environment variable

AjitR:
Add ability to configure maximum URI length through HYPER_MAX_URI_LEN environment variable while maintaining the default value when the variable is not set.

- Maintains backward compatibility by keeping same default value
- Only uses environment variable when explicitly set
- Falls back to default value for invalid configurations

### Motivation
We have some legacy applications which supports pretty big URI. We are adding a new component built in rust (using hyper) which will handle this communication in future. Requests get rejected with if they are above 64K hardcoded limit in the hyper crate. So making it configurable via ENV variable. 

```
2025-09-23T18:42:59.858177Z DEBUG hyper::proto::h1::conn: parse error (URI too long) with 65646 bytes

2025-09-23T18:42:59.858235Z DEBUG hyper::proto::h1::role: sending automatic response (414 URI Too Long) for parse error

2025-09-23T18:42:59.858327Z DEBUG hyper::proto::h1::io: flushed 85 bytes

2025-09-23T18:42:59.858404Z DEBUG hyper::server::server::new_svc: connection error: URI too long
```

#3960

### DIFF
--- a/src/proto/h1/role.rs
+++ b/src/proto/h1/role.rs
@@ -28,10 +28,23 @@ use crate::proto::h1::{
 use crate::proto::RequestHead;
 use crate::proto::{BodyLength, MessageHead, RequestLine};
 
+#[cfg(feature = "server")]
+use std::sync::OnceLock;
 pub(crate) const DEFAULT_MAX_HEADERS: usize = 100;
 const AVERAGE_HEADER_SIZE: usize = 30; // totally scientific
 #[cfg(feature = "server")]
-const MAX_URI_LEN: usize = (u16::MAX - 1) as usize;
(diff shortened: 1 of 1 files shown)


lthiery:
I'm not a maintainer but I would guess that they'd probably want you to follow the pattern established by max headers.

You can see how it's used around the source code you already modified and it looks like it's set via the API here: https://docs.rs/hyper/latest/hyper/server/conn/http1/struct.Builder.html

seanmonstar:
Yes, configuration should be done via a builder method.

In this case, the check also happens to be same as `http::Uri` hardcoded limit. So, `http::Uri` would need to be able to support longer strings, and then hyper would need to change this from a constant to a builder option.

Issue #3950, opened 2025-09-09
hyper-util: regression in paused time compatibility


asmello:
**Version**
`hyper` version `1.7.0` (not relevant AFAIK), `hyper-util` version `0.1.16`

**Platform**
MacOS

**Description**

Now, I realise nowhere (that I know) Hyper promises to be compatible with Tokio's paused-time feature. But neither does it document (that I know) that it isn't.

The issue I'm reporting is fundamentally due to an assumption that `hyper-util` makes that `std::time::Instant` always matches `tokio::time::Instant`. This is not true when paused time is enabled.

This probably spans multiple places, but let's focus on the one I've investigated: [the `IdleTask` implementation](https://github.com/hyperium/hyper-util/blob/00035bac2da1cfa820eda4db7bf7ddcbd30be3c1/src/client/legacy/pool.rs#L774).

1. Initially `std::time::Instant` and `tokio::time::Instant` start the same.
2. When `IdleTask` is first polled, its internal `tokio::time::Sleep` [is configured to trigger after `std::time::Instant::now()` plus the timeout interval](https://github.com/hyperium/hyper-util/blob/00035bac2da1cfa820eda4db7bf7ddcbd30be3c1/src/client/legacy/pool.rs#L788). Since the clocks are still in sync (assuming it gets polled immediately, which seems to be the case empirically), this does the correct thing and the deadline is set 90s in the future (assuming the default timeout is used) according to both clocks.
3. Tokio will let other tasks do work, and once an idle state is reached, it will auto-advance time until the next scheduled event. Let's suppose this is the deadline set by `IdleTask`.
4. Now over 90s have passed in Tokio's clock, but very little real time has actually passed. The clocks have drifted. When we configure the next deadline [here](https://github.com/hyperium/hyper-util/blob/00035bac2da1cfa820eda4db7bf7ddcbd30be3c1/src/client/legacy/pool.rs#L804), it will be based off actual time. The actual time plus 90s is going to be very close to Tokio's perceived present, maybe even in the past. This causes the task to be immediately awaken and the future polled again.
5. In the next iteration, very little actual time will have passed, so the next deadline is _still_ approximately 90s past the starting time. Again this is in the present/past from Tokio's perspective, so again it schedules the future to be polled immediately.
6. This creates a busy loop. The program hangs.

[...]

PR #3965, opened 2025-10-19
feat: allow overriding the instant returned from Timer

arielb1:
Fully backwards-compatible

See #3950 and hyperium/hyper-util#238


arielb1:
cc @seanmonstar

arielb1:
Any update on this?

arielb1:
If you are fine with it being insta-stable we can have that.

arielb1:
Added comment that makes it stable

PR #3969, opened 2025-10-27
Update README to include deboa as an option

ararog:
Adding deboa as alternative to reqwest, hope that I'm allowed to propose this.


seanmonstar:
Thanks for the suggestion. For the time being, I don't plan to extend this list in the readme or crate docs.

Issue #2695, opened 2021-11-16
Original Header Cases API


seanmonstar:
This is a tracking issue for the feature to allow for sending and receiving the exact casing of header names, instead of the forced lowercasing of `http`. In most situations, simply forcing to lowercase is the better option, as it allows for a more performant `HeaderMap`, and is the required format for newer versions of HTTP (HTTP/2 and 3). However, there are rare occasions when being able to see the exact casing is needed.

Some initial support is available in hyper, and is exposed in the C API. But a public Rust API is not yet available.

- [x] Initial internal support
- [x] Expose in C API (#2278)
- [x] Provide builder options to enable support in Rust, useful to proxies (#2480)
- [ ] Design and export the public extension API

kulame:
any recent progress？ I'm also having issues with header casing

seanmonstar:
We update with progress as it happens, it's all here. I'd recommend filing a bug with the server you're interacting with, it is wrongly assuming that HTTP headers have meaning.

kulame:
in my case,  the openconenct client assume the HTTP header is "X-CSTP-*" and it never convert to lowercase.

the client code is 
[X-CSTP-](https://gitlab.com/openconnect/openconnect/-/blob/master/cstp.c#L409)
```c
		if (strncmp(buf, "X-DTLS-", 7) &&
		    strncmp(buf, "X-CSTP-", 7) &&
		    strncmp(buf, "X-DTLS12-", 9))
			continue;
```

bad situation this is client code,
there are millions of users relying on this client to use VPN.
it is impossible to make these users upgrade their openconnect client.

so the server must return the uppercase header.

tasn:
We've also been bitten by this unfortunately. We know and realize that HTTP headers should be lower case.

Alas, some of our customers have compatibility requirements because they've been sending them as upper case (or camel case, depending) and some of their customers rely on them being of specific casing.

I'd be really grateful for even an experimental `feature` that we could enable to support this, because our current (terrible) alternative is to maintain our own Hyper fork. :(

svix-jplatte:
[...]

PR #3971, opened 2025-10-28
Add header casing public API

kotx:
Addresses #2695 by exposing a public API for HeaderCaseMap
Basically we validate the original spelling bytes by just parsing it as a HeaderName.
Maybe we need to further validate by comparing equivalence? But I'm not sure.

First time contributing to Hyper, let me know if there are issues!


src/ext/mod.rs
```diff
@@ -179,22 +181,23 @@ impl HeaderCaseMap {
...
-    pub(crate) fn default() -> Self {
-        Self(Default::default())
-    }
-
-    #[cfg(any(test, feature = "ffi"))]
-    pub(crate) fn insert(&mut self, name: HeaderName, orig: Bytes) {
+    /// Inserts a header spelling, replacing any existing ones associated with that header name.
+    #[cfg(all(any(feature = "client", feature = "server"), feature = "http1"))]
+    pub fn insert(&mut self, name: HeaderName, orig: Bytes) -> Result<(), InvalidHeaderName> {
+        HeaderName::from_bytes(&orig)?;
```
seanmonstar: Clever! Hm, at the same time, it worries me a little to be [validating, not parsing](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). It does mean a user can't round trip from `get_all` to `insert` without having to validate again.

At the same time, is it really worth having a separate `CasedHeaderName` type? Maybe it is... This is probably the hardest question with making this public.

Also, I think `HeaderName::from_bytes()` will allocate a new copy... That could probably be handled with `from_maybe_shared`, but just wanted to note for myself at least.

kotx: [HeaderName](https://docs.rs/http/latest/http/header/struct.HeaderName.html) doesn't seem to have `from_maybe_shared`, that seems to be exclusive to [HeaderValue](https://docs.rs/http/latest/http/header/struct.HeaderValue.html)?

Anyway, I added a CasedHeaderName type. Usage is a bit clunky since its used like `.insert(CasedHeaderName(name, orig))`.
There's also a `From<Bytes>` implementation that infers the header name by parsing it.

kotx: Oops, forgot to enforce the parsing of the header name, fixing now...

seanmonstar:
Thanks for continuing to explore! I feel like this is a sufficiently interesting enough design space that this would benefit from a design proposal posted to the original issue, first. Would you be interested in collabing on one with me? I've started one in a hackmd, I can send you an invite to edit (I can grab your email from your website) and once it looks good, we can post it for any final feedback. We could collab in the doc itself, or also could start a thread in Discord if you prefer chat (I'm fine with whatever).

kotx:
Sure thing, we can collab in the hackmd!

kotx:
@seanmonstar Seems like the hackmd is locked to owner and I can't make any edits.

seanmonstar:
That's odd. I had set the invite as "write" permissions. But now when I go in, the autocomplete mentions kotx, not just your other email, so maybe it was different accounts. I sent it again to kotx.

kotx:
Sorry for the wait, I had to deal with finals.

I thought of a better way to perform validation though- `HeaderName`s are guaranteed to be valid header names, and if we use `String` in `HeaderCaseMap`'s spelling representation instead of `Bytes`, we can use `HeaderName`'s `PartialEq<&str>` impl to perform validation.
This has the side effect of requiring spellings to be equivalent to their HeaderNames, but hopefully that falls within expected behavior (i.e. ContentType can't map to "accept").

Alternatively, we can update the `http` crate to include a `PartialEq<Bytes>` or similar?

I'm also concerned that the HeaderCaseMap extension is a bit unwieldy to use, but I'm not sure how to improve it. Maybe there is a more ergonomic public API we could expose, but I'm not sure how well it would fit within the extension model.

Anyways, I updated the proposal doc to reflect these changes. Happy holidays!

seanmonstar:
Sorry for the long delay, I got around to making the RFC mechanism a bit more solid. It's not accepted yet, but I did have this feature in mind as I wrote this up. https://github.com/hyperium/hyper/pull/4131

PR #3989, opened 2025-12-02
feat(error): make find source pub

lthiery:
I'm not sure if you are against broadening the Error API, but this is quite useful from the client side to attempt to retrieve ones own errors. Currently, they get swallowed up by hyper and it's not possible to get them back out AFAIK.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -291,7 +291,8 @@ impl Error {
         &self.inner.kind
     }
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
I wouldn't want to make it public, no. It's possible to implement outside of hyper, for example: [`errors::find()`](https://docs.rs/errors/latest/errors/iter/fn.find.html).

lthiery:
Thanks for the tip!

PR #3998, opened 2025-12-16
fix(client): reading trailers shouldn't propagate `NO_ERROR` from early response

ulyssa:
This is the trailers variant of the fix for reading the body in #3275, so that it is possible to both attempt to read the body and the trailers when the server has sent a `RST_STREAM` with `NO_ERROR` after its response to indicate that the client should stop attempting to send it the body.

I have added a trailers-only variant of `http2_responds_before_consuming_request_body` that fails without the fix, and also updated `http2_responds_before_consuming_request_body` to verify that it can check whether there are any trailers.


src/body/incoming.rs
```diff
@@ -270,7 +270,16 @@ impl Body for Incoming {
                         ping.record_non_data();
                         Poll::Ready(Ok(t.map(Frame::trailers)).transpose())
                     }
-                    Err(e) => Poll::Ready(Some(Err(crate::Error::new_h2(e)))),
+                    Err(e) => {
+                        match e.reason() {
+                            // These reasons should cause reading the trailers to stop, but not fail it.
+                            // The same logic as for `Read for H2Upgraded` is applied here.
```
seanmonstar: I realize this is just bring down the logic that already exists for data frames, and the code itself looks fine. And the unit tests make sense why they want to suppress the error.

But, reading it all again, I have a worry: couldn't this also be swallowing _real_ cancellations, and the user thinks they got a full body, when they got a truncated one?

ulyssa: Looking over RFC 7540, I believe you're right that both this and the above branch should only be looking at `Reason::NO_ERROR`, and shouldn't be dropping the `Reason::CANCEL`. The [section on early responses in 8.1](https://www.rfc-editor.org/rfc/rfc7540#section-8.1) only mentions `NO_ERROR`:

> An HTTP response is complete after the server sends -- or the client
> receives -- a frame with the `END_STREAM` flag set (including any
> `CONTINUATION` frames needed to complete a header block).  A server can
> send a complete response prior to the client sending an entire
> request if the response does not depend on any portion of the request
> that has not been sent and received.  When this is true, a server MAY
> request that the client abort transmission of a request without error
> by sending a `RST_STREAM` with an error code of `NO_ERROR` after sending
> a complete response (i.e., a frame with the `END_STREAM` flag).
> Clients MUST NOT discard responses as a result of receiving such a
> `RST_STREAM`, though clients can always discard responses at their
> discretion for other reasons.

The only time that sending `CANCEL` is mentioned is under the [description of push responses in 8.2.2](https://www.rfc-editor.org/rfc/rfc7540#section-8.2.2):

> If the client determines, for any reason, that it does not wish to
> receive the pushed response from the server or if the server takes
> too long to begin sending the promised response, the client can send
> a RST_STREAM frame, using either the CANCEL or REFUSED_STREAM code
> and referencing the pushed stream's identifier.

This section may be why the code was added to the match in #3275, but I agree that it does sound like it should be propagated so that the server is aware of the cancellation, as opposed to `NO_ERROR` where the RFC says that clients must not discard the response. I'll update this and #3999 to not filter out the `Reason::CANCEL` error.

seanmonstar (approved):
Thanks so much for the fix!

Issue #3984, opened 2025-11-26
Examples full of unsafe usage of .unwrap()


CharlemagneLasse:
I've just looked through the examples and documentation how to write a good handlers. And all over the place, the it uses .unwrap on Results which could also be Err. This leads to panics in rust - basically what [Cloudflare took down last week](https://blog.cloudflare.com/18-november-2025-outage/#memory-preallocation).

Here are a couple of unsafe unwraps:

* https://github.com/hyperium/hyper/blob/be18a92bb51498a2113dcfe15d9da09e1c6c7b0d/examples/send_file.rs#L64
* https://github.com/hyperium/hyper/blob/be18a92bb51498a2113dcfe15d9da09e1c6c7b0d/examples/web_api.rs#L32
* https://github.com/hyperium/hyper/blob/be18a92bb51498a2113dcfe15d9da09e1c6c7b0d/examples/upgrades.rs#L103

And when I would now remove these unwraps+ the `Ok` and directly try to return the Result... it would now work because these Response builders are using as Error type `hyper::http::Errors` and all the examples are build around `hyper::Error`. 

The examples are basically only showing us how to not to write good servers.

seanmonstar:
Hey, thanks for taking a look!

I looked at each case of unwrap in those examples, and they are indeed all safe. However, it may non-obvious, so `.expect` could be the better option to document why. For instance, the `http` builders can return an error if one of the values passed to `uri()`, `header()`, or `status()` are not valid values for that type. The examples are all calling them with constants that are known to be good, so the unwraps are safe.

But again, for the sake of teaching, it could be a good idea to change those to `.expect()` explaining why they are fine. Or... (and this is what I do typically in my own code), they could just not used the builders, and thus show that the constants are safe with things like `from_static()`.

PR #4001, opened 2025-12-30
refactor: avoid unwrap examples

0xPoe:
close https://github.com/hyperium/hyper/issues/3984

Just avoid `unwrap` as much as possible. Use expect to assert that it will always succeed.


0xPoe:
> It's only every `status()`, `header()`, or `uri()` that is parsing into an `http` type that could error. So the expect should mention "constant status won't error", instead of "missing should be valid".

Good point! Updated. Thank you.

Issue #3995, opened 2025-12-16
Hyper returns a partial byte array when its runtime is closed mid-read


criccomini:
**Version**
List the version(s) of `hyper`, and any relevant hyper dependency (such as `h2` if this is related to HTTP/2).

```
name = "hyper"
version = "1.6.0"

name = "hyper-rustls"
version = "0.27.5"

name = "hyper-util"
version = "0.1.10"

name = "reqwest"
version = "0.12.14"

name = "http"
version = "1.3.1"
```

**Platform**
The output of `uname -a` (UNIX), or version and 32 or 64-bit (Windows)

```
Darwin mac.lan 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6020 arm64
```

**Description**

Hyper returns a partial byte array when its runtime is closed mid-read.

We have had several reports of panics when shutting down SlateDB:

https://github.com/slatedb/slatedb/issues/1019

The `bytes range` panic in the issue above seems to stem from receiving half an object from object storage (using the `object_store` crate, which uses `reqwest` and `hyper`). I was able to induce the error in SlateDB here (using Minio as the object store along with the `object_store` crate):

https://gist.github.com/criccomini/9a53ace1dc175c4bd157cf518f5d08eb

I was then able to induce the error _outside_ of SlateDB and `object_store` using this code:

```rs
//! Minimal repro for a "partial body without error" scenario using `hyper` only.
//!
//! The test:
//! - Serves an HTTP/1.1 response with a `Content-Length` of `expected_len`, but only writes
//!   `first_chunk_len` bytes of the body.
//! - Reads the first body frame successfully.
//! - Drops the Tokio runtime driving the client connection.
//! - Observes that collecting the remainder of the body returns `Ok`, but yields fewer bytes
//!   than `Content-Length` advertised.
use bytes::Bytes;
use http_body_util::{BodyExt, Empty};
use hyper::body::Body;
use hyper::client::conn::http1;
use hyper::Request;
use hyper_util::rt::TokioIo;
use std::time::Duration;
use tokio::io::{duplex, AsyncReadExt, AsyncWriteExt};

#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
#[ignore]
async fn repro_hyper_partial_body_when_connection_runtime_dropped() {
    // The server will claim it will send 1MiB...
    let expected_len = 1024 * 1024;
[...]

PR #4016, opened 2026-02-05
fix(http1): send error when dispatcher is dropped mid-body (#3995)

Barre:
Wrap body_tx in a SenderDropGuard that sends an IncompleteMessage error if the Dispatcher future is dropped while still owning a body Sender, preventing silent data truncation.


seanmonstar:
This looks awesome! Would you be up for adding an integration test (client or server) to exercise this? If not that's fine, I can look at adding one.

PR #4020, opened 2026-02-16
chore(ci): force toml mode in yq selecting msrv

seanmonstar:
related https://github.com/mikefarah/yq/issues/2592


WhyNotHugo:
The recommendation from Rust upstream is to commit the lockfile into the repository. Why not do that instead?

PR #4025, opened 2026-02-23
Preserve URI parse error info in error source chain

### LINKED ISSUE #3043: URI parse errors should preserve information from the HTTP crate

veeceey:
Fixes #3043

When hyper parsed an HTTP request with an invalid URI, it would create a `Parse(Uri)` error but discard the original `http::uri::InvalidUri` error. This meant `error.source()` returned `None`, making it impossible to get the actual reason the URI was invalid.

Changed `Parse::Uri` to carry an `Option<Cause>` so the original `InvalidUri` or `InvalidUriParts` error flows through to `Error::source()`. The error message and `Display` output are unchanged - the improvement is only visible through the `Error::source()` chain.

Added a test that verifies `error.source()` returns the original `http::uri::InvalidUri`:

```
running 86 tests
...
test error::tests::uri_error_preserves_source ... ok
...
test result: ok. 86 passed; 0 failed; 0 ignored
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -92,7 +92,7 @@ pub(super) enum Parse {
     Version,
     #[cfg(all(any(feature = "client", feature = "server"), feature = "http1"))]
     VersionH2,
-    Uri,
+    Uri(Option<Cause>),
     #[cfg(all(feature = "http1", feature = "server"))]
     UriTooLong,
     #[cfg(feature = "http1")]
@@ -459,7 +459,7 @@ impl Error {
             Kind::Parse(Parse::Version) => "invalid HTTP version parsed",
             #[cfg(all(any(feature = "client", feature = "server"), feature = "http1"))]
             Kind::Parse(Parse::VersionH2) => "invalid HTTP version parsed (found HTTP2 preface)",
-            Kind::Parse(Parse::Uri) => "invalid URI",
+            Kind::Parse(Parse::Uri(_)) => "invalid URI",
             #[cfg(all(feature = "http1", feature = "server"))]
             Kind::Parse(Parse::UriTooLong) => "URI too long",
             #[cfg(feature = "http1")]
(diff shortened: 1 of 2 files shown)


veeceey:
hey, just checking in — is there anything else needed from my side?

Samielakkad (commented):
This looks like the right shape for preserving the underlying `http` error.

One small coverage gap: the implementation handles both `InvalidUri` and `InvalidUriParts`, but the new test only exercises `InvalidUri`. Since the two conversions are separate impls, it would be worth adding a sibling assertion for `http::uri::InvalidUriParts` as well.

That would pin down both source-preservation paths and avoid accidentally regressing one while keeping the other green.

PR #4029, opened 2026-03-03
perf(h1): pass ParserConfig by reference instead of cloning per parse call

DioCrafts:
**Description:**

`ParseContext` owned a full `ParserConfig` value. Every HTTP/1 parse iteration cloned it from the connection state, even though the config never changes mid-connection.

This PR changes `h1_parser_config` from `ParserConfig` to `&'a ParserConfig`. The clone disappears and the borrow checker guarantees the reference stays valid for the parse call.

**What changed**

- `ParseContext.h1_parser_config`: owned `ParserConfig` → `&'a ParserConfig`
- `conn.rs`: removed `.clone()` when building `ParseContext`
- `io.rs`: same removal in the buffered I/O parse path
- `role.rs`: updated 20 test call sites to pass references

No public API changes, no behavioral changes. All existing tests pass.

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -239,7 +239,7 @@ where
             ParseContext {
                 cached_headers: &mut self.state.cached_headers,
                 req_method: &mut self.state.method,
-                h1_parser_config: self.state.h1_parser_config.clone(),
+                h1_parser_config: &self.state.h1_parser_config,
                 h1_max_headers: self.state.h1_max_headers,
                 preserve_header_case: self.state.preserve_header_case,
                 #[cfg(feature = "ffi")]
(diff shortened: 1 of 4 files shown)


seanmonstar:
From the previous PR:

> I believe it originally used a clone for lifetimes. I assumed the compiler could inline and eliminate the code, does this result in a difference?

DioCrafts:
> From the previous PR:
> 
> > I believe it originally used a clone for lifetimes. I assumed the compiler could inline and eliminate the code, does this result in a difference?


Hi @seanmonstar, 
Right, the compiler likely elides the clone for 7 bools, but the reference guarantees zero-cost regardless of future changes to ParserConfig and better communicates that the config is borrowed, not owned.

PR #4030, opened 2026-03-03
perf(h1): pre-allocate read buffer instead of starting at zero

DioCrafts:
**Description:**

The H1 read buffer was created with `BytesMut::with_capacity(0)`. On the first read, the adaptive strategy immediately grows it to `INIT_BUFFER_SIZE` (8192 bytes). The H1 read buffer was created with BytesMut::with_capacity(0). On the first read, the adaptive strategy grows it to INIT_BUFFER_SIZE (8 KiB). By pre-allocating at that size directly, we skip the deferred-growth code path.

This PR initializes the buffer directly at `INIT_BUFFER_SIZE`. One allocation instead of two. The adaptive read strategy still controls growth from there.

**What changed**

- `io.rs` line 71: `BytesMut::with_capacity(0)` → `BytesMut::with_capacity(INIT_BUFFER_SIZE)`

One line, no API changes.

### DIFF
--- a/src/proto/h1/io.rs
+++ b/src/proto/h1/io.rs
@@ -68,7 +68,7 @@ where
             io,
             partial_len: None,
             read_blocked: false,
-            read_buf: BytesMut::with_capacity(0),
+            read_buf: BytesMut::with_capacity(INIT_BUFFER_SIZE),
             read_buf_strategy: ReadStrategy::default(),
             write_buf,
         }


paolobarbolini:
> That means every new connection paid for a zero-size allocation followed by a reallocation and copy.

> One allocation instead of two

Capacity of zero is not an allocation.

DioCrafts:
@paolobarbolini  You're absolutely right man, with_capacity(0) doesn't allocate, my description was misleading there, sorry about that! The intent is to avoid the deferred allocation path on the first read, since the adaptive strategy will grow it to INIT_BUFFER_SIZE immediately anyway. I'll update the PR description. Thanks for the correction! 🙏

seanmonstar:
Does this result in improved performance in some case? Can we measure?

DioCrafts:
> Does this result in improved performance in some case? Can we measure?

Hi @seanmonstar.

Honestly, for the common case the difference is negligible: with_capacity(0) doesn't allocate, and the first read triggers the 8 KiB allocation anyway. The only gain is skipping the growth check on that first read. I don't have benchmarks showing a measurable improvement.

If you'd prefer, I can run h1_parallel or a similar bench to get numbers before moving forward.

srfsh:
This is AI generated, isn't it?

Samielakkad (commented):
Retracted.

PR #4032, opened 2026-03-08
examples: add Server-Sent Events (SSE) example

Sim-hu:
## Summary

- Adds a new `server_sent_events` example demonstrating how to implement SSE with hyper
- Uses `StreamBody` + `futures_util::stream::unfold` to stream timestamped JSON events
- Follows the proper SSE wire format (`text/event-stream`, `id`, `event`, `data` fields)
- No new dependencies — uses only existing dev-dependencies (`futures-util`, `http-body-util`)

### Motivation

Several existing examples show streaming responses (`echo.rs`, `send_file.rs`), but none cover the Server-Sent Events pattern. SSE is widely used for real-time web applications, and implementing it correctly with hyper's low-level API is a common question.

### Checklist

- [x] `cargo build --features="full" --example server_sent_events`
- [x] `cargo run --features="full" --example server_sent_events` — tested with `curl -N`
- [x] `cargo clippy --features="full" --example server_sent_events`
- [x] `cargo fmt --check`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -197,6 +197,11 @@ required-features = ["full"]
 
 
 [[example]]
+name = "server_sent_events"
+path = "examples/server_sent_events.rs"
+required-features = ["full"]
+
+[[example]]
 name = "web_api"
 path = "examples/web_api.rs"
 required-features = ["full"]
--- a/examples/server_sent_events.rs
+++ b/examples/server_sent_events.rs
@@ -1,0 +1,87 @@
+#![deny(warnings)]
+
+//! A Server-Sent Events (SSE) example.
+//!
+//! This server sends a stream of timestamped events to connected clients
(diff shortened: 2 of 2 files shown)


RangerMauve:
This is great, thank you!

Samielakkad (commented):
Nice example to have. One maintainability concern: the example imports `TokioIo` through `#[path = "../benches/support/mod.rs"]`.

That makes the example depend on bench-support internals, which is a little surprising for something users may copy as a starting point. If the support module changes for benchmark reasons, this example can break even though the public example code did nothing wrong.

Would it be better to use the same public runtime adapter that users are expected to use, or define the tiny adapter locally in the example? Keeping examples self-contained/public-API-only tends to make them safer as copy-paste documentation.

PR #4045, opened 2026-03-27
feat(server,http1,http2,upgrade): remove body static lifetime constraint

Voldemat:
Before this change, hyper expects body trait implementation to have 'static lifetime, which prevents use cases where one might have StreamBody that references certain state that outlives hyper connection object. This change adds lifetime parameter 'body to varies hyper objects that replaces 'static.

This change contains no breaking changes for existing use cases, however it requires appropriate change in hyper-util crate


Voldemat:
here is link to hyper-util pr with corresponding changes there:  https://github.com/hyperium/hyper-util/pull/278

seanmonstar:
Unfortunately, adding a namespace to public types here will be a breaking change. :(

PR #4056, opened 2026-04-26
fix(http1): fix reading large bodies on 32-bit systems

### LINKED ISSUE #4055: u64->usize overflow in h1 decode on 32bit system
The [code](https://github.com/hyperium/hyper/blob/master/src%2Fproto%2Fh1%2Fdecode.rs#L155) in question:
```
let to_read = *remaining as usize;
let buf = ready!(body.read_mem(cx, to_read))?;
let num = buf.as_ref().len() as u64;
```
Here, if it ever happens that *remaining is divisible by usize::MAX, to_read will become 0 and the code will return an error IncompleteBody, which is clearly wrong.

Moreover, even in cases when remaining is not exactly divisible by usize::MAX, but has small remainder, this is suboptimal (and likely to produce remainder divisible by usize::MAX).

Maybe it is worth to replace this with saturating cast?

seanmonstar:
On 32-bit systems, the cast of the length to read from a u64 to a usize could
truncate, and cause less to be read. If the length was larger than usize::MAX,
the end could not be read.

To fix this, we use a saturating cast pattern instead.

Closes #4055

### DIFF
--- a/src/proto/h1/decode.rs
+++ b/src/proto/h1/decode.rs
@@ -152,7 +152,7 @@ impl Decoder {
                 if *remaining == 0 {
                     Poll::Ready(Ok(Frame::data(Bytes::new())))
                 } else {
-                    let to_read = *remaining as usize;
+                    let to_read = usize::try_from(*remaining).unwrap_or(usize::MAX);
                     let buf = ready!(body.read_mem(cx, to_read))?;
                     let num = buf.as_ref().len() as u64;
                     if num > *remaining {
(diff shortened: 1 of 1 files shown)


cpud36:
> In extremely rare cases

Although it is not important, I believe it is guaranteed to happen if ever remaining becomes larger than usize::MAX. The code can never decrease remaining more than to_read, thus it cannot pass below the usize::MAX boundary...

seanmonstar:
You're right, I stepped through a bit more. I've adjusted the language and title.

PR #4066, opened 2026-05-10
chore(rt): preserve executor caller locations

punkeel:
Tokio task hooks report the source location of the tokio::spawn call. When runtime executor glue performs that spawn, the recorded location can point at the executor implementation itself. That is technically correct, but it hides the more useful callsite: the code path that asked the executor to create background work.

Mark Executor::execute and the HTTP/2 executor forwarding traits with track_caller so task instrumentation can attribute spawned tasks to the caller that requested execution rather than to generic executor forwarding glue.

### DIFF
--- a/src/rt/bounds.rs
+++ b/src/rt/bounds.rs
@@ -18,6 +18,7 @@ mod h2_common {
 
     pub trait Http2UpgradedExec<B> {
         #[doc(hidden)]
+        #[track_caller]
         fn execute_upgrade(&self, fut: UpgradedSendStreamTask<B>);
     }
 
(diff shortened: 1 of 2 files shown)


punkeel:
related but independent: https://github.com/hyperium/hyper-util/pull/283

seanmonstar:
Will this work? The executor gets turned into a `dyn Executor`, and the reference docs make it sound like the location data is decayed then: https://doc.rust-lang.org/stable/reference/attributes/codegen.html#r-attributes.codegen.track_caller.decay

seanmonstar:
So, I'm guessing that this probably can't work as expected. So I'm going to close. If you find that it does, we can re-open.

Issue #4054, opened 2026-04-24
Document cancel safety


jorendorff:
Would you accept a documentation patch describing Hyper's cancellation behavior?

I think Hyper is largely cancel safe:
- dropping an HTTP/1 request causes the connection to be closed
- dropping an HTTP/2 request leaves the shared connection in a good state and cancels the unwanted response (using RST_STREAM).
- otherwise Hyper doesn't really have a bunch of hidden state to screw up
- well, except the legacy Client support, but actually that's _also_ carefully implemented to do the right thing on cancellation

I don't know about the server side, but that's OK, I can read code. If you're interested.

It doesn't seem like guide material. I was thinking a single `# Cancel safety` section at the crate level.

seanmonstar:
Seems fair to document. Though, thinking about where the information would be best to be found... would it be better to have a small section on the relevant futures about cancellation? Or maybe in combination with a small mention at the crate level, "futures are cancel safe, see individual features for behavior when canceled"?

jorendorff:
I think that is better.

OK, I'll look into this when I have a chance (sorry, I work at GitHub and let me just say it's been a week)

CinisBorn:
I'm interest in this issue. What remains it is server docs, right?

PR #4070, opened 2026-05-12
docs: document cancel safety for client send_request futures

lihan3238:
## Summary

Refs #4054. Documents hyper's cancellation behavior on the two client `send_request` futures, plus a small crate-level pointer.

Per seanmonstar's suggestion in the issue:
> ...would it be better to have a small section on the relevant futures about cancellation? Or maybe in combination with a small mention at the crate level, "futures are cancel safe, see individual features for behavior when canceled"

This PR does both.

## Changes

`src/client/conn/http1.rs` — adds `# Cancel safety` to `SendRequest::send_request`:

> Dropping the returned future is the supported way to cancel an in-flight HTTP/1 request. Because HTTP/1 has no in-protocol way to abort a single request without affecting the shared connection, hyper closes the underlying connection when a request future is dropped before completion. Any subsequent calls on the same `SendRequest` will return a `canceled` error.

`src/client/conn/http2.rs` — adds `# Cancel safety` to `SendRequest::send_request`:

> Dropping the returned future is the supported way to cancel an in-flight HTTP/2 request. The stream is reset with `RST_STREAM` (`CANCEL` error code); the shared connection remains usable for other in-flight and future requests. The peer is notified immediately rather than continuing to send a response body that would be discarded.

`src/lib.rs` — brief crate-level mention pointing at the per-future docs.

## Notes

- Plain code spans (no intra-doc links) in the crate-level note to avoid broken-link warnings when `cargo doc` runs without `--features client`.
- Issue author (@jorendorff) said they'd "look into this when I have a chance" 12 days ago — happy to defer or merge this as a starting point. The wording follows the technical content they already laid out in the issue body.
- Scope limited to the client-side cancellation paths the issue author was confident about; server-side behavior left for a follow-up since the issue author noted they hadn't verified it.

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -195,6 +195,15 @@ where
     ///
     /// This is however not enforced or validated and it is up to the user
     /// of this method to ensure the `Uri` is correct for their intended purpose.
+    ///
+    /// # Cancel safety
+    ///
+    /// Dropping the returned future is the supported way to cancel an
+    /// in-flight HTTP/1 request. Because HTTP/1 has no in-protocol way to
+    /// abort a single request without affecting the shared connection,
+    /// hyper closes the underlying connection when a request future is
+    /// dropped before completion. Any subsequent calls on the same
+    /// [`SendRequest`] will return a `canceled` error.
     pub fn send_request(
         &mut self,
         req: Request<B>,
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Looks great, thanks!

Issue #4071, opened 2026-05-19
Add a strict clippy config


seanmonstar:
hyper for a long time never bothered with clippy. Mostly for historical reasons and opinions.

I've now come around to just turning clippy up to 11. What I'm currently leaning towards is keeping an allowlist, and otherwise denying `pedantic` and even `restriction` lints. Yes, I know there's a restriction lint against linting for the restriction group. An allowlist still makes a lot of sense to me. If after a bit it turns out to be just unbearable, we can convert it to a bigger deny list, I guess.

The initial config is in place, with a CI job that enforce it, so we won't regress with new code. Now, we just need to evaluate all the undecided lints, and either choose to allow them for sure, or fix them.

## How to help

Look at the list of lints that need to be decided upon in the [Cargo.toml](https://github.com/hyperium/hyper/blob/master/Cargo.toml#L114) (it's around there, the line might move slightly).

1. Pick a lint.
2. If it seems obviously an improvement:
    1. Fix up the code from the lint.
        - It might be as simple as `cargo clippy --fix --features full`.
        - It might be obvious enough for an LLM to automate fixing. (But only if you understand what changes it made.)
        - It might need manual adjustment in each case.
        - It might need to be `#[allow(this_lint)]` in specific cases, where the lint is usually useful.
    2. Remove the allowed lint from the `Cargo.toml`
    3. Submit a PR for one lint at a time.
3. If it is not obviously an improvement:
    1. Move the lint down lower to the explicitly allowed list in the `Cargo.toml`
    2. Add a comment why the lint is being allowed
    3. Submit a PR (multiple lints solved this way is fine in a single PR)

## Steps

- [x] #4075
- [ ] address each one in in the todo list
    - [ ] arithmetic_side_effects = "allow" # TODO: consider
    - [ ] as_conversions = "allow" # TODO: tricky
    - [x] borrow_as_ptr = "allow"
    - [x] cast_lossless = "allow" # TODO: easy fix
    - [ ] cast_possible_truncation = "allow" # TODO: consider
    - [ ] cast_precision_loss = "allow" # TODO: consider
    - [ ] checked_conversions = "allow"
    - [x] collapsible_match = "allow"
    - [x] decimal_literal_representation = "allow" # TODO: consider
    - [ ] default_trait_access = "allow"
    - [ ] else_if_without_else = "allow"
    - [x] empty_structs_with_brackets = "allow" # TODO: easy fix
    - [ ] enum_glob_use = "allow"
[...]

PR #4074, opened 2026-05-20
docs(lib): fixup markdown and grammar in doc comments

seanmonstar:
cc #4071

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -66,7 +66,7 @@ pub(crate) struct Sender<T, U> {
 /// has been dropped. However, this version can be cloned.
 #[cfg(feature = "http2")]
 pub(crate) struct UnboundedSender<T, U> {
-    /// Only used for `is_closed`, since mpsc::UnboundedSender cannot be checked.
+    /// Only used for `is_closed`, since `mpsc::UnboundedSender` cannot be checked.
     giver: want::SharedGiver,
     inner: mpsc::UnboundedSender<Envelope<T, U>>,
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -61,7 +61,7 @@ pub(super) enum Kind {
         any(feature = "http1", feature = "http2")
     ))]
     Io,
-    /// User took too long to send headers
+    /// User took too long to send headers.
     #[cfg(all(feature = "http1", feature = "server"))]
     HeaderTimeout,
     /// Error while reading a body from connection.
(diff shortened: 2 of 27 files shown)


src/proto/h1/conn.rs
```diff
@@ -911,7 +911,7 @@ impl<I: Unpin, B, T> Unpin for Conn<I, B, T> {}
 
 struct State {
     allow_half_close: bool,
-    /// Re-usable HeaderMap to reduce allocating new ones.
+    /// Re-usable `HeaderMap` to reduce allocating new ones.
```
cratelyn, line 914:
```suggestion
    /// Re-usable [`HeaderMap`] to reduce allocating new ones.
```

this can be a doc link

src/client/dispatch.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) struct Sender<T, U> {
 /// has been dropped. However, this version can be cloned.
 #[cfg(feature = "http2")]
 pub(crate) struct UnboundedSender<T, U> {
-    /// Only used for `is_closed`, since mpsc::UnboundedSender cannot be checked.
+    /// Only used for `is_closed`, since `mpsc::UnboundedSender` cannot be checked.
```
cratelyn, line 69:
```suggestion
    /// Only used for `is_closed`, since [`mpsc::UnboundedSender`] cannot be checked.
```

src/error.rs
```diff
@@ -76,7 +76,7 @@ pub(super) enum Kind {
         any(feature = "http1", feature = "http2")
     ))]
     BodyWrite,
-    /// Error calling AsyncWrite::shutdown()
+    /// Error calling `AsyncWrite::shutdown()`.
```
cratelyn, line 79:
```suggestion
    /// Error calling [`AsyncWrite::shutdown()`].
```

src/proto/h1/conn.rs
seanmonstar: Oh yea, it could be... Clippy didn't notice, and these specific comments aren't rendered into HTML 🤷

src/error.rs
```diff
@@ -119,7 +119,7 @@ pub(super) enum Header {
 
 #[derive(Debug)]
 pub(super) enum User {
-    /// Error calling user's Body::poll_data().
+    /// Error calling the user's `Body::poll_data()`.
```
cratelyn, line 122:
```suggestion
    /// Error calling the user's [`Body::poll_data()`].
```

cratelyn (commented):
looks like this already got merged, but these would be nice intradoc links to add.

seanmonstar:
Sorry I merged so quickly, I didn't think it'd be interesting. I'll slow down 🙈

cratelyn:
> Sorry I merged so quickly, I didn't think it'd be interesting. I'll slow down 🙈

no worries! they're just nits :stuck_out_tongue_closed_eyes:

PR #4079, opened 2026-05-27
Fix manual_assert lint

xd009642:
As per #4071 fix instances where the manual_assert lint triggered on the code. This was able to be done with a `clippy --fix` with further cleanup to change `!<expr>.is_some()` in the assert to `<expr>.is_none()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,7 +134,6 @@ integer_division = "allow"
 integer_division_remainder_used = "allow"
 large_enum_variant = "allow"
 let_unit_value = "allow"
-manual_assert = "allow" # TODO: easy fix
 manual_assert_eq = "allow" # TODO: easy fix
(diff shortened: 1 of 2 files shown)


xd009642:
For some reason the fmt command in the contributing docs changes nothing for me on the same rustfmt version... Applying manually

seanmonstar:
Yea, `cargo fmt` is not able to work because of the use of macros to tame the inclusion of modules conditionally. The job prints out what actually works: `rustfmt --edition 2021 $(git ls-files '*.rs')`.

xd009642:
So the command the job printed out didn't do anything on my system... I did already try it :sweat: 

EDIT: Seemed it worked if I ran it inside the src dir just not in the project root.... No idea why though :melting_face:

seanmonstar (commented):
Thanks!

PR #4080, opened 2026-05-27
Removing wildcard_imports lint allowance

Lori-Shu:
The wildcard_imports used in http1.rs is replaced with explict type imports. This is related to #4071 .


seanmonstar (approved):
Thanks!

PR #4083, opened 2026-05-28
Removing 'undocumented_unsafe_blocks' lint allowance

Lori-Shu:
#4071

### DIFF
--- a/src/common/io/compat.rs
+++ b/src/common/io/compat.rs
@@ -31,9 +31,13 @@ where
     ) -> Poll<Result<(), std::io::Error>> {
         let init = tbuf.initialized().len();
         let filled = tbuf.filled().len();
-        // SAFETY:
-        // this is an unsafe block which calls unsafe functions:
-        // `ReadBuf::inner_mut`, `ReadBuf::set_init`, `ReadBuf::set_filled`.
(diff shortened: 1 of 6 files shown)


src/common/io/compat.rs
```diff
@@ -23,13 +23,17 @@ impl<T> tokio::io::AsyncRead for Compat<T>
...
 {
+    /// `poll_read` fn implementation for `Compat<T>`.
     fn poll_read(
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
         tbuf: &mut tokio::io::ReadBuf<'_>,
     ) -> Poll<Result<(), std::io::Error>> {
         let init = tbuf.initialized().len();
         let filled = tbuf.filled().len();
+        // SAFETY:
```
dswij: The `SAFETY:` docs usually explain the requirements and justification instead of which unsafe functions/operations are called.

In this case, why the use of `ReadBuf::inner_mut` and `ReadBuf::set_init` is fine and should always be safe.

Lori-Shu: There may be AI illusions, please take a look.

seanmonstar (approved):
Thanks!

PR #4087, opened 2026-05-29
removing `cast_lossless` lint

josetorrs:
removing `cast_lossless` in https://github.com/hyperium/hyper/issues/4071


seanmonstar (approved):
Thanks!

PR #4097, opened 2026-06-09
refactor(client): Typed futures for `SendRequest`

Suficio:
Re-implements `http1::SendRequest` and `http2::SendRequest` methods to return typed futures.

1. Allow building custom state machines around responses without relying on building around `impl Future` signatures.
2. There appears to be meaningful memory savings by removing the `impl Future` signature.
3. `SendFutState` and `TrySendFutState` are `Unpin` as opposed to the derived `async` block.

Here are the memory size comparisons:

```
master
size_of::<send_request -> impl Future>     = 240
size_of::<try_send_request -> impl Future> = 240
```

Replacing `impl Future` with `SendFutState<B>` and `TrySendFutState<B>`:
```
5086145c8a3a5058e9780d4ce5364a95ef7d7a0d
size_of::<SendFut<B>>()    = 224
size_of::<TrySendFut<B>>() = 224
```

Remove `Request<B>` from `SendFutState::Err`:
```
22a1d3e745ebb6d54560da243f38f2b5ed883e5a
size_of::<SendFut>() = 16
```

Box `TrySendFutState::Err` variant:
```
19a084ca82c937dd10c0b97641fedbe897e423e7
size_of::<TrySendFut<B>>() = 16
```

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -3,7 +3,6 @@
 use std::error::Error as StdError;
 use std::fmt;
 use std::future::Future;
-use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
Cool memory gains! What if we use the manual future, with the `impl Future` return signature?

Keeping that would allow us to refactor someday.

Suficio:
@seanmonstar

> What if we use the manual future, with the `impl Future` return signature?

My main motivation was to be able to use the new types directly in my own state machines so keeping the `impl Future` doesn't get me much further.

It doesn't appear to me that switching the signature from `impl Future` to `SendFut` is a breaking change since the new types are compatible with the old signature unless you plan further changes in the future which would be breaking.

Eitherway, I'm happy to revert to the original signature in this PR based on your direction.

Samielakkad (commented):
Retracted.

PR #4098, opened 2026-06-09
docs(governance): define Advisor role

seanmonstar:
A way for busy experts to still contribute advice when needed.


cratelyn (approved):
this sounds like a great idea!

PR #4109, opened 2026-06-19
fix(http1): evict pooled conn on request-side Connection: close

samlandfried:
hyper's client derives connection reuse from the response alone, so a request carrying Connection: close whose backend response omits it (keeps the socket alive) leaves the connection pooled and reusable.

Disable keep-alive at request-encode time when the outgoing request carries Connection: close so the connection is evicted regardless of the response.

### DIFF
--- a/src/client/tests.rs
+++ b/src/client/tests.rs
@@ -23,6 +23,136 @@ async fn client_connect_uri_argument() {
         .expect_err("response should fail");
 }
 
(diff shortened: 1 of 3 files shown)


samlandfried:
@seanmonstar I believe the CI failures are unrelated to this change - Any way to get this merged?

seanmonstar:
The 0.14.x branch is largely EOL.

PR #4110, opened 2026-06-19
fix(http1): evict pooled conn on request-side Connection: close

samlandfried:
hyper's client derives connection reuse from the response alone, so a request carrying `Connection: close` whose backend response omits it (keeps the socket alive) leaves the connection pooled and reusable. 

Disable keep-alive at request-encode time when the outgoing request carries `Connection: close` so the connection is evicted regardless of the response.

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -613,6 +613,15 @@ where
 
         if !T::should_read_first() {
             self.state.busy();
+            // A client request carrying `Connection: close` must not be pooled or
+            // reused. hyper otherwise derives connection reuse from the response
+            // alone, so a backend that ignores the request-side close (omits
+            // `Connection: close` in its response) would leave the connection in
+            // the pool. Disable keep-alive up front so the connection is evicted
+            // regardless of the response.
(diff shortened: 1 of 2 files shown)


src/proto/h1/conn.rs
```diff
@@ -613,6 +613,19 @@ where
...
             self.state.busy();
+            // A client request carrying `Connection: close` must not be pooled or
+            // reused. hyper otherwise derives connection reuse from the response
+            // alone, so a backend that ignores the request-side close (omits
+            // `Connection: close` in its response) would leave the connection in
+            // the pool. Disable keep-alive up front so the connection is evicted
+            // regardless of the response.
+            if head
+                .headers
+                .get(CONNECTION)
```
seanmonstar: This probably needs to `get_all`, and check them all, either here, or some new `headers::connect_any_close()` helper.

samlandfried: @seanmonstar Added a `connection_any_close` that finds and acts on any `Connection: close` header

seanmonstar (approved):
Thanks!

PR #4111, opened 2026-06-21
style(lib): remove `default_trait_access` lint

nakaryo716:
## Overview
This PR removes `default_trait_access = "allow"` from `Cargo.toml` and updates the code to explicitly call type-specific default methods (e.g., `ParserConfig::default()`) instead of `Default::default()`.

## Changes
- Removed `default_trait_access = "allow"` from `Cargo.toml`.
- Replaced `Default::default()` with explicit type-default calls in code.
- Refactored the httparse import in `src/server/conn/http1.rs` to align with the client-side style.

### Full path for h2 config
In `src/client/conn/http2.rs` and `src/server/conn/http2.rs`, I used fully qualified paths to avoid `Default::default()` while preventing name conflicts:
```rust
h2_builder: proto::h2::client::Config::default(),
```
*Question*:   
Does this look good to you, or  would you prefer introducing type aliases (e.g., `use proto::h2::client::Config as H2ClientConfig`) at the top of the file to keep it shorter?

### Import refactoring in `src/server/conn/http1.rs`
I changed `use httparse;` to `use httparse::ParserConfig;` 

If you prefer keeping the original import style to minimize the diff, please let me know. I will revert this specific part.

### Untouched test codes
I noticed that there are still many `Default::default()` occurrences in the test codes (e.g., `src/proto/h1/role.rs`).   Currently, the CI clippy job does not check test targets because it runs without --all-targets:

```YAML
cargo clippy --features full -- -D warnings
```
As a result, the CI passes with this PR.

*Question*:  
 Should I also update the test codes, or should we keep this PR focused only on the library code?

## Related issue
Part of a #4071 

Sorry for the long description! 
If there are any changes needed, please let me know. Otherwise, feel free to merge this PR if everything looks good to you.🙏

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,6 @@ cast_precision_loss = "allow" # TODO: consider
 checked_conversions = "allow"
 collapsible_match = "allow"
 decimal_literal_representation = "allow" # TODO: consider
-default_trait_access = "allow"
 else_if_without_else = "allow"
 enum_glob_use = "allow"
 float_arithmetic = "allow"
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -332,7 +332,7 @@ impl Builder {
             h09_responses: false,
             h1_writev: None,
             h1_read_buf_exact_size: None,
-            h1_parser_config: Default::default(),
+            h1_parser_config: ParserConfig::default(),
             h1_title_case_headers: false,
             h1_preserve_header_case: false,
             h1_max_headers: None,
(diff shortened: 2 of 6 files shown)


seanmonstar:
> Currently, the CI clippy job does not check test targets because it runs without --all-targets:

Yea, that's on purpose. There's a couple lints that are much less important in tests, like unwrap, panic, etc, and when I enabled `--all-targets`, it couldn't figure that out. If it can eventually, I'd be happy to enable it for all targets.

seanmonstar (approved):

PR #4117, opened 2026-06-26
style(proto): fix `decimal_literal_representation` lint

josetorrs:
Part of https://github.com/hyperium/hyper/issues/4071

> What it does
Warns if there is a better representation for a numeric literal.

> Why restrict this?
Especially for big powers of 2, a hexadecimal representation is usually more readable than a decimal representation.

from the [docs](https://rust-lang.github.io/rust-clippy/master/index.html#decimal_literal_representation) though TBH I think `65_535` is more readable here


seanmonstar:
Hm, I agree, the 65k feels clearer the FFFF. Perhaps not as useful of a lint...

josetorrs:
> Hm, I agree, the 65k feels clearer the FFFF. Perhaps not as useful of a lint...

added it to the explicitly allowed list

seanmonstar (approved):

PR #4118, opened 2026-06-26
fix(h1): append duplicate trailer values when encoding (match #4107)

mahirhir:
#4107 fixed the decode side: `decode_trailers` switched from `HeaderMap::insert` to `append`, so repeated trailer values received on the wire are preserved. The encode side is the unfixed sibling: `Encoder::encode_trailers` (`src/proto/h1/encode.rs`) builds `allowed_trailers` with `insert`, so when a user sets the same trailer field multiple times, only the last value reaches the wire.

The one-line fix mirrors #4107 (`insert` → `append`). Verified both ways with `cargo test --features full`: a regression test fails on the old code (only the second value reaches the wire) and passes after; encode module 10/10, full lib suite 100 passed, 0 regression.

### DIFF
--- a/src/proto/h1/encode.rs
+++ b/src/proto/h1/encode.rs
@@ -181,7 +181,7 @@ impl Encoder {
 
                     if allowed_set.contains(name) {
                         if is_valid_trailer_field(name) {
-                            allowed_trailers.insert(name, value);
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks!

PR #4120, opened 2026-06-27
remove arithmetic_side_effects

yanns:
part of https://github.com/hyperium/hyper/issues/4071

In general, I've added quite a lot of `#[allow(clippy::arithmetic_side_effects)]`. I still think that this lint is useful. It's easier to search for this and to review arithmetic operations. It can also prevent some issues in the future.

There are some parts where I've used some saturating operations, which can have an impact of performance. This needs a careful review.

If this is too complex to review, I can remove the new saturating operations and add the `allow` in those places. I could later open smaller PRs trying to address some of those.

### DIFF
--- a/src/rt/io.rs
+++ b/src/rt/io.rs
@@ -300,7 +300,13 @@ impl<'data> ReadBuf<'data> {
     }
 
     #[inline]
+    #[allow(clippy::arithmetic_side_effects)]
     fn remaining(&self) -> usize {
+        debug_assert!(
+            self.capacity() >= self.filled,
+            "filled must fit in capacity()"
+        );
+        // Cannot overflow, asserted above
         self.capacity() - self.filled
     }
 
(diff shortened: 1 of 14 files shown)


src/rt/io.rs
```diff
@@ -300,7 +300,13 @@ impl<'data> ReadBuf<'data> {
     }
 
     #[inline]
+    #[allow(clippy::arithmetic_side_effects)]
     fn remaining(&self) -> usize {
+        debug_assert!(
+            self.capacity() >= self.filled,
```
yanns, line 306: This should be true. We could use an `assert!` as in `put_slice`, but this would have some performance impact in release mode. I've opted for a `debug_assert`. Let me know if you think it's the right thing to do.
An alternative is to use `saturating_sub` which would have a performance cost.

PR #4126, opened 2026-07-09
Fix typos in comments and documentation

maxtaran2010:
Fix several minor typos found by codespell in source comments and documentation strings.


seanmonstar:
Thanks for the PR! But I think the words are fine the way they are.

PR #4131, opened 2026-07-17
hyper Improvement Proposals

seanmonstar:
This introduces an RFC-like process: hyper Improvement Proposals.

The reason I started working on this is because I realized there are several large features semi-proposed[^1] that need proper scrutiny, and the existing mechanisms to do so were inadequate. I also hope to put into writing more of the process of maintaining, with the hopes of empowering other collaborators to make progress.

They have many similarities with other RFC processes, so they won't be alien. But they also define specific values that are more important to hyper specifically. This includes HIP-0001, and also a template for future HIPs. And, just as the lifecycle section mentions submitting with bullet points, this proposal has them. They will be turned into prose before merging, if accepted.

[Rendered](https://github.com/hyperium/hyper/blob/master/docs/hips/0001-hyper-improvement-proposals.md)

[^1]: Big features I'm aware of: header casing, Buf providers, metrics, early hints, HTTP/3, header order.

> [!IMPORTANT]
> Please make line comments, instead of comments on the PR as a whole. It makes discussion much easier to keep untangled.

### DIFF
--- a/docs/hips/0000-template.md
+++ b/docs/hips/0000-template.md
@@ -1,0 +1,37 @@
+# HIP-0000: Template
+
+## Summary
+
+(The tl;dr. Concise statement of the reason for the document and recommendation.)
+
+## Tenets
+
+(List of tenets, values, guiding principles that are used to evaluate the proposal.)
+
+## Motivation
+
+(Clearly explain the problem.)
+
+## Recommendation
+
+### User Experience
+
+### Implementation
+
+### Testing Plan
+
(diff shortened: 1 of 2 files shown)


cratelyn (approved):

dswij (approved):
I'm almost always in favor of written proposals, so that's a +1 from me.

I wonder if we want an announcement, or at least a link in the main README to explain the HIP in detail, as in HIP-0001.

PR #4132, opened 2026-07-18
fix: preserve body order in vectored writes

### LINKED ISSUE #3650: Bug with response using struct implementing bytes::Buf
**Version**
hyper: 1.3.1
hyper-util 0.1.3
bytes: 1.6.0
tokio: 1.37,0

**Platform**
Linux LAPTOP-CSN0P74T 5.15.146.1-microsoft-standard-WSL2

**Summary**
For a project I'm building I wanted to provide serialization of some internal structures and send them out as responses.
To do this I implemented a struct that implementes hyper::body::Body and implemented bytes::Buf for my internal structure.
When sending the structure I found that parts of the Buf had been replaced by different data.
Specifically I found that not all parts of the Buf were read and instead different data was inserted.

**Minimal Viable Example**
Here is a minimal example that implements the Buf interface over a Vec<u8> but only returns single byte slices.

```
use std::{cmp::min, convert::Infallible, net::SocketAddr};

use hyper::{body::Frame, server::conn::http1, service::service_fn, Error, Request, Response};
use tokio::{net::TcpListener, spawn};

const TEST_MESSAGE: &str =
    "test message that needs to be a few bytes long for the sake of demonstration";

struct MinimalBytes {
    current: usize,
    data: Vec<u8>,
}

impl bytes::Buf for MinimalBytes {
        return self.data.len() - self.current;
    }

    fn advance(&mut self, cnt: usize) {
        if self.current + cnt > self.data.len() {
            panic!("Attempted to advance past end of MinimalBytes");
        }
        self.current += cnt;
    }

    fn chunk(&self) -> &[u8] {
        let end = min(self.current + 1, self.data.len());
        return &self.data[self.current..end];
    }
}

struct MinimalBody {
    frame: Option<MinimalBytes>,
}

impl hyper::body::Body for MinimalBody {
    type Data = MinimalBytes;
    type Error = Error;
    fn poll_frame(
        self: std::pin::Pin<&mut Self>,
        _cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<Option<Result<hyper::body::Frame<Self::Data>, Self::Error>>> {
        std::task::Poll::Ready(
            self.get_mut()
                .frame
                .take()
                .and_then(|min_bytes| Some(Ok(Frame::data(min_bytes)))),
        )
    }
}

async fn test_service(
    _: Request<hyper::body::Incoming>,
) -> Result<Response<MinimalBody>, Infallible> {
    return Ok(Response::new(MinimalBody {
        frame: Some(MinimalBytes {
            current: 0,
            data: String::from(TEST_MESSAGE).into_bytes(),
        }),
    }));
}

#[tokio::main]
async fn main() {
    // start server
    let addr = SocketAddr::from(([0, 0, 0, 0], 8080));
    let listener = TcpListener::bind(addr).await.unwrap();
    loop {
        let (stream, _) = listener.accept().await.unwrap();

        let io = hyper_util::rt::TokioIo::new(stream);

        spawn(async move {
            if let Err(err) = http1::Builder::new()
                .serve_connection(io, service_fn(test_service))
                .await
            {
                eprintln!("Error serving connection: {:?}", err);
            }
        });
    }
}
```

What I would expect to happen is a response that contains the test message. What happens instead is that after every byte of that message it inserts the 7 bytes `\r\n0\r\n\r\n` and then advances past 7 bytes in the actual buffer without every asking for those chunks.
I have checked it with a few different steps sizes, and it always seems to insert these specific 7 bytes if a chunk is returned that is not equal to the total size of remaining.
If the chunk length is equal to remaining then everything works.

Let me know if there is some additional information I could provide or something I could easily look into to help fix this.

snowyukitty:
`Buf::chunks_vectored` can expose a non-contiguous view when `chunk()` only
returns a partial prefix. A `Chain` can append slices from its tail even though
its head was not fully represented. Hyper then advances the logical buffer by
the number of bytes written, silently skipping body bytes.

This change:

- adds an internal HTTP/1 vectored-buffer contract that only composes slices
  known to form a contiguous prefix;
- exposes user-provided buffers in Exact, Limited, and Chunked encodings through
  `chunk()`, while retaining vectored composition for hyper-owned framing;
- stops HTTP/1 queue packing at the first buffer that is not fully covered; and
- keeps HTTP/2 `SendBuf` on `Buf`'s single-chunk default instead of forwarding
  arbitrary user `chunks_vectored` output.

The HTTP/2 change addresses the same bug class and is kept in a separate commit,
so it can be split out if preferred.

For common `Bytes` bodies, the iovec layout and cross-frame batching are
unchanged. Custom body frames backed by multiple chunks may require additional
writes; avoiding that without copying would require a stronger guarantee than
the current `Buf` contract provides.

Regression tests cover:

- nested partial buffers in Chunked, Exact, and Limited/`Take` encodings;
- real HTTP/1 `poll_flush` behavior with partial writes and queued frames;
- unchanged `Bytes` batching; and
- the equivalent HTTP/2 `SendBuf` path.

This follows the contract clarification discussed in
tokio-rs/bytes#701 and tokio-rs/bytes#754.

Closes #3650.


seanmonstar:
Oh yea, I forgot all about this. Thanks for looking into it! (I think `bytes::buf::Chain` is still wrong, too...)

One of my concerns when originally filing this was with the performance loss of having to defensively check all this. Do you how much this "costs"?

snowyukitty:
Yeah, I think you're right about `Chain`. `tokio-rs/bytes#754` removed the "all remaining slices" promise, but `Chain` can still append `b` after `a` has exposed only a prefix—so `[b"h", b"\r\n"]` while `b"ello"` is still waiting in `a` remains possible.

I did a small release-mode A/B probe, comparing the PR base (`540fff91`) with this branch (`13af40df`). The honest answer is that there are two quite different costs here.

The defensive bookkeeping itself appears to be tiny. This patch doesn't walk or sum arbitrary slices; at each hyper-owned boundary it takes one `chunk()`, compares its length with `remaining()`, and only crosses the boundary if that chunk covered the component completely.

In local gather microbenchmarks on my Windows machine, the common `Bytes` path moved from 15.63 to 15.85 ns for an Exact body, and from 35.81 to 36.62 ns for eight queued chunked frames—about 1.4% and 2.3%, both under one nanosecond per gather. My loopback end-to-end runs were noisy enough that I don't think they support a credible throughput claim in either direction.

The cost that can actually hurt is a well-behaved custom `Buf` that exposes several valid slices. This patch deliberately stops forwarding that multi-slice view. In an `8 × 1 KiB` segmented test, assuming each offered write completes, the base exposes all eight segments in one write attempt, while this branch exposes one segment per attempt (`1 → 8`). Ordinary `Bytes` bodies don't get that amplification: the body, framing, and queued frames still produce the same iovec layout.

So, it isn't free—but it also isn't a blanket tax on every body. The PR chooses a provable invariant—everything offered to I/O is exactly the prefix that `advance()` will consume—over multi-slice batching for arbitrary user buffers. That closes the skipped/repeated-byte corruption in both HTTP/1 and HTTP/2 while preserving the common `Bytes` batching path.

With the current `Buf` interface and current `Chain` behavior, hyper can't distinguish a safe multi-slice buffer from a gapped one without copying or getting a stronger upstream signal. That's the slightly awkward shape of the trade-off here.

If the segmented-buffer regression is too steep, I think the clean way to recover that optimization is upstream—a fixed `Chain`, and perhaps some sort of completeness/capability signal—rather than another length heuristic in hyper. I'd be happy to turn this probe into a reproducible Linux benchmark if that would make the decision easier.

snowyukitty:
I followed up on the Linux benchmark I offered. There is one correction to the scope of my earlier Windows numbers.

I ran paired, interleaved release builds of base `540fff91` and this branch on WSL2 Ubuntu 24.04:

- Exact 1 KiB `Bytes`: 15.371 -> 15.378 ns/gather. The paired delta was +0.007 ns (95% CI -0.050 to +0.066), so there was no measurable difference.
- Eight queued 1 KiB chunked `Bytes` frames: 38.858 -> 43.167 ns/gather, or +4.3 ns. A second run measured +4.0 ns. Both builds still produce the same 24 iovecs and one write attempt, so the 11% microbenchmark difference should not be read as an 11% HTTP throughput loss.
- An 8-slice custom `Buf`: the base used one `writev` attempt per 8 KiB, while this branch used eight. `strace` confirmed 1,000 versus 8,000 `writev` calls for 1,000 messages.

So my earlier "under one nanosecond" result holds for the Exact/`Bytes` path here, but not for the queued case on this toolchain. The practical trade-off is still fairly narrow: ordinary `Bytes` keeps the same write batching, while a custom `Buf` exposing multiple slices loses that batching in order to preserve the correct body order.

I also ran a WSL2 Unix-socket stress test. It made the extra syscall cost visible, but since this was not bare Linux TCP, I don't think its throughput ratio would be useful to generalize.

Given that scope, this still looks like a reasonable correctness/performance trade-off to me. No action needed; I just wanted to close the loop and correct the platform-specific part of my earlier answer. I have the reproducible harness and raw paired samples if they would be useful.

snowyukitty:
Quick maintenance update: I reproduced the HTTP/1 and HTTP/2 ordering regression against current unmodified master (`fca07bd3937d4a3a4822372acdf766d0640796af`) and prepared a regular merge of current master into the branch, without rebasing or force-updating it. The resolution preserves upstream's HTTP/1 cleanup and HTTP/2 capacity changes together with the `SendBuf` contiguous-prefix invariant.

The focused buffer/HTTP/1/HTTP/2 tests, the full full-feature library suite, clippy/check/build coverage, targeted rustfmt, and a release-mode performance sanity check all pass locally. I haven't pushed the merge yet, so the public head remains `0771af4b708f430ba09d0ed57f6895fcc2f283ef`.

snowyukitty:
Correction to my last message: I said the merge was not pushed and the head was still `0771af4b`. It was pushed — the merge commit is dated 16 August, two days after that message — and I did not come back to say so. Sorry for leaving that standing.

The branch had gone into conflict against `master` again since. I have merged current `master` into it; the head is now `5c54986d` and the pull request is mergeable again. Against `master` the branch is 4 files, +398/-36.

The only upstream commit that has touched those four files since this branch's base is #4173, which restyled the `chunks_vectored` overrides this branch removes, so upstream's side of each conflict was that method under a new lifetime name. I kept the `VectoredBuf` trait and `chunks_vectored_prefix`, and took #4173's convention with it: the `'t` lifetime in the code this branch adds is now `'data`. Across those four files that rename is the only change from the previously published head — eight lines, all signatures.

At `5c54986d`:

- `cargo test --lib --features full` — 122 passed, 0 failed, including `send_buf_vectored_is_contiguous_prefix` and `send_buf_bytes_stays_vectored_with_framing`
- `rustfmt --check` on the four files — clean
- `cargo clippy --features full --lib` — one warning, `unknown lint: clippy::unused_async_trait_impl`, which is identical on unmodified `master`

I could not run the integration targets here: `tests/support/tokiort.rs` is a symlink that this Windows checkout materialised as a text file, so they fail to compile before any of this branch's code is reached. That is my environment rather than the branch.

seanmonstar:
Checking in here, I do appreciate the desire to shore up this problem. But, if I'm understanding this correctly, the proposed fix here essentially disables vectored writes on custom `Buf`s. That can drastically hurt performance.

Considering that custom `Buf`s, and even `Chain`, can have this problem, the solution is likely something within the `bytes` crate. Either by redesigning that trait method (hard), or by providing some util function that can be called by all custom `chunks_vectored` implementations, to make sure the underlying bufs are not limited, and if they are, to stop collecting.

PR #4133, opened 2026-07-19
fix(http1): return 414 when request line overflows max_buf_size

### LINKED ISSUE #3874: Hyper returns status 431 instead of 414 if url overflows max_buf_size
**Version**
Hyper 0.14.30

**Platform**
ALL

**Description**

In my user case I use hyper as server and config it with a fixed http1_max_buf_size
```
hyper::Server::builder(incoming)
        .http1_max_buf_size(64 * 1024)
        ..serve(hyper::service::make_service_fn(... )
```

If the incomming url is greater than 64k, hyper returns `431 Request Header Fields Too Large` instead of  `414 URL TOO LONG`. And since hyper will close connection afterwards, my service_fn is unable to catch this exception.(Or maybe it's possible but I just don't know how to)

I inspected the code briefly and found out the problem is here: [here](https://github.com/hyperium/hyper/blob/72ebcffb7d82cda15aa74507b2cc522ca2a7a94d/src/proto/h1/io.rs#L234)

if url overflows max_buf_size hyper just returns Error::new_too_large() and parsed as `431 Request Header Fields Too Large`

I also inspect the [parse_headers()](https://github.com/hyperium/hyper/blob/72ebcffb7d82cda15aa74507b2cc522ca2a7a94d/src/proto/h1/io.rs#L187) implememtation, it seems that hyper must consume the whole buf to parse both request line and headers correctly. If the buf overflows, the current implemementation is unable to tell the difference between url too long or header too large. 

I am wondering if this is a bug or any special consideration for the current implemementation/

[code sample that causes the bug]

I expected to see this happen:  `414 URL TOO LONG`

Instead, this happened: `431 Request Header Fields Too Large`

deeferentleeg:
## Summary

When an HTTP/1 server is configured with `http1_max_buf_size` and receives a request whose start-line (method + URI + version) is so large that it fills the read buffer **before a newline is even seen**, hyper reported the failure as `431 Request Header Fields Too Large`. Per [RFC 7231 §6.5.12](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.12), the more specific status for an oversized URI is `414 URI Too Long`.

Closes #3874.

## Root cause

In `Buffered::parse` (`src/proto/h1/io.rs`), when `max_buf_size` is reached the code unconditionally returns `Error::new_too_large()` → `Parse::TooLarge` → `431`. There is no distinction between "the request line itself is too long" and "the header fields are too large."

This is exactly the fix direction @seanmonstar sketched in the issue:

> It's possible hyper could add some more code to better determine if the Start-Line never finished, when seeing a `TooLarge` error. Is that sufficient?

## Fix

Detect the case in `Buffered::parse`: when `max_buf_size` is reached on a **server** and **no `\n`** has appeared in the buffer yet, the overflow is caused by the request line (typically the URI) rather than the header fields. Return `Parse::UriTooLong` (→ `414`) instead of `Parse::TooLarge` (→ `431`).

The existing `Parse::UriTooLong` variant already maps to `StatusCode::URI_TOO_LONG` in `Server::on_error` (`src/proto/h1/role.rs`), and its `Display` string ("URI too long") and `is_parse_too_large()` already cover it — so no status-code mapping or display changes are needed.

### Changes

- **`src/error.rs`** — add `Error::new_uri_too_long()` constructor (gated `#[cfg(all(feature = "server", feature = "http1"))]`, matching the `Parse::UriTooLong` variant's cfg).
- **`src/proto/h1/io.rs`** — in `parse()`, when `max_buf_size` is hit on a server with no `\n` in the buffer, return `new_uri_too_long()` instead of `new_too_large()`. Added two unit tests:
  - `parse_overlong_uri_is_uri_too_long` — request line with no newline overflows buffer → expects `UriTooLong`.
  - `parse_overlong_headers_is_too_large` — complete request line + oversized headers → still expects `TooLarge` (regression guard).
- **`tests/server.rs`** — updated the `max_buf_size` integration test: it sends `POST /` + a large payload (no newline), so the expected response is now `414` instead of `431`.

The existing `max_buf_size_split_header_boundary` test (which sends a complete request line *then* overflows on headers) continues to expect `431` and is unaffected.

## AI disclosure

This change was developed with AI assistance. The human author reviewed every line, verified the root cause against the source code, and owns the change.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -372,6 +372,11 @@ impl Error {
         Error::new(Kind::Parse(Parse::TooLarge))
     }
 
+    #[cfg(all(feature = "server", feature = "http1"))]
+    pub(super) fn new_uri_too_long() -> Error {
+        Error::new(Kind::Parse(Parse::UriTooLong))
+    }
+
     #[cfg(all(any(feature = "client", feature = "server"), feature = "http1"))]
     pub(super) fn new_version_h2() -> Error {
         Error::new(Kind::Parse(Parse::VersionH2))
--- a/src/proto/h1/io.rs
+++ b/src/proto/h1/io.rs
@@ -199,6 +199,14 @@ where
                 let curr_len = self.read_buf.len();
                 if curr_len >= max {
                     debug!("max_buf_size ({}) reached, closing", max);
+                    // When parsing a request and the start-line (request
+                    // line) hasn't even finished — i.e. no newline has been
+                    // seen yet — the overflow is caused by an oversized URI
+                    // rather than the header fields, so report `414 URI Too
+                    // Long` instead of `431 Request Header Fields Too Large`.
+                    if S::is_server() && !self.read_buf.contains(&b'\n') {
+                        return Poll::Ready(Err(crate::Error::new_uri_too_long()));
+                    }
                     return Poll::Ready(Err(crate::Error::new_too_large()));
                 }
                 if curr_len > 0 {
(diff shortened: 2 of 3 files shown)


Samielakkad (commented):
Retracted.

cratelyn:
hello, thank you for opening a pull request. i took a look through the discussion in https://github.com/hyperium/hyper/issues/3874 and it sounds like there is not consensus about doing this. i could not identify any real-world use cases that this proposal solves.

if that discussion progresses further and we come to decision, i'd welcome this being reopened, but in the meantime i think the best way to help would be identifying any real-world applications that cannot work around the existing status code behavior.

PR #4135, opened 2026-07-24
headers: optimize from_digits using try_fold for functional digit accumulation

Aditya-9-6:
### Summary
This PR optimizes `from_digits` in `src/headers.rs` by replacing the imperative `for &b in bytes` loop with a functional `try_fold` iterator pipeline.

### Rationale & Performance Benefit
`from_digits` is executed on every HTTP request and response containing numerical headers like `Content-Length`.

Using `try_fold`:
- Streamlines register allocation and allows LLVM to generate branch-free digit accumulation.
- Lowers single header parsing overhead by ~45.8%.

### DIFF
--- a/src/headers.rs
+++ b/src/headers.rs
@@ -68,30 +68,16 @@ pub(super) fn content_length_parse_all_values(values: ValueIter<'_, HeaderValue>
 }
 
 fn from_digits(bytes: &[u8]) -> Option<u64> {
-    // cannot use FromStr for u64, since it allows a signed prefix
-    let mut result = 0u64;
-    const RADIX: u64 = 10;
-
     if bytes.is_empty() {
         return None;
     }
 
-    for &b in bytes {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! Are you sure of the improvement? Exploring the ASM of both versions shows nearly identical output. I would expect LLVM to be able to vectorize the regular loop too.

Aditya-9-6:
I compiled both versions with `rustc -O --emit=asm` to inspect the generated machine instructions:

```assembly
.LBB_loop:
	movzbl	(%rcx,%rsi), %ebx
	addb	$-48, %bl           # '0' subtraction
	cmpb	$9, %bl             # Digit check (<= 9)
	ja	.LBB_err
	mulq	%r11                # * 10
	addq	%rdx, %rax          # + digit

seanmonstar:
Ok, thanks. There's no real difference then. Let's keep it the way it is.

PR #4141, opened 2026-08-01
refactor(error/upgrade): move from the use of ref patterns to `&` and remove allow for lint `ref_patterns`

logan-bobo:
**This PR aims to** 
- Remove `allow` for `ref_patterns` lint, that will now be caught by a `warn` on `restriction`
```
restriction = { level = "warn", priority = -2 }
```
- Move from the use of ref patterns to `&`

**References** 
- https://doc.rust-lang.org/reference/patterns.html?highlight=ref#binding-modes

progresses https://github.com/hyperium/hyper/issues/4071

### DIFF
--- a/src/body/incoming.rs
+++ b/src/body/incoming.rs
@@ -172,15 +172,12 @@ impl Incoming {
 
     #[cfg(feature = "ffi")]
     pub(crate) fn as_ffi_mut(&mut self) -> &mut crate::ffi::UserBody {
-        match self.kind {
-            Kind::Ffi(ref mut body) => return body,
-            _ => {
-                self.kind = Kind::Ffi(crate::ffi::UserBody::new());
-            }
+        if !matches!(self.kind, Kind::Ffi(_)) {
+            self.kind = Kind::Ffi(crate::ffi::UserBody::new());
         }
 
-        match self.kind {
-            Kind::Ffi(ref mut body) => body,
+        match &mut self.kind {
+            Kind::Ffi(body) => body,
             _ => unreachable!(),
         }
     }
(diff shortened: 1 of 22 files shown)


tests/server.rs
```diff
@@ -3563,7 +3563,7 @@ impl Drop for Serve {
         drop(self.thread.take());
         /*
         let r = self.thread.take().unwrap().join();
-        if let Err(ref e) = r {
+        if let Err(e) = &r {
             println!("{:?}", e);
         }
         r.unwrap();
```
logan-bobo, line 3569: I wonder if all of this code can be dropped?

src/body/incoming.rs
```diff
@@ -172,15 +172,12 @@ impl Incoming {
...
+            self.kind = Kind::Ffi(crate::ffi::UserBody::new());
         }
 
-        match self.kind {
-            Kind::Ffi(ref mut body) => body,
+        match &mut self.kind {
+            Kind::Ffi(body) => body,
             _ => unreachable!(),
         }
     }
```
logan-bobo, line 183: I want to call this one out as the only place we could not do a straight `ref` -> `&` change. 

The behaviour however is still the same.
1. If the kind is ffi return a mutable reference to the body
2. if the kind is not ffi set self.kind to `Kind::Ffi(crate::ffi::UserBody::new())`
3. After creating `Kind::Ffi(crate::ffi::UserBody::new())` return as a mutable reference.

tests/server.rs
dswij: We can drop this one

src/body/incoming.rs
dswij: Looks fine to me :+1:

dswij (approved):

PR #4148, opened 2026-08-09
docs(body): add streaming read examples

### LINKED ISSUE #2201: Expand 'body' documentation to explain body streams
The docs of `hyper::body` should be expanded to explain that request and response bodies in hyper are streamed, not buffered. And we should show some examples on that page about how to read the body data.

fzlzjerry:
## Summary

- document how to read data and trailer frames from an incoming body
- explain how frame-by-frame processing preserves connection back-pressure
- show intentional whole-body collection and warn about unbounded buffering

## Testing

- `cargo fmt --check`
- `cargo test --doc --features full`
- `cargo test --features full`
- `cargo clippy --features full -- -D warnings`
- `RUSTDOCFLAGS='-D rustdoc::broken-intra-doc-links' cargo doc --no-deps --features full`

Closes #2201

### DIFF
--- a/src/body/mod.rs
+++ b/src/body/mod.rs
@@ -52,6 +52,7 @@
 //! use http_body_util::BodyExt as _;
 //! use hyper::body::{Bytes, Incoming};
 //!
+//! /// Consider using `Limited` if the body is untrusted.
 //! async fn read_entire_body(body: Incoming) -> Result<Bytes, hyper::Error> {
 //!     Ok(body.collect().await?.to_bytes())
 //! }
--- a/src/body/mod.rs
+++ b/src/body/mod.rs
@@ -17,7 +17,51 @@
 //! There are additional implementations available in [`http-body-util`][],
 //! such as a `Full` or `Empty` body.
 //!
+//! ## Reading a body
+//!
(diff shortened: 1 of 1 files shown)


src/body/mod.rs
```diff
@@ -17,7 +17,51 @@
...
+//! polling for the next one preserves back-pressure on the connection.
+//!
+//! If a body is known to be small, it can be collected into memory instead:
+//!
+//! ```
+//! use http_body_util::BodyExt as _;
+//! use hyper::body::{Bytes, Incoming};
+//!
+//! async fn read_entire_body(body: Incoming) -> Result<Bytes, hyper::Error> {
+//!     Ok(body.collect().await?.to_bytes())
```
seanmonstar, line 57: I know people copy and past examples without reading the surrounding text. Should we include a comment so that its more likely they copy the recommendation to use a limit?

```
/// Consider using `Limited` if the body is untrusted.
```

fzlzjerry: Good point. I added the warning directly above the collecting function so it stays with the example when copied. The updated snippet passes `cargo test -p hyper --doc --features full`.

seanmonstar (approved):

PR #4150, opened 2026-08-10
fix(http1): synchronize dispatch shutdown with sends

### LINKED ISSUE #4122: Body dropped too late when an h1 handshake hits an error
### Version

hyper 1.10.1

### Platform

Linux x86_64 7.0.0-27-generic

### Summary

Wasmtime hit a spurious failure in a test recently in CI that I think I've diagnosed to some code in hyper. The test in question is a wasm guest that issues an HTTP/1.1 request to a server that only works with HTTP/2, and the expectation is that the wasm guest gets a `HttpProtocolError` within wasm. This test ended up timing out in CI and getting a timeout failure instead.

After some investigation it looks like this is due hyper's use of tokio's channels and the precise synchronization/sequencing of sending a request from one channel to another. The code in question attached here is a small program which showcases this race where the intention of the test is:

* No actual I/O is performed, it's just in-memory data structures.
* An http1 handshake is done against a connection that reads corrupt data (sort of mimicking what an http2 server might initiate with)
* The code in question sends a request, completes the connection I/O, and then asserts that the request was dropped.
* There's various bits and pieces of the program intended to stress scheduling to make the bug appear more often.

### Code Sample

```rust
use bytes::Bytes;
use http_body::Body;
use hyper::rt::{Read, ReadBufCursor, Write};
use std::future::Future;
use std::io;
use std::pin::Pin;
use std::sync::atomic::{AtomicBool, Ordering};
use std::sync::{Arc, Barrier};
use std::task::{Context, Poll, Waker};

fn main() {
    // start up some background work to stress scheduling a bit more.
    for _ in 0..16 {
        std::thread::spawn(|| loop {
            std::hint::black_box(std::time::Instant::now());
        });
    }

    let rt = tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()
        .unwrap();
    let _guard = rt.enter();

    for _ in 0..2_000_000 {
        // Handshake against the mock IO. handshake() itself does no I/O.
        let (mut sender, mut conn) = match rt
            .block_on(hyper::client::conn::http1::handshake::<_, TrackedBody>(
                MockIo::default(),
            )) {
            Ok(p) => p,
            Err(_) => continue,
        };

        let dropped = Arc::new(AtomicBool::new(false));
        let req = hyper::Request::connect("http://x/")
            .body(TrackedBody {
                dropped: dropped.clone(),
            })
            .unwrap();

        let barrier = Arc::new(Barrier::new(2));

        // Thread B: dispatch the request the instant the barrier releases.
        let b2 = barrier.clone();
        let sender_thread = std::thread::spawn(move || {
            b2.wait();
            let fut = sender.send_request(req);
            (fut, sender)
        });

        barrier.wait();
        let waker = Waker::noop();
        let mut cx = Context::from_waker(&waker);
        for _ in 0..4 {
            if Pin::new(&mut conn).poll(&mut cx).is_ready() {
                break;
            }
        }
        drop(conn);

        let (fut, sender2) = sender_thread.join().unwrap();
        drop(fut);

        // At this point the future from `send_request` is gone, the receiver
        // through `conn` is gone, so it should be the case that our body in the
        // request that was sent was dropped. However this isn't always true so
        // sometimes this will print out.
        if !dropped.load(Ordering::SeqCst) {
            println!("body wasn't dropped");
        }
        drop(sender2);
        assert!(
            dropped.load(Ordering::SeqCst),
            "request must be reclaimed once the SendRequest drops"
        );
    }
}

#[derive(Default)]
struct MockIo {
    gave: bool,
}

impl Read for MockIo {
    fn poll_read(
        mut self: Pin<&mut Self>,
        _cx: &mut Context<'_>,
        mut buf: ReadBufCursor<'_>,
    ) -> Poll<io::Result<()>> {
        // read enough bytes that the connection should terminate with an error.
        if !self.gave {
            self.gave = true;
            buf.put_slice(&[0u8; 9]);
        }
        Poll::Ready(Ok(()))
    }
}

impl Write for MockIo {
    fn poll_write(
        self: Pin<&mut Self>,
        _: &mut Context<'_>,
        _buf: &[u8],
    ) -> Poll<io::Result<usize>> {
        // shouldn't be called for h1
        unreachable!()
    }
    fn poll_flush(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<io::Result<()>> {
        Poll::Ready(Ok(()))
    }
    fn poll_shutdown(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<io::Result<()>> {
        Poll::Ready(Ok(()))
    }
}

/// An empty body that sets `dropped` to true when dropped.
struct TrackedBody {
    dropped: Arc<AtomicBool>,
}

impl Body for TrackedBody {
    type Data = Bytes;
    type Error = std::convert::Infallible;
    fn poll_frame(
        self: Pin<&mut Self>,
        _: &mut Context<'_>,
    ) -> Poll<Option<Result<http_body::Frame<Bytes>, Self::Error>>> {
        Poll::Ready(None)
    }
}

impl Drop for TrackedBody {
    fn drop(&mut self) {
        self.dropped.store(true, Ordering::SeqCst);
    }
}
```

### Expected Behavior

My expectation of this program is that the body of the request is always dropped by the time that the only live value is the sender itself. Effectively this should never print `body wasn't dropped`. 

### Actual Behavior

Locally ~2m iterations show ~20 instances of the body not being dropped before the sender is fully dropped. This is how in Wasmtime's embedding it ended up showing as a timeout because we rely on the body being dropped to register when the request is sent and such (or an error occurred).

### Additional Context

Some poking and prodding around things shows that one possible culprit here is [this line](https://github.com/hyperium/hyper/blob/e0d14d19a0a87962efe92acdaa029253be54a612/src/client/dispatch.rs#L202) which is a bit more robust using Tokio's `try_recv` (my guess is that `try_recv` was added after this was authored). Even with that, however, the program above still doesn't drop the body 100% of the time, and this seems due to the fact that even if `close` has been called on a channel it's possible for `try_recv` to return `Empty` (as opposed to `Disconnected`) if there's an outstanding permit of a message being sent. I think this is basically still a race between the sender/receiver where `try_recv` isn't enough, hyper would have to loop over `try_recv`.

Overall this is where I figured it'd be best to file an issue. I'd want to confirm the expectation that this program should reliably drop the body, and then also see what others' thoughts are on this.

fzlzjerry:
Fixes #4122.

## Summary

- close the HTTP/1 dispatch receiver before teardown
- continue polling the closed receiver until a concurrent synchronous send has either published its envelope or observed the closure
- add a regression test that exercises the close/send race 10,000 times

## Background

Tokio's unbounded channel reserves message capacity before it publishes the
envelope into the channel list. Receiver shutdown could close the channel and
observe a pending receive during that window. The send would then finish after
shutdown, leaving the request owned by the channel until the remaining
`SendRequest` was dropped.

After closing the receiver, `Pending` means that such a pre-close send is still
in flight. Retrying the existing single-poll receive until it yields the
envelope or reports the closed channel removes that window without introducing
an additional synchronization object. The HTTP/2 path remains unchanged.

## Validation

- `cargo test --features full` (309 passed, 10 ignored)
- `cargo +1.97.1 clippy --features full -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo check --no-default-features --features client,http1`
- `cargo check --no-default-features --features client,http2`
- `cargo check --no-default-features --features client,http1,http2`
- minimal dependency resolution followed by `cargo check --features full`

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -192,16 +192,18 @@ impl<T, U> Receiver<T, U> {
     }
 
     #[cfg(feature = "http1")]
-    pub(crate) fn close(&mut self) {
+    pub(crate) fn close_and_recv(&mut self) -> Option<(T, Callback<T, U>)> {
         self.taker.cancel();
         self.inner.close();
-    }
-
-    #[cfg(feature = "http1")]
-    pub(crate) fn try_recv(&mut self) -> Option<(T, Callback<T, U>)> {
-        match crate::common::task::now_or_never(self.inner.recv()) {
-            Some(Some(mut env)) => env.0.take(),
-            _ => None,
+        loop {
+            match crate::common::task::now_or_never(self.inner.recv()) {
+                Some(Some(mut env)) => return env.0.take(),
+                Some(None) => return None,
+                // A send can reserve capacity immediately before close() and
+                // publish its envelope immediately afterwards. Once closed,
+                // Pending means that such a synchronous send is still in flight.
+                None => std::hint::spin_loop(),
+            }
         }
     }
 }
@@ -394,10 +396,63 @@ mod tests {
     use std::pin::Pin;
     use std::task::{Context, Poll};
 
+    #[cfg(feature = "http1")]
+    use std::sync::atomic::{AtomicBool, Ordering};
+    #[cfg(feature = "http1")]
+    use std::sync::{mpsc as std_mpsc, Arc, Barrier};
+
     use super::{channel, Callback, Receiver};
 
     #[derive(Debug)]
     struct Custom(#[allow(dead_code)] i32);
+
+    #[cfg(feature = "http1")]
+    struct TrackDrop(Arc<AtomicBool>);
+
+    #[cfg(feature = "http1")]
+    impl Drop for TrackDrop {
+        fn drop(&mut self) {
+            self.0.store(true, Ordering::SeqCst);
+        }
+    }
+
+    #[cfg(feature = "http1")]
+    #[test]
+    fn receiver_shutdown_reclaims_concurrent_send() {
+        let barrier = Arc::new(Barrier::new(2));
+        let worker_barrier = barrier.clone();
+        let (work_tx, work_rx) = std_mpsc::channel::<(super::Sender<TrackDrop, ()>, TrackDrop)>();
+        let (done_tx, done_rx) = std_mpsc::channel();
+        let worker = std::thread::spawn(move || {
+            while let Ok((mut tx, val)) = work_rx.recv() {
+                worker_barrier.wait();
+                drop(tx.send(val));
+                worker_barrier.wait();
+                done_tx.send(tx).unwrap();
+            }
+        });
+
+        // Tokio's unbounded channel reserves capacity before publishing the
+        // envelope. Repeat enough times to exercise shutdown in that window.
+        for _ in 0..10_000 {
+            let (tx, mut rx) = channel::<TrackDrop, ()>();
+            let dropped = Arc::new(AtomicBool::new(false));
+            work_tx.send((tx, TrackDrop(dropped.clone()))).unwrap();
+            barrier.wait();
+            drop(rx.close_and_recv());
+            drop(rx);
+            barrier.wait();
+            let tx = done_rx.recv().unwrap();
+            assert!(
+                dropped.load(Ordering::SeqCst),
+                "value remained queued after the receiver was dropped"
+            );
+            drop(tx);
+        }
+
+        drop(work_tx);
+        worker.join().unwrap();
+    }
 
     impl<T, U> Future for Receiver<T, U> {
         type Output = Option<(T, Callback<T, U>)>;
--- a/src/proto/h1/dispatch.rs
+++ b/src/proto/h1/dispatch.rs
@@ -728,8 +728,7 @@ cfg_client! {
                         }));
                         Ok(())
                     } else if !self.rx_closed {
-                        self.rx.close();
-                        if let Some((req, cb)) = self.rx.try_recv() {
+                        if let Some((req, cb)) = self.rx.close_and_recv() {
                             trace!("canceling queued request with connection error: {}", err);
                             // in this case, the message was never even started, so it's safe to tell
                             // the user that the request was completely canceled


src/client/dispatch.rs
```diff
@@ -51,6 +64,10 @@ pub(crate) struct Sender<T, U> {
     /// without notice.
     #[cfg(feature = "http1")]
     buffered_once: bool,
+    /// Synchronizes HTTP/1 sends with receiver shutdown so an envelope cannot
+    /// be published after the receiver has finished draining the channel.
+    #[cfg(feature = "http1")]
+    send_lock: Arc<Mutex<()>>,
```
seanmonstar: An additional Arc Mutex just for synchronizing this feels heavy handed. I'd rather a solution that doesn't need one.

seanmonstar (changes requested):

src/client/dispatch.rs
fzlzjerry: Agreed—the extra `Arc<Mutex>` was broader than needed. I replaced it in [`42dfc554`](https://github.com/hyperium/hyper/pull/4150/commits/42dfc554fb28c9f49ede431191c1372efee7e3da) with a close-and-drain loop using Tokio's existing `UnboundedReceiver::try_recv`. After `close()`, `Empty` represents the narrow window where a synchronous send reserved capacity but has not published its envelope yet, so the receiver retries until it receives that envelope or observes `Disconnected`. I also removed the now-unused no-op-waker helper and rebased onto current `master`. Validation: `cargo test --features full` (309 passed, 10 ignored), `cargo clippy --features full -- -D warnings`, and `cargo fmt --check`.

fzlzjerry:
Updated the implementation in [`333d33a`](https://github.com/hyperium/hyper/pull/4150/commits/333d33af999b1367735b1c19d7c5ec63434c6215) to remove the added `Arc<Mutex>`. Shutdown now closes the receiver and polls its existing receive future until a synchronous send that already reserved capacity either publishes its envelope or the channel terminates.

The new run passes the race regression, stable/beta/nightly tests, MSRV, minimal versions, Miri, features, docs, and the remaining checks. The sole red Linter job is Rust 1.98 reporting three new lints in untouched files (`src/body/incoming.rs`, `src/client/conn/http1.rs`, and `src/client/conn/http2.rs`); `cargo +1.97.1 clippy --features full -- -D warnings` passes locally. The PR diff remains limited to the dispatch shutdown path and its regression test.

fzlzjerry:
Rebased this onto current `master` (`v1.11.1`), which also picks up the Rust 1.98 lint fixes from the untouched files. The no-`Arc<Mutex>` implementation remains unchanged. Full tests and Clippy pass on Rust 1.98, and the nightly bench targets pass as well. @seanmonstar, could you take another look when convenient?

PR #4154, opened 2026-08-10
fix missing_errors_doc lint

CinisBorn:
fix `missing_errors_doc` lint warnings and remove lint exception from `Cargo.toml`.

Refs: [#4071](https://github.com/hyperium/hyper/issues/4071#issue-4481161595)

# Note 

I don't have a broad experience in the codebase. Some error messages may be too general.

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -74,6 +74,10 @@ pub struct Builder<Ex> {
 ///
 /// This is a shortcut for `Builder::new(exec).handshake(io)`.
 /// See [`client::conn`](crate::client::conn) for more.
+///
+/// # Errors
+///
+/// Returns an error if the HTTP/2 connection handshake fails.
 pub async fn handshake<E, T, B>(
     exec: E,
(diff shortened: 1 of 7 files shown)


CinisBorn:
This is a surprise! I will fix it right now.

src/client/conn/http2.rs
```diff
@@ -147,6 +153,15 @@ where
...
     /// would be discarded.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the connection is not ready or if an error occurs while
+    /// processing the request.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the internal dispatch channel is dropped without returning an error. This indicates a bug.
```
seanmonstar: Could we only document errors in this PR? I'm not sure about documenting internals like this, they could change.

CinisBorn: Sure! I will fix it right now.

seanmonstar (approved):
Thanks!

PR #4164, opened 2026-08-23
feat(http2): expose h2 data_frame_budget on the client conn builder

LukasKorba:
h2 0.4.16 added a client-side guard against floods of small DATA frames (connection closed with
`GOAWAY ENHANCE_YOUR_CALM` / `too_many_data_frames`), and h2 0.4.18 added the tuning knob for it:
`client::Builder::data_frame_budget`. hyper's `client::conn::http2::Builder` doesn't expose it, so
callers whose peers legitimately send many small frames have no mitigation. Real-world case: gRPC
server-streaming APIs that emit one small message per frame — e.g. Zcash's lightwalletd streams
one ~600-byte block per DATA frame, and a 10,000-item range trips the default budget partway
through, deterministically, on released hyper 1.11.0 + h2 0.4.18 (minimal tonic-only reproducer
attached to, and full context in, zcash/lightwalletd#593:
https://github.com/zcash/lightwalletd/issues/593).

This PR adds `Builder::data_frame_budget` following the exact pass-through pattern of
`max_concurrent_reset_streams` / `reset_stream_duration`, and bumps the h2 floor to 0.4.18 (the
release that introduced the method). Verified: `cargo test --features full` all green; and with
this passthrough plumbed further up through tonic, the reproducer's failing range completes when
the budget is raised, and fails immediately when the budget is set artificially low — the setting
demonstrably reaches h2 in both directions.

A sibling tonic PR exposing this on `Endpoint` is filed as a draft blocked on this one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ atomic-waker = { version = "1.1.2", optional = true }
 futures-channel = { version = "0.3", optional = true }
 futures-core = { version = "0.3.31", optional = true }
 futures-util = { version = "0.3", default-features = false, features = ["alloc"], optional = true }
-h2 = { version = "0.4.14", optional = true }
+h2 = { version = "0.4.18", optional = true }
 http-body-util = { version = "0.1", optional = true }
 httparse = { version = "1.9", optional = true }
 httpdate = { version = "1.0", optional = true }
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -392,6 +392,21 @@ where
         self
     }
 
+    /// Sets a connection-level budget for limiting memory overhead from
+    /// received small DATA frames.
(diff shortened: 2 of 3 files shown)


seanmonstar:
A new release on Monday for h2 will try to determine a better budget size based on what you configured for connection window. If that is sufficient, I'd rather not have to add configs elsewhere.

LukasKorba:
Confirmed sufficient: with h2 0.4.19 at default config, the reproducer from [zcash/lightwalletd#593](https://github.com/zcash/lightwalletd/issues/593) completes 10,000/10,000 blocks (3/3 runs); a same-hour control on 0.4.18 still fails with GoAway(too_many_data_frames). The window-derived budget covers this case — closing as superseded. Thanks for the fast root fix.

PR #4166, opened 2026-08-25
feat: make H2Upgraded public for HTTP/2 upgrade downcasting

### LINKED ISSUE #3987: Make `H2Upgraded` public to allow downcasting `Upgraded` for `Sync` access to underlying IO
Hi, I'd like to request that `hyper::proto::h2::H2Upgraded<T>` be made `pub`, so users can downcast an `hyper::upgrade::Upgraded` object to access the underlying I/O type for HTTP/2 connections.

### Problem
Currently, `Upgraded` is **not `Sync`**, which prevents sending it across threads or storing it in `Sync` contexts. While HTTP/1 upgraded connections can be downcast using `.downcast::<Tokio<TcpStream >>()`, but the underlying I/O type for HTTP/2 `H2Upgraded<T>` is `private`, making it impossible to downcast and recover the inner IO (which is often `Sync`).

```rust
// This fails to compile — `H2Upgraded` is private
upgraded.downcast::<hyper::proto::h2::H2Upgraded<bytes::Bytes>>()
```

### Note
- The construction of H2Upgraded:
  - client side: https://github.com/hyperium/hyper/blob/be18a92bb51498a2113dcfe15d9da09e1c6c7b0d/src/proto/h2/client.rs#L591-L593
  - server side: https://github.com/hyperium/hyper/blob/be18a92bb51498a2113dcfe15d9da09e1c6c7b0d/src/proto/h2/server.rs#L495-L503 
- This issue is partly similar to https://github.com/hyperium/hyper/issues/3587, but it doesn't solve the problem in the HTTP/2 scenario.

Thanks for all the great work on Hyper! I'd be happy to help implement this if you agree.

atheerium:
Fixes hyperium/hyper#3987 — Users currently cannot downcast upgraded HTTP/2 connections to access the underlying IO. H2Upgraded was hidden behind pub(crate) modules and pub(super) visibility. This change makes H2Upgraded public with documentation and Debug impl, and makes the module path public.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -144,7 +144,11 @@ pub mod ffi;
 
 cfg_proto! {
     mod headers;
-    mod proto;
+    #[cfg_attr(docsrs, doc(cfg(all(
+        any(feature = "http1", feature = "http2"),
+        any(feature = "client", feature = "server"),
+    ))))]
+    pub mod proto;
 }
 
 cfg_feature! {
--- a/src/proto/h2/mod.rs
+++ b/src/proto/h2/mod.rs
@@ -14,7 +14,8 @@ use pin_project_lite::pin_project;
 use crate::body::Body;
 
 pub(crate) mod ping;
-pub(crate) mod upgrade;
+/// HTTP/2 connection upgrade handling.
+pub mod upgrade;
 
 cfg_client! {
     pub(crate) mod client;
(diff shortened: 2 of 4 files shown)


cratelyn:
hi @atheerium. i'll note @seanmonstar's comment here: https://github.com/hyperium/hyper/issues/3987#issuecomment-3597251323

> I don't think I'd necessarily want to make `H2Upgraded` public.
>
> But `Upgraded` is `Send`, so you can send it to another thread. By not being `Sync`, it just means you need to put it in a synchronizing type before you allow thread-safe clones. So `Arc<Upgraded>` isn't allowed, but `Arc<Mutex<Upgraded>>` is perfectly legal.

i don't think this PR is a change that we're interested in making to hyper's API.

PR #4167, opened 2026-08-26
clippy: remove lints removed in newer clippy versions

tangaac:
## Summary

Recent clippy versions have removed several lints that were previously
allowed in `Cargo.toml`. With `-D warnings` (or the default
`-W unknown-lints`), these produce `unknown lint` errors that break CI on
stable and newer toolchains.

Removed lints:

- `clippy::unnecessary_rest_pattern`
- `clippy::rest_pattern_accessible_field`
- `clippy::definition_in_module_root`

These were all set to `"allow"` with TODO comments indicating they were
already under consideration.

## Verification

`cargo clippy --features full --lib` no longer emits `unknown lint`
warnings after this change.


seanmonstar (approved):

dswij:
These lints are not removed, though. It hasn't landed on stable yet.

See 

1. https://rust-lang.github.io/rust-clippy/master/index.html?search=unnecessary_rest_p
2. https://rust-lang.github.io/rust-clippy/master/index.html?search=rest_pattern_acces
3. https://rust-lang.github.io/rust-clippy/master/index.html?search=definition_in

seanmonstar:
Yea, I thought they were newly added when I fixed up CI a couple days ago. And CI seemed fine that I had added the rule... But CI was green on this PR too. (I got very little sleep last night, so might be why I'm more confused.)

dswij:
Seems like the CI job did emit the warnings in master, but it came from rustc instead of clippy, so it's not denied :sweat_smile:  

https://github.com/hyperium/hyper/actions/runs/32725621086/job/97426119277

seanmonstar:
OK, I see. So then these are perhaps new lints to add with the MSRV bump I guess. 😵

PR #4173, opened 2026-08-28
style: replace single-character lifetime names with descriptive names

dox4:
Removes the `single_char_lifetime_names` allow lint from the strict clippy config and replaces all single-character lifetime names with descriptive names.

Part of #4071.

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -408,7 +408,7 @@ mod tests {
     }
 
     /// Helper to check if the future is ready after polling once.
-    struct PollOnce<'a, F>(&'a mut F);
+    struct PollOnce<'func, F>(&'func mut F);
 
     impl<F, T> Future for PollOnce<'_, F>
     where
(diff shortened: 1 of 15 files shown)


src/client/dispatch.rs
```diff
@@ -408,7 +408,7 @@ mod tests {
     }
 
     /// Helper to check if the future is ready after polling once.
-    struct PollOnce<'a, F>(&'a mut F);
+    struct PollOnce<'func, F>(&'func mut F);
```
cratelyn:
```suggestion
    struct PollOnce<'fut, F>(&'fut mut F);
```

this might be a better symbol, given that `F` is a future.

cratelyn (approved):
thank you! i have one small nit about one of the lifetime names used here.

dox4:
> thank you! i have one small nit about one of the lifetime names used here.

thank you.

dswij (approved):

PR #4182, opened 2026-09-03
feat(http1): support expect: 100-continue on the client

### LINKED ISSUE #3833: Update Hyper Client to support handling `expect: 100-continue` header

bountis:
Handle the `Expect: 100-continue` header fully.
https://www.rfc-editor.org/rfc/rfc9110.html#name-expect

This adds the missing handling logic of the header. The client didn't handle the header previously.

If there's the `Expect: 100-continue` header present the client only writes the head and parks the body.
If the server returns `100 Continue` then go on and write the body.
If the server returns any non-1xx response the client drops the body and closes the connection.
Empty body ignores the header completely.

Also add a configurable timer to the client, client sends the body after the timeout, only if there's a timer setup, if not then it waits indefinitely.

Closes #3833

### DIFF
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -234,7 +234,10 @@ where
             }
         }
 
-        let msg = match self.io.parse::<T>(
+        #[cfg(feature = "client")]
+        let mut seen_continue = false;
+
+        let parse_result = self.io.parse::<T>(
             cx,
             ParseContext {
                 cached_headers: &mut self.state.cached_headers,
@@ -247,8 +250,17 @@ where
                 h09_responses: self.state.h09_responses,
                 #[cfg(feature = "client")]
                 on_informational: &mut self.state.on_informational,
+                #[cfg(feature = "client")]
+                seen_continue: &mut seen_continue,
             },
-        ) {
+        );
+
+        #[cfg(feature = "client")]
+        if seen_continue {
+            self.release_continue();
+        }
+
+        let msg = match parse_result {
             Poll::Ready(Ok(msg)) => msg,
             Poll::Ready(Err(e)) => return self.on_read_head_error(e),
             Poll::Pending => {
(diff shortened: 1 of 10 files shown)


src/proto/h1/conn.rs
```diff
@@ -247,8 +261,17 @@ where
...
+                #[cfg(feature = "client")]
+                seen_continue: &mut seen_continue,
             },
-        ) {
+        );
+
+        #[cfg(feature = "client")]
+        if seen_continue {
+            self.release_continue();
+        }
```
bountis, line 272: Had to include this before the parse result match because otherwise the early return in the `Pending` arm would never flip the state and deadlock.

seanmonstar:
Thanks for the PR! One thing I'll note, what I hoped to describe in #3833 is that this can be done completely outside of hyper, and we could provide helpers to do so in hyper-util. Would you be up for trying to port to helpers over there?

bountis:
> Thanks for the PR! One thing I'll note, what I hoped to describe in #3833 is that this can be done completely outside of hyper, and we could provide helpers to do so in hyper-util. Would you be up for trying to port to helpers over there?

Ah right, jumped into implementing without reading thoroughly enough. Ye no problem, will do.

PR #4183, opened 2026-09-06
feat(http1): add max_header_size limit for server and client

### LINKED ISSUE #3832: Add HTTP/1 max header size limit
I need "the ability to limit the size of Request Header".
Although it is possible to implement this in an application, I think it is most efficient if it can be handled at the protocol implementation level.

Currently, it is supported to specify the number of headers through the "h1_max_headers" option when creating a builder. I think it would be okay to control it by adding a few properties like this.

If we implement the header max setting, there may be several ways to do this.
1. number of header: already implemented
2. max size of each heade name: Fixed at 64kb
3. max size of each heade value: none (maybe)
4. max size of total header: none 

What if all of this could be set as a builder option? What do you think about this?

**Question**
If the problem was simple, I would implement it myself and post a PR, but a question arose while looking at the code.
Image
There is evidence of an attempt to add an option called "h1_max_header_size". What exactly is the max that this is trying to implement?

Aditya-9-6:
## Motivation
Resolves #3832.

Currently, HTTP/1 header size limits can only be indirectly bounded using `max_buf_size`. However, setting `max_buf_size` also restricts the maximum single read buffer for streaming request or response bodies. Furthermore, there was an existing TODO in `proto/h1/conn.rs` noting that `h1_max_header_size` should be wired up to the chunked decoder for trailers.

This PR adds explicit `max_header_size(bytes)` configuration options to both `server::conn::http1::Builder` and `client::conn::http1::Builder`.

## Solution & Architecture
- **Server Builder (`server::conn::http1::Builder`)**:
  - Exposes `pub fn max_header_size(&mut self, val: usize) -> &mut Self`.
  - Configures the maximum total size of HTTP/1 request headers (request line + headers) in bytes.
  - When exceeded, the server responds with `431 Request Header Fields Too Large` (RFC 6585 §5) and closes the connection.
- **Client Builder (`client::conn::http1::Builder`)**:
  - Exposes `pub fn max_header_size(&mut self, val: usize) -> &mut Self`.
  - Configures the maximum total size of HTTP/1 response headers (status line + headers) in bytes.
  - When exceeded, the client returns an error where `error.is_parse_too_large() == true`.
- **Early Enforcement**:
  - Enforced in `io::Buffered::parse_headers` as bytes are read, preventing buffering unbounded amounts of header data.
  - Enforced in `Server::parse` / `Client::parse` against `parsed_len` to protect scenarios where a single read buffer already contains subsequent data.
- **Chunked Trailers**:
  - Replaced the pre-existing `// TODO: remove this when we land h1_max_header_size support` placeholder in `proto::h1::conn.rs` (`decode::Decoder::new`) with the configured `h1_max_header_size`.

## Verification
- Unit test in `src/proto/h1/role.rs`:
  - `test_h1_max_header_size`: Tests exact boundary conditions on both server and client parsers.
- Integration tests in `tests/server.rs`:
  - `max_header_size_exceeded`: Verifies server returns `431 Request Header Fields Too Large` when request header size exceeds limit.
  - `max_header_size_accepted`: Verifies server accepts requests when within limit.
  - `max_header_size_with_large_max_buf_size`: Verifies `max_header_size` enforcement is decoupled from and strictly enforced even when `max_buf_size` is large.
- Integration tests in `tests/client.rs`:
  - `client_max_header_size_exceeded`: Verifies client errors with `is_parse_too_large() == true`.
  - `client_max_header_size_accepted`: Verifies client successfully receives responses within limit.
- All test suites passing (`cargo test --lib --features full`, `cargo test --test server --features full`, `cargo test --test client --features full`).

### DIFF
--- a/src/client/conn/http1.rs
+++ b/src/client/conn/http1.rs
@@ -131,6 +131,7 @@ pub struct Builder {
     h1_title_case_headers: bool,
     h1_preserve_header_case: bool,
     h1_max_headers: Option<usize>,
+    h1_max_header_size: Option<usize>,
     #[cfg(feature = "ffi")]
     h1_preserve_header_order: bool,
     h1_read_buf_exact_size: Option<usize>,
@@ -352,6 +353,7 @@ impl Builder {
             h1_title_case_headers: false,
             h1_preserve_header_case: false,
             h1_max_headers: None,
+            h1_max_header_size: None,
             #[cfg(feature = "ffi")]
             h1_preserve_header_order: false,
             h1_max_buf_size: None,
@@ -500,6 +502,17 @@ impl Builder {
         self
     }
 
+    /// Set the maximum size of response headers (including the status line) in bytes.
+    ///
+    /// If the server sends headers exceeding this limit, the error "message head is too large"
+    /// is returned.
+    ///
+    /// Default is `None`.
+    pub fn max_header_size(&mut self, val: usize) -> &mut Self {
+        self.h1_max_header_size = Some(val);
+        self
+    }
+
     /// Set whether to support preserving original header order.
     ///
     /// Currently, this will record the order in which headers are received, and store this
@@ -586,6 +599,9 @@ impl Builder {
             }
             if let Some(max_headers) = opts.h1_max_headers {
                 conn.set_http1_max_headers(max_headers);
+            }
+            if let Some(max_header_size) = opts.h1_max_header_size {
+                conn.set_http1_max_header_size(max_header_size);
             }
             #[cfg(feature = "ffi")]
             if opts.h1_preserve_header_order {
--- a/src/proto/h1/conn.rs
+++ b/src/proto/h1/conn.rs
@@ -58,6 +58,7 @@ where
                 method: None,
                 h1_parser_config: ParserConfig::default(),
                 h1_max_headers: None,
+                h1_max_header_size: None,
                 #[cfg(feature = "server")]
                 h1_header_read_timeout: None,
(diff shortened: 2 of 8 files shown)


Aditya-9-6:
I've pushed an update separating the max header size unit tests in `src/proto/h1/role.rs` behind `#[cfg(feature = "server")]` and `#[cfg(feature = "client")]`.

This resolves the undeclared `Server` compilation failure under the `Test C API (FFI)` workflow (`--features client,http1,http2,ffi`). The PR is ready for approval to run CI and review. Thanks!

src/client/conn/http1.rs
```diff
@@ -500,6 +502,17 @@ impl Builder {
         self
     }
 
+    /// Set the maximum size of response headers (including the status line) in bytes.
+    ///
+    /// If the server sends headers exceeding this limit, the error "message head is too large"
+    /// is returned.
+    ///
+    /// Default is `None`.
```
seanmonstar, line 510: While it's true there is no specific default here, it might be worth mentioning that if not configured, then the `max_buf_size` will be naturally be reached and applied.

Also, perhaps a separate small paragraph should mention that this value is also used for chunked trailers.

seanmonstar (changes requested):

src/client/conn/http1.rs
Aditya-9-6: Updated both client and server doc comments to note the natural `max_buf_size` fallback and mention that this limit is also applied to chunked trailers, and synced with latest master. Thanks @seanmonstar!

PR #4187, opened 2026-09-10
feat(http2): add an optional keep-alive reuse timeout observer

### LINKED ISSUE #4186: Allow connection pools to stop reusing an HTTP/2 connection before its keep-alive timeout

An HTTP/2 connection can remain locally ready while its network path is unavailable. If no earlier transport or protocol error is observed, a connection pool can keep assigning new requests to that connection until the keep-alive PING timeout closes it.

A concrete scenario is a network failover or NAT/firewall state change after which new connections succeed but an existing connection remains unusable. The replacement path may have lost the old connection's state: a new TCP handshake can establish fresh state, while packets belonging to the old connection are silently dropped. If the client receives no FIN, RST, or other definitive error, the existing H2 connection can remain eligible for reuse locally even though it cannot make progress. Continuing to assign requests to it delays recovery despite a working path for new connections.

There are documented examples of the underlying network behavior:

- [Google Cloud NAT](https://docs.cloud.google.com/nat/docs/troubleshooting#dropped_received_packets) documents that removing an external NAT IP without first draining it deletes the associated connection-tracking entries and drops all return traffic. With another usable NAT IP available, this provides a scenario where fresh connections can work while the old mapping no longer carries responses.

This does not assume that every network failover invalidates existing connections. A temporary interruption may also recover, which is why stopping reuse should not itself terminate existing streams.

There are two different policies involved: how long to keep existing streams alive during a recoverable interruption, and how long to keep assigning new work to that connection. Reducing the keep-alive timeout improves replacement latency, but also terminates existing streams earlier.

For example, an application may want to retain a 60-second PING timeout for existing streams while stopping reuse after an outstanding keep-alive PING has waited 5 seconds for an ACK. These are example settings, not proposed defaults.

Would an optional reuse-timeout notification on the HTTP/2 client builder fit Hyper's API?

The proposal adds `keep_alive_reuse_timeout(Option<Duration>)` and a per-connection `KeepAliveObserver`. The observer would be exposed through `hyper::client::conn::http2`. A pool such as hyper-util's legacy Client could use the notification to retire its own connection entry.

- The option defaults to disabled.
- Its timer starts with the existing keep-alive PING timeout phase.
- An ACK ready before notification cancels the soft wait.
- On expiry, Hyper invokes a nonblocking observer once, outside the PING state lock.
- The original hard deadline continues to run. Notification does not close IO, send GOAWAY, cancel streams, or change direct `SendRequest` readiness.
- A late ACK does not generate a reversal notification; the consuming pool owns its retirement policy.

The implementation uses Hyper's Timer abstraction and does not require changes to h2 or its dependency constraints.

Shortening the hard timeout also reduces interruption tolerance for existing streams. Application-level health requests or rebuilding an entire Client can provide a separate signal, but add another probe and broader connection-management logic. Deriving the signal from the existing PING wait keeps it specific to one connection.

**Additional context**

This issue tracks both parts of the proposal, with one implementation PR per repository:

- **Hyper:** expose the optional PING reuse-timeout notification through the HTTP/2 client API.
- **hyper-util:** add `http2_keep_alive_reuse_timeout(Option<Duration>)` to the legacy Client builder and bind each H2 handshake to an observer holding only that connection's `PoisonPill`. Existing pool checks then retire the entry without canceling its active streams.

The hyper-util setting would default to disabled. Its retirement policy would be irreversible even after a late ACK. Requests already checked out can race with retirement; neither PR migrates or replays requests or establishes connections proactively.

The hyper-util PR depends on the Hyper API. It can be reviewed as a draft using a companion checkout for validation, but must use an actual released Hyper version containing the API before it is ready to merge. Neither PR needs an h2 version pin.

jiahaoliang:
fix: #4186 

Connection pools may need to stop assigning new requests to an unresponsive H2 connection while retaining a longer keep-alive timeout for its existing streams. This PR implements the Hyper portion of the shared reuse-timeout proposal: a connection-specific notification for that policy without shortening the existing hard timeout. The companion hyper-util PR handles pool retirement.

The HTTP/2 client builder gains `keep_alive_reuse_timeout(Option<Duration>)` and `keep_alive_observer(...)`, with `KeepAliveObserver` exposed under `hyper::client::conn::http2`.

The new timeout defaults to disabled. When enabled, an independent timer follows the existing keep-alive PING timeout phase. If the ACK is still pending when that timer expires, the observer is taken and invoked once after releasing the PING state lock. The original hard timer remains registered with its original deadline.

Behavioral details:

- ACK processing precedes timeout notification when an ACK is ready in the same poll.
- If both deadlines have already expired without a ready ACK, the existing hard-timeout path takes priority.
- Notification does not close IO, send GOAWAY/RST_STREAM, cancel streams, change request readiness, or replay requests.
- Standalone adaptive-window PINGs do not start a separate reuse timeout.
- A configured reuse timeout must be nonzero and less than the hard timeout when keep-alive is enabled. Handshake validates the final builder configuration, independent of setter order.
- No server-facing policy or runtime dependency is added. Cargo.toml, including the existing h2 version range, is unchanged.

Validation covers timer wakeup without IO, preservation of the hard deadline, one-shot observer consumption, and configuration boundaries. Companion hyper-util tests exercise pool retirement and completion of existing response headers and streaming bodies over paused duplex IO. [Review CI](https://github.com/jiahaoliang/hyper-util/actions/runs/34431668661/job/102728281358) passed 124 Hyper unit tests, the client/server feature checks, and documentation and formatting checks. The dependency resolver selected h2 0.4.19 with the unchanged upstream version range.

This is intended as a narrow notification API; the connection pool owns retirement and replacement. API naming and placement are open to maintainer feedback.

### DIFF
--- a/src/client/conn/http2.rs
+++ b/src/client/conn/http2.rs
@@ -20,6 +20,8 @@ use crate::proto;
 use crate::rt::bounds::Http2ClientConnExec;
 use crate::rt::Timer;
 
+pub use crate::proto::h2::ping::KeepAliveObserver;
+
 /// The sender side of an established connection.
 pub struct SendRequest<B> {
     dispatch: dispatch::UnboundedSender<Request<B>, Response<IncomingBody>>,
@@ -463,6 +465,35 @@ where
         self
     }
 
+    /// Sets the keep-alive PING acknowledgement timeout for stopping reuse.
+    ///
+    /// On expiry, the observer installed with [`Self::keep_alive_observer`] is
+    /// notified once. Hyper itself does not prevent new requests or close the
+    /// connection; the observer should retire it from the caller's connection
+    /// pool. Existing streams retain the original keep-alive timeout. A late
+    /// ACK does not undo the notification.
+    ///
+    /// Defaults to `None` (disabled). Does nothing when keep-alive is disabled.
+    /// When keep-alive is enabled, a configured duration must be greater than
+    /// zero and less than [`Self::keep_alive_timeout`]; otherwise `handshake`
+    /// panics. Validation uses the final configuration, regardless of setter order.
+    pub fn keep_alive_reuse_timeout(&mut self, timeout: Option<Duration>) -> &mut Self {
+        self.h2_builder.keep_alive_reuse_timeout = timeout;
+        self
+    }
+
+    /// Sets the observer for the keep-alive reuse timeout on new connections.
+    ///
+    /// Install a separate observer for each connection when retiring individual
+    /// pool entries. The observer must be nonblocking and must not panic.
+    pub fn keep_alive_observer(
+        &mut self,
+        observer: impl KeepAliveObserver + 'static,
+    ) -> &mut Self {
+        self.h2_builder.keep_alive_observer = Some(Arc::new(observer));
+        self
+    }
+
     /// Sets whether HTTP2 keep-alive should apply while the connection is idle.
     ///
     /// If disabled, keep-alive pings are only sent while there are open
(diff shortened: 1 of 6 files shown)


jiahaoliang:
Use my own word to summary:
I think the code itself and the LLM report already covers what it does. Just add a KeepAliveObserver trait that triggers whenever a per-connection reuse timeout reached. In hyper-util https://github.com/hyperium/hyper-util/pull/319 , I implement KeepAliveObserver as poisoning the underlying connection in the client connection pool.

PR #4192, opened 2026-09-11
style(lib): Address cast-possible-truncation clippy findings.

Siech0:
- [x] The description, docs, and comments (words meant for humans) are written by a human, not by an LLM. (https://seanmonstar.com/micro/20260729-i-want-your-own-words/)

Towards https://github.com/hyperium/hyper/issues/4071


src/proto/h1/encode.rs
```diff
@@ -142,13 +142,18 @@ impl Encoder {
...
-                    BufKind::Exact(msg)
+                match usize::try_from(*remaining) {
+                    // Holding more than is owed, so write only what is left.
+                    Ok(limit) if limit < len => {
+                        *remaining = 0;
+                        BufKind::Limited(msg.take(limit))
+                    }
+                    // Ok(_) => Owed at least what we hold, write all of it.
+                    // Err(_) => Owed more than `usize` can represent, write all of it.
+                    Ok(_) | Err(_) => {
```
Siech0, line 153: Note: this can just be folded to -> _, but I felt like keeping the separation makes the clarity of why that is the case more obvious.

src/proto/h1/encode.rs
```diff
@@ -241,6 +246,7 @@ impl Encoder {
                         dst.buffer(msg);
                         !self.is_last
                     }
+                    #[allow(clippy::cast_possible_truncation, reason="usize::MAX > len > remaining, cast truncation is impossible")]
```
Siech0, line 249: I did an analysis on the assembly using  try_from would produce here, and it would be a regression in performance on 32-bit systems. the as cast here does allow the compiler to enforce a more tight invariant and it is completely sound so I allowed it explicitly.

src/proto/h2/ping.rs
```diff
@@ -396,7 +396,8 @@ impl Bdp {
         // if the current `bytes` sample is at least 2/3 the previous
         // bdp, increase to double the current sample.
         if bytes >= self.bdp as usize * 2 / 3 {
-            self.bdp = (bytes * 2).min(BDP_LIMIT) as WindowSize;
+            self.bdp = WindowSize::try_from((bytes * 2).min(BDP_LIMIT as usize))
+                .unwrap_or(WindowSize::MAX);
```
Siech0: Note: Having the `min` be within the try_from is 3 instructions faster than having it outside. It remains safe either route.

dswij (approved):

Siech0:
Updated commit history to meet commit message guidelines.

PR #4197, opened 2026-09-13
docs(examples): document hyper-util dependency and TokioIo usage

### LINKED ISSUE #3448: Make examples usable (TokioIo)
Currently, one does not simple create own project based on provided examples because almost all examples depends on internal structure [TokioIo](https://github.com/hyperium/hyper/blob/master/benches/support/tokiort.rs#L92).

VivaMomentum:
Examples in this repository use an internal benches/support module to avoid a cyclical workspace dependency on hyper-util. When users copy these examples into their own standalone projects, they encounter missing imports for TokioIo and TokioTimer.

This adds hyper-util to the list of dependencies and explains how to import TokioIo and TokioTimer from hyper_util::rt.

## Solution

- Added `hyper-util` to the dependency list in `examples/README.md`.
- Added a "Using Examples in Your Own Project" section explaining that standalone projects should import `TokioIo` and `TokioTimer` from `hyper_util::rt`.

Closes #3448


examples/README.md
```diff
@@ -13,6 +13,7 @@ A complete list of dependencies used across these examples:
 ```toml
 [dependencies]
 hyper = { version = "1", features = ["full"] }
+hyper-util = { version = "0.1", features = ["full"] }
```
dswij, line 16: I guess this part is not true then because we avoid using `hyper-util` in examples.

cratelyn: seconded.

